PR #1, opened 2023-05-12
Fix CICD

aholstrup1:


.github/AL-Go-Settings.json
```diff
@@ -37,5 +37,5 @@
   "CICDPushBranches": [
     "main"
   ],
-  "rulesetFile": "..\\Build\\Rulesets\\module.ruleset.json"
+  "rulesetFile": "..\\..\\Build\\Rulesets\\module.ruleset.json"
```
mazhelez, line 40: Ruleset file paths are relevant to what now?

PR #21, opened 2023-05-17
[main] Update baseline package handling

aholstrup1:
Update baseline package handling to not depend on BCArtifacts anymore. 

CLEAN mode should grab the latest version out of main as the baseline 
Default/Translated mode will use the package defined in Packages.Json as the baseline. This should always be a build out of release/22

Here's a test run of the CICD workflow out of this branch:
https://github.com/microsoft/BCApps/actions/runs/5001796789

### DIFF
--- a/Build/Scripts/EnlistmentHelperFunctions.psm1
+++ b/Build/Scripts/EnlistmentHelperFunctions.psm1
@@ -137,7 +137,7 @@ function Get-PackageLatestVersion() {
 .Synopsis
     Installs a package from a NuGet.org feed
 .Parameter PackageName
-    The name of the package to look for in the Packages config
+    The name of the package to install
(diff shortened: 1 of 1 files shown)


Build/Scripts/EnlistmentHelperFunctions.psm1
```diff
@@ -139,39 +133,15 @@ function Get-PackageLatestVersion() {
...
-    }
-
-    return $updatedBaseline
-}
-
 <#
 .Synopsis
     Installs a package from a NuGet.org feed
 .Parameter PackageName
     The name of the package to look for in the Packages config
```
mazhelez: It's longer needed to look in the Packages config, right?

```suggestion
    The name of the package to install
```

PR #38, opened 2023-05-30
Fetch correct baseline package 

aholstrup1:
Right now when we update packages, we fetch version 23.x.x.x for Microsoft-Dynamics-BusinessCentral-BCApps. However, our baselines should come from the previously release and not from an earlier build out of master.

### DIFF
--- a/Build/Scripts/UpdatePackageVersions.ps1
+++ b/Build/Scripts/UpdatePackageVersions.ps1
@@ -45,7 +45,7 @@ foreach($packageName in $packageNames)
 if ($updatesAvailable) {
     # Create branch and push changes
     Set-GitConfig -Actor $Actor
-    $BranchName = New-TopicBranch -Category "UpdatePackageVersions"
+    $BranchName = New-TopicBranch -Category "UpdatePackageVersions/$TargetBranch"
     $title = "[$TargetBranch] Update package versions"
     Push-GitBranch -BranchName $BranchName -Files @("Build/Packages.json") -CommitMessage $title
 


Build/Packages.json
```diff
@@ -1,8 +1,9 @@
 {
     "Microsoft.Dynamics.BusinessCentral.Translations":  {
-                                                            "Version":  "23.0.20230511.4"
+                                                            "Version":  "23.0.20230529.3"
                                                         },
     "Microsoft-Dynamics-BusinessCentral-BCApps":  {
-                                                      "Version":  "22.2.31"
+                                                      "Version":  "22.2.45",
+                                                      "UsePreviousMajorMinor":  true
```
mazhelez: Wouldn't it be better to be explicit on the base version?
E.g. instead of `UsePreviousMajorMinor`, add `baseVersion` (name is TBD).

So, in the _main_ branch, it will be `22.x`, and in the case of _release/22_, it will be... again `22.x`.

aholstrup1: It just means we'll have another configuration we'll have to remember to update when we do a version change. The fewer of these the better IMO

mazhelez: With your approach there'll be an "automatic" change of the baseline one we branch a new release branch, which is not ideal as well.

The same way we control the version of the baselines (we don't automatically pick the latest one, but instead have an automation that updates the version we use every now and then), we might want to control... well, the version of the baseline for the _main_ branch.

aholstrup1: What about MaxVersion? Once we want the baseline to be a 23.x.x.x version, we update MaxVersion to 24

mazhelez: Either `BaseVersion` or `MaxVersion` would do.

With `BaseVersion` (e.g. 22), we take the max version where the major is 22.
With `MaxVersion` (e.g. 23) we take the max version where the major is less than 23 (e.g. 22). 
So, no real difference IMO.

Build/Scripts/UpdatePackageVersions.ps1
```diff
@@ -22,7 +22,8 @@ foreach($packageName in $packageNames)
 {
     $currentPackage = Get-ConfigValue -Key $packageName -ConfigType Packages
     $currentVersion = $currentPackage.Version
-    $latestVersion = Get-PackageLatestVersion -PackageName $packageName
+
+    $latestVersion = Get-PackageLatestVersion -PackageName $packageName -MaxVersion $currentPackage.MaxVersion
```
mazhelez: Wouldn't that call fail if `$currentPackage.MaxVersion` isn't defined?

Build/Scripts/EnlistmentHelperFunctions.psm1
```diff
@@ -116,19 +116,24 @@ function Set-ConfigValue() {
     For example, if the repo version is 1.2, the function will look for the latest version of the package that has major.minor = 1.2.
 .Parameter PackageName
     The name of the package
+.Parameter MaxVersion
+    The maximum version to look for. If not specified, the function will use the `repoVersion` setting.
```
mazhelez: It wouldn't use `repoVersion` as a max version, but as a "base version" 😄

Build/Scripts/UpdatePackageVersions.ps1
aholstrup1: It didn't fail for me, but I guess it's best to check this :) Added a check for it now

Build/Scripts/UpdatePackageVersions.ps1
```diff
@@ -40,7 +45,7 @@ foreach($packageName in $packageNames)
 if ($updatesAvailable) {
     # Create branch and push changes
     Set-GitConfig -Actor $Actor
-    $BranchName = New-TopicBranch -Category "UpdatePackageVersions/$TargetBranch"
+    $BranchName = New-TopicBranch -Category "UpdatePackageVersions"
```
mazhelez: If there's no `TargetBranch` in the branch name, don't we risk a branch collision?
The automation for updating the versions runs in parallel for all branches, so there is a chance that the update branches are created in the same minute, no?

aholstrup1: Good catch - That change wasn't intentional. I'll bring it back

PR #52, opened 2023-06-19
Adding Repository Sync Action

aholstrup1:


.github/workflows/PostGate.yaml
```diff
@@ -0,0 +1,62 @@
...
+    steps:
+    - name: Checkout mirror
+      uses: actions/checkout@v3
+      with:
+        ref: ${{ env.HEAD_BRANCH }}
+        repository: ${{ secrets.MIRRORREPONAME }}
+        token: ${{ secrets.SYNCREPOSITORYTOKEN }}
+        fetch-depth: 0
+
+    - name: Push latest changes to mirror
```
mazhelez, line 47: Oh, do we really want to expose that there is a mirror? There should be a way to sync repos "behind the scene"

PR #67, opened 2023-07-12
Add Sync Pipeline 

aholstrup1:
This PR sets up an ADO pipeline that can sync changes between GitHub and ADO


Build/Scripts/SyncMirror.ps1
```diff
@@ -0,0 +1,49 @@
...
+    } else {
+        az login
+    }
+    return (az account get-access-token | ConvertFrom-Json)
+}
+
+$Branch = $Branch -replace "refs/heads/", ""
+$MIAccessToken = Get-AccessToken -ManagedIdentityAuth:$ManagedIdentityAuth
+
+git clone "https://$($MIAccessToken.accessToken)@$TargetRepository" BCApps
```
tomasmsft, line 26: consider cloning to <FullPathTo\BCApps> - we "don't know" what current directory is.
On the other hand git clone will fail if path exits

PR #80, opened 2023-08-03
Create tag when CI/CD is requested

aholstrup1:
Rather than waiting for the CI/CD to complete before tagging a commit, we can create the tag as soon as the CI/CD workflow is submitted. This will help with the sync mechanism so it the syncing also includes the tag of a recent commit.

### DIFF
--- a/.azuredevops/SyncMirroredRepository.yml
+++ b/.azuredevops/SyncMirroredRepository.yml
@@ -29,7 +29,9 @@ jobs:
     displayName: Sync mirror with upstream
     inputs:
       filePath: Build\Scripts\SyncMirror.ps1
+      # If pipeline is triggered by a branch we sync the branch with latest changes from source repository
       ${{ if startsWith(variables['Build.SourceBranch'], 'refs/heads') }}:
         arguments: '-SourceRepository "https://$(GitHubPAT)@github.com/microsoft/BCApps.git" -TargetRepository $(TargetRepository) -Branch $(Build.SourceBranch) -ManagedIdentityAuth'
+      # If pipeline is triggered by a new tag we only sync the tags
       ${{ if startsWith(variables['Build.SourceBranch'], 'refs/tags') }}:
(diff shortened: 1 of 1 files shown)


Build/Scripts/SyncMirror.ps1
```diff
@@ -31,19 +30,23 @@ RunAndCheck git reset HEAD --hard
...
 
-if (RunAndCheck git ls-remote origin $branch) {
-    RunAndCheck git checkout origin/$branch --track
-    RunAndCheck git pull origin $branch
-}
-else {
-    RunAndCheck git checkout upstream/$branch --track
+if ($Branch -match "refs/heads/") {
+    $Branch = $Branch -replace "refs/heads/", ""
+    if (RunAndCheck git ls-remote origin $branch) {
```
mazhelez: Is this how we differentiate between tags and branches? A comment or two in here would help a lot.

.github/workflows/CreateBuildTag.yaml
```diff
@@ -3,46 +3,43 @@ name: 'Create Build Tag'
...
     runs-on: windows-latest
     steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Get Build Version
+        id: GetBuildVersion
+        run: |
+          $event = Get-Content $env:GITHUB_EVENT_PATH -Raw | ConvertFrom-Json
+          $buildNumber = $event.workflow_run.run_number
```
mazhelez: Minor: might simply use `github.event.workflow_run.run_number` as `$buildNumber`

E.g.
```
...
    env: 
       buildNumber: ${{ github.event.workflow_run.run_number }}
...
```

Build/Scripts/SyncMirror.ps1
aholstrup1: Yes it is - I've refactored this a bit and added comments + logging

PR #97, opened 2023-08-18
Use specific version of the bcinsider artifact

mazhelez:
- Use specific version of the _bcinsider_ artifact and add a workflow to regularly uptake the newest version
- Use app baselines from a BC artifact
- Disable breaking checks for Clean mode as it's failing (known issue)
- Disable Translated mode for System Application as it's failing due to an issue with translations

### DIFF
--- a/Build/Packages.json
+++ b/Build/Packages.json
@@ -1,6 +1,6 @@
 {
     "Microsoft.Dynamics.BusinessCentral.Translations":  {
-                                                            "Version":  "23.0.20230821.5",
+                                                            "Version":  "23.0.20230605.5",
                                                             "Source": "NuGet.org"
                                                         },
     "AppBaselines-BCArtifacts":  {


Build/Scripts/EnlistmentHelperFunctions.psm1
```diff
@@ -116,24 +116,60 @@ function Set-ConfigValue() {
...
 .Parameter PackageName
     The name of the package
-.Parameter MaxVersion
-    The maximum version to look for
 #>
 function Get-PackageLatestVersion() {
     param(
         [Parameter(Mandatory=$true)]
-        [string] $PackageName,
-        [string] $MaxVersion
```
mazhelez, line 126: Max version should probably stay.

Build/Scripts/EnlistmentHelperFunctions.psm1
```diff
@@ -116,24 +116,76 @@ function Set-ConfigValue() {
...
+function Get-LatestBCArtifactVersion
+(
+    [Parameter(Mandatory=$true)]
+    $minimumVersion
+)
+{
+    $artifactUrl = Get-BCArtifactUrl -type Sandbox -country base -version $minimumVersion -select Latest
+
+    if(-not $artifactUrl) {
+        #Fallback to bcinsider. Should go away soon. Haha, yeah right.
```
mazhelez: Fix!

Build/Scripts/EnlistmentHelperFunctions.psm1
mazhelez: Not really, no.

.github/workflows/UpdateBCArtifactVersion.yaml
```diff
@@ -0,0 +1,52 @@
...
+      UpdateBranches: ${{ steps.GetBranches.outputs.UpdateBranches }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Get Branches
+        id: GetBranches
+        run: |
+          git fetch
+          $branches = @(git for-each-ref --format="'%(refname:short)'" refs/remotes/origin/release/ | % { $_ -replace 'origin/', '' })
```
aholstrup1, line 26: I think for the sake of keeping it consistent, I think we should name our branches releases/xxx

mazhelez: I thought _release**s**/22_ doesn't work with AL-Go?

mazhelez: Anyway, let's have at it separately (not this PR).

PR #134, opened 2023-08-30
Trigger sync daily and don't include source code on package

aholstrup1:


.azuredevops/SyncMirroredRepository.yml
```diff
@@ -1,15 +1,13 @@
...
-  batch: true
+trigger: none
+schedules:
+- cron: '0 8 * * *'
+  displayName: Daily Sync
   branches:
     include:
     - main
-    - release/*
-  tags:
```
mazhelez, line 10: If syncing isn't triggered on tags, lines around 34 aren't needed anymore.
Also, Build\Scripts\SyncMirror.ps1 might need some cleaning as well.

aholstrup1: I've cleaned up SyncMirroredRepository.yml. I think it's fine to keep the if(branch) statement in SyncMirror.ps1 though

PR #137, opened 2023-08-30
Add performance toolkit project and split System Application project

aholstrup1:


Projects/Business Central Performance Toolkit/.AL-Go/settings.json
```diff
@@ -0,0 +1,14 @@
+{
+    "appFolders":  [
+        "..\\..\\Apps\\DevTools\\BusinessCentralPerformanceToolkit",
+        "..\\..\\Apps\\DevTools\\BusinessCentralPerformanceToolkit Tests"
```
mazhelez: Shouldn't "BusinessCentralPerformanceToolkit Tests" be in `testFolders`?

Projects/Business Central Performance Toolkit/.AL-Go/settings.json
```diff
@@ -0,0 +1,14 @@
...
+    ],
+    "testFolders":  [
+    ],
+    "buildModes": [
+        "Translated"
```
mazhelez: I don't think it should be translated. At least in projects.json the `hasTranslations` prop isn't set to `true`.

Projects/System Application Tests/.AL-Go/RunTestsInBcContainer.ps1
```diff
@@ -0,0 +1,6 @@
+Param(
```
mazhelez, line 1: Same file should probably be removed from "System Application\\.AL-Go"

Projects/Business Central Performance Toolkit/.AL-Go/settings.json
```diff
@@ -0,0 +1,14 @@
+{
```
mazhelez: The app is called `Performance Toolkit`. Not sure why the folder is _BusinessCentralPerformanceToolkit_.
Can we use at least call the AL-Go project that way? _Business Central Performance Toolkit_ is super long...

PR #201, opened 2023-10-02
Allow obsolete tags up to MajorVersion + 3 when building in CLEAN mode

aholstrup1:
When building in CLEAN mode we will sometimes encounter obsolete tags that are greater than the current version in master. 
https://github.com/microsoft/BCApps/actions/runs/6352615458/job/17256197076

Allow obsolete tags up to MajorVersion + 3 when building in CLEAN mode


Build/Scripts/GuardingV2ExtensionsHelper.psm1
```diff
@@ -185,6 +187,15 @@ function Update-AppSourceCopVersion
         $obsoleteTagAllowedVersions += "$i.0"
     }
 
+    # Add 3 versions for tasks built with CLEANpreProcessorSymbols
+    if ($BuildMode -eq "Clean")
```
mazhelez: What about just adding the following before the loop above?

```
# Add 3 versions for tasks built with CLEANpreProcessorSymbols
if ($BuildMode -eq "Clean") {
    $maxAllowedObsoleteVersion += 3
}
```

PR #206, opened 2023-10-02
Add workflow to verify AL-Go version

mazhelez:
- [ ] Doc
- [ ] Unit tests?

### DIFF
--- a/.github/workflows/VerifyALGoVersion.yaml
+++ b/.github/workflows/VerifyALGoVersion.yaml
@@ -25,5 +25,7 @@ jobs:
 
       - name: Verify AL-Go version
         id: verifyALGoVersion
+        env:
+          GH_TOKEN: ${{ github.token }}
         run: |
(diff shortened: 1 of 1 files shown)


Build/Scripts/VerifyALGoVersion.ps1
```diff
@@ -0,0 +1,35 @@
+<#
+.SYNOPSIS
+    Verifies the currently used AL-Go version is approved.
+#>
+Param(
```
aholstrup1, line 5: Remove?

.github/workflows/VerifyALGoVersion.yaml
```diff
@@ -0,0 +1,31 @@
+name: 'Verify AL-Go Version'
```
aholstrup1, line 1: I wonder if the naming should be more general, so we could also add e.g. Pester tests to this workflow. Maybe "CI Tests" or something like that

PR #230, opened 2023-10-09
Bug 486975: Remove unused using directives.

JesperSchulz:
Remove using directives, which won't be needed when CLEAN tags have been applied.

### DIFF
--- a/src/System Application/App/Azure AD Plan/src/Plan Configuration/PlanConfiguration.Codeunit.al
+++ b/src/System Application/App/Azure AD Plan/src/Plan Configuration/PlanConfiguration.Codeunit.al
@@ -5,7 +5,9 @@
 
 namespace System.Azure.Identity;
 
+#if not CLEAN22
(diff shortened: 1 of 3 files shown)


src/System Application/App/Azure AD Plan/src/Plan Configuration/PlanConfiguration.Codeunit.al
```diff
@@ -5,7 +5,9 @@
 
 namespace System.Azure.Identity;
 
+#if not CLEAN22
```
AndreasMoth, line 8: Could you also turn the warning into an error for clean builds to avoid this in the future?

JesperSchulz: That is already done. In order to uptake the new platform, where this is an error, we need to address all these issues. Hence this PR.

PR #232, opened 2023-10-09
Set AS0008 action to None in AppSourceCop.ruleset.json

samyhaff:
This rule should be disabled here. It checks that we the no Microsoft or System namespaces are defined.

### DIFF
--- a/src/rulesets/PTECop.ruleset.json
+++ b/src/rulesets/PTECop.ruleset.json
@@ -83,7 +83,8 @@
                   },
                   {
                       "id":  "PTE0021",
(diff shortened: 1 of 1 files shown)


src/rulesets/AppSourceCop.ruleset.json
```diff
@@ -31,7 +31,7 @@
                   },
                   {
                       "id":  "AS0008",
-                      "action":  "Error"
+                      "action":  "None"
```
grobyns: pls add the rule description as justification.

PR #233, opened 2023-10-09
Set PTE0021 action to None in PTECop.ruleset.json

samyhaff:
This rule should be disabled here. It checks that we the no Microsoft or System namespaces are defined.


src/rulesets/PTECop.ruleset.json
```diff
@@ -83,7 +83,7 @@
                   },
                   {
                       "id":  "PTE0021",
-                      "action":  "Error"
+                      "action":  "None"
```
grobyns: pls add the rule description as justification.

PR #234, opened 2023-10-09
Trigger full build on workflow and rulesets changes

mazhelez:
Full Build should be triggered if the rulesets are changed, as well as the workflows responsible for the build.

### DIFF
--- a/.github/AL-Go-Settings.json
+++ b/.github/AL-Go-Settings.json
@@ -54,7 +54,10 @@
     "skipUpgrade":  true,
     "PartnerTelemetryConnectionString":  "InstrumentationKey=403ba4d3-ad2b-4ca1-8602-b7746de4c048;IngestionEndpoint=https://swedencentral-0.in.applicationinsights.azure.com/",
     "fullBuildPatterns":  [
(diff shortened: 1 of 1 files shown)


.github/AL-Go-Settings.json
```diff
@@ -54,7 +54,10 @@
     "skipUpgrade":  true,
     "PartnerTelemetryConnectionString":  "InstrumentationKey=403ba4d3-ad2b-4ca1-8602-b7746de4c048;IngestionEndpoint=https://swedencentral-0.in.applicationinsights.azure.com/",
     "fullBuildPatterns":  [
-                              "build/*"
+                              "build/*",
+                              "src/rulesets/*",
+                              ".github/workflows/PullRequestHandler.yaml",
```
aholstrup1, line 59: Why not everything under the .github folder?

mazhelez: Well, AL-Go-Settings.json is included by default. And for the rest of the workflows, it doesn't matter.

If a PR changes, say CICD.yaml, that has no effect on the PR Build whatsoever.

PR #240, opened 2023-10-11
Make sure feature telemetry state updates can be done in quick succession

IhorHandziuk:
**Problem:**
If feature telemetry state update calls are done quicker than it takes to start a session (a few milliseconds), then the update might not go through correctly.

**Solution:**
If the same feature telemetry instance needs to update the state of a feature more than once in a row, then make sure that the previous session had time to complete.

### DIFF
--- a/src/System Application/App/Telemetry/src/Feature Telemetry/FeatureUptakeStatusImpl.Codeunit.al
+++ b/src/System Application/App/Telemetry/src/Feature Telemetry/FeatureUptakeStatusImpl.Codeunit.al
@@ -145,6 +145,14 @@ codeunit 8705 "Feature Uptake Status Impl."
         exit(false);
     end;
 
+    /// <summary>
+    /// The updates to Feature Uptake states must be done in a separate session (e. g. to not interfere with subsequent Page.RunModal() calls).
+    /// Normally we don't need to wait for this separate session to finish, as feature uptake states are expected to not change rapidly, and it only takes a few milliseconds for the session to complete.
+    /// But if the feature registers uptake states one right after another, then the session may not have enough time to start and acquire a lock on the "Feature Uptake" table by the time the next uptake state is registered.
+    /// For such cases, we can wait for this previous session to finish.
+    /// </summary>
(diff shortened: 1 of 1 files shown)


src/System Application/App/Telemetry/src/Feature Telemetry/FeatureUptakeStatusImpl.Codeunit.al
```diff
@@ -139,4 +144,24 @@ codeunit 8705 "Feature Uptake Status Impl."
...
+    local procedure WaitForStartedUpdateFeatureUptakeSession()
+    var
+        StartDateTime: DateTime;
+        Timeout: Duration;
+    begin
+        if StartedSessionId = 0 then
+            exit;
+
+        Timeout := 100; // wait for up 0.1 seconds
+        StartDateTime := CurrentDateTime;
```
stkillen: Minor: Parentheses missing

IhorHandziuk: Added

src/System Application/App/Telemetry/src/Feature Telemetry/FeatureUptakeStatusImpl.Codeunit.al
```diff
@@ -13,6 +13,11 @@ codeunit 8705 "Feature Uptake Status Impl."
     TableNo = "Feature Uptake";
     Permissions = tabledata "Feature Uptake" = rimd;
 
+    var
+        StartedSessionId: Integer;
```
AndreasMoth, line 17: Would it make sense to make this codeunit single instance instead? That way the benefit is applied to the entire session.

IhorHandziuk: If we were to go down that path, we would probably want to add some additional checks to make sure we are not waiting for a session that updates an unrelated feature.
Overall, I don't think this bug fix is worth that big of an investment.

PR #244, opened 2023-10-12
AL0792 - Decrease severity temporarily so that violations can be fixed

BazookaMusic:
**Issue**
Due to a bug in how this rule was promoted from a ruleset, there are some places where the base app has violated it. 

**Solution**
Temporarily demote it to a warning in order to allow BaseApp to build.

### DIFF
--- a/src/rulesets/Compiler.ruleset.json
+++ b/src/rulesets/Compiler.ruleset.json
@@ -3167,7 +3167,7 @@
         },
         {
             "id": "AL0792",
-            "action": "Warning"
+            "action": "Error"
         },
         {
             "id": "AL0793",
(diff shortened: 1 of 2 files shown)


src/rulesets/Compiler.ruleset.json
```diff
@@ -3167,7 +3167,7 @@
         },
         {
             "id": "AL0792",
-            "action": "Error"
+            "action": "Warning"
```
aholstrup1: Could you put it under ruleset.json? I think that's usually where we put ruleset exceptions. Could you also include a justification?

BazookaMusic: Yep, changed it

PR #254, opened 2023-10-17
Introduce "User Details" module

IhorHandziuk:
**Problem:**
There is a lot of information that would be valuable to see on the `Users` page in a list view for search and filtering purposes (such as details about user plans, permission sets, user properties etc). But it's not possible to have table extensions for system tables, so such fields couldn't have been added directly.

**Solution:**
- Introduce a new temporary table `User Details` that fetches information from the `User` table and other related tables.
- Introduce a new page that surfaces the information from the `User Details` table in a list view.

**Note:**
The new page will be referenced in the `Navigate` pane of the `Users` page.

### DIFF
--- a/src/System Application/App/Azure AD Plan/src/User Details/PlanUserDetails.Codeunit.al
+++ b/src/System Application/App/Azure AD Plan/src/User Details/PlanUserDetails.Codeunit.al
@@ -29,7 +29,10 @@ codeunit 776 "Plan User Details"
             end;
 
         UserDetails."User Plans" := CopyStr(UserPlansTextBuilder.ToText().TrimEnd(' ; '), 1, MaxStrLen(UserDetails."User Plans"));
-        UserDetails."Is Delegated" := AzureADPlan.IsPlanAssignedToUser(PlanIds.GetDelegatedAdminPlanId(), UserSecID) or AzureADPlan.IsPlanAssignedToUser(PlanIds.GetHelpDeskPlanId(), UserSecID) or AzureADPlan.IsPlanAssignedToUser(PlanIds.GetD365AdminPartnerPlanId(), UserSecID);
+        UserDetails."Is Delegated" := AzureADPlan.IsPlanAssignedToUser(PlanIds.GetDelegatedAdminPlanId(), UserSecID) or
+                                      AzureADPlan.IsPlanAssignedToUser(PlanIds.GetHelpDeskPlanId(), UserSecID) or
+                                      AzureADPlan.IsPlanAssignedToUser(PlanIds.GetD365AdminPartnerPlanId(), UserSecID);
+
         UserDetails."Has M365 Plan" := AzureADPlan.IsPlanAssignedToUser(PlanIds.GetMicrosoft365PlanId(), UserSecId);
 
         UsersInPlans.SetFilter(Plan_Name, '*Essential*');
--- a/src/System Application/App/User Details/Permissions/UserDetailsView.PermissionSet.al
+++ b/src/System Application/App/User Details/Permissions/UserDetailsView.PermissionSet.al
@@ -4,6 +4,7 @@
 // ------------------------------------------------------------------------------------------------
 
 namespace System.Security.User;
+
 using System.Security.AccessControl;
 using System.Environment.Configuration;
 
--- a/src/System Application/App/User Details/src/UserDetails.Codeunit.al
+++ b/src/System Application/App/User Details/src/UserDetails.Codeunit.al
@@ -16,11 +16,11 @@ codeunit 774 "User Details"
     /// Retrieves the details of a user.
     /// </summary>
     /// <param name="UserDetails">The user details record to be populated.</param>
-    procedure GetUserDetails(var UserDetails: Record "User Details")
+    procedure Get(var UserDetails: Record "User Details")
     var
         UserDetailsImpl: Codeunit "User Details Impl.";
     begin
-        UserDetailsImpl.GetUserDetails(UserDetails);
+        UserDetailsImpl.Get(UserDetails);
     end;
 
     /// <summary>
--- a/src/System Application/App/User Details/src/UserDetails.Page.al
+++ b/src/System Application/App/User Details/src/UserDetails.Page.al
@@ -97,7 +97,7 @@ page 774 "User Details"
         UserDetails: Codeunit "User Details";
     begin
         IsSaaS := EnvironmentInformation.IsSaaS();
-        UserDetails.GetUserDetails(Rec);
+        UserDetails.Get(Rec);
     end;
 
     protected var
--- a/src/System Application/App/User Details/src/UserDetailsImpl.Codeunit.al
+++ b/src/System Application/App/User Details/src/UserDetailsImpl.Codeunit.al
@@ -4,6 +4,7 @@
 // ------------------------------------------------------------------------------------------------
 
 namespace System.Security.User;
+
 using System.Security.AccessControl;
 
 codeunit 775 "User Details Impl."
@@ -12,7 +13,7 @@ codeunit 775 "User Details Impl."
     InherentEntitlements = X;
     InherentPermissions = X;
 
-    procedure GetUserDetails(var UserDetails: Record "User Details")
+    procedure Get(var UserDetails: Record "User Details")
     var
         User: Record User;
         LocalUserDetails: Record "User Details";
--- a/src/System Application/Test Library/User Details/src/UserDetailsTestLibrary.Codeunit.al
+++ b/src/System Application/Test Library/User Details/src/UserDetailsTestLibrary.Codeunit.al
@@ -19,18 +19,18 @@ codeunit 132001 "User Details Test Library"
     /// <summary>
     /// Saves the user details inside this instance, so that they can be access later.
     /// </summary>
-    procedure FetchUserDetails()
+    procedure GetUserDetails()
     var
         UserDetails: Codeunit "User Details";
     begin
-        UserDetails.GetUserDetails(UserDetailsRec);
+        UserDetails.Get(UserDetailsRec);
     end;
 
     /// <summary>
     /// Gets the value of "Has SUPER permission set" for the user.
     /// </summary>
     /// <param name="UserSID">The user security ID</param>
-    procedure GetHasSuperPermissionSet(UserSID: Guid): Boolean
+    procedure HasSuperPermissionSet(UserSID: Guid): Boolean
     begin
         if not UserDetailsRec.Get(UserSID) then
             Error(UserDoesNotExistErr, UserSID);
@@ -43,7 +43,7 @@ codeunit 132001 "User Details Test Library"
     /// Gets the value of "User Plans" for the user.
     /// </summary>
     /// <param name="UserSID">The user security ID</param>
-    procedure GetUserPlans(UserSID: Guid): Text
+    procedure UserPlans(UserSID: Guid): Text
     begin
         if not UserDetailsRec.Get(UserSID) then
             Error(UserDoesNotExistErr, UserSID);
@@ -55,7 +55,7 @@ codeunit 132001 "User Details Test Library"
     /// Gets the value of "Is Delegated" for the user.
     /// </summary>
     /// <param name="UserSID">The user security ID</param>
-    procedure GetIsDelegated(UserSID: Guid): Boolean
+    procedure IsDelegated(UserSID: Guid): Boolean
     begin
         if not UserDetailsRec.Get(UserSID) then
             Error(UserDoesNotExistErr, UserSID);
@@ -67,7 +67,7 @@ codeunit 132001 "User Details Test Library"
     /// Gets the value of "Has M365 Plan" for the user.
     /// </summary>
     /// <param name="UserSID">The user security ID</param>
-    procedure GetHasM365Plan(UserSID: Guid): Boolean
+    procedure HasM365Plan(UserSID: Guid): Boolean
     begin
         if not UserDetailsRec.Get(UserSID) then
             Error(UserDoesNotExistErr, UserSID);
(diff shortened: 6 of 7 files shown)


src/System Application/App/User Details/src/UserDetailsImpl.Codeunit.al
```diff
@@ -0,0 +1,61 @@
...
+            repeat
+                UserDetails."User Security ID" := User."User Security ID";
+                FillInUserDetails(UserDetails);
+                UserDetails.Insert();
+            until User.Next() = 0;
+    end;
+
+    local procedure FillInUserDetails(var UserDetails: Record "User Details")
+    var
+        AzureADUserManagement: Codeunit "Azure AD User Management";
```
grobyns: I'm not sure why this is a separate module. Why is getting user info out of AAD (ups Entra) not simply part of one of the AAd modules we already have?

src/System Application/App/User Details/src/UserDetails.Page.al
```diff
@@ -0,0 +1,137 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Security.User;
+
+using System.Environment;
+
+page 774 "User Details"
+{
+    ApplicationArea = Basic, Suite;
```
darjoo, line 17: Should this not be `ApplicationArea = All;`?

src/System Application/App/User Details/src/UserDetails.Codeunit.al
```diff
@@ -0,0 +1,25 @@
...
+/// </summary>
+codeunit 774 "User Details"
+{
+    Access = Public;
+
+    /// <summary>
+    /// Retrieves the details of a user.
+    /// </summary>
+    /// <param name="UserDetails">The user details record to be populated.</param>
+    procedure GetUserDetails(var UserDetails: Record "User Details")
```
darjoo: The table has almost all fields as internal, what's the point of this public facing API?
Should partners be using this at all?

IhorHandziuk: The idea is that partners can make a table extension and then a page extension to surface new fields on the page. Setting most of the fields as internal, as there are other ways to get that data, so they don't need to be exposed.

src/System Application/App/User Details/src/UserDetails.Page.al
IhorHandziuk: I followed the application area of the `Users` page. It would make sense for the 2 to be the same.

src/System Application/App/User Details/src/UserDetailsImpl.Codeunit.al
IhorHandziuk: It's not only data from Entra, in the future all sorts of data can be added to the table. Ideally, the introduced table and page would only have `User Security ID` (+ fields from system tables), and then it would be up to other modules to add table extensions and page extension with the relevant data. But since we ship System Application as a single app, and it's impossible to have the base table and table extensions in the same app, the Entra stuff has to go on the main table.

src/System Application/App/Azure AD Plan/src/User Details/PlanUserDetails.PageExt.al
```diff
@@ -0,0 +1,70 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Azure.Identity;
+
+using System.Security.User;
+
+pageextension 774 "Plan User Details" extends "User Details"
+{
```
grobyns, line 14: missing docs no?

src/System Application/App/Azure AD Plan/src/User Details/PlanUserDetails.TableExt.al
```diff
@@ -0,0 +1,65 @@
...
+using System.Security.User;
+
+/// <summary>
+/// Adds information about user plans to the user details module.
+/// </summary>
+tableextension 774 "Plan User Details" extends "User Details"
+{
+    fields
+    {
+        /// <summary>
```
grobyns, line 17: documenting internals and not public stuff? seems odd :P

src/rulesets/PTECop.ruleset.json
```diff
@@ -7,7 +7,8 @@
                   },
                   {
                       "id":  "PTE0002",
-                      "action":  "Error"
+                      "action":  "None",
```
grobyns: please don't edit this ruleset. Edit/Add it in ruleset.json instead.

grobyns (dismissed):

src/System Application/App/User Details/src/UserDetails.Page.al
duichwer, line 1: Personally I would explicit set InsertAllowed, Delete Allowed and Modify Allowed to false additionally.

The Editable property can be changed by Page Extensions. The other properties can't be changed.

src/System Application/App/User Details/src/UserDetails.Table.al
```diff
@@ -0,0 +1,135 @@
...
+            Editable = false;
+            FieldClass = FlowField;
+            Access = Internal;
+        }
+        /// <summary>
+        /// An ID assigned to the user in Microsoft Entra.
+        /// </summary>
+        field(22; "Authentication Object ID"; Text[80])
+        {
+            Caption = 'Entra Object ID';
```
pri-kise: Normally all the properties are sorted (check out e.g. _table 8889 "Sent Email"_).
If you're using AL Code Outline, then there's a code action to align this with other objects.

IhorHandziuk: Didn't know about this extension. Seems to have a lot of nice features - maybe we'll start using it internally.

pri-kise: I think one of the best command is CleanUp on Uncommitted Files in Project.
You have to set the task in the settings.json (https://github.com/anzwdev/al-code-outline#settings) first, but then every developer can perform this task before committing and publishing.
image

src/System Application/App/User Details/Permissions/UserDetailsView.PermissionSet.al
```diff
@@ -0,0 +1,19 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Security.User;
```
AndreasMoth, line 6: Minor: Newline between namespace and using.

src/System Application/App/User Details/src/UserDetailsImpl.Codeunit.al
```diff
@@ -0,0 +1,33 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Security.User;
```
AndreasMoth, line 6: Minor: Newline between namespace and using.

src/System Application/App/Azure AD Plan/src/User Details/PlanUserDetails.Codeunit.al
```diff
@@ -0,0 +1,43 @@
...
+    begin
+        UsersInPlans.SetRange(User_Security_ID, UserSecId);
+        if UsersInPlans.Open() then
+            while UsersInPlans.Read() do begin
+                UserPlansTextBuilder.Append(UsersInPlans.Plan_Name);
+                UserPlansTextBuilder.Append(' ; ');
+            end;
+
+        UserDetails."User Plans" := CopyStr(UserPlansTextBuilder.ToText().TrimEnd(' ; '), 1, MaxStrLen(UserDetails."User Plans"));
+        UserDetails."Is Delegated" := AzureADPlan.IsPlanAssignedToUser(PlanIds.GetDelegatedAdminPlanId(), UserSecID) or AzureADPlan.IsPlanAssignedToUser(PlanIds.GetHelpDeskPlanId(), UserSecID) or AzureADPlan.IsPlanAssignedToUser(PlanIds.GetD365AdminPartnerPlanId(), UserSecID);
```
AndreasMoth: Could you multiline this and align spacing such that it's easy to see the difference between each call (the PlanIds reference). That would make this much more readable.

src/System Application/App/User Details/src/UserDetails.Codeunit.al
```diff
@@ -0,0 +1,35 @@
...
```
AndreasMoth: This means you call UserDetails.GetUserDetails(UserDetails), maybe it should just be called Get? Then you would call UserDetails.Get(UserDetails); since the codeunit is already called UserDetails this seems a bit excessive to also name it that.

src/System Application/App/User Details/src/UserDetails.Page.al
```diff
@@ -0,0 +1,105 @@
...
+/// </summary>
+page 774 "User Details"
+{
+    AboutText = 'View the additional information about users in a list view, which allows for easy searching and filtering.';
+    AboutTitle = 'About the users detailed view';
+    ApplicationArea = Basic, Suite;
+    Caption = 'Users';
+    DeleteAllowed = false;
+    Editable = false;
+    InsertAllowed = false;
```
AndreasMoth, line 21: No need for InsertAllowed = false, ModifyAllowed = false and DeleteAllowed = false when you set Editable = false. These are automatically set to false in this case. Only needed if you don't set Editable = false and want special logic.

src/System Application/App/User Details/src/UserDetails.Page.al
```diff
@@ -0,0 +1,105 @@
...
+    var
+        EnvironmentInformation: Codeunit "Environment Information";
+        UserDetails: Codeunit "User Details";
+    begin
+        IsSaaS := EnvironmentInformation.IsSaaS();
+        UserDetails.GetUserDetails(Rec);
+    end;
+
+    protected var
+        IsSaaS: Boolean;
```
AndreasMoth, line 104: Not sure why IsSaaS is protected.. if we want a var they can change then it should reflect that like "ShowAuthentication" or something. Otherwise we are confusing things if partners set this to true/false in onPrem/SaaS to get some behaviour.. we think we are/are not on SaaS but we actually are/are not.

src/System Application/Test Library/User Details/src/UserDetailsTestLibrary.Codeunit.al
```diff
@@ -0,0 +1,113 @@
...
+codeunit 132001 "User Details Test Library"
+{
+    var
+        UserDetailsRec: Record "User Details";
+        UserDoesNotExistErr: Label 'The user with security ID %1 does not exist', Locked = true;
+
+    /// <summary>
+    /// Saves the user details inside this instance, so that they can be access later.
+    /// </summary>
+    procedure FetchUserDetails()
```
AndreasMoth: "GetUserDetails" to align with the interface function name. Also I would return the record to allow for more usages. I know it's public but if you both want to call functions in here and do something additional with the record..

src/System Application/Test Library/User Details/src/UserDetailsTestLibrary.Codeunit.al
```diff
@@ -0,0 +1,113 @@
...
+        UserDetails: Codeunit "User Details";
+    begin
+        UserDetails.GetUserDetails(UserDetailsRec);
+    end;
+
+    /// <summary>
+    /// Gets the value of "Has SUPER permission set" for the user.
+    /// </summary>
+    /// <param name="UserSID">The user security ID</param>
+    procedure GetHasSuperPermissionSet(UserSID: Guid): Boolean
```
AndreasMoth: I would just name all these "HasSuperPermissionSet", obviously we do a get but seems a bit odd to mention it for every function. If we want something in front then perhaps Check- or Test-HasSuperPermissionSet

src/System Application/Test Library/User Details/src/UserDetailsTestLibrary.Codeunit.al
```diff
@@ -0,0 +1,113 @@
...
+    begin
+        if not UserDetailsRec.Get(UserSID) then
+            Error(UserDoesNotExistErr, UserSID);
+
+        UserDetailsRec.CalcFields("Has SUPER permission set");
```
AndreasMoth, line 38: If you instead set auto-calcfield on this field before doing a GET, you automatically get this value evaluated without having to do a second database call here. I know this isn't super important for tests, just wanted to mention this speedup :)

src/System Application/App/User Details/src/UserDetails.Page.al
pri-kise: Editable = false can be overwritten by PageExtensions.

InsertAllowed and the other properties aren't changeable by PageExtensions.

Since this page is extensible I'd vote for ensuring that no extension should be able to make changes, that users can insert any value on this page.

src/System Application/App/User Details/src/UserDetails.Page.al
IhorHandziuk: It's because most page extensions would need to use this variable to control the visibility. Ideally, we would declare it as `const` or `final`, but we don't have such concepts in AL.

src/System Application/Test Library/User Details/src/UserDetailsTestLibrary.Codeunit.al
IhorHandziuk: I actually tried that originally, but it didn't work for some reason. As this is very monor - resolving.

AndreasMoth (approved):

JesperSchulz (approved):

PR #255, opened 2023-10-17
AppSource app preview support

qutreson:
**Content:**
Adding support for installing preview versions of AppSource extensions using a PreviewKey. The installation is triggered using:  `https://businesscentral.dynamics.com/[TenantID]/?noSignUpCheck=1&filter='ID' IS '[AppID]' AND 'PREVIEWKEY' IS '[PreviewKey]'&page=2503`

### DIFF
--- a/src/System Application/App/Extension Management/permissions/ExtensionManagementObjects.PermissionSet.al
+++ b/src/System Application/App/Extension Management/permissions/ExtensionManagementObjects.PermissionSet.al
@@ -31,5 +31,6 @@ permissionset 2504 "Extension Management - Objects"
                   Page "Extn Deployment Status Detail" = X,
                   Page "Marketplace Extn Deployment" = X,
                   Page "Upload And Deploy Extension" = X,
-                  Page "Extension Setup Launcher" = X;
+                  Page "Extension Setup Launcher" = X,
+                  Table "Extension Installation" = X;
 }


src/System Application/App/Extension Management/src/ExtensionInstallation.Table.al
```diff
@@ -0,0 +1,36 @@
...
+        /// The app identifier. Should uniquely identify the application and remains static across versions.
+        /// </summary>
+        field(1; "ID"; Guid)
+        {
+            Caption = 'ID';
+        }
+        /// <summary>
+        /// The preview key to be used when installing a preview version of an AppSource app.
+        /// </summary>
+        field(2; PREVIEWKEY; Text[2048])
```
EmilDamsbo, line 32: Did we get this spec from Partner Center? Is there any chance the preview key from Partner Center is longer?

src/System Application/App/Extension Management/src/ExtensionMarketplace.Codeunit.al
```diff
@@ -216,11 +216,17 @@ codeunit 2501 "Extension Marketplace"
         if IsNullGuid(ApplicationId) then
             Error(TelemetryExtensionNotFoundErr, ApplicationId);
 
+        if PreviewKey <> '' then begin
```
EmilDamsbo, line 219: This path should probably be covered in the extension management tests:
https://github.com/microsoft/BCApps/tree/eb3e653b1dcce398d456547d3c62f7675700d937/src/System%20Application/Test/Extension%20Management

src/System Application/App/Extension Management/src/ExtensionInstallation.Table.al
qutreson: Discussing off-line. 2048 should be enough to cover all preview keys defined. I will add it in our work item regarding documentation of the feature.

src/System Application/App/Extension Management/src/ExtensionMarketplace.Codeunit.al
qutreson: A good test would be to mock the "Extension Operation Impl" codeunit to verify which parameters are passed to the DeployExtension method. That would require to refactor the "Extension Marketplace" codeunit to use interfaces for its implementation codeunits so that we can then mock the calls done.

qutreson: As discussed, linked to work item 467161 on the backlog

PR #256, opened 2023-10-17
[Copilot] Marketing Text Adjustments

juliamakulec:
This PR brings a few fixes to the Marketing Text and AI module.

AI SDK:
- `CurrPage.Update()` is passed `false` in the AI SDK, fixing a bug with unwanted changes saved to the copilot capabilities

Marketing Text:
- `CanSuggest()` and `IsEnabled()` checks are modified to correctly manage visibility and actions for Entity/Marketing Text. This is also reflected in the `GenerateSuggestion`, to ensure that all required checks pass before a call to AI service is allowed.

### DIFF
--- a/src/System Application/App/Entity Text/src/EntityTextImpl.Codeunit.al
+++ b/src/System Application/App/Entity Text/src/EntityTextImpl.Codeunit.al
@@ -46,7 +46,7 @@ codeunit 2012 "Entity Text Impl."
         UserPrompt: Text;
         Suggestion: Text;
     begin
-        if not IsEnabled(false) then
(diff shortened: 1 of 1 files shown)


src/System Application/App/Entity Text/src/EntityTextImpl.Codeunit.al
```diff
@@ -45,6 +46,11 @@ codeunit 2012 "Entity Text Impl."
         UserPrompt: Text;
         Suggestion: Text;
     begin
+        if not IsEnabled(false) then
```
juliamakulec: Silent should be true

PR #263, opened 2023-10-19
Label pull requests by area and from forks

aholstrup1:
Label PRs to look like this (for example)
image

### DIFF
--- a/.github/workflows/PullRequestLabeler.yaml
+++ b/.github/workflows/PullRequestLabeler.yaml
@@ -27,4 +27,4 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          gh api /repos/${{ github.repository }}/issues/${{ github.event.pull_request.number }}/labels  -f "labels[]=Community Contribution" -H "Accept: application/vnd.github.v3+json" -H "X-GitHub-Api-Version: 2022-11-28"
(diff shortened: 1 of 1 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,10 @@
+"Area: Build":
```
mazhelez: .yaml instead?

.github/workflows/PullRequestLabeler.yaml
```diff
@@ -0,0 +1,30 @@
...
+        with:
+            repo-token: '${{ secrets.GITHUB_TOKEN }}'
+            sync-labels: true
+      
+      - name: Label community contribution
+        if: github.event.pull_request.head.repo.full_name != github.repository
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh api /repos/${{ github.repository }}/issues/${{ github.event.pull_request.number }}/labels  -f "labels[]=Community Contribution" -H "Accept: application/vnd.github.v3+json" -H "X-GitHub-Api-Version: 2022-11-28"
```
mazhelez: How about "External Contribution"?

.github/labeler.yml
```diff
@@ -0,0 +1,10 @@
...
+- build/**
+- .github/**
+- .azuredevops/**
+
+"Area: System Application":
```
mazhelez: What about something like:

> "Build: AL-Go"
"Build: automation"

and 

> "AL: System Application"
"AL: Developer Tools"

or

> "App: System Application"
> "App: Developer Tools"

aholstrup1: I like that :)

.github/workflows/PullRequestLabeler.yaml
JesperSchulz: Let's discuss.

PR #269, opened 2023-10-22
Improve the performance of retention policies filtering

IhorHandziuk:
**Problem:**
Retention policy runs often take a very long time, especially if the tables they operate on have a lot of records.

**Solution:**
- Maintain a hash set of system IDs of expired records, which allows for efficient `Count()` calls when checking if the maximum number of records to be deleted has been reached.
- Prepare the cache of table filter views outside the main loop of marking / unmarking records, as reading BLOBs is an expensive operation.

**Extra changes:**
1. Fix the behavior of the `GetOldestRecordDate` function in the case when there are no records in the table. Currently in this case an error will be surfaced in the notification: "There are no records within the filter..."
2. Remove the filtering logic from the retention policy deletion implementation, as:
- it should not be the responsibility of this object to perform filtering;
- the deletion is normally not the bottleneck in retention policy runs - filtering is, so there is no reason to skip the deletion of already computed expired records.
- we currently have a bug in the implementation that essentially ensures that the limit is not applied for tables that have indirect delete permissions (which are likely most of the tables that are available in retention policies).

**Note:**
The potential performance improvements have been identified using the in-client performance profiler, as well as PerfView profiler.

### DIFF
--- a/src/System Application/App/Retention Policy/src/Apply Retention Policy/RetenPolFilteringImpl.Codeunit.al
+++ b/src/System Application/App/Retention Policy/src/Apply Retention Policy/RetenPolFilteringImpl.Codeunit.al
@@ -97,7 +97,10 @@ codeunit 3915 "Reten. Pol. Filtering Impl." implements "Reten. Pol. Filtering"
         if YoungestExpirationDate >= Yesterday() then
             YoungestExpirationDate := Yesterday();
         OldestRecordDate := GetOldestRecordDate(RetentionPolicySetup);
-        NumberOfDays := YoungestExpirationDate - OldestRecordDate;
+        if OldestRecordDate = 0D then
+            NumberOfDays := 0
+        else
+            NumberOfDays := YoungestExpirationDate - OldestRecordDate;
 
         if NumberOfDays <= 0 then begin
             RetentionPolicyLog.LogInfo(LogCategory(), StrSubstNo(OldestRecordYoungerThanExpirationLbl, RetentionPolicySetup."Table Id", RetentionPolicySetup."Table Caption"));


src/System Application/App/Retention Policy/src/Apply Retention Policy/RetenPolFilteringImpl.Codeunit.al
```diff
@@ -77,20 +77,30 @@ codeunit 3915 "Reten. Pol. Filtering Impl." implements "Reten. Pol. Filtering"
...
+        SelectedSystemIds: Dictionary of [Guid, Boolean]; // using a dictionary here, as there is no native hash set type
+        RetentionPolicySetupLineTableFilters: Dictionary of [Guid, Text];
     begin
         RecordRef.Open(RetentionPolicySetup."Table ID");
         RecordReference.Initialize(RecordRef, RecordReferenceIndirectPermission);
 
         RetentionPolicySetupLine.SetRange("Table ID", RetentionPolicySetup."Table Id");
         RetentionPolicySetupLine.SetRange(Enabled, true);
-        if RetentionPolicySetupLine.IsEmpty then
+        if not RetentionPolicySetupLine.FindSet(false) then
```
stkillen, line 88: Is FindSet more efficient than the IsEmpty check, or what is the reason for this change?

src/System Application/App/Retention Policy/src/Apply Retention Policy/RetenPolDeleteImpl.Codeunit.al
```diff
@@ -33,51 +31,14 @@ codeunit 3916 "Reten. Pol. Delete. Impl." implements "Reten. Pol. Deleting"
...
 
         if not RecordReferenceIndirectPermission.ReadPermission(RecordRef) then
-            RetentionPolicyLog.LogWarning(LogCategory(), StrSubstNo(MissingReadPermissionLbl, RecordRef.Number, RecordRef.Caption))
-        else
-            if not RecordReferenceIndirectPermission.IsEmpty(RecordRef) then
-                if (RecordReferenceIndirectPermission.Count(RecordRef) > (RetenPolDeletingParam."Max. Number of Rec. To Delete" + NumberOfRecordsToDeleteBuffer())) then begin
-                    RetentionPolicyLog.LogWarning(LogCategory(), StrSubstNo(TooManyRecordsToDeleteLbl, RetenPolDeletingParam."Total Max. Nr. of Rec. to Del."));
-                    LimitRecordsToBeDeleted(RecordRef, RecordReferenceIndirectPermission, RetenPolDeletingParam."Skip Event Rec. Limit Exceeded", RetenPolDeletingParam."Max. Number of Rec. To Delete", RetenPolDeletingParam."Total Max. Nr. of Rec. to Del.");
-                end;
+            RetentionPolicyLog.LogWarning(LogCategory(), StrSubstNo(MissingReadPermissionLbl, RecordRef.Number, RecordRef.Caption));
```
stkillen, line 34: So this warning is logged every time records are deleted using the event - doesn't seem right, but was apparently also the case before.

src/System Application/App/Retention Policy/src/Apply Retention Policy/RetenPolFilteringImpl.Codeunit.al
IhorHandziuk: I've added a `repeat ... until` loop underneeth, so a `FindSet` was needed anyway. I decided to remove the `IsEmpty` call, as it was redundant.

src/System Application/App/Retention Policy/src/Apply Retention Policy/RetenPolDeleteImpl.Codeunit.al
IhorHandziuk: Yes, not sure why this is logged - but just leaving this as is for now.

PR #271, opened 2023-10-23
Let automation update existing pull requests if possible

aholstrup1:
Over the weekend we had three automated pull requests created trying to update the same thing. 
image

This PR changes that behavior such that subsequent runs of "Update BC Artifact Version" will try to update the existing PR if possible. That way we don't end up with multiple PRs. 

I'm also changing the schedule, so it doesn't run over the weekend, as the PR most likely won't get merged until Monday anyway.


build/scripts/UpdateBCArtifactVersion.ps1
```diff
@@ -50,15 +50,15 @@ function UpdateBCArtifactVersion() {
     return $false
 }
 
+$pullRequestTitle = "[$TargetBranch] Update BC Artifact version"
+$BranchName = New-TopicBranchIfNeeded -Category "UpdateBCArtifactVersion/$TargetBranch" -PullRequestTitle $pullRequestTitle
```
mazhelez, line 54: If we allow only one branch per category/target branch, then the PR title isn't needed for the filtering.

build/scripts/UpdateBCArtifactVersion.ps1
```diff
@@ -50,15 +50,15 @@ function UpdateBCArtifactVersion() {
...
 $updatesAvailable = UpdateBCArtifactVersion
 
 if ($updatesAvailable) {
     # Create branch and push changes
     Set-GitConfig -Actor $Actor
-    $BranchName = New-TopicBranch -Category "UpdateBCArtifactVersion/$TargetBranch"
-    $title = "[$TargetBranch] Update BC Artifact version"
-    Push-GitBranch -BranchName $BranchName -Files @(".github/AL-Go-Settings.json") -CommitMessage $title
-
+    Push-GitBranch -BranchName $BranchName -Files @(".github/AL-Go-Settings.json") -CommitMessage $pullRequestTitle
```
mazhelez, line 61: Does it make sense to update the branch if it's not new? E.g. pull _main_ into it.

aholstrup1: Mmh I don't think we need to do we? The PR Build runs on the "merge" branch so we are testing the changes with the latest changes from the parent branch.

build/scripts/UpdateBCArtifactVersion.ps1
aholstrup1: Actually thinking about it.. What if we don't clean up the branch after PRs are merged?

build/scripts/UpdateBCArtifactVersion.ps1
mazhelez: Just seems a bit weird to work on an old branch that's probably behind from the target branch.

build/scripts/UpdateBCArtifactVersion.ps1
mazhelez: You mean intentionally keeping it by design or if the branch was not deleted after merging the PR?

PR #281, opened 2023-10-25
[Permission Sets] Enable inclusion and exclusion of unassignable permission sets

stkillen:
### Summary
It's only possible to include _assignable_ permission sets when creating a composed permission set. However, by cloning a permission set using the copy action it is possible to have unassignable permission sets (it is only the UI that doesn't support it).

This is due to the current implementation using the Aggregate Permission Set table for all lookups and this table only includes assignable permission sets.

This PR enables usage of unassignable permission sets by replacing usages of Aggregate Permission Set with Tenant Permission Set and Metadata Permission Set.

- Replace all usages of Aggregate Permission Set in permission set inclusion/exclusion
- Ensure that the permission set name is looked up from the Tenant / Metadata permission set tables

_New lookup page_

image

### Work Item(s)
Fixes AB#473508


src/System Application/App/Permission Sets/src/PermissionSetLookupList.Page.al
```diff
@@ -0,0 +1,87 @@
...
+namespace System.Security.AccessControl;
+
+/// <summary>
+/// Look up page for selecting a permission set from Tenant Permission Set or Metadata Permission Set.
+/// </summary>
+page 9878 "Permission Set Lookup List"
+{
+    Caption = 'Permission Set Lookup';
+    Editable = false;
+    DeleteAllowed = false;
```
AndreasMoth: No need to disable delete, insert and modify if the page isn't editable :)

src/System Application/App/Permission Sets/src/PermissionSetLookupList.Page.al
```diff
@@ -0,0 +1,87 @@
...
+
+    var
+        SelectedRecord: Record "PermissionSet Buffer";
+
+    internal procedure GetSelectedRecord(var CurrSelectedRecord: Record "PermissionSet Buffer")
+    begin
+        CurrSelectedRecord := SelectedRecord;
+    end;
+
+    internal procedure GetSelectedRecords(var CurrSelectedRecords: Record "PermissionSet Buffer")
```
AndreasMoth, line 72: We should either clear all content in this record first or make sure the caller does so, right now the caller does not and we don't here either. I would personally say it's the responsibility of this page to only set the currently selected records, I.E. clear the record.

src/System Application/App/Permission Sets/src/PermissionSetRelationImpl.Codeunit.al
```diff
@@ -364,6 +398,22 @@ codeunit 9856 "Permission Set Relation Impl."
         end;
     end;
 
+    local procedure GetPermissionSetName(AppId: Guid; RoleId: Code[30]; Scope: Option System,Tenant): Text[30];
```
AndreasMoth, line 401: Minor: Could we make the scope option an enum? Not just for permission set, but in general. After all it's used everywhere: System, Tenant.

src/System Application/App/Permission Sets/src/PermissionSetRelationImpl.Codeunit.al
```diff
@@ -364,6 +398,22 @@ codeunit 9856 "Permission Set Relation Impl."
...
+    begin
+        if Scope = Scope::Tenant then
+            if TenantPermissionSet.Get(AppId, RoleId) then
+                exit(TenantPermissionSet.Name);
+
+        if Scope = Scope::System then
+            if MetaDataPermissionSet.Get(AppId, RoleId) then
+                exit(MetaDataPermissionSet.Name);
+
+        exit('');
```
AndreasMoth, line 414: Should we return something like "Unavailable"? or something that indicates this permission set does not exist to the user...

AndreasMoth: Maybe this should be handled on the page actually.

src/System Application/App/Permission Sets/src/PermissionSetRelationImpl.Codeunit.al
```diff
@@ -398,7 +448,22 @@ codeunit 9856 "Permission Set Relation Impl."
         exit(10);
     end;
 
-    procedure LookupPermissionSet(AllowMultiselect: Boolean; var AggregatePermissionSet: Record "Aggregate Permission Set"): Boolean
+    procedure LookupAllPermissionSets(AllowMultiselect: Boolean; var PermissionSetBuffer: Record "PermissionSet Buffer"): Boolean
+    var
+        PermissionSetLookupList: Page "Permission Set Lookup List";
+    begin
+        PermissionSetLookupList.LookupMode(true);
+        if PermissionSetLookupList.RunModal() = Action::LookupOK then begin
```
AndreasMoth, line 456: It seems a bit odd to me that the page itself doesn't know if multi-select is allowed. That sounds like something it should know and tell the user. Otherwise the user may try to multi-select when it's not allowed and vice versa. Would be nice info to the user :)

AndreasMoth (approved):
Last comment, could you add a screenshot of how the page looks as well?

src/System Application/App/Permission Sets/src/PermissionSetLookupList.Page.al
stkillen: Removed

src/System Application/App/Permission Sets/src/PermissionSetRelationImpl.Codeunit.al
stkillen: This could be done in a separate PR for the entire module, if we want to prioritize that. This is how it's done currently in the entire module and the permission set tables themselves are also using an option.

src/System Application/App/Permission Sets/src/PermissionSetRelationImpl.Codeunit.al
stkillen: By definition it should always be in one of the two tables. However, in the unknown scenario where we can't get it in either, I believe an empty string is a decent fallback in that scenario.

src/System Application/App/Permission Sets/src/PermissionSetRelationImpl.Codeunit.al
stkillen: I'm not sure how to communicate that to the user. Multi select is allowed when using the "Select permission sets" action and it's not allowed when adding a single permission set through the drilldown - a similar pattern is applied on other pages in the application

darjoo (approved):

JesperSchulz (approved):

PR #290, opened 2023-10-30
Update Copilot with latest changes

darjoo:
Changes:

- Add metaprompt
- Update Copilot pages with new text, telemetry, formatting
- Allow disabling of Copilot on tenant, including 3rd party extensions
- Update Entity Text caption to `Marketing text suggestions`

### DIFF
--- a/src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIChatMessagesImpl.Codeunit.al
+++ b/src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIChatMessagesImpl.Codeunit.al
@@ -144,7 +144,7 @@ codeunit 7764 "AOAI Chat Messages Impl"
             exit;
 
         Initialize();
-        CheckandAddMetaprompt();
+        CheckandAddPrimarySystemMessage();
 
         if SystemMessage.Unwrap() <> '' then begin
             MessageJsonObject.Add('role', Format(Enum::"AOAI Chat Roles"::System));
@@ -190,7 +190,7 @@ codeunit 7764 "AOAI Chat Messages Impl"
     end;
 
     [NonDebuggable]
-    local procedure CheckandAddMetaprompt()
+    local procedure CheckandAddPrimarySystemMessage()
     var
         AzureOpenAIImpl: Codeunit "Azure OpenAI Impl";
         Telemetry: Codeunit Telemetry;
--- a/src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIChatMessagesImpl.Codeunit.al
+++ b/src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIChatMessagesImpl.Codeunit.al
@@ -25,8 +25,8 @@ codeunit 7764 "AOAI Chat Messages Impl"
         IsSystemMessageSet: Boolean;
         MessageIdDoesNotExistErr: Label 'Message id does not exist.';
         HistoryLengthErr: Label 'History length must be greater than 0.';
-        TelemetryMetapromptSetbutEmptyTxt: Label 'Metaprompt was set but is empty.', Locked = true;
-        TelemetryMetapromptEmptyTxt: Label 'Metaprompt was not set.', Locked = true;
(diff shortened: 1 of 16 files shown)


src/System Application/App/AI/src/Copilot/CopilotTelemetry.Codeunit.al
```diff
@@ -18,30 +18,45 @@ codeunit 7775 "Copilot Telemetry"
...
+    var
+        CopilotCapabilitiesImpl: Codeunit "Copilot Capability Impl";
+        FeatureTelemetry: Codeunit "Feature Telemetry";
+        Language: Codeunit Language;
+        SavedGlobalLanguageId: Integer;
+    begin
+        CopilotCapabilitiesImpl.AddTelemetryDimensions(CopilotCapability, AppId, CustomDimensions);
+
+        SavedGlobalLanguageId := GlobalLanguage();
+        GlobalLanguage(Language.GetDefaultApplicationLanguageId());
```
IhorHandziuk: These should either be placed above `CopilotCapabilitiesImpl.AddTelemetryDimensions`, or removed completely.
`Feature Telemetry` has the same calls performed for the common custom dimensions it adds automatically.

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
```diff
@@ -166,6 +167,15 @@ codeunit 7774 "Copilot Capability Impl"
         Notification.Send();
     end;
 
+    procedure ShowCapabilitiesNotAvailableOnPremNotification()
+    var
+        Notification: Notification;
+    begin
+        Notification.Id(CreateGuid());
```
IhorHandziuk: We have the same guid here (e. g. saved as a label) - otherwise users may be spammed with multiple instances of the same notification.

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIChatMessagesImpl.Codeunit.al
```diff
@@ -182,4 +188,19 @@ codeunit 7764 "AOAI Chat Messages Impl"
...
+    var
+        AzureOpenAIImpl: Codeunit "Azure OpenAI Impl";
+        Telemetry: Codeunit Telemetry;
+    begin
+        if SystemMessage.Unwrap().Trim() = '' then begin
+            if IsSystemMessageSet then
+                Telemetry.LogMessage('0000LO9', TelemetryMetapromptSetbutEmptyTxt, Verbosity::Normal, DataClassification::SystemMetadata)
+            else
+                Telemetry.LogMessage('0000LOA', TelemetryMetapromptEmptyTxt, Verbosity::Normal, DataClassification::SystemMetadata);
+            SetPrimarySystemMessage(AzureOpenAIImpl.GetMetaprompt());
```
IhorHandziuk, line 203: How is metaprompt different from the system message (here they seem to be used interchangeably)? If they are the same, we should only use one term. If they are indeed different, we probably want to add code documentation to explain the difference.

IhorHandziuk (changes requested):

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIChatMessagesImpl.Codeunit.al
darjoo: Changing this function's name, but I won't change telemetry label or `GetMetaprompt`. 

`GetMetaprompt` in `AzureOpenAIImpl` is also used in Text Completions and metaprompts are metaprompts there. Only reason it is System here, is because of how system messages work and we have the `SetPrimarySystemMessage` function

IhorHandziuk (dismissed):

JesperSchulz (dismissed):

JesperSchulz (dismissed):

IhorHandziuk (dismissed):

stkillen (approved):

PredragMaricic (approved):

JesperSchulz (approved):

PR #306, opened 2023-11-06
Using PS classes for GitHub-related objects (PRs and issues) in work item validation

mazhelez:


build/scripts/GitHub/GitHubIssue.class.psm1
```diff
@@ -0,0 +1,48 @@
...
+    $Repository
+    $Issue
+
+    #Constructor
+    hidden GitHubIssue([int] $IssueId, [string] $Repository) {
+        $this.IssueId = $IssueId
+        $this.Repository = $Repository
+
+        $gitHubIssue = gh api "/repos/$Repository/issues/$IssueId" -H ([GitHubAPI]::AcceptJsonHeader) -H ([GitHubAPI]::GitHubAPIHeader)| ConvertFrom-Json
+        if (-not $gitHubIssue) {
```
mazhelez: This does not work as `$gitHubIssue` is still an object when the issue is not found.

.github/workflows/WorkitemValidation.yaml
```diff
@@ -1,6 +1,7 @@
 name: Work Item Validation
 on:
   pull_request_target:
+    types: [opened, synchronize, reopened, edited]
```
mazhelez, line 4: Limit target branches

build/scripts/PullRequestValidation/ValidateIssuesForPullRequest.ps1
```diff
@@ -0,0 +1,64 @@
...
+
+    $issue = [GitHubIssue]::Get($issueId, $this.Repository)
+
+    if ($issue) {
+
+        Write-Host "Issue $issueId found: $(ConvertTo-Json $issue))"
+
+        $Comment = "Issue $($issue.html_url) is not approved. Please make sure the issue is approved before continuing with the pull request"
+        if (-not $issue.IsApproved()) {
+            $this.PullRequest.AddComment($Comment)
```
mazhelez: There is no $this` here.

build/scripts/PullRequestValidation/ValidateIssuesForPullRequest.ps1
```diff
@@ -0,0 +1,64 @@
...
+
+            Write-Warning "::Warning:: $Comment"
```
mazhelez: Workflow should fail here.

.github/workflows/WorkitemValidation.yaml
```diff
@@ -1,6 +1,8 @@
...
+    branches: [main]
```
aholstrup1, line 5:
```suggestion
    branches: [main, releases/*]
```

build/scripts/PullRequestValidation/ValidateIssuesForPullRequest.ps1
```diff
@@ -0,0 +1,47 @@
...
+    )
+
+# Set error action
+$ErrorActionPreference = "Stop"
+
+Write-Host "Validating PR $PullRequestNumber"
+
+$pullRequest = [GitHubPullRequest]::Get($PullRequestNumber, $Repository)
+$prDescription = $pullRequest.GetBody()
+$issuesStartingPoint = "Fixes #"
```
aholstrup1: Fixes is only one of the keywords you can use (https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword). Why not set issuesStartingPoint to be "# Work Items"?  

E.g. 
$workItemsStartingIndex = $description.IndexOf("# Work Item(s)")

build/scripts/PullRequestValidation/ValidateIssuesForPullRequest.ps1
```diff
@@ -0,0 +1,47 @@
...
+
+$issueStartingIndex = -1
+if($prDescription) {
```
aholstrup1: Consider doing the opposite check earlier on. `if (-not $prDescription) { throw "" }`. Then you don't need to set $issueStartingIndex = -1. IndexOf will return -1 anyway if it doesn't find anything.

build/scripts/PullRequestValidation/ValidateIssuesForPullRequest.ps1
```diff
@@ -0,0 +1,47 @@
...
+
+foreach ($issueId in $issueIds) {
+    Write-Host "Validating issue $issueId"
+    $issue = [GitHubIssue]::Get($issueId, $Repository)
+
+    $Comment = "Issue $($issue.html_url) is not approved. Please make sure the issue is approved before continuing with the pull request."
+    if (-not $issue.IsApproved()) {
+        $pullRequest.AddComment($Comment)
+
+        throw "$Comment"
```
aholstrup1: Bit of an edge case.. But would it make sense to let the foreach loop keep running and throw at the end? That way we can create a comment for all non-approved issues in case there are multiple

build/scripts/GitHub/GitHubPullRequest.class.psm1
```diff
@@ -0,0 +1,68 @@
...
+
+    [string] GetBody() {
+        return $this.PullRequest.body
+    }
+
+    <#
+        Removes a comment from the pull request if it exists.
+    #>
+    RemoveComment($Message) {
+        $existingComments = gh api "/repos/$($this.Repository)/issues/$($this.PRNumber)/comments" -H $this.AcceptJsonHeader -H $this.GitHubAPIHeader  | ConvertFrom-Json
```
aholstrup1, line 66: Since we're calling the issues api, wouldn't it make sense to place the "comment" functions in the GitHubIssue class? 
Maybe the RemoveComment method in this class can just create a GitHubIssue object and call RemoveComment on that?

build/scripts/PullRequestValidation/ValidateIssuesForPullRequest.ps1
mazhelez: If it's a whole section, it might be hard to determine where it ends and what exactly to scan for.

I say we scan for whatever we put in the PR template. What do you think?

build/scripts/PullRequestValidation/ValidateIssuesForPullRequest.ps1
mazhelez: Potato, potato (potayto, potahto) 😆

build/scripts/PullRequestValidation/ValidateIssuesForPullRequest.ps1
mazhelez: Good one, will do.

.github/workflows/WorkitemValidation.yaml
mazhelez: Let's get this as a separate effort as in some places there's "release/" and other - "release**s**/" (see [here](https://github.com/microsoft/BCApps/blob/2530d5583e175e79c13e0d114b0bb388377cf68b/.github/AL-Go-Settings.json#L45) for instance).

build/scripts/PullRequestValidation/ValidateIssuesForPullRequest.ps1
mazhelez: Done

build/scripts/GitHub/GitHubPullRequest.class.psm1
mazhelez: It's the issue endpoint because issues and PRs on GitHub share a lot of functionality, e.g. the comment section.

In this case I have a strong opinion that this should stay in the PR class.

build/scripts/PullRequestValidation/ValidateIssuesForPullRequest.ps1
```diff
@@ -0,0 +1,58 @@
...
+$issueRegex = "Fixes #(\d+)"
+
+if(-not $prDescription) {
+    throw "Could not find pull request description. Please make sure the pull request description contains a line that contains 'Fixes #' followed by the issue number being fixed."
+}
+
+# Get all issue matches
+$issueMatches = Select-String $issueRegex -InputObject $prDescription -AllMatches
+
+if(-not $matches) {
```
aholstrup1:
```suggestion
if(-not $issueMatches.Matches) {
```

$matches doesn't exist for me if I run it locally. Does it work for you?

build/scripts/PullRequestValidation/ValidateIssuesForPullRequest.ps1
```diff
@@ -0,0 +1,58 @@
...
+    throw "Could not find issues section in the pull request description. Please make sure the pull request description contains a line that contains 'Fixes #' followed by the issue number being fixed."
```
aholstrup1: Could we create a comment on the pull request with this message + throw an error?

build/scripts/PullRequestValidation/ValidateIssuesForPullRequest.ps1
mazhelez: Good catch, I forgot to rename that one.

build/scripts/PullRequestValidation/ValidateIssuesForPullRequest.ps1
mazhelez: Added.

PR #317, opened 2023-11-09
Stop inserting newer versions of Guided Experience Item and provide the functionality to cleanup

haoranpb:
Instead of inserting a new `Guided Experience Item` (e.g. after editing Title), we modify the latest version instead (Note: only when `Version > 0`, [rename](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/methods-auto/record/record-rename-method) will not work when value is 0)

Also provide enhanced functionality to cleanup duplicated records, Job Queue implementation will be implemented once we uptake this change in NAV repository

Fixes [AB#485933](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/485933)

- [x] Automation Tests
- [x] Manual testing is needed with team


src/System Application/App/Guided Experience/src/Guided Experience/GuidedExperienceItemCleanup.Codeunit.al
```diff
@@ -0,0 +1,134 @@
...
+
+namespace System.Environment.Configuration;
+
+using System.Apps;
+
+codeunit 1994 "Guided Experience Item Cleanup"
+{
+    InherentEntitlements = X;
+    InherentPermissions = X;
+    Permissions =
```
Groenbech96, line 15: internal

haoranpb: Good catch

haoranpb: Actually, just realized that I made it public on purpose... Note: all the procedures are either `local` or `internal`

Because we will need to access this codeunit to be able to schedule the Job Queue in BaseApp `Codeunit::"Guided Expereince Item Cleanup"`

Even thought Job Queue is planned to move to System App, we still need to do this for now

VolodySyn (approved):

src/System Application/Test/Guided Experience/src/DuplicatedGuidedExpItem.Codeunit.al
```diff
@@ -0,0 +1,128 @@
...
+        GuidedExperienceItem: Record "Guided Experience Item";
+        GuidedExperience: Codeunit "Guided Experience";
+        Limit, Counter : Integer;
+    begin
+        // [SCENARIO] Insert multiple Guided Experience Items with the same Code
+        PermissionsMock.Set('Guided Exp Edit');
+
+        GuidedExperienceItem.DeleteAll();
+
+        Assert.AreEqual(0, GuidedExperienceItem.Count(), 'No records should exists in the Guided Experience Item table before the test starts.');
```
Groenbech96, line 35: Labels

Groenbech96 (changes requested):

src/System Application/Test/Guided Experience/src/DuplicatedGuidedExpItem.Codeunit.al
haoranpb: Not sure what this the convention here tbh, those messages don't need to be translated. I think it is ok for now

Groenbech96 (approved):
LGTM

grobyns (approved):

PR #336, opened 2023-11-14
New Compare Methods for Time and DateTime Comparison

### LINKED ISSUE #335: [BC Idea]: New Compare Methods for Time and DateTime Comparison
### BC Idea Link

https://experience.dynamics.com/ideas/idea/?ideaid=59c5f0fb-0683-ee11-a81c-000d3adc4046

### Description

**🕮 Describe the issue**
The comparison of time values does not work if you compare the time value before writing to the database with the time value that was written to the database.
The cause of the problem is that in BC DateTime and Time values are rounded by SQL Server ([See MS Docs](https://learn.microsoft.com/en-us/sql/t-sql/data-types/datetime-transact-sql?view=sql-server-ver16)). This means that a DateTime / Time value can change just because it was written to the database, the comparison with the original value then runs on error.

For DateTimes there is a method "CompareDateTime" in the codeunit TypeHelper to compare DateTimes with a small threshold value to compensate for the rounding of the DateTime values. However, the threshold seems to be unnecessarily large.
Also, there is no method to compare Times.

**🔧 To Reproduce**

Run the follwowing code:
```
codeunit 50100 "Insert Time"
{
    trigger OnRun()
    var
        Contact: Record Contact;
        NewLastModifiedTime: Time;
        TimeText: Text;
    begin
        TimeText := '12:40:22.468Z';
        Contact.Get('KT100006');
        Evaluate(NewLastModifiedTime, TimeText, 9);
        Contact."Last Time Modified" := NewLastModifiedTime;
        Contact.Modify();

       Message('Before Write to Database %1 = %2', Format(Contact."Last Time Modified", 0, 9), Format(NewLastModifiedTime, 0, 9));

        Commit();
        Contact.Get('KT100006');
        Message('After Write to Database %1 <> %2', Format(Contact."Last Time Modified", 0, 9), Format(NewLastModifiedTime, 0, 9));
    end;
}

```
The code will output the following Messages:

Before Write to Database 12:40:22.468Z = 12:40:22.468Z

After Write to Database 12:40:22.467Z <> 12:40:22.468Z

**📣 Expected behavior**
There should be methods to compare DateTime / Time values that take care of the sql rounding.

**Additional context**
Previous Disscussion to this topic: https://github.com/microsoft/BusinessCentralApps/issues/483
Yammer Thread: https://www.yammer.com/dynamicsnavdev/#/threads/show?threadId=2506817338753024

JohnnyUndercover:
# Summary 
Add Methods to for DateTimes and Times:
- Compare
- IsGreater
- IsLess
- IsEqual
- RoundToSQLAccuracy

# Please give feedback for the following points

- [ ] Do we really need and want a procedure that only tells us whether two (date)times are very close? Would we use the procedure also if the SQL rounding issue never existed? (See https://github.com/microsoft/BusinessCentralApps/issues/483#issuecomment-1805345350)
- [ ] CompareTimes is now missing the case, when both times are guaranteed unrounded (See https://github.com/microsoft/BusinessCentralApps/issues/483#issuecomment-1804541820)
- [ ] I don't like the Codeunit Naming "Date and Time Helper", any better ideas?

# Work Item(s) 
Fixes #335

### DIFF
--- a/src/System Application/App/Date and Time/src/DateAndTimeHelper.Codeunit.al
+++ b/src/System Application/App/Date and Time/src/DateAndTimeHelper.Codeunit.al
@@ -9,6 +9,51 @@ codeunit 8722 "Date and Time Helper"
     InherentPermissions = X;
 
     /// <summary>
+    /// Compares two DateTime values and returns true if the FirstDateTime is greater than the SecondDateTime.
+    /// The DateTimes are rounded to SQL Server accuracy before comparison,
+    /// to account for the fact that SQL Server rounds DateTime values to the nearest 0, 3, or 7 milliseconds.
+    /// </summary>
+    /// <param name="FirstDateTime">The first DateTime value to compare.</param>
+    /// <param name="SecondDateTime">The second DateTime value to compare</param>
+    /// <returns>true if the FirstDateTime is greater than the SecondDateTime; otherwise, false.</returns>
+    procedure IsGreater(FirstDateTime: DateTime; SecondDateTime: DateTime): Boolean
+    var
+        DateandTimeHelperImpl: Codeunit "Date and Time Helper Impl.";
+    begin
+        exit(DateandTimeHelperImpl.Compare(FirstDateTime, SecondDateTime) = 1);
+    end;
+
+    /// <summary>
+    /// Compares two DateTime values and returns true if the FirstDateTime is less than the SecondDateTime.
+    /// The DateTimes are rounded to SQL Server accuracy before comparison,
+    /// to account for the fact that SQL Server rounds DateTime values to the nearest 0, 3, or 7 milliseconds.
+    /// </summary>
+    /// <param name="FirstDateTime">The first DateTime value to compare.</param>
+    /// <param name="SecondDateTime">The second DateTime value to compare</param>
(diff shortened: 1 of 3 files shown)


NKarolak (commented):
Just ensuring that the earlier discussion points do not get lost:

From https://github.com/microsoft/BusinessCentralApps/issues/483#issuecomment-1805345350:

> Do we really need and want a procedure that only tells us whether two (date)times are very close? Would we use the procedure also if the SQL rounding issue never existed?
> 
> Where do you actually need to use it? I can only think of test apps.
> And when I search [where Microsoft uses TypeHelper.CompareDateTime](https://github.com/search?q=repo%3AStefanMaron%2FMSDyn365BC.Code.History+TypeHelper.CompareDateTime&type=code), there are only a few more results.
> Does the CRM Integration or the Master Data Mgt. need the threshold, or would it still work correctly with a more precise CompareDateTime? I don't know ...

From https://github.com/microsoft/BusinessCentralApps/issues/483#issuecomment-1804541820:

> CompareTimes is now missing the case, when both times are guaranteed unrounded: Then no rounding should be applied.
> E.g., comparing 23:59:59.992 with 23:59:59.994 (both would be rounded to 3) should return -1, but with your current code, it would return 0.

PR #341, opened 2023-11-15
Add Azure File Services API module

### LINKED ISSUE #340: [BC Idea]: Azure File Share API Module
### BC Idea Link

None

### Description

It would be helpful to add a module to handle Azure File Share service, that's available on the Azure Cloud, in the same way the Azure Blob Storage service is handled currently in the system.

The module was already written in the old repository (https://github.com/microsoft/ALAppExtensions/pull/23710) and this issue is only to port the pull request to the new repository.

Krajzys:
# Summary 

This is a port of the pull request from the old repository, most of the convsersation and reviews are also there (https://github.com/microsoft/ALAppExtensions/pull/23710).
The goal of the PR is to add a module that facilitates working with Azure File Share service (which is a part of the Azure Cloud).
The module is heavily based on the Azure Blob Storage Service API module, and most of the architecture is similair.

There are also tests added for the module, but they should be ran manually, as they require passing credentials.

Fixes #340 

Work item: [AB#477758](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/477758)

### DIFF
--- a/src/System Application/App/Azure File Services API/src/Helper/AFSFormatHelper.Codeunit.al
+++ b/src/System Application/App/Azure File Services API/src/Helper/AFSFormatHelper.Codeunit.al
@@ -3,6 +3,10 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 // ------------------------------------------------------------------------------------------------
 
+namespace System.Azure.Storage.Files;
+
+using System.Utilities;
+
 codeunit 8955 "AFS Format Helper"
 {
     Access = Internal;
@@ -68,7 +72,7 @@ codeunit 8955 "AFS Format Helper"
     local procedure FormatDateTime(MyDateTime: DateTime; FormatSpecifier: Text): Text
     var
         DateTimeAsXmlString: Text;
-        DateTimeDotNet: DotNet DateTime;
+        DateTimeDotNet: DotNet System.DateTime;
     begin
         DateTimeAsXmlString := Format(MyDateTime, 0, 9); // Format as XML, e.g.: 2020-11-11T08:50:07.553Z
         exit(DateTimeDotNet.Parse(DateTimeAsXmlString).ToUniversalTime().ToString(FormatSpecifier));
--- a/src/System Application/App/Azure File Services API/src/Helper/AFSFormatHelper.Codeunit.al
+++ b/src/System Application/App/Azure File Services API/src/Helper/AFSFormatHelper.Codeunit.al
@@ -1,0 +1,76 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+codeunit 8955 "AFS Format Helper"
+{
+    Access = Internal;
+    InherentEntitlements = X;
+    InherentPermissions = X;
+
+    [NonDebuggable]
+    procedure AppendToUri(var UriText: Text; ParameterIdentifier: Text; ParameterValue: Text)
+    var
+        UriBuilder: Codeunit "Uri Builder";
+        Uri: Codeunit Uri;
+    begin
+        UriBuilder.Init(UriText);
+        if ParameterIdentifier <> '' then
+            UriBuilder.AddQueryParameter(ParameterIdentifier, ParameterValue)
+        else
+            UriBuilder.AddQueryFlag(ParameterValue);
+        UriBuilder.GetUri(Uri);
+        UriText := Uri.GetAbsoluteUri();
+    end;
+
+    [NonDebuggable]
+    procedure RemoveCurlyBracketsFromString("Value": Text): Text
+    begin
+        exit(DelChr("Value", '=', '{}'));
+    end;
+
+    procedure ConvertToDateTime(PropertyValue: Text): DateTime
+    var
+        NewDateTime: DateTime;
+    begin
+        NewDateTime := 0DT;
+        // PropertyValue is something like the following: 'Mon, 24 May 2021 12:25:27 GMT'
+        // 'Evaluate' converts these correctly
+        Evaluate(NewDateTime, PropertyValue);
+        exit(NewDateTime);
+    end;
+
+    procedure ConvertToInteger(PropertyValue: Text): Integer
+    var
+        NewInteger: Integer;
+    begin
+        if Evaluate(NewInteger, PropertyValue) then
+            exit(NewInteger);
+    end;
+
+    procedure ConvertToEnum(FieldName: Text; PropertyValue: Text): Variant
+    begin
+        if FieldName = 'Resource Type' then
+            case LowerCase(PropertyValue) of
+                LowerCase(Format(Enum::"AFS File Resource Type"::File)):
+                    exit(Enum::"AFS File Resource Type"::File);
+                LowerCase(Format(Enum::"AFS File Resource Type"::Directory)):
+                    exit(Enum::"AFS File Resource Type"::Directory);
+            end;
+    end;
+
+    procedure GetRfc1123DateTime(MyDateTime: DateTime): Text
+    begin
+        exit(FormatDateTime(MyDateTime, 'R')); // https://go.microsoft.com/fwlink/?linkid=2210384
+    end;
+
+    local procedure FormatDateTime(MyDateTime: DateTime; FormatSpecifier: Text): Text
+    var
+        DateTimeAsXmlString: Text;
+        DateTimeDotNet: DotNet DateTime;
+    begin
+        DateTimeAsXmlString := Format(MyDateTime, 0, 9); // Format as XML, e.g.: 2020-11-11T08:50:07.553Z
+        exit(DateTimeDotNet.Parse(DateTimeAsXmlString).ToUniversalTime().ToString(FormatSpecifier));
+    end;
+}
--- a/src/System Application/App/Azure File Services API/src/Helper/AFSHttpContentHelper.Codeunit.al
+++ b/src/System Application/App/Azure File Services API/src/Helper/AFSHttpContentHelper.Codeunit.al
@@ -3,6 +3,8 @@
 // Licensed under the MIT License. See License.txt in the project root for license information.
 // ------------------------------------------------------------------------------------------------
 
+namespace System.Azure.Storage.Files;
+
 codeunit 8953 "AFS HttpContent Helper"
 {
     Access = Internal;
--- a/src/System Application/App/Azure File Services API/src/Helper/AFSHttpContentHelper.Codeunit.al
+++ b/src/System Application/App/Azure File Services API/src/Helper/AFSHttpContentHelper.Codeunit.al
@@ -1,0 +1,114 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+codeunit 8953 "AFS HttpContent Helper"
+{
+    Access = Internal;
+    InherentEntitlements = X;
+    InherentPermissions = X;
+
+    var
+        ContentLengthLbl: Label '%1', Comment = '%1 = Length', Locked = true;
+        RangeLbl: Label 'bytes=%1-%2', Comment = '%1 = Range Start, %2 = Range End', Locked = true;
+
+    /// <summary>
+    /// Checks if the HttpContent is empty
+    /// </summary>
+    /// <param name="HttpContent">The HttpContent to check.</param>
+    /// <returns>true if the content is not empty, false otherwise.</returns>
+    [NonDebuggable]
+    procedure ContentSet(HttpContent: HttpContent): Boolean
+    var
+        VarContent: Text;
+    begin
+        HttpContent.ReadAs(VarContent);
+        if StrLen(VarContent) > 0 then
+            exit(true);
+
+        exit(VarContent <> '');
+    end;
+
+    /// <summary>
+    /// Retrieves the length of the given stream (used for "HttpContent-Length" header in PUT-operations)
+    /// </summary>
+    /// <param name="SourceInStream">The InStream for Request Body.</param>
+    /// <returns>The length of the current stream</returns>
+    [NonDebuggable]
+    procedure GetContentLength(var SourceInStream: InStream): Integer
+    var
+        Length: Integer;
+    begin
+        Evaluate(Length, Format(SourceInStream.Length()));
+        exit(Length);
+    end;
+
+    /// <summary>
+    /// Retrieves the length of the given stream (used for "HttpContent-Length" header in PUT-operations)
+    /// </summary>
+    /// <param name="SourceText">The Text for Request Body.</param>
+    /// <returns>The length of the current stream</returns>
+    [NonDebuggable]
+    procedure GetContentLength(SourceText: Text): Integer
+    var
+        Length: Integer;
+    begin
+        Length := StrLen(SourceText);
+        exit(Length);
+    end;
+
+    [NonDebuggable]
+    procedure AddFilePutContentHeaders(AFSOperationPayload: Codeunit "AFS Operation Payload"; ContentLength: Integer; ContentType: Text; RangeStart: Integer; RangeEnd: Integer)
+    var
+        HttpContent: HttpContent;
+    begin
+        AddFilePutContentHeaders(HttpContent, AFSOperationPayload, ContentLength, ContentType, RangeStart, RangeEnd);
+    end;
+
+    [NonDebuggable]
+    procedure AddFilePutContentHeaders(var HttpContent: HttpContent; AFSOperationPayload: Codeunit "AFS Operation Payload"; var SourceInStream: InStream; RangeStart: Integer; RangeEnd: Integer)
+    var
+        Length: Integer;
+    begin
+        // Do this before calling "GetContentLength", because for some reason the system errors out with "Cannot access a closed Stream."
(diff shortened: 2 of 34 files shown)


pri-kise:
We need now Namespaces (and optional using statements)

I'd suggest something like `namespace System.Azure.Storage.Files.Shares;`

Krajzys:
Thanks for the tip. How about just `System.Azure.Storage.Files`? Adding shares might be a bit redundant (the only file service on the Azure storage is File Share).

Regarding second point, yes I think we can remove the AFS from most files, but should I then replace it with more descriprive name or just remove the prefix? e.g. AFS Handle -> File Handle.

JesperSchulz:
Uh, naming and namespace discussions. I'm curious what @AndreasMoth would suggest 😁

DevHarmonize:
@JesperSchulz 

Hi Jesper,
we spoke during Directions Lyon about this PR, Is it still possible to have this in the product in one of the next minors?

kr

src/System Application/Test/Azure File Services API/src/AFSTestLibrary.Codeunit.al
```diff
@@ -0,0 +1,52 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+codeunit 132515 "AFS Test Library"
```
JesperSchulz: Should this file not be in the "Test Library" folder?

src/System Application/Test/Azure File Services API/src/AzureTestLibrary.Codeunit.al
```diff
@@ -0,0 +1,78 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+/// <summary>
+/// Provides common authorization functionality for using Azure storage accounts in tests for Azure Storage Services.
+/// </summary>
+codeunit 132516 "Azure Test Library"
```
JesperSchulz: Should this file not be in the "Test Library" folder?

src/System Application/Test/Azure File Services API/src/AFSFileClientTest.Codeunit.al
```diff
@@ -0,0 +1,303 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
```
JesperSchulz, line 5: Even test codeunits need namespaces in the System Application.

JesperSchulz:
Security Review scheduled for Tuesday, 12th of December.

JesperSchulz:
Security review conducted with @darjoo. Darrick will comment on outcome.

darjoo:
After performing the security review with Jesper, the following is the outcome that needs to be fixed.
- Telemetry for all usages

Fx, telemetry when calling `CreateFile`
- OnCall, OnSuccess, OnError

JesperSchulz:
@Krajzys, if we get the missing telemetry added next week, we can release this module with 23.3 in January!

Krajzys:
@JesperSchulz great! I will get on to it. I should add it through the LogMessage procedure right? I have a questions then. How should I detemine the EventIDs?

Krajzys:
I have added the telemetry messages, for now with temporary eventIDs (0000AFS) that need to be changed. Please let me know if the telemetry is enough and how to proceed with the eventIDs 😃.

src/System Application/Test/DisabledTests/AFSFileClient.json
```diff
@@ -0,0 +1,8 @@
+[
+    {
+        "bug":  "",
+        "codeunitId":  132517,
```
JesperSchulz, line 4: Why are the tests disabled? Does that mean we've got 0% test coverage?

JesperSchulz (changes requested):

JesperSchulz:
> @JesperSchulz great! I will get on to it. I should add it through the LogMessage procedure right? I have a questions then. How should I detemine the EventIDs?

We've got a script which determines the IDs. I'll assign the fitting IDs to your PR! However, we're currently discussing that file names and folder names are customer content and hence cannot get emitted to telemetry. But without that information, what good does telemetry then do? We'll post the outcome of that discussion to this PR. If worst comes to worst, we temporarily revert your last commit and submit the module to the repo in its former state. Then we can create a chaser PR, if we come up with meaningful telemetry!

The 2 things pending for sure though are the disabled tests and the Test Libraries in the Test project. Once those have been fixed, I'm tempted to push the "merge button" 😎

src/System Application/Test/DisabledTests/AFSFileClient.json
pri-kise: I think the problem was, when the tests were created, that @darjoo didn't allow any IsHandled pattern for Testing Purpose only.

Therefore those tests require a real Azure File Share Instance for testing.

Azurite doesn't support Azure File Share yet: https://github.com/Azure/Azurite/issues/113

The module maybe needs to be refactored to use the RestClient and give the possibility to specify a HttpClientHanlder.  -> At least for testing purpose. Then it might be possible to Mock Web Requests.

src/System Application/Test/Azure File Services API/src/AzureTestLibrary.Codeunit.al
JesperSchulz: @grobyns just pointed out, that these probably aren't libraries - in the sense that they would come in handy in other modules too. They are probably rather internal test helpers. Maybe a change of name would suffice here. I'd like to reserve the word Library for proper test libraries, if possible.

darjoo:
@Krajzys wonderful job with adding the telemetry. A little too good :D 
Due to privacy reasons, we cannot log any variable potentially containing customer content, EUPI etc etc.
As such, filename, filepath, directory name and all cannot be logged. Please remove those bits.

Krajzys:
@darjoo @JesperSchulz Okay, I updated the telemetry messages and renamed the library codeunits to helper.
Regardind the disabled tests. What @pri-kise wrote was correct. We cannot mock requests with current architecture so the module would need to use RestClient. Otherwise we would need to have a regular azure file share setup only for testing purposes and somehow get the credentials for authorizing requests without hardcoding them.

src/System Application/Test/Azure File Services API/src/AFSTestHelper.Codeunit.al
```diff
@@ -0,0 +1,56 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Test.Azure.Storage.Files;
+
+using System.Azure.Storage;
+
+codeunit 132515 "AFS Test Helper"
```
grobyns: This name doesn't tell me what the code does... I'd name this 
AFS Get Test Storage Authorization

or a variation/abbreviation thereof... 

also, just FYI, I very much frown on terms like Helper and Management because they do not add anything. You might as well not add them to the name and you'd get the same amount of info... I think we can name things aptly with a little effort.

src/System Application/Test/Azure File Services API/src/AzureTestHelper.Codeunit.al
```diff
@@ -0,0 +1,82 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Test.Azure.Storage.Files;
+
+using System.Azure.Storage.Files;
+
+/// <summary>
+/// Provides common authorization functionality for using Azure storage accounts in tests for Azure Storage Services.
```
grobyns: does it provide common auth functionaliy? As far as i see most of the code clears the fileshare (i.e. a reset or init for tests).

src/System Application/Test/Azure File Services API/src/AzureTestHelper.Codeunit.al
```diff
@@ -0,0 +1,82 @@
...
+/// </summary>
+codeunit 132516 "Azure Test Helper"
```
grobyns: Maybe something like 
AFS Init. Test Storage

?

JesperSchulz:
Alright: so here's the plan. Address the remaining comments made by Gert and we'll merge. All HttpClient related code seems to be nicely wrapped in the internal codeunit 8954 "AFS Web Request Helper", which we afterwards can refactor to use the AL Rest Client - which again enables us to better test this.
Does that sound reasonable?

src/System Application/Test/Azure File Services API/src/AFSTestHelper.Codeunit.al
Krajzys: Okay, I agree. Changed to AFS Get Test Storage Auth.

src/System Application/Test/Azure File Services API/src/AzureTestHelper.Codeunit.al
Krajzys: Right, I updated the summary.

src/System Application/Test/Azure File Services API/src/AzureTestHelper.Codeunit.al
Krajzys: Yes, that sounds better. Updated.

Krajzys:
@JesperSchulz Yes, that sounds great. I fixed things pointed out by @grobyns.

JesperSchulz:
Awesome! You're REALLY responsive - wow! Let's do this 🥳
Starting CI!

JesperSchulz (approved):

src/System Application/App/Azure File Services API/src/Helper/AFSFormatHelper.Codeunit.al
```diff
@@ -0,0 +1,80 @@
...
+        UriText := Uri.GetAbsoluteUri();
+    end;
+
+    [NonDebuggable]
+    procedure RemoveCurlyBracketsFromString("Value": Text): Text
+    begin
+        exit(DelChr("Value", '=', '{}'));
+    end;
+
+    procedure ConvertToDateTime(PropertyValue: Text): DateTime
```
pri-kise, line 37: How will that DateTime Method will work in different languages?

What is the format of DateTime Values, that are returned by the API?
I couldn't find the API Documentation that fast.

Krajzys: This should be language agnostic since it should only be called on the values from the API.
The function you mentioned is used in List Directories calls, and documentation (https://learn.microsoft.com/en-us/rest/api/storageservices/list-directories-and-files#datetime-format-and-api-version-for-timestamp-fields) says it uses the following formats:
ISO8601
RFC1123

In my tests they were evaluated correctly.

JesperSchulz:
CI passed! I got a few more comments internally, which I'll address - nothing major. I expect to get this merged before 2024 and in time for the 2023 Update 3 in January.

src/System Application/App/Azure File Services API/src/Helper/AFSHttpContentHelper.Codeunit.al
```diff
@@ -0,0 +1,116 @@
...
+    end;
+
+    /// <summary>
+    /// Retrieves the max range avaialble to update with PutRange request.
+    /// </summary>
+    /// <returns>The max range available to update.</returns>
+    [NonDebuggable]
+    procedure GetMaxRange(): Integer
+    begin
+        exit(4 * Power(2, 20));
```
JesperSchulz, line 114: What's that number? Why 4 * 2 to the power of 20?

pri-kise: I guess this should be maximum ChunkSize for File Upload: 4 MiB -> 4194304 Bytes

Krajzys: Yes, my intention was for the procedure name to define this value, so this would be max range that can be uploaded at once. I can change it to what pri-kise proposed.

JesperSchulz: A comment describing this "magic" number would be fine :-)

JesperSchulz: But I'll take care of that. I'll push a few updates to this PR in the coming days. Nothing major at all.

JesperSchulz:
@Krajzys, I have addressed some minor code review comments in my internal 23.3 pull request, which I'd like to push to this branch before merging. Could you give me write access to the branch? When I've pushed my changes, I'd love for you to take one final look, and then we merge. But I've made sure this releases in January 🥳

darjoo (approved):

JesperSchulz:
I'll just merge and submit a chaser pack :-)

JesperSchulz:
@Krajzys, here a small post-merge request 😋
Any chance you could provide some very simple code samples on how to use this module for the release notes? I could create them myself and get inspired by the tests you've written, but maybe you have some great code already which would show just how simple it is to work with Azure Files?

Krajzys:
@JesperSchulz yes, no problem, I could provide some examples. I will try to send them here before the end of today.

Krajzys:
Sorry for the delay, I prepared three simple sample pieces of code.
```al
procedure GetFileAsStream(StorageAccount: Text; FileShare: Text; Authorization: Interface "Storage Service Authorization"; FilePath: Text)
var
    AFSFileClient: Codeunit "AFS File Client";
    AFSOperationResponse: Codeunit "AFS Operation Response";
    FileContents: InStream;
begin
    AFSFileClient.Initialize(StorageAccount, FileShare, Authorization);
    AFSOperationResponse := AFSFileClient.GetFileAsStream(FilePath, FileContents);
    if not AFSOperationResponse.IsSuccessful() then
        Error(AFSOperationResponse.GetError());

    DownloadFromStream(FileContents, '', '', '', FilePath);
end;

procedure SendStreamFile(StorageAccount: Text; FileShare: Text; Authorization: Interface "Storage Service Authorization"; FilePath: Text; var FileContent: InStream)
var
    AFSFileClient: Codeunit "AFS File Client";
    AFSOperationResponse: Codeunit "AFS Operation Response";
begin
    AFSFileClient.Initialize(StorageAccount, FileShare, Authorization);
    AFSOperationResponse := AFSFileClient.CreateFile(FilePath, FileContent);
    if not AFSOperationResponse.IsSuccessful() then
        Error(AFSOperationResponse.GetError());

    AFSOperationResponse := AFSFileClient.PutFileStream(FilePath, FileContent);
    if not AFSOperationResponse.IsSuccessful() then
        Error(AFSOperationResponse.GetError());
end;

procedure GetFileNamesInDirectory(StorageAccount: Text; FileShare: Text; Authorization: Interface "Storage Service Authorization"; DirectoryPath: Text[2048]): List of [Text[2048]];
var
    AFSDirectoryContent: Record "AFS Directory Content";
    AFSFileClient: Codeunit "AFS File Client";
    AFSOperationResponse: Codeunit "AFS Operation Response";
    FileNames: List of [Text[2048]];
begin
    AFSFileClient.Initialize(StorageAccount, FileShare, Authorization);
    AFSOperationResponse := AFSFileClient.ListDirectory(DirectoryPath, AFSDirectoryContent);
    if not AFSOperationResponse.IsSuccessful() then
        Error(AFSOperationResponse.GetError());

    AFSDirectoryContent.SetRange("Resource Type", AFSDirectoryContent."Resource Type"::File);
    if not AFSDirectoryContent.FindSet() then
        exit;
    repeat
        FileNames.Add(AFSDirectoryContent."Full Name");
    until AFSDirectoryContent.Next() = 0;

    exit(FileNames);
end;
```

Here is also how I call them:
```al
action(SendFile)
{
    ApplicationArea = All;
    Caption = 'Send File';
    Image = SendAsPDF;
    ToolTip = 'Sends a file to the FileShare';

    trigger OnAction()
    var
        StorageServiceAuthorization: Codeunit "Storage Service Authorization";
        FileStream: InStream;
        FileName: Text;
    begin
        if not UploadIntoStream('', '', '', FileName, FileStream) then
            exit;
        SampleAFSOperations.SendStreamFile(StorageAccount, FileShare, StorageServiceAuthorization.UseReadySAS(SASToken), Path, FileStream);
        Message('File %1 sent succesfully.', FileName);
    end;
}
action(GetFileAsStream)
{
    ApplicationArea = All;
    Caption = 'Get File as Stream';
    ToolTip = 'Gets file as a stream.';
    Image = Download;

    trigger OnAction()
    var
        StorageServiceAuthorization: Codeunit "Storage Service Authorization";
    begin
        SampleAFSOperations.GetFileAsStream(StorageAccount, FileShare, StorageServiceAuthorization.UseReadySAS(SASToken), Path);
    end;
}
```

PJohnstonSysco:
@Krajzys  @JesperSchulz  Hi, I'm messing about with this in the most recent release and it works great with SAS Token but when I try to use a Shared Key for authentication I run into issues uploading a file to the File Share. I have attached the error. Any ideas? Authentication using shared key works fine for downloading and listing directory requests.

image

DevHarmonize:
I also receive this error when using shared key for upload operation.
But otherwise it works perfectly.

Thanks for your effort in making this @Krajzys !

PR #366, opened 2023-11-23
[Copilot] Add approximation token function to AI SDK

darjoo:
#### Summary 
The Bank Rec app has the token approximation function which should be in the AI SDK for all to use. 

#### Work Item(s) 
Fixes [AB#492097](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/492097)

### DIFF
--- a/src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
+++ b/src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
@@ -27,6 +27,7 @@ codeunit 7772 "Azure OpenAI Impl"
         TextCompletionsAOAIAuthorization: Codeunit "AOAI Authorization";
         EmbeddingsAOAIAuthorization: Codeunit "AOAI Authorization";
         FeatureTelemetry: Codeunit "Feature Telemetry";
+        Telemetry: Codeunit Telemetry;
         InvalidModelTypeErr: Label 'Selected model type is not supported.';
         GenerateRequestFailedErr: Label 'The request did not return a success status code.';
         CompletionsFailedWithCodeErr: Label 'The text completions generation failed.';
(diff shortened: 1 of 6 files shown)


src/System Application/App/User Settings/src/UserPersonalization.Page.al
```diff
@@ -135,7 +135,10 @@ page 9214 "User Personalization"
         UserSettingsImpl.HideUsersDependingOnPermissions(Rec);
         UserSettingsImpl.HideExternalUsers(Rec);
         CurrPage.Caption := UserSettingsTok;
+    end;
```
darjoo: This should not be here

JesperSchulz (dismissed):

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
```diff
@@ -480,4 +492,17 @@ codeunit 7772 "Azure OpenAI Impl"
...
 
+    [NonDebuggable]
+    procedure ApproximateTokenCount(Input: Text): Decimal
+    var
+        AverageWordsPerToken: Decimal;
+        TokenCount: Integer;
+        WordsInInput: Integer;
+    begin
+        AverageWordsPerToken := 0.6; // Based on OpenAI estimate
+        WordsInInput := Input.Split(' ', ',', '.', '!', '?', ';', ':', '/n').Count;
```
AndreasMoth, line 503: Should we also add '-', '_', '/', '\' etc.? Or maybe we can give a comment on why only these. Often a - combines two words.
Also, we should probably remove all empty results, like "test, abc" I assume will give the result "test", "" (between ',' and ' ') and "abc". I.E. a count of 3 words

darjoo: We certainly can but this as the name specifies, is just an approximate. If we want a more definite way which covers everything, we should look into https://github.com/openai/tiktoken at a future date. We can create a new function `TokenCount` when we use that.

EvgenijKorovin (approved):

JesperSchulz (approved):

Issue #314, opened 2023-11-08
[BCIdea]: Add reference "FindBlobsByTags" from the "ABS Client Impl." codeunit  to the "ABS Blob Client" codeunit


lorenzfleerackers:
### BC Idea Link

https://experience.dynamics.com/ideas/idea/?ideaid=6c231259-217e-ee11-a81c-000d3ae53364

### Description

I want to filter the blobs by their tags.

The function is allready created in the implementation codeunit but is not added to the client.

![image](https://github.com/microsoft/BCApps/assets/28091845/2a130ce7-da0c-4ae2-a81a-447fcb9e565b)

Can I create the pull request myself?

Kinds regards
Lorenz <br /><br />Internal work item: AB#492545

JesperSchulz:
Just a little more patience. We're currently setting the right automations up between GitHub <-> DevOps <-> BCIdeas. Shortly we'll be ready to triage and approve ideas!

JesperSchulz:
Approved! Please create a PR.

PR #368, opened 2023-11-23
Add reference "FindBlobsByTags" from the "ABS Client Impl." codeunit to the "ABS Blob Client"

lorenzfleerackers:
#### Summary 
BC Idea Link
https://experience.dynamics.com/ideas/idea/?ideaid=6c231259-217e-ee11-a81c-000d3ae53364

Description
I want to filter the blobs by their tags.

The function is allready created in the implementation codeunit but is not added to the client.

In this pull request, I included the function in the client codeunit.

#### Work Item(s) 
Add reference "FindBlobsByTags" from the "ABS Client Impl." codeunit to the "ABS Blob Client" codeunit #314 
https://github.com/microsoft/BCApps/issues/314

### DIFF
--- a/src/System Application/App/Azure Blob Services API/src/ABSBlobClient.Codeunit.al
+++ b/src/System Application/App/Azure Blob Services API/src/ABSBlobClient.Codeunit.al
@@ -570,6 +570,18 @@ codeunit 9053 "ABS Blob Client"
     end;
 
     /// <summary>
+    /// The Find Blobs By Tags operation retrieves blobs based on user-defined tags for the specified blob, represented as one or more key-value pairs.
+    /// see: https://go.microsoft.com/fwlink/?linkid=2211407
+    /// </summary> 
+    /// <param name="SearchTags">A Dictionary of [Text, Text] with tags to search on.</param>   
+    /// <param name="FoundBlobs">An XmlDocument containing the results of the identified blobs.</param>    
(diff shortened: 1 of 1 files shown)


JesperSchulz (dismissed):

mazhelez (dismissed):

src/System Application/App/Azure Blob Services API/src/ABSBlobClient.Codeunit.al
```diff
@@ -569,6 +569,18 @@ codeunit 9053 "ABS Blob Client"
         exit(ABSClientImpl.SetBlobTags(BlobName, Tags));
     end;
 
+    /// <summary>
+    /// The Find Blobs By Tags operation retrieves blobs based on user-defined tags for the specified blob, represented as one or more key-value pairs.
+    /// see: https://go.microsoft.com/fwlink/?linkid=2211407
```
JesperSchulz: That link actually points to "Set Blob Tags" and not to "Find Blobs By Tags". Minor thing, but maybe worth fixing before we merge?

lorenzfleerackers: I changed the URL to the Find Blobs By Tags documentation.

JesperSchulz (approved):

darjoo (approved):

PR #379, opened 2023-11-28
Add verify execute permissions

aholstrup1:
#### Summary 
Porting an internal check to verify that objects are given the appropriate permissions

#### Work Item(s) 
Fixes AB#492320

### DIFF
--- a/build/scripts/VerifyExecutePermissions.ps1
+++ b/build/scripts/VerifyExecutePermissions.ps1
@@ -1,0 +1,91 @@
+param(
+    [Parameter(Mandatory = $true)]    
+    [string] $ModulesDirectory
+)
+
+Set-StrictMode -Version Latest
+Write-Host "Verifying execute permissions on all objects in $ModulesDirectory"
+
+$includefilter = @('*.Table.al', '*.Report.al', '*.Codeunit.al', '*.Page.al', '*.XMlPort.al', '*.Query.al')
+$AlObjects = @{}
+
+foreach ($folder in (Get-ChildItem -path $ModulesDirectory -Directory)) {
+    $files = @(Get-ChildItem -path $folder.Fullname -Recurse -File -Filter *.al -Include $includefilter)
+
+    # Find all objects
+    foreach ($file in $files) {
+        $type = $file.BaseName.Split('.', 2)[1]
+        $content = (Get-Content -Path $file.FullName) | Where-Object -FilterScript { $_ -notmatch '^\s*//' } # filter out comment lines
+        $ObjectName = ($content -match "^$($type)\s\d*\s").Split(" ", 3)[2]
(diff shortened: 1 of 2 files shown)


build/scripts/VerifyExecutePermissions.ps1
```diff
@@ -0,0 +1,91 @@
...
+    [string] $ModulesDirectory
+)
+
+Set-StrictMode -Version Latest
+Write-Host "Verifying execute permissions on all objects in $ModulesDirectory"
+
+$includefilter = @('*.Table.al', '*.Report.al', '*.Codeunit.al', '*.Page.al', '*.XMlPort.al', '*.Query.al')
+$AlObjects = @{}
+
+foreach ($folder in (Get-ChildItem -path $ModulesDirectory -Directory)) {
```
mazhelez:
```suggestion
foreach ($folder in (Get-ChildItem -Path $ModulesDirectory -Directory)) {

```

mazhelez: I don't get why it's going to every folder in a module. Shouldn't it just scan recursively for .al files in the module's folder?

aholstrup1: It probably should 😆 Let me adjust that

build/scripts/VerifyExecutePermissions.ps1
```diff
@@ -0,0 +1,87 @@
+param(
+    [Parameter(Mandatory = $true)]
+    [string] $ModulesDirectory
```
mazhelez: I know we only call it for the system modules now, but technically `$ModulesDirectory` is just an app project folder.
I suppose this check would be added for other apps as well, so let's adjust that.

build/scripts/VerifyExecutePermissions.ps1
```diff
@@ -0,0 +1,87 @@
...
+                $ObsoletePending = $content.IndexOf($ObsoletePending[0]) -lt $content.IndexOf($brackets[1])
+            }
+        }
+        $ObsoleteRemoved = $ObsoleteRemoved -and (-not $ObsoletePending)
+    }
+
+    # check if inherent execute entitlements exist on object level
+    $InherentDefined = ($content -match '(\s)InherentEntitlements.+X')
+
+    if (!($ObsoleteRemoved) -AND !($InherentDefined)) {
```
mazhelez:
```suggestion
    if (!($ObsoleteRemoved) -and !($InherentDefined)) {

```

mazhelez (approved):
Not a fan of this parsing around, but I admit we have nothing better at the moment. So, let's do it!

mazhelez (approved):

tomasmsft (approved):

PR #401, opened 2023-12-01
#387 BCPT TestSuiteLineExists

### LINKED ISSUE #387: [Bug]: Missing function BCPT TestSuiteLineExists
### Describe the issue

Currently there are no way of knowing if a TestSuiteLine that runs an object already exists. 

Since removing the option Access = Internal is out of the option (#361 ) we have to create a function that returns true if a line with that objects exists.

An override function can also search for the combination of object and the parameter field.

To make the searching better a key should be created
  key(Key3; "BCPT Code", "Codeunit ID", Parameters)
     

### Expected behavior

Possible to see if a BCPT Suite line already exists.

### Steps to reproduce

n/a

### Additional context

I can fix it.

NAVFreak:
#### Summary 
I have added functions to see if Test Suite Lines already exists. On two functions it also returns the lineno.

#### Work Item(s) 
Fixes #387 

<br /><br />Internal work item: AB#493728

### DIFF
--- a/src/Tools/Performance Toolkit/App/src/BCPTLine.Table.al
+++ b/src/Tools/Performance Toolkit/App/src/BCPTLine.Table.al
@@ -260,6 +260,9 @@ table 149001 "BCPT Line"
         {
             SumIndexFields = "No. of Sessions";
         }
+
+        key(Key3; "BCPT Code", "Codeunit ID", Parameters)
+        { }
     }
 
     var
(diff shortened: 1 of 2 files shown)


JesperSchulz:
@NAVFreak, please fix compilation issues, then we'll get this reviewed and in it goes 🥳

Seem to be fairly straight forward:
Error: AA0150 Parameter 'SuiteCode' is declared by reference but never changed in method 'SetBCPTLineCodeunitFilter'.
Error: AA0150 Parameter 'CodeunitID' is declared by reference but never changed in method 'SetBCPTLineCodeunitFilter'.

ChethanT (approved):

PR #402, opened 2023-12-01
#398 GetSeed for CU Any

### LINKED ISSUE #398: Add "GetSeed" to "Library - Random"
### Describe the request

When working with random values it could be more helpful if you can output the current seed to the error message by getting it:

image

(Or would it be possible to output the current seed by default at the stored error message in the test tool?)

It would help at analyzing errors by knowing which seed got used to reproduce the error (by setting it for example at the beginning of the test and then debug)

### Additional context

see above

TKapitan:
#### Summary 
GetSeed() procedure for CU Any. It required a small refactoring for the SetDefaultSeed() procedure to explicitly define the seed instead of the implicit definition using Randomize(). The value used as seed is still the same.

#### Work Item(s) 
Fixes #398

<br /><br />Internal work item: AB#493044

### DIFF
--- a/src/Tools/Test Framework/Test Libraries/Any/src/Any.Codeunit.al
+++ b/src/Tools/Test Framework/Test Libraries/Any/src/Any.Codeunit.al
@@ -252,14 +252,9 @@ codeunit 130500 "Any"
     /// Setting this value will change the numbers returned.
     /// </summary>
     procedure SetDefaultSeed()
-    var
-        DateTime1, DateTime2 : DateTime;
     begin
-        DateTime1 := CreateDateTime(Today(), 000000T);
-        DateTime2 := CurrentDateTime();
-
         SeedSet := true;
-        SetSeed(DateTime2 - DateTime1);
+        SetSeed(Time() - 0T);
     end;
 
     /// <summary>.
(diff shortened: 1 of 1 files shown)


src/Tools/Test Framework/Test Libraries/Any/src/Any.Codeunit.al
```diff
@@ -238,14 +238,28 @@ codeunit 130500 "Any"
...
     procedure SetDefaultSeed()
+    var
+        DateTime1, DateTime2 : DateTime;
     begin
+        DateTime1 := CreateDateTime(Today(), 000000T);
+        DateTime2 := CurrentDateTime();
+
         SeedSet := true;
-        Randomize();
+        SetSeed(DateTime2 - DateTime1);
```
grobyns: The default seed is the the current time (# ms since midnight). Date doesn't come into play afaik so why not simply do:
        SetSeed(Time() - 0T);

JesperSchulz: What do you think, @TKapitan. Do you want to make that change?

TKapitan: > The default seed is the the current time (# ms since midnight). Date doesn't come into play afaik so why not simply do: SetSeed(Time() - 0T);

You got me. I somehow forgot that anything that returns Duration can be used. Thanks @grobyns

IhorHandziuk (approved):

PR #404, opened 2023-12-04
[Copilot reminder texts] New enum values and events to support feature-specific prompt, capability and saving the accepted suggestion

dcenic:
Entity Text has a hardcoded completion prompt and Copilot Capability for 'Generate Marketing Text'. It also does the saving of accepted proposal in a very specific way (directly to Entity Text system table).
With this change I am:
- introducing overrides based on Entity Text Scenario, that enables Entity Text logic to be run for Reminder Text as well as for Marketing Text
- new tone and emphasis values.

<br /><br />Internal work item: AB#493390

### DIFF
--- a/src/System Application/App/Entity Text/src/EntityTextCapability.EnumExt.al
+++ b/src/System Application/App/Entity Text/src/EntityTextCapability.EnumExt.al
@@ -7,5 +7,9 @@ enumextension 2015 "Entity Text Capability" extends "Copilot Capability"
     {
         Caption = 'Marketing text suggestions';
     }
+    value(2016; "Reminder Text")
+    {
+        Caption = 'Reminder text suggestions';
+    }
 
 }
--- a/src/System Application/App/Entity Text/src/EntityTextCapability.EnumExt.al
+++ b/src/System Application/App/Entity Text/src/EntityTextCapability.EnumExt.al
@@ -3,13 +3,13 @@ using System.AI;
 
 enumextension 2015 "Entity Text Capability" extends "Copilot Capability"
 {
+    value(2014; "Reminder Text")
+    {
+        Caption = 'Reminder text suggestions';
+    }
     value(2015; "Entity Text")
     {
         Caption = 'Marketing text suggestions';
-    }
-    value(2016; "Reminder Text")
-    {
-        Caption = 'Reminder text suggestions';
     }
 
 }
--- a/src/System Application/App/Entity Text/src/EntityTextEmphasis.Enum.al
+++ b/src/System Application/App/Entity Text/src/EntityTextEmphasis.Enum.al
@@ -67,4 +67,13 @@ enum 2012 "Entity Text Emphasis"
     {
         Caption = 'Speed';
     }
+
+    /// <summary>
+    /// Emphasizes consequences.
+    /// </summary>
+    value(7; Consequences)
+    {
+        Caption = 'Consequences';
+    }
+
 }
--- a/src/System Application/App/Entity Text/src/EntityTextImpl.Codeunit.al
+++ b/src/System Application/App/Entity Text/src/EntityTextImpl.Codeunit.al
@@ -226,7 +226,7 @@ codeunit 2012 "Entity Text Impl."
     begin
         PromptSecretName := PromptObjectKeyTxt;
 
-        exit(GetPromptObject(PromptObjectKeyTxt))
+        exit(GetPromptObject(PromptSecretName))
     end;
 
     [NonDebuggable]
--- a/src/System Application/App/Entity Text/src/EntityTextImpl.Codeunit.al
+++ b/src/System Application/App/Entity Text/src/EntityTextImpl.Codeunit.al
@@ -221,7 +221,11 @@ codeunit 2012 "Entity Text Impl."
 
     [NonDebuggable]
     local procedure GetPromptInfo(): JsonObject
+    var
+        PromptSecretName: Text;
     begin
+        PromptSecretName := PromptObjectKeyTxt;
+
         exit(GetPromptObject(PromptObjectKeyTxt))
     end;
 
--- a/src/System Application/App/Entity Text/src/EntityTextImpl.Codeunit.al
+++ b/src/System Application/App/Entity Text/src/EntityTextImpl.Codeunit.al
@@ -29,14 +29,19 @@ codeunit 2012 "Entity Text Impl."
         exit(AzureOpenAI.IsEnabled(Enum::"Copilot Capability"::"Entity Text", Silent));
     end;
 
-    procedure CanSuggest(): Boolean
+    procedure CanSuggest(SourceScenario: Enum "Entity Text Scenario"): Boolean
     var
         EntityTextAOAISettings: Codeunit "Entity Text AOAI Settings";
     begin
         if not EntityTextAOAISettings.IsEnabled(true) then
             exit(false);
 
-        exit(HasPromptInfo());
+        exit(HasPromptInfo(SourceScenario));
+    end;
+
+    procedure CanSuggest(): Boolean
+    begin
+        exit(CanSuggest(0));
     end;
 
     [NonDebuggable]
@@ -68,10 +73,8 @@ codeunit 2012 "Entity Text Impl."
     procedure InsertSuggestion(SourceTableId: Integer; SourceSystemId: Guid; SourceScenario: Enum "Entity Text Scenario"; SuggestedText: Text)
     var
         EntityText: Record "Entity Text";
-        Handled: Boolean;
     begin
-        OnInsertSuggestion(SourceTableId, SourceSystemId, SourceScenario, SuggestedText, Handled);
-        if Handled then
+        if SourceScenario = Enum::"Entity Text Scenario"::"Reminder Text" then
             exit;
 
         InsertSuggestion(SourceTableId, SourceSystemId, SourceScenario, SuggestedText, EntityText);
@@ -212,6 +215,12 @@ codeunit 2012 "Entity Text Impl."
     end;
 
     [NonDebuggable]
+    local procedure GetPromptInfo(): JsonObject
+    begin
+        GetPromptInfo(0);
+    end;
+
+    [NonDebuggable]
     local procedure GetPromptInfo(var SourceScenario: Enum "Entity Text Scenario"): JsonObject
     var
         AzureKeyVault: Codeunit "Azure Key Vault";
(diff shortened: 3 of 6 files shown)


src/System Application/App/Entity Text/src/EntityTextImpl.Codeunit.al
```diff
@@ -50,19 +56,24 @@ codeunit 2012 "Entity Text Impl."
...
 
         exit(Suggestion);
     end;
 
     procedure InsertSuggestion(SourceTableId: Integer; SourceSystemId: Guid; SourceScenario: Enum "Entity Text Scenario"; SuggestedText: Text)
     var
         EntityText: Record "Entity Text";
+        Handled: Boolean;
     begin
+        OnInsertSuggestion(SourceTableId, SourceSystemId, SourceScenario, SuggestedText, Handled);
```
darjoo: I don't agree with this event. 

Using events will allow any extension to potentially break existing copilots. We should refactor to have a way to control this before we even get here. Such that Marketing Text Suggestions or any copilot that uses `GenerateSuggestion` will always have their intended flow. 


Side note, the event is in the Impl. It can't be subscribed by any extensions, is that by design? If so, the above holds less weight.

src/System Application/App/Entity Text/src/EntityTextImpl.Codeunit.al
```diff
@@ -201,14 +212,19 @@ codeunit 2012 "Entity Text Impl."
...
     var
         AzureKeyVault: Codeunit "Azure Key Vault";
         PromptObject: JsonObject;
         PromptObjectText: Text;
+        PromptSecretName: Text;
+        Handled: Boolean;
     begin
-        if not AzureKeyVault.GetAzureKeyVaultSecret(PromptObjectKeyTxt, PromptObjectText) then
-            Error(PromptNotFoundErr);
+        OnGetPromptSecretName(PromptSecretName, SourceScenario, Handled);
```
darjoo: Same with this event.

src/System Application/App/Entity Text/src/EntityTextImpl.Codeunit.al
dcenic: Event removed in latest iteration

src/System Application/App/Entity Text/src/EntityTextImpl.Codeunit.al
dcenic: Event removed in last iteration

src/System Application/App/Entity Text/src/EntityTextCapability.EnumExt.al
```diff
@@ -3,6 +3,10 @@ using System.AI;
 
 enumextension 2015 "Entity Text Capability" extends "Copilot Capability"
 {
+    value(2014; "Reminder Text")
```
Jose-agg, line 6: I don't think that it make sense to have this value here, as the "Entity Text" app should be agnostic to the different implementations - and in this situation you are adding some feature specific code to a generic tool.

src/System Application/App/Entity Text/src/EntityTextImpl.Codeunit.al
```diff
@@ -63,6 +79,9 @@ codeunit 2012 "Entity Text Impl."
     var
         EntityText: Record "Entity Text";
     begin
+        if SourceScenario = Enum::"Entity Text Scenario"::"Reminder Text" then
```
Jose-agg, line 82: Same comment as before, this is adding a dependency in a "generic" tool to a feature, is this something we really want to do? Wouldn't it be better to leave this code agnostic to the feature that use it and write the code for reminders on top of it?

src/System Application/App/Entity Text/src/EntityTextImpl.Codeunit.al
```diff
@@ -202,12 +221,35 @@ codeunit 2012 "Entity Text Impl."
...
+
+        exit(GetPromptObject(PromptSecretName))
+    end;
+
+    [NonDebuggable]
+    local procedure GetPromptInfo(SourceScenarioName: Text): JsonObject
+    var
+        PromptSecretName: Text;
+    begin
+        if SourceScenarioName = Format(Enum::"Entity Text Scenario"::"Reminder Text") then
```
Jose-agg, line 237: Same as before.

src/System Application/App/Entity Text/src/EntityTextImpl.Codeunit.al
```diff
@@ -423,7 +473,12 @@ codeunit 2012 "Entity Text Impl."
             AzureOpenAI.SetAuthorization(Enum::"AOAI Model Type"::"Chat Completions", Endpoint, Deployment, ApiKey);
         end;
 
-        AzureOpenAI.SetCopilotCapability(Enum::"Copilot Capability"::"Entity Text");
+        if SourceScenarioName = Format(Enum::"Entity Text Scenario"::"Reminder Text") then
```
Jose-agg, line 476: Same as before

src/System Application/App/Entity Text/src/EntityTextImpl.Codeunit.al
```diff
@@ -451,6 +506,7 @@ codeunit 2012 "Entity Text Impl."
         Endpoint: Text;
         Deployment: Text;
         ApiKey: SecretText;
+        ReminderTextPromptObjectKeyTxt: Label 'AOAI-Prompt-GenerateReminderText', Locked = true;
```
Jose-agg, line 509: Same as before

src/System Application/App/Entity Text/src/EntityTextScenarioExt.EnumExt.al
```diff
@@ -0,0 +1,9 @@
+namespace System.Text;
+
+enumextension 2014 "Entity Text Scenario Ext." extends "Entity Text Scenario"
```
Jose-agg, line 3: I believe that this extension should life within BaseApp. For the same reasons I mentioned before, and in this particular case to follow the design on the marketing text functionallity (the extension belongs to BaseApp)

Jose-agg (commented):
Please review the comments I left about making the tool agnostic to the different implementations.

src/System Application/App/Entity Text/src/EntityTextCapability.EnumExt.al
dcenic: it cannot be agnostic, this must be hardcoded, because we don't want to raise events. there was strong pushback on raising any events, so we decided to have 'marketing text' and 'reminder text' support  hardcoded in Entity Text module.

src/System Application/App/Entity Text/src/EntityTextImpl.Codeunit.al
dcenic: Same answer as before - it is a conscious decision not to raise events

src/System Application/App/Entity Text/src/EntityTextEmphasis.Enum.al
```diff
@@ -67,4 +67,13 @@ enum 2012 "Entity Text Emphasis"
     {
         Caption = 'Speed';
     }
+
+    /// <summary>
+    /// Emphasizes consequences.
+    /// </summary>
+    value(7; Consequences)
```
nikolakukrika, line 74: Is consequences the best term? Maybe warnings?

src/System Application/App/Entity Text/src/EntityTextScenarioExt.EnumExt.al
dcenic: Same answer as before - it is a conscious decision not to raise events. it must live inside Entity Text so that I can make an if statement instead of raising an event and asking the subscriber to provide the prompt secret name or the copilot capability

src/System Application/App/Entity Text/src/EntityTextImpl.Codeunit.al
dcenic: Same answer as before - it is a conscious decision not to raise events, but hardcode Reminder Text inside Entity Text. We must not let subscribers provide their own prompt.

dcenic: Same answer as before - it is a conscious decision not to raise events, but hardcode Reminder Text inside Entity Text. We must not let subscribers provide their own prompt.

src/System Application/App/Entity Text/src/EntityTextScenarioExt.EnumExt.al
nikolakukrika, line 3: Extensions are rather limiting. You can have 1 per app. Keep this in mind. Maybe this was changed with recent modern dev work, but I believe it was not, try to define two enum extensions in the same app and see if it will compile.

nikolakukrika: Reminders are a base app object, thus the extension should go to the base app.

src/System Application/App/Entity Text/src/EntityTextImpl.Codeunit.al
```diff
@@ -41,20 +51,26 @@ codeunit 2012 "Entity Text Impl."
 
     [NonDebuggable]
     procedure GenerateSuggestion(Facts: Dictionary of [Text, Text]; Tone: Enum "Entity Text Tone"; TextFormat: Enum "Entity Text Format"; TextEmphasis: Enum "Entity Text Emphasis"; CallerModuleInfo: ModuleInfo): Text
+    begin
+        exit(GenerateSuggestion(Facts, Tone, TextFormat, TextEmphasis, CallerModuleInfo, ''));
+    end;
+
+    [NonDebuggable]
```
nikolakukrika, line 58: NonDebuggable is too wide in this case. It should only cover the methods that are sensitive. 
IsEnabled and CanSuggest should be debuggable, then from BuildPrompts and GenerateAndReviewCompletion it should be non-debuggable

src/System Application/App/Entity Text/src/EntityTextImpl.Codeunit.al
```diff
@@ -41,20 +51,26 @@ codeunit 2012 "Entity Text Impl."
...
+    procedure GenerateSuggestion(var Facts: Dictionary of [Text, Text]; var Tone: Enum "Entity Text Tone"; var TextFormat: Enum "Entity Text Format"; var TextEmphasis: Enum "Entity Text Emphasis"; var CallerModuleInfo: ModuleInfo; SourceScenarioName: Text): Text
     var
         SystemPrompt: Text;
         UserPrompt: Text;
```
nikolakukrika, line 62: UserPrompt should be moved within the codeunit and made non-debuggable, so codeunit hides the internal value. Then we could have it as a global variable and not pass it around. This way we can have more code debuggable.

dcenic:
Closed - Reminder Text logic will live in Base App. If I need to copy & paste something from Entity Text, I will do it.

PR #413, opened 2023-12-06
Add SecretText overload

SBalslev:
When working with Azure KeyVault it should be possible to get secrets as SecretText.

Fixes [AB#493425](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/493425)

### DIFF
--- a/src/System Application/App/Secrets/src/AppKeyVaultSecretPrImpl.Codeunit.al
+++ b/src/System Application/App/Secrets/src/AppKeyVaultSecretPrImpl.Codeunit.al
@@ -66,4 +66,15 @@ codeunit 3801 "App Key Vault Secret Pr. Impl."
         // It will return false in these cases, and set SecretValue=''.
         exit(NavAzureKeyVaultAppSecretProvider.TryGetAzureKeyVaultSecret(SecretName, SecretValue));
     end;
+
+    [NonDebuggable]
+    procedure GetSecret(SecretName: Text; var SecretValue: SecretText): Boolean
+    var
+        SecretValueText: Text;
+    begin
+        if not GetSecret(SecretName, secretValueText) then
+            exit(false);
+        SecretValue := SecretValueText;
+        exit(true);
+    end;
 }
--- a/src/System Application/App/Secrets/src/AppKeyVaultSecretPrImpl.Codeunit.al
+++ b/src/System Application/App/Secrets/src/AppKeyVaultSecretPrImpl.Codeunit.al
@@ -72,7 +72,7 @@ codeunit 3801 "App Key Vault Secret Pr. Impl."
     var
         SecretValueText: Text;
     begin
-        if not GetSecret(SecretName, secretValueText) then
+        if not GetSecret(SecretName, SecretValueText) then
             exit(false);
         SecretValue := SecretValueText;
         exit(true);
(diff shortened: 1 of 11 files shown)


pri-kise:
I like this PR, but those methods are all Scope OnPrem.

What could be very useful is to add such methods to a `interface "Secret Provider V2"` and implement this in `codeunit 3800 "App Key Vault Secret Provider"`.
Are there already plans to improve those codeunits, too?

See also : https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-app-key-vault

darjoo:
We should obsolete the old function and move towards recommending using only the SecretText function.

SBalslev:
> We should obsolete the old function and move towards recommending using only the SecretText function.

At this point I disagree; we do not know the data stored in KeyVault so we cannot make assumptions that all consumers will use the data in a context that can use SecretText. An example is configuration data like Microsoft Entra application client id.

src/System Application/App/Secrets/src/AppKeyVaultSecretPrImpl.Codeunit.al
```diff
@@ -66,4 +66,15 @@ codeunit 3801 "App Key Vault Secret Pr. Impl."
         // It will return false in these cases, and set SecretValue=''.
         exit(NavAzureKeyVaultAppSecretProvider.TryGetAzureKeyVaultSecret(SecretName, SecretValue));
     end;
+
+    [NonDebuggable]
+    procedure GetSecret(SecretName: Text; var SecretValue: SecretText): Boolean
+    var
+        SecretValueText: Text;
+    begin
+        if not GetSecret(SecretName, secretValueText) then
```
christianbraeunlich:
```suggestion
        if not GetSecret(SecretName, SecretValueText) then
```

src/System Application/App/Secrets/src/InMemorySecretProvImpl.Codeunit.al
```diff
@@ -37,4 +37,16 @@ codeunit 3803 "In Memory Secret Prov Impl."
 
         exit(false);
     end;
+
+    [NonDebuggable]
+    procedure GetSecret(SecretName: Text; var SecretValue: SecretText): Boolean
+    var
+        SecretValueText: Text;
+    begin
+        if not GetSecret(SecretName, secretValueText) then
```
christianbraeunlich:
```suggestion
        if not GetSecret(SecretName, SecretValueText) then
```

christianbraeunlich (changes requested):

src/System Application/App/Secrets/src/SecretProviderv2.Interface.al
```diff
@@ -0,0 +1,28 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Security;
+
+/// <summary>
+/// Abstraction for secret providers.
+/// </summary>
+interface "Secret Provider v2"
```
darjoo, line 11: Last bit, missing tests

JesperSchulz (approved):

BazookaMusic (approved):
Looks good

PR #417, opened 2023-12-08
Add a script to create local development environment

mazhelez:
#### Summary 
Create a script to create local development environment for working with apps from the repository.

The NewDevEnv.ps1 script creates a docker container (if it doesn't exist), builds the apps, based on the provided parameters, and publishes the apps to the container.

Test run: https://github.com/mazhelez/BCApps/actions/runs/7225774768/job/19689897656

Note:
- The created container contains only apps from the BCApps repository.

To be considered for future enhancement:
- The created container contains all BC apps (by Microsoft): the ones from this repo are rebuilt, the rest are taken from the BC artifact.
- Rebuild apps functionality. Now the rebuild happens only after manual clean up the cache folder.

#### Work Item(s) 
Fixes AB#492033


.github/AL-Go-Settings.json
```diff
@@ -5,7 +5,7 @@
     "runs-on":  "windows-latest",
     "cacheImageName":  "",
     "UsePsSession":  false,
-    "artifact":  "bcinsider//24.0.14476.0/base",
+    "artifact":  "https://bcinsider.azureedge.net/sandbox/24.0.14476.0/base",
```
mazhelez: Opened a separate PR for this change: #418

.gitignore
```diff
@@ -24,6 +24,7 @@ bld/
 [Oo]bj/
 [Ll]og/
 [Oo]ut/
+.artifactscache
```
mazhelez:
```suggestion
.artifactsCache
```

build/scripts/DevEnv/NewDevEnv.ps1
```diff
@@ -0,0 +1,189 @@
...
+[pscredential] $script:credential = New-Object System.Management.Automation.PSCredential ($userName, $(ConvertTo-SecureString $password -AsPlainText -Force))
+$script:baseFolder = Get-BaseFolder
+$script:packageCacheFolder = ResolveFolder -folder $packageCacheFolder
+$script:containerName = $containerName
+
+Write-Host "Loading BCContainerHelper module" -ForegroundColor Yellow
+InstallBCContainerHelper
+
+Write-Host "Creating container $script:containerName" -ForegroundColor Yellow
+CreateBCContainer
```
aholstrup1: Do you need the container when building the app? Setting up a container is... slow. So if there's a compilation issue with my app, it'd be nice to know about that before spinning up a container

build/scripts/DevEnv/NewDevEnv.ps1
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    Import-Module "BCContainerHelper" -DisableNameChecking
+}
+
+function CreateBCContainer {
+    $containerExist = $null -ne $(docker ps -q -f name="$script:containerName")
+
+    if ($containerExist) {
+        Write-Host "Container $script:containerName already exists."
```
aholstrup1: This is nice! 👏 The only assumption is that you're using the same username/password I suppose?

build/scripts/DevEnv/NewDevEnv.ps1
```diff
@@ -0,0 +1,189 @@
+using module .\AppProjectInfo.class.psm1
```
aholstrup1, line 1: Don't be shy with this script (NewDevEnv.ps1)😄 Should we put it in the root folder instead?

build/scripts/DevEnv/NewDevEnv.ps1
```diff
@@ -0,0 +1,234 @@
...
+Write-Host "Creating container $script:containerName" -ForegroundColor Yellow
+CreateBCContainer
+
+foreach($currentProjectPath in $projectPaths) {
+    Write-Host "Building app in $currentProjectPath" -ForegroundColor Yellow
+    $appFile = BuildApp -appProjectFolder $currentProjectPath -packageCacheFolder $script:packageCacheFolder
+
+    Write-Host "Publishing app $appFile to $script:containerName" -ForegroundColor Yellow
+    PublishApp -appFile $appFile
+}
```
aholstrup1: I ran the script and got a BC environment with only the System Application installed. 
image


I guess this is by design but is that what developers would expect? Wouldn't it be better to get a full bc environment with baseapp and first party apps + The system application from my branch?

build/scripts/DevEnv/NewDevEnv.ps1
mazhelez: Yeah, didn't want to overthink it. There might be a way to get the credentials from the existing docker container, but I don't see why anyone would use different credentials for different containers.

build/scripts/DevEnv/NewDevEnv.ps1
mazhelez: That's because I'm still not sure if script is the better option here. _NewDevEnv_ is a script now because _localDevEnv_ is, but wouldn't a proper cmdlet be better? E.g. `New-BCDevEnv` 😎

build/scripts/DevEnv/NewDevEnv.ps1
mazhelez: That's definitely a valid scenario! I'll see how it can be supported.

build/scripts/DevEnv/NewDevEnv.ps1
aholstrup1: I think that's fine 😄

build/scripts/DevEnv/NewDevEnv.ps1
aholstrup1: Yeah perhaps. But then we'd need some kind of init script in the root? To init the New-BCDevEnv command

build/scripts/DevEnv/NewDevEnv.ps1
mazhelez: That order exists to potentially support compiling the apps in the container (depending on whether `useCompilerFolder` setting is used). Creating a container is needed for the dev env anyway. If an app fails to compile, the container will not be re-created on a subsequential run. 

If we explicitly decide to compiler the apps outside the container, then maybe creating the container and compiling the apps can be run in parallel. What do you think? 🤔

build/scripts/DevEnv/NewDevEnv.ps1
```diff
@@ -0,0 +1,234 @@
...
+[CmdletBinding(DefaultParameterSetName = 'ProjectPaths')]
+param(
+    [Parameter(Mandatory = $false)]
+    [string] $containerName = "BC-$(Get-Date -Format 'yyyyMMdd')",
+
+    [Parameter(Mandatory = $false)]
+    [string] $userName = 'admin',
+
+    [Parameter(Mandatory = $false)]
+    [string] $password = 'P@ssword1',
```
jonasblunck: I don't think we should default password parameters. If something need to be protected by a password, I'd suggest making the user provide it.

mazhelez: Of course. I only placed it there because, when testing, I keep forgetting what password I used (or simply mistyped it).

mazhelez: Fixed.

build/scripts/DevEnv/NewDevEnv.ps1
mazhelez: Latest approach: apps are compiled outside the container and the container itself is being created as a background job (while the apps are compiling). Hope that makes sense.

build/scripts/DevEnv/NewDevEnv.ps1
mazhelez: Not going to be in this iteration. Future enhancement it is 😉

build/scripts/DevEnv/AppProjectInfo.class.psm1
```diff
@@ -0,0 +1,75 @@
...
+        $this.AppJson = $_appJson
+    }
+
+    static [AppProjectInfo] Get([string] $appProjectFolder) {
+        $appInfo = [AppProjectInfo]::new($appProjectFolder, 'app')
+
+        return $appInfo
+    }
+
+    static [AppProjectInfo] Get([string] $appProjectFolder, $type) {
```
aholstrup1:
```suggestion
    static [AppProjectInfo] Get([string] $appProjectFolder, [string] $type) {
```

build/scripts/DevEnv/AppProjectInfo.class.psm1
```diff
@@ -0,0 +1,75 @@
...
+#>
+class AppProjectInfo {
+    [string] $AppProjectFolder
+    [string] $Id
+    [ValidateSet('app', 'test')]
+    [string] $Type
+    [PSCustomObject] $AppJson
+
+
+    hidden AppProjectInfo([string] $appProjectFolder, $type = 'app') {
```
aholstrup1:
```suggestion
    hidden AppProjectInfo([string] $appProjectFolder, [string] $type = 'app') {
```

build/scripts/DevEnv/NewDevEnv.ps1
```diff
@@ -0,0 +1,110 @@
...
+# Build apps
+$appFiles = @()
+$buildingAppsStats = Measure-Command {
+    Write-Host "Building apps..." -ForegroundColor Yellow
+    [Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSUseDeclaredVarsMoreThanAssignments', '', Justification = 'false-postiive')]
+    $appFiles = BuildApps -projectPaths $projectPaths -packageCacheFolder $packageCacheFolder
+}
+
+Write-Host "Building apps took $($buildingAppsStats.TotalSeconds) seconds"
+Write-Host "Apps files: $($appFiles -join [Environment]::NewLine)" -ForegroundColor Green
```
aholstrup1:
```suggestion
Write-Host "App files: $($appFiles -join [Environment]::NewLine)" -ForegroundColor Green
```

build/scripts/DevEnv/NewDevEnv.psm1
```diff
@@ -0,0 +1,311 @@
+using module .\AppProjectInfo.class.psm1
+using module .\ALGoProjectInfo.class.psm1
+
+Import-Module "BCContainerHelper" -DisableNameChecking
+Import-Module "$PSScriptRoot\..\EnlistmentHelperFunctions.psm1" -DisableNameChecking
+
+$script:allApp = @()
```
aholstrup1:
```suggestion
$script:allApps = @()
```

build/scripts/DevEnv/NewDevEnv.psm1
```diff
@@ -0,0 +1,311 @@
...
+                    $appInfos += $testAppInfo
+                }
+            }
+        }
+
+        $script:allApps = $appInfos
+    }
+
+    return $script:allApps
+}
```
aholstrup1: Are you missing _Export-ModuleMember_ or was that intentional? Not sure if it matters

mazhelez: Is all functions exported by default? I can definitely add them explicitly.

Also, how about following the `<Verb>-<Something>` pattern for function names?

TKapitan:
@mazhelez @JesperSchulz 

I think that the first point from future enhancements will be needed for the base app repository. The last update with v24 broke the current script (https://github.com/microsoft/BusinessCentralApps/issues/546), so now the base app repository can't be recreated (or even used)

build/scripts/DevEnv/NewDevEnv.psm1
EmilDamsbo: 1. I'd definitely agree with going for `<verb>-<thingitdoes>`
2. If a module doesn't specify any exports, its functions are still available after importing the module, but any variables are not available; if you intend to expose all functions here and not the variables, then no changes are required.

aholstrup1 (approved):

.gitignore
```diff
@@ -24,6 +24,7 @@ bld/
 [Oo]bj/
 [Ll]og/
 [Oo]ut/
+.artifactsCache
```
EmilDamsbo, line 27: This is the default value of the `NewDevEnv.ps1`, but what if someone's using a non-default cache path? Is it feasible to blanket exclude `*.app` files from being uploaded to source control? That's how the AL-Go templates handle it:
* https://github.com/microsoft/AL-Go-PTE/blob/main/.gitignore
* https://github.com/microsoft/AL-Go-AppSource/blob/main/.gitignore

mazhelez: *.app files are already excluded (see line 305).

If someone decides to use a use a non-default cache path that is the repo, they'll need to bear the consequences. The cache folder doesn't store only the .app files anyway.

mazhelez:
> @mazhelez @JesperSchulz
> 
> I think that the first point from future enhancements will be needed for the base app repository. The last update with v24 broke the current script ([microsoft/BusinessCentralApps#546](https://github.com/microsoft/BusinessCentralApps/issues/546)), so now the base app repository can't be recreated (or even used)

Interesting. That might be a separate issue.

The current PR aims to solve the problem with creating a local dev env when there are project dependencies.

build/scripts/DevEnv/NewDevEnv.ps1
mazhelez: Keeping it as is right now. We can always restructure later.

EmilDamsbo (approved):

aholstrup1 (approved):

PR #419, opened 2023-12-11
Skip work item validation on forks

mazhelez:
#### Summary 
Skip work item validation on forks
#### Work Item(s) 
Fixes AB#492651

### DIFF
--- a/.github/workflows/WorkitemValidation.yaml
+++ b/.github/workflows/WorkitemValidation.yaml
@@ -15,6 +15,7 @@ defaults:
 
 jobs:
   GitHubIssueValidation:
+    if: github.event.pull_request.head.repo.full_name == 'microsoft/BCApps'
     name: 'Validate link to issues'
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/WorkitemValidation.yaml
```diff
@@ -15,6 +15,7 @@ defaults:
 
 jobs:
   GitHubIssueValidation:
+    if: github.event.pull_request.head.repo.full_name == 'microsoft/BCApps'
```
aholstrup1, line 18: Wait.. what's the head/base repo in this case? Isn't the head repo where the PR is coming from. I.e. if someone opens up a PR into microsoft/BCApps from a fork, the head repo is user/BCApps and the base repo is microsoft/BCApps?

```suggestion
    if: github.event.pull_request.base.repo.full_name == 'microsoft/BCApps'
```

mazhelez: Oh wow, yes, nice catch. Totally swapped these two.

PR #439, opened 2023-12-18
[Copilot] Hide status field when data movement is not enabled

darjoo:
#### Summary 
**Problem** 
It is currently misleading that copilots are active/enabled when the data movement toggle is off.
If the toggle is off and copilots show active, the copilots are actually deactivated due to data movement not being allowed.

**Solution**
Hide the Status column whenever data movement toggle is off.

image

#### Work Item(s) 
Fixes AB#491279

### DIFF
--- a/src/System Application/App/AI/src/Copilot/CopilotCapabilitiesGA.Page.al
+++ b/src/System Application/App/AI/src/Copilot/CopilotCapabilitiesGA.Page.al
@@ -165,7 +165,6 @@ page 7774 "Copilot Capabilities GA"
 
     internal procedure SetDataMovement(Value: Boolean)
     begin
-        LogMessage('12345', 'SetDataMovement', Verbosity::Error, DataClassification::SystemMetadata, TelemetryScope::All, 'Data', Format(Value));
         DataMovementEnabled := Value;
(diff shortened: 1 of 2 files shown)


src/System Application/App/AI/src/Copilot/CopilotCapabilitiesGA.Page.al
```diff
@@ -164,6 +165,7 @@ page 7774 "Copilot Capabilities GA"
 
     internal procedure SetDataMovement(Value: Boolean)
     begin
+        LogMessage('12345', 'SetDataMovement', Verbosity::Error, DataClassification::SystemMetadata, TelemetryScope::All, 'Data', Format(Value));
```
JesperSchulz: Doesn't look right ;-)

IhorHandziuk (approved):

JesperSchulz (approved):

src/System Application/App/AI/src/Copilot/CopilotCapabilitiesGA.Page.al
darjoo: Went away :)

PR #460, opened 2024-01-03
Update extension installation flow for AppSource preview - part 2 

qutreson:
#### Summary 
- Propagate the 'responseUrl' when installing an app through page 2503
- Rename `ApplicationId` to `MarketplaceApplicationId` in "Extension Marketplace" codeunit to distinguish between the partner center application id (text with different tokens) and the AL application id (Guid from app.json).
- Remove code related to support of installation of apps by Package ID since it wasn't supported fully.

#### Work Item(s) 
Fixes [AB#486086](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/486086)


src/System Application/App/Extension Management/src/ExtensionInstallation.Table.al
```diff
@@ -33,5 +33,12 @@ table 2503 "Extension Installation"
         {
             Caption = 'Preview Key';
         }
+        /// <summary>
+        /// The telemetry URL for the Partner Center ingestion call back to be used when installing a version of an AppSource app.
+        /// </summary>
+        field(3; RESPONSEURL; Text[2048])
```
darjoo: Does `RESPONSEURL` and `PREVIEWKEY` need to be all caps?

darjoo: Is the caption also really `Telemetry URL`? :D

qutreson: I will double check because it has been some time since I added PREVIEWKEY, but AFAIR the table field's casing has to be matching the one in URL filter for the installation link on page 2503. Since it was using ID all upper case, I used PREVIEWKEY too. For the response URL, this isn't surfaced in the install link (at least for now), but I made it upper case for consistency. I agree, it looks a bit weird (hence my comment on the table header), but in AL, it doesn't really matter since it's not case sensitive so you can reference it with PascalCase in code. 

Regarding the caption, I can update it - not sure why I made it inconsistent.

src/System Application/App/Extension Management/src/MarketplaceExtnDeployment.Page.al
```diff
@@ -153,5 +153,5 @@ page 2510 "Marketplace Extn Deployment"
         ActiveUsersLbl: Label 'Note: There might be other users working in the system.';
         WarningLbl: Label 'Installing extensions during business hours will disrupt other users.';
         RefreshInfoLbl: Label 'After installation, your session will refresh, and you can set up your extension.';
-        PreviewInfoLbl: Label 'Note: You are about to install a preview version of the extension.';
+        PreviewInfoLbl: Label 'Note: A preview key was provided for the installation. If there is no higher public version for your environment, a preview version of the extension matching the specified preview key will be installed.';
```
qutreson: Missing MultiLine = true on corresponding field

qutreson: Added

src/System Application/App/Extension Management/src/ExtensionInstallation.Table.al
qutreson: Caption was updated. I also verified that the casing must match the one of the table. Waiting on decision regarding the casing, but likely going to stay this way.

qutreson: Discussed. Casing changed to PreviewKey!

darjoo (approved):

thpeder (approved):

PR #461, opened 2024-01-03
Remove dependency from Guided Experience to Entity Text

haoranpb:
#### Summary 

Remove dependency from Guided Experience to Entity Text

The Marketing Text Tour will show up on Eval companies, so no need to filter anymore. The Marketing Text FactBox will always show up in Eval Company (tho additional setup might be needed to use the functionality)

#### Work Item(s) 
Fixes [AB#495295](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/495295)

### DIFF
--- a/src/System Application/App/Guided Experience/src/Checklist Banner/ChecklistBanner.Codeunit.al
+++ b/src/System Application/App/Guided Experience/src/Checklist Banner/ChecklistBanner.Codeunit.al
@@ -41,15 +41,4 @@ codeunit 3728 "Checklist Banner"
     internal procedure OnOpenChecklistBannerPage(var SkipWelcomeState: Boolean; IsEvaluationCompany: Boolean)
     begin
     end;
-
-    /// <summary>
-    /// Integration Event to provide the ability to filter on the checklist items before it is loaded to the checklist banner.
-    /// Do NOT use complex logic in the subscriber as it will impact the performance of opening company.
-    /// </summary>
-    /// <param name="SkipChecklistItem">Whether to skip the checklist item or not </param>
-    /// <param name="SpotlightTourType">Specifies the checklist's Spotlight Tour type if any</param>
-    [IntegrationEvent(false, false)]
-    internal procedure OnBeforeAddChecklistItemToBuffer(var SkipChecklistItem: Boolean; SpotlightTourType: Enum "Spotlight Tour Type")
-    begin
-    end;
 }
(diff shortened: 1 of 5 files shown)


src/System Application/App/Guided Experience/src/Checklist Banner/ChecklistBanner.Codeunit.al
```diff
@@ -41,4 +41,15 @@ codeunit 3728 "Checklist Banner"
     internal procedure OnOpenChecklistBannerPage(var SkipWelcomeState: Boolean; IsEvaluationCompany: Boolean)
     begin
     end;
+
+    /// <summary>
+    /// Integration Event to provide the ability to filter on the checklist items before it is loaded to the checklist banner.
+    /// Do NOT use complex logic in the subscriber as it will impact the performance of opening company.
```
grobyns: seems like we shouldn't have this event at all... Why create the possibility of things going wrong?

src/System Application/App/Guided Experience/src/Checklist Banner/ChecklistBanner.Codeunit.al
```diff
@@ -41,4 +41,15 @@ codeunit 3728 "Checklist Banner"
...
+    /// </summary>
+    /// <param name="SkipChecklistItem">Whether to skip the checklist item or not </param>
+    /// <param name="SpotlightTourType">Specifies the checklist's Spotlight Tour type if any</param>
+    [IntegrationEvent(false, false)]
+    internal procedure OnBeforeAddChecklistItemToBuffer(var SkipChecklistItem: Boolean; SpotlightTourType: Enum "Spotlight Tour Type")
```
darjoo: Missing the Checklist Item itself. Any subscriber would not know what they are skipping except for the type.

src/System Application/App/Guided Experience/src/Checklist Banner/ChecklistBanner.Codeunit.al
haoranpb: Agree, trying a few things to try to get rid of it

src/System Application/App/Guided Experience/src/Checklist Banner/ChecklistBanner.Codeunit.al
haoranpb: Privacy deployment failed, so asked Sam directly.

The factbox is shown when there is read access to entity text, which is granted via `"System Application - View"`

So I guess it's safe to say in Eval Company, the Text Box we are looking for is always there. I vote we just remove the logic as the PR's current state

darjoo (approved):

src/System Application/App/Guided Experience/src/Checklist Banner/ChecklistBanner.Codeunit.al
haoranpb: Removed event after swtiching approach

grobyns (approved):

PR #463, opened 2024-01-04
Turn AW0009 and AS0059 warnings into errors

darjoo:
#### Summary 
Turn codecop AW0009 and AS0059 into errors.

Using pragmas instead of fixing due to following reasons:
AS0059: Table needs to be changed to temporary. Codecop was added due to their specific apparently.

AW0009: The places where the bitmaps are used, aren't one of the supported places for media/mediaset fields.
Fx, list of records. The only time it works, is if it is showing in a tile view, which for Emails, we are not.
The MediaCleanup uses Tenant Media record directly, which is a blob.

#### Work Item(s) 
Fixes [AB#487172](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/487172)

### DIFF
--- a/src/rulesets/ruleset.json
+++ b/src/rulesets/ruleset.json
@@ -31,12 +31,12 @@
   "rules": [
     {
       "id": "AS0059",
-      "action": "Warning",
+      "action": "Error",
       "justification": "Will be fixed. Reserved database tables are read-only in a multi-tenant environment"
     },
     {
       "id": "AW0009",
-      "action": "Warning",
+      "action": "Error",
       "justification": "Using a Blob with subtype Bitmap on a page field is deprecated. Instead use the Media/MediaSet data types."
     },
     {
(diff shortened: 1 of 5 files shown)


src/rulesets/ruleset.json
```diff
@@ -31,12 +31,12 @@
   "rules": [
     {
       "id": "AS0059",
```
IhorHandziuk: AFAIK, the way we enable errors is by deleting the entries from this file (as Error is the default)

IhorHandziuk (dismissed):

IhorHandziuk (approved):

JesperSchulz (approved):
Discussed why we need to pragma our way around the remaining two warnings, and the reasoning was legit. We cannot live up to those rules without considerable investment, which isn't justified compared to the gain we get out of it. Hence approved.

PR #464, opened 2024-01-04
[Document Sharing] Add cancelled parameter and use in word template edit after merge

stkillen:
#### Summary 
For document sharing, add a state parameter that is set when import of file is cancelled.

Use the state parameter to cancel execution in Word Templates `EditDocumentAfterMerge` if import is cancelled. This will ensure that attachments are not set to unique in Interaction Templates whenever the import is cancelled.

#### Work Item(s) 
Fixes [AB#494806](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/494806)


src/System Application/App/Document Sharing/src/DocumentSharing.Table.al
```diff
@@ -121,6 +121,15 @@ table 9560 "Document Sharing"
...
             InitValue = Ask;
         }
+
+        /// <summary>
+        /// Specifies whether the action was cancelled.
+        /// </summary>
+        field(13; Cancelled; Boolean)
+        {
+            DataClassification = SystemMetadata;
+            InitValue = false;
```
NKarolak:
```suggestion
```

IhorHandziuk (dismissed):

src/System Application/App/Document Sharing/src/DocumentSharingImpl.Codeunit.al
```diff
@@ -134,7 +134,8 @@ codeunit 9561 "Document Sharing Impl."
                         Handled := false;
                         Sleep(2000); // This sleep is to ensure the OneDrive clears the lock on the file after the user saves and closes.
                         DocumentSharingCodeunit.OnGetFileContents(DocumentSharing, Handled);
-                    end;
+                    end else
+                        DocumentSharing.Cancelled := true;
```
msft-sam: I wonder if we could also get this for the share dialog too, unfortunately we don't open it as a modal so returning this information might be tricky, but I know we do get events from the dialog to know what has been selected (or not).

Could also be nice to have a state enum or something instead of the bool (that way we can just extend the enum rather than adding more fields to the rec)

On the other side, is there a reason we don't just `Error('')` here (also would be nice to emit telemetry)?

stkillen: Implementing this for the share dialog seems non-trivial without changing the flow or making the codeunit _single instance_. Once the page has been opened we cannot retrieve information back from it. Thus, keeping this PR simple and only solving it for the specific scenario required by Word Templates.

I added a state enum to make it easier to extend in the future. This will currently always default to success for any other scenario than this.

I don't believe `Error('')` here would be the correct solution. In my opinion it should be up to the consumer to react appropriately on the user cancelling. For the scenario in Word Templates` Error('')` is the correct response, but there might be other scenarios where you want to act differently.

src/System Application/App/Word Templates/src/WordTemplateImpl.Codeunit.al
```diff
@@ -609,6 +609,10 @@ codeunit 9988 "Word Template Impl."
         TempDocumentSharing.Insert();
 
         DocumentSharing.Share(TempDocumentSharing);
+
+        if TempDocumentSharing.State = TempDocumentSharing.State::Cancelled then
```
StefanSosic, line 613: Would be nice to cover also this one scenario in tests
\src\System Application\Test\Word Templates\src\*
I would say codeunit 130443 "Word Templates Test" is good fit for it

PR #465, opened 2024-01-04
[Copilot] Europe geo check addition

darjoo:
#### Summary 
1.  Move geo calls into a TryFunction. If the dotnet call fails, the geo check falls back to False as default.
2. Add Europe geo check which sets the toggle for allowing data movement if the region does not have resources. 
      - Europe region will always show the toggle and this is opt out.
      - In geo (With resources), no toggle
      - Out of geo (Without resources), toggle will show and it is opt in.

EU
image
Out of EU
image

#### Work Item(s) 
Fixes [AB#491306](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/491306) [AB#495385](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/495385)

### DIFF
--- a/src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
+++ b/src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
@@ -109,10 +109,11 @@ codeunit 7772 "Azure OpenAI Impl"
 
     local procedure CheckPrivacyNoticeState(Silent: Boolean; Capability: Enum "Copilot Capability"): Boolean
     var
+        CopilotCapabilityImpl: Codeunit "Copilot Capability Impl";
         PrivacyNotice: Codeunit "Privacy Notice";
         CopilotNotAvailable: Page "Copilot Not Available";
-        ALCopilotFunctions: DotNet ALCopilotFunctions;
         WithinGeo: Boolean;
+        WithinEuropeGeo: Boolean;
     begin
         case PrivacyNotice.GetPrivacyNoticeApprovalState(CopilotCapabilityImpl.GetAzureOpenAICategory(), false) of
             Enum::"Privacy Notice Approval State"::Agreed:
@@ -129,7 +130,8 @@ codeunit 7772 "Azure OpenAI Impl"
             else begin
                 // Privacy notice not set, we will not cross geo-boundries
                 if not Silent then begin
-                    WithinGeo := ALCopilotFunctions.IsWithinGeo() or ALCopilotFunctions.IsEuropeGeo();
+                    CopilotCapabilityImpl.CheckGeo(WithinGeo, WithinEuropeGeo);
+                    WithinGeo := WithinGeo or WithinEuropeGeo;
                     if not WithinGeo then begin
                         CopilotNotAvailable.SetCopilotCapability(Capability);
                         CopilotNotAvailable.Run();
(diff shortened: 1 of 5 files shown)


EvgenijKorovin (approved):

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
```diff
@@ -127,13 +127,15 @@ codeunit 7772 "Azure OpenAI Impl"
                     exit(false);
                 end;
             else begin
-                WithinGeo := ALCopilotFunctions.IsWithinGeo();
                 // Privacy notice not set, we will not cross geo-boundries
-                if not Silent then
+                if not Silent then begin
+                    CopilotCapabilityImpl.CheckGeo(WithinGeo, WithinEuropeGeo);
+                    WithinGeo := WithinGeo or WithinEuropeGeo;
```
msft-sam, line 133: I think it's fine to just check `WithinGeo`, if we are `WithinGeo` it should already be true if `WithinEuropeGeo` is also true

src/System Application/App/AI/src/Copilot/CopilotAICapabilities.Page.al
```diff
@@ -54,7 +53,7 @@ page 7775 "Copilot AI Capabilities"
             {
                 ShowCaption = false;
                 InstructionalText = 'Generative AI capabilities are deactivated because Azure OpenAI Service is not available in your region. By allowing data movement, you agree to data being stored and processed by the Azure OpenAI Service outside of your environment''s geographic region or compliance boundary.';
-                Visible = not InGeo;
+                Visible = not WithinGeo;
 
                 field(DataMovement; AllowDataMovement)
```
msft-sam, line 58: Should we still show this if we are within Europe geo?
We might need a new group though because the instructional text doesn't really apply for the Europe case.
e.g. 
`WithinEuropeGeo` = Show toggle (`WithinGeo` will be true, so default value is true)
`WithinGeo && !WithinEuropeGeo` = Show `AlwaysConnected` group
`!WithinGeo` = Show `NotAlwaysConnected` group (and instructional text talks about not in region)

darjoo: I don't think we need to have a separate group, it still applies. As it's the compliance boundary that we're checking. 
For Europe, if a country returns `WithinGeo=true` that would mean the resources we need are within the compliance boundary. 
Unless we're saying for all NEU/WEU (by compliance boundaries), DE etc are separate, we'll always show this separate text + the toggle. Mike is OOF, so this decision can't be made atm. I'll leave this PR open till he's back for us to discuss. I'm on the side that this is fine, but he may propose otherwise.

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
darjoo: Not necessarily true. If Europe doesn't have a certain resource, `WithinGeo` would be false.
Very unlikely to happen of course.

juliamakulec (dismissed):

EvgenijKorovin (approved):

JesperSchulz (approved):

PR #474, opened 2024-01-09
Address upcoming PTECop rule in test enum extension

EmilDamsbo:
#### Summary 
This AL compiler issue https://github.com/microsoft/AL/issues/7547 is being addressed with the introduction of a new rule validated enum ordinal values falls within the allocated id ranges. This should fix this violation

#### Work Item(s)
Fixes [AB#487991](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/487991)


grobyns (changes requested):
This change breaks the test. I'll let you investigate on your own but ping me if you want a hint... :-P

EmilDamsbo:
> This change breaks the test. I'll let you investigate on your own but ping me if you want a hint... :-P

That's what happens when you change stuff without knowing how it works. I've ignored the rule instead.

grobyns:
> > This change breaks the test. I'll let you investigate on your own but ping me if you want a hint... :-P
> 
> That's what happens when you change stuff without knowing how it works. I've ignored the rule instead.

Your fix wasn't bad, the only thing you had to do was go the test page and update the enum ordinals in the AutoFormat property to match... don't disable the rule, this is a good change... you just missed 2 lines...

EmilDamsbo:
> > > This change breaks the test. I'll let you investigate on your own but ping me if you want a hint... :-P
> > 
> > 
> > That's what happens when you change stuff without knowing how it works. I've ignored the rule instead.
> 
> Your fix wasn't bad, the only thing you had to do was go the test page and update the enum ordinals in the AutoFormat property to match... don't disable the rule, this is a good change... you just missed 2 lines...

That's not really true though, it would also be affecting the

EmilDamsbo:
Closed by mistake

grobyns (approved):

JesperSchulz (approved):

PR #479, opened 2024-01-11
Allow apostrophes in page field names for edit in excel

DenLilleMand:
#### Summary 

The issue

Edit-In Excel module does not externalize page field names the same way the platform does, when the field name contains an apostrophe.
For a page field defined as:
```
field("new vendor's name"; rec."New Vendor's Name")
{
	ApplicationArea = All;
}
```

The platform will externalize this name as `new_vendor_x0027_s_name` while the Edit-In Excel module will externalize it as `new_vendor_s_name`.

Expected behavior

Within reason the platform and the Edit-In Excel module should externalize fields the same way to avoid discrepancies between the API metadata and the Edit-In Excel field bindings.

Steps to reproduce

Create a page field with an apostrophe in its name on a card page, then try to export its corresponding list page using Edit-In Excel and observe the error dialog in the Excel side pane where add-ins are loaded.

#### Work Item(s) 
Fixes [AB#423505](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/423505/)

### DIFF
--- a/src/System Application/App/Edit in Excel/src/EditinExcelImpl.Codeunit.al
+++ b/src/System Application/App/Edit in Excel/src/EditinExcelImpl.Codeunit.al
@@ -15,7 +15,7 @@ using System.Reflection;
 
 codeunit 1482 "Edit in Excel Impl."
 {
-    Access = Internal;
+    Access = Public;
     InherentEntitlements = X;
     InherentPermissions = X;
 
@@ -46,6 +46,7 @@ codeunit 1482 "Edit in Excel Impl."
         DialogTitleTxt: Label 'Export';
         ExcelFileNameTxt: Text;
         XmlByteEncodingTok: Label '_x00%1_%2', Locked = true;
+        XmlByteEncoding2Tok: Label '%1_x00%2_%3', Locked = true;
 
     procedure EditPageInExcel(PageCaption: Text[240]; PageId: Integer; EditinExcelFilters: Codeunit "Edit in Excel Filters"; FileName: Text)
     var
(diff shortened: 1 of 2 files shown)


src/System Application/App/Edit in Excel/src/EditinExcelImpl.Codeunit.al
```diff
@@ -729,15 +733,22 @@ codeunit 1482 "Edit in Excel Impl."
...
-                if CurrentPosition > 1 then begin
-                    if ConvertedName[CurrentPosition - 1] = '_' then begin
-                        ConvertedName := DelStr(ConvertedName, CurrentPosition, 1);
-                        CurrentPosition -= 1;
+            if ConvertedName[CurrentPosition] = '''' then begin
+                ByteValue := Convert.ToByte(ConvertedName[CurrentPosition]);
+                StartStr := CopyStr(ConvertedName, 1, CurrentPosition - 1);
+                EndStr := CopyStr(ConvertedName, CurrentPosition + 1);
+                ConvertedName := StrSubstNo(XmlByteEncoding2Tok, StartStr, Convert.ToString(ByteValue, 16), EndStr);
+                CurrentPosition += 6;
```
encimita, line 742: Maybe leave a comment to explain the number, something like 
// difference in length between '_' and `_x00nn_`
or
// length of `_x00nn_` minus one that will be added later

or similar

encimita (dismissed):

mazhelez (approved):

encimita (approved):

PR #495, opened 2024-01-19
#437 Expose an event to skip suite errors

### LINKED ISSUE #437: [BC Idea]: BCPT Handled event in AddLogEntry
### BC Idea Link

https://experience.dynamics.com/ideas/idea/?ideaid=15563e0c-b492-ee11-a81c-000d3a7e6e50

### Description

[In codeunit 149005 "BCPT Line"](https://github.com/microsoft/BCApps/blob/main/src/Tools/Performance%20Toolkit/App/src/BCPTLine.Codeunit.al) , there is the function:

```
 internal procedure AddLogEntry(var BCPTLine: Record "BCPT Line"; Operation: Text; ExecutionSuccess: Boolean; Message: Text; NumSQLStatements: Integer; StartTime: DateTime; EndTime: Datetime)
```

I would like that we put a handled integration event in the beginning of this function.

```
OnBeforeAddLogEntry(BCPTLine, Operation, ExecutionSuccess, Message, handled);

    if Handled then
      exit;
...
// Event code (wrong copy/paste in bc idea)

    local procedure OnBeforeAddLogEntry(var BCPTLine: Record "BCPT Line"; var Operation: Text; var ExecutionSuccess: Boolean; var Message: Text; var Handled: Boolean)
    begin
    end;

```

Reason:
Many of us run BCPT flow with random customer data to get a better representation of the data that is used.
The problem with doing that is sometimes some items/customers/vendors have very special setting that makes a confirmation or message popup. In order to tackle that we do message/confirmation handlers.

But since this does not always occur we get an error for every time we didn't use the handler. An event like this can give us the possibility to skip that error if we want it.

I have seen some argument that is up to the developer to create clean test cases where the handlers are always used. To some extent that is valid. But the other side of the coin is that we have to do a lot of code cloning to make those test distinct. But the worst part is that will pollute the the actual performance test. Because now we have to filter out the records which has these particular settings. Settings that wasn't meant to be filtered upon. This will lead to that the slowest functions are those that prepares the data in the test and that makes the performance test quite useless since we can't measure the realistic flows.

NAVFreak:
#### Summary 
This PR exposes an event to make it possible to skip an unwanted error when running a BCPT suit.
The idea was to expose an event i BCPTLine codeunit.

However, during the implementation I discovered that [BCPTLine codeunit has access internal] ([https://github.com/microsoft/BCApps/blob/faeca5adabaa2e2b7644d744e21f9a1f97d1d971/src/Tools/Performance%20Toolkit/App/src/BCPTLine.Codeunit.al#L12](https://eur02.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fmicrosoft%2FBCApps%2Fblob%2Ffaeca5adabaa2e2b7644d744e21f9a1f97d1d971%2Fsrc%2FTools%2FPerformance%2520Toolkit%2FApp%2Fsrc%2FBCPTLine.Codeunit.al%23L12&data=05%7C02%7CMarcus.Nordlund%40navet.se%7C20e35a07443c45d23d3408dc187791b1%7C56bbfa1c071347e3957f22760ff6118d%7C0%7C0%7C638412149035646962%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=ojrRAsw6wNK0d%2BBlFSVyE4U3I0p6fyBGZXkiD%2FHjrkU%3D&reserved=0)).

To solve this I created a wrapper subscriber in BCPTTestSuite and calls another event that is public.

I hope this workaround is acceptable.

#### Work Item(s) 
Fixes #437

Fixes [AB#498190](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/498190)


src/Tools/Performance Toolkit/App/src/BCPTLine.Codeunit.al
```diff
@@ -346,4 +351,11 @@ codeunit 149005 "BCPT Line"
         Parm := format(FldRef.Value, 0, 9);
         exit(true);
     end;
+
+    local procedure OnBeforeAddLogEntry(var BCPTLine: Record "BCPT Line"; var Operation: Text; var ExecutionSuccess: Boolean; var Message: Text; var Handled: Boolean)
```
grobyns: it's not an event so the name is misleading. If its sole purpose is to raise the event, then I would call it RaiseOnBlaBlaBla.
but I still wonder why you're not raising the event directly...

src/Tools/Performance Toolkit/App/src/BCPTLine.Codeunit.al
```diff
@@ -346,4 +351,11 @@ codeunit 149005 "BCPT Line"
...
+    var
+        BCPTTestSuite: Codeunit "BCPT Test Suite";
+    begin
+        BCPTTestSuite.OnBeforeBCPTLineAddLogEntryWrapper(BCPTLine, Operation, ExecutionSuccess, Message, Handled);
```
grobyns: not your issue but what does the 'Wrapper' bit on the end mean?

grobyns: oh, it is your issue... damn... yeah, what extra info does the word wrapper convey here?

src/Tools/Performance Toolkit/App/src/BCPTTestSuite.Codeunit.al
```diff
@@ -411,4 +411,36 @@ codeunit 149006 "BCPT Test Suite"
...
+    /// <summary>
+    /// Wrapper for OnBeforeBCPTLineAddLogEntry since it is not possible subscribe to an event in a object with scope internal
+    /// This one converts BCPTLine to variables instead of a record.
+    /// </summary>
+    /// <param name="BCPTLine"></param>
+    /// <param name="Operation"></param>
+    /// <param name="ExecutionSuccess"></param>
+    /// <param name="Message"></param>
+    /// <param name="Handled"></param>
+    internal procedure OnBeforeBCPTLineAddLogEntryWrapper(var BCPTLine: Record "BCPT Line"; var Operation: Text; var ExecutionSuccess: Boolean; var Message: Text; var Handled: Boolean)
```
grobyns:
```suggestion
    internal procedure RaiseOnBeforeBCPTLineAddLogEntry(var BCPTLine: Record "BCPT Line"; var Operation: Text; var ExecutionSuccess: Boolean; var Message: Text; var Handled: Boolean)
```

grobyns: I don't understand the purpose of this method. Why would an outside caller need to raise the event directly and out of context?

grobyns (dismissed):
I don't see why you wouldn't just raise the event directly on the facade. You don't need the 'wrapper' functions and without them, you also don't have a confusing naming issue...

src/Tools/Performance Toolkit/App/src/BCPTLine.Codeunit.al
```diff
@@ -164,7 +164,12 @@ codeunit 149005 "BCPT Line"
     var
         BCPTLogEntry: Record "BCPT Log Entry";
         BCPTRoleWrapperImpl: Codeunit "BCPT Role Wrapper"; // single instance
+        Handled: Boolean;
     begin
+        OnBeforeAddLogEntry(BCPTLine, Operation, ExecutionSuccess, Message, handled);
```
TKapitan: Standard in BC for these variables is IsHandled, not Handled.

src/Tools/Performance Toolkit/App/src/BCPTLine.Codeunit.al
NAVFreak: Yeah, this is misleading but I have a reason for that.
The ENTIRE object has access = internal. 
Why sole idea was to raise this event directly! But since the object has access = internal you can't subscribe to the event from another app which is the purpose of the entire change.

Hence why I did this wrapper thingy workaround. Another solution would have been to do the event directly, subscribe to it in "BCPT Tes Suite" and then call another event which can be reached from other apps since "BCPT Test Suite" does not have internal accesss. But I thought that option would be confusing for people who looks for events and reads the code. There would be no direct calling thread to the event in "BCPT Test  Suite" which takes away the readability of the change.

src/Tools/Performance Toolkit/App/src/BCPTLine.Codeunit.al
```diff
@@ -164,7 +164,12 @@ codeunit 149005 "BCPT Line"
...
+
+        if Handled then
+            exit;
```
TKapitan: You also need to handle the return value from this procedure; now you are returning always false.

src/Tools/Performance Toolkit/App/src/BCPTLine.Codeunit.al
NAVFreak: I don't know if the wrapper is proper naming convention. It is just tool to make an event available for other apps and for the programmer who reads the code can follow the thread easily to which event he/she should really subscribe on.

src/Tools/Performance Toolkit/App/src/BCPTTestSuite.Codeunit.al
```diff
@@ -411,4 +411,36 @@ codeunit 149006 "BCPT Test Suite"
...
+    begin
+        OnBeforeBCPTLineAddLogEntry(BCPTLine."BCPT Code", BCPTLine."Codeunit ID", BCPTLine.Description, Operation, ExecutionSuccess, Message, Handled);
+    end;
+
+    /// <summary>
+    /// This event is raised before a log entry is added to the BCPT Line table.
+    /// It can be used to skip errors which are not relevant for the test suite. Like unused handler functions.
+    /// </summary>
+    /// <param name="SuiteCode"></param>
+    /// <param name="CodeunitId"></param>
```
TKapitan: Add explanation/info for all parameters exposed in publisher

TKapitan (changes requested):

src/Tools/Performance Toolkit/App/src/BCPTTestSuite.Codeunit.al
NAVFreak: As I mentioned and try to document, you CAN'T subscribe to an event from ANOTHER app if the event is located in an object with access = internal. Thats why I wrote in the PR that the original plan with a direct raising of event doesn't work and why I had to to this weird workaround.

TKapitan: I'm not sure whether it's currently an acceptable solution, but I saw many events in the base app/system apps that are internal. If it's allowed, I believe it would be better than having a wrapper method.

grobyns: I think you are mixing up two things. No, you cannot subscribe to an event in an internal codeunit, but your event does not have to be on an internal codeunit. Almost none of our events are. The event publisher is placed on the public facade codeunit (which is where you have it too) as an internal method (and yes, you can subscribe to that). 
Now, you can call that method directly from the internal object and everyone can subscribe to it. You do not need any of the wrapper methods you created.

src/Tools/Performance Toolkit/App/src/BCPTLine.Codeunit.al
NAVFreak: I don't understand what you mean. Function  AddLogEntry does not have a return value.
Or do you mean Handled (isHandled) ? That should be taken care of the subscriber to the event in BCPTTestSuite.Codeunit.

NAVFreak:
@TKapitan , @grobyns  I have now done the desired changes that you wanted

src/Tools/Performance Toolkit/App/src/BCPTLine.Codeunit.al
TKapitan: Sorry, my bad! 

image

I somehow missed that the exit true/false lines are from completely different procedure 200 lines below :)

TKapitan: It seems I can't mark it as resolved, please close it.

src/Tools/Performance Toolkit/App/src/BCPTTestSuite.Codeunit.al
```diff
@@ -411,4 +411,20 @@ codeunit 149006 "BCPT Test Suite"
...
+    /// <summary>
+    /// This event is raised before a log entry is added to the BCPT Line table.
+    /// It can be used to skip errors which are not relevant for the test suite. Like unused handler functions.
+    /// </summary>
+    /// <param name="SuiteCode">The test suite code</param>
+    /// <param name="CodeunitId">The id of the test codunit that is being run</param>
+    /// <param name="Description">Description of the test on the "BCPT Line"</param>
+    /// <param name="Operation">Which operation that is currently executed</param>
+    /// <param name="ExecutionSuccess">If the test succedded or not</param>
+    /// <param name="Message"></param>
```
grobyns: missing a doc text on the message param

src/Tools/Performance Toolkit/App/src/BCPTLine.Codeunit.al
```diff
@@ -346,4 +351,11 @@ codeunit 149005 "BCPT Line"
         Parm := format(FldRef.Value, 0, 9);
         exit(true);
     end;
+
+    local procedure RaiseOnBeforeBCPTLineAddLogEntry(var BCPTLine: Record "BCPT Line"; var Operation: Text; var ExecutionSuccess: Boolean; var Message: Text; var IsHandleld: Boolean)
```
grobyns: I still do no understand why you have this method. You don't need it.

src/Tools/Performance Toolkit/App/src/BCPTTestSuite.Codeunit.al
```diff
@@ -411,4 +411,20 @@ codeunit 149006 "BCPT Test Suite"
...
+    /// <param name="IsHandled">Handled pattern parameter</param>
+    [IntegrationEvent(false, false)]
+    procedure OnBeforeBCPTLineAddLogEntry(SuiteCode: Code[10]; CodeunitId: Integer; Description: Text; var Operation: Text; var ExecutionSuccess: Boolean; var Message: Text; var IsHandled: Boolean)
```
grobyns: why are any of these parameter var? As far as i can tell, the only way it could have an effect is if you changed the values but did not set IsHandled. If you did that you would change what is logged. If you set IsHandled you would simply not log anything and changing the values would not have any effect. 

If what you want is to change the values that are written to the log then I think there should be an indication that the log was altered by a subscriber.

src/Tools/Performance Toolkit/App/src/BCPTLine.Codeunit.al
```diff
@@ -164,7 +164,12 @@ codeunit 149005 "BCPT Line"
     var
         BCPTLogEntry: Record "BCPT Log Entry";
         BCPTRoleWrapperImpl: Codeunit "BCPT Role Wrapper"; // single instance
+        IsHandled: Boolean;
     begin
+        RaiseOnBeforeBCPTLineAddLogEntry(BCPTLine, Operation, ExecutionSuccess, Message, IsHandled);
```
grobyns:
```suggestion
        BCPTTestSuite.OnBeforeBCPTLineAddLogEntry(BCPTLine."BCPT Code", BCPTLine."Codeunit ID", BCPTLine.Description, Operation, ExecutionSuccess, Message, IsHandled);
```

src/Tools/Performance Toolkit/App/src/BCPTLine.Codeunit.al
grobyns: not that it matters, but there's also a typo in IsHandled

src/Tools/Performance Toolkit/App/src/BCPTTestSuite.Codeunit.al
grobyns:
```suggestion
    procedure OnBeforeBCPTLineAddLogEntry(SuiteCode: Code[10]; CodeunitId: Integer; Description: Text; Operation: Text; ExecutionSuccess: Boolean; Message: Text; var IsHandled: Boolean)
```

NAVFreak: I have been running this code local for while but it was not until now I realised that that skipping this function is not a good scenario because what we want to achieve is to change from a "faulty log" to a correct log. Which means changing the values without setting IsHandled. Otherwise we won't receive a complete log for that scenario 😏

Example: 
Three scenarios:
1. Is having a unused handled event which gives error at the end of execution
2. Is having a handled unused handled event which wont't give an error at the end of execution
3. A normal test that completes successfully for comparison

The event code:
image

Result:
image

Edit:
I added a boolean which the subscriber can let the publisher know if something is changed

src/Tools/Performance Toolkit/App/src/BCPTTestSuite.Codeunit.al
NAVFreak: Fixed

src/Tools/Performance Toolkit/App/src/BCPTLine.Codeunit.al
NAVFreak: Removed metod

src/Tools/Performance Toolkit/App/src/BCPTTestSuite.Codeunit.al
NAVFreak: Want the var:s as mentioned earlier

grobyns: Ok. Do you still need the IsHandled to skip writing the log completely? Wouldn't you always want a log entry? 

Regarding the ValuesAreChanged Boolean, I think that should be automatic. It should not require anything from the subscriber. So, you would need to save the values before the event, compare after the event, and if they are changed then write that to the log. You might consider passing the original values to the event as non-var and the new ones as var. That way if there are multiple subscribers they can check whether the log was already modified when they were called.

You could use a boolean, but given that you need to old and new values anyway to check, you could just add fields to the log entry like 
"Orig. Operation"
"Orig. Message"
"Orig. Exec. Status"

now the subscriber can edit the values and all will be recorded so you'll be able to track the changes in the log.

NAVFreak: Have added these parameters, plus initialization and handling of them.

src/Tools/Performance Toolkit/App/src/BCPTLine.Codeunit.al
```diff
@@ -238,6 +239,23 @@ codeunit 149005 "BCPT Line"
...
+    begin
+        if ModifiedExecutionSuccess <> ExecutionSuccess then
+            ExecutionSuccess := ModifiedExecutionSuccess;
+        if ModifiedMessage <> Message then
+            Message := ModifiedMessage;
+        if ModifiedOperation <> Operation then
+            Operation := ModifiedOperation;
+    end;
+
+    local procedure InitValuesSubscriberCanModify(var Operation: Text; var ExecutionSuccess: Boolean; var Message: Text; var ModifiedExecutionSuccess: Boolean; var ModifiedMessage: Text; var ModifiedOperation: Text)
```
grobyns:
```suggestion
    local procedure InitValuesSubscriberCanModify(Operation: Text; ExecutionSuccess: Boolean; Message: Text; var ModifiedExecutionSuccess: Boolean; var ModifiedMessage: Text; var ModifiedOperation: Text)
```

src/Tools/Performance Toolkit/App/src/BCPTLine.Codeunit.al
```diff
@@ -164,13 +164,14 @@ codeunit 149005 "BCPT Line"
...
         BCPTTestSuite: Codeunit "BCPT Test Suite";
+        ModifiedMessage, ModifiedOperation : Text;
+        ModifiedExecutionSuccess: Boolean;
     begin
-        BCPTTestSuite.OnBeforeBCPTLineAddLogEntry(BCPTLine."BCPT Code", BCPTLine."Codeunit ID", BCPTLine.Description, Operation, ExecutionSuccess, Message, IsHandled, ValuesAreChanged);
-        if IsHandled then
-            exit;
+        InitValuesSubscriberCanModify(Operation, ExecutionSuccess, Message, ModifiedExecutionSuccess, ModifiedMessage, ModifiedOperation);
+        BCPTTestSuite.OnBeforeBCPTLineAddLogEntry(BCPTLine."BCPT Code", BCPTLine."Codeunit ID", BCPTLine.Description, Operation, ExecutionSuccess, Message, ModifiedOperation, ModifiedExecutionSuccess, ModifiedMessage);
+        UpdateValuesFromSubscribers(Operation, ExecutionSuccess, Message, ModifiedExecutionSuccess, ModifiedMessage, ModifiedOperation);
```
grobyns: why not (simply)
"Original Operation" := Operation;
"Original Message" := Message;
"Original Execution Success" := ExecutionSuccess;
BCPTTestSuite.OnBeforeBCPTLineAddLogEntry(BCPTLine."BCPT Code", BCPTLine."Codeunit ID", BCPTLine.Description, "Original Operation", "Original Execution Success", "Original Message", Operation, ExecutionSuccess, Message);

grobyns: this does the same and you do not need to compare the values afterwards.

src/Tools/Performance Toolkit/App/src/BCPTLine.Codeunit.al
```diff
@@ -164,13 +164,14 @@ codeunit 149005 "BCPT Line"
     var
         BCPTLogEntry: Record "BCPT Log Entry";
         BCPTRoleWrapperImpl: Codeunit "BCPT Role Wrapper"; // single instance
-        IsHandled: Boolean;
         ValuesAreChanged: Boolean;
         BCPTTestSuite: Codeunit "BCPT Test Suite";
+        ModifiedMessage, ModifiedOperation : Text;
```
grobyns: You should add fields to the BCPT Log Entry table instead and use those.

src/Tools/Performance Toolkit/App/src/BCPTLine.Codeunit.al
NAVFreak: Bah! 🤦‍♂️ That is the third time I make change, test, it doesn't work, fix the error and then get feedback that would have been better to refactor the change instead of fixing the error. I need star thinking of the complete solution instead of just fixing the error 🤪
Fix done

src/Tools/Performance Toolkit/App/src/BCPTLine.Codeunit.al
NAVFreak: Of course... 🤦‍♂️

henrikfrovst (dismissed):

src/Tools/Performance Toolkit/App/src/BCPTLine.Codeunit.al
```diff
@@ -231,6 +235,13 @@ codeunit 149005 "BCPT Line"
             Dimensions)
     end;
 
+    local procedure InitValuesSubscriberCanModify(var Operation: Text; var ExecutionSuccess: Boolean; var Message: Text; var ModifiedExecutionSuccess: Boolean; var ModifiedMessage: Text; var ModifiedOperation: Text)
```
ChethanT: This looks like an unused method. Will you delete this?

src/Tools/Performance Toolkit/App/src/BCPTLine.Codeunit.al
```diff
@@ -164,7 +164,11 @@ codeunit 149005 "BCPT Line"
     var
         BCPTLogEntry: Record "BCPT Log Entry";
         BCPTRoleWrapperImpl: Codeunit "BCPT Role Wrapper"; // single instance
+        ValuesAreChanged: Boolean;
+        BCPTTestSuite: Codeunit "BCPT Test Suite";
     begin
+        BCPTTestSuite.OnBeforeBCPTLineAddLogEntry(BCPTLine."BCPT Code", BCPTLine."Codeunit ID", BCPTLine.Description, Operation, ExecutionSuccess, Message, Operation, ExecutionSuccess, Message);
```
ChethanT: From you discussion with Gert, I can see that the solution we are looking for is something like the below

1. Add fields to "BCPT Log Entry" table to capture the original values
        _field(20; "Orig. Operation"; Text[100])
        {
              Caption = 'Orig. Operation';
        }
        field(21; "Orig. Status"; Option)
        {
              Caption = 'Orig. Status';
              OptionMembers = Success,Error;
        }
        field(22; "Orig. Message"; Text[250])
        {
              Caption = 'Orig. Message';
        }
        field(23; "Log was Modified"; Boolean)
        {
              Caption = 'Log was Modified';
        }_

2. Change AddLogEntry in "BCPT Line" codeunit to capture both original values and modified values
_internal procedure AddLogEntry(var BCPTLine: Record "BCPT Line"; Operation: Text; ExecutionSuccess: Boolean; Message: Text; NumSQLStatements: Integer; StartTime: DateTime; EndTime: Datetime)
    var
        BCPTLogEntry: Record "BCPT Log Entry";
        BCPTRoleWrapperImpl: Codeunit "BCPT Role Wrapper"; // single instance
        ValuesAreChanged: Boolean;
        BCPTTestSuite: Codeunit "BCPT Test Suite";
        ModifiedOperation: Text;
        ModifiedExecutionSuccess: Boolean;
        ModifiedMessage: Text;
        EntryWasModified: Boolean;
    begin
        ModifiedOperation := Operation;
        ModifiedExecutionSuccess := ExecutionSuccess;
        ModifiedMessage := Message;
        BCPTTestSuite.OnBeforeBCPTLineAddLogEntry(BCPTLine."BCPT Code", BCPTLine."Codeunit ID", BCPTLine.Description, Operation, ExecutionSuccess, Message, ModifiedOperation, ModifiedExecutionSuccess, ModifiedMessage);
        if (Operation <> ModifiedOperation) or (ExecutionSuccess <> ModifiedExecutionSuccess) or (Message <> ModifiedMessage) then
            EntryWasModified := true;_

3. Store both original and modified values in the log entry table
        _BCPTLogEntry.Operation := copystr(ModifiedOperation, 1, MaxStrLen(BCPTLogEntry.Operation));
        BCPTLogEntry."Orig. Operation" := copystr(Operation, 1, MaxStrLen(BCPTLogEntry."Orig. Operation"));
        if ModifiedExecutionSuccess then
            BCPTLogEntry.Status := BCPTLogEntry.Status::Success
        else begin
            BCPTLogEntry.Status := BCPTLogEntry.Status::Error;
            BCPTLogEntry."Error Call Stack" := CopyStr(GetLastErrorCallStack, 1, MaxStrLen(BCPTLogEntry."Error Call Stack"));
        end;
        if ExecutionSuccess then
            BCPTLogEntry."Orig. Status" := BCPTLogEntry.Status::Success
        else
            BCPTLogEntry."Orig. Status" := BCPTLogEntry.Status::Error;
        BCPTLogEntry.Message := copystr(ModifiedMessage, 1, MaxStrLen(BCPTLogEntry.Message));
        BCPTLogEntry."Orig. Message" := copystr(Message, 1, MaxStrLen(BCPTLogEntry."Orig. Message"));_

ChethanT (changes requested):

src/Tools/Performance Toolkit/App/src/BCPTLine.Codeunit.al
NAVFreak: No, not all.

The discussion is that I want an event to change some of the parameters that comes into function AddLogEntry.
But Gert wanted a way for other subscribers to know if the value has been changed hence why we added the "Original value" parameter.
Example, the operation and the "Orig. Operation" parameters.
Subscribers can now check if the Operation parameter has already been changed by other subscribers by doing following code:
`If "Orig. Operation" <> Operation then ....`
image

src/Tools/Performance Toolkit/App/src/BCPTLine.Codeunit.al
NAVFreak: You are right, apparently my code analyzers didn't caught that 🤔

src/Tools/Performance Toolkit/App/src/BCPTLine.Codeunit.al
ChethanT: Wouldn't the changes I proposed resolve the concerns? Subscribers have access to original and modified content and content is also stored in the BCPT Log Entry table for future auditing if needed.

NAVFreak: Well, yes. 
I just don't see the need for it to be stored in table. But overall it is a more correct way if someone would be interested in that audit.

NAVFreak: Do you want that as well or can we skip it?

ChethanT: Yes please. It is important to log the original result as the subscriber can completely overwrite the origin result without a trace:-).

NAVFreak: @ChethanT 
I'm not sure about this though?
```
if (Operation <> ModifiedOperation) or (ExecutionSuccess <> ModifiedExecutionSuccess) or (Message <> ModifiedMessage) then
EntryWasModified := true;

```
Since we are not using `EntryWasModified`.

Edit: Later with a sharper I see what you want. I have added so this field is also being set in table.

NAVFreak:
@ChethanT do you think this can go into next release? Meaning do you have time to merge it before the break? 😬

src/Tools/Performance Toolkit/App/src/BCPTLine.Codeunit.al
```diff
@@ -173,18 +186,25 @@ codeunit 149005 "BCPT Line"
         BCPTLogEntry."BCPT Line No." := BCPTLine."Line No.";
         BCPTLogEntry.Version := BCPTHeader.Version;
         BCPTLogEntry."Codeunit ID" := BCPTLine."Codeunit ID";
-        BCPTLogEntry.Operation := copystr(Operation, 1, MaxStrLen(BCPTLogEntry.Operation));
+        BCPTLogEntry.Operation := copystr(ModifiedOperation, 1, MaxStrLen(BCPTLogEntry.Operation));
```
ChethanT:
```suggestion
        BCPTLogEntry.Operation := CopyStr(ModifiedOperation, 1, MaxStrLen(BCPTLogEntry.Operation));
```

src/Tools/Performance Toolkit/App/src/BCPTLine.Codeunit.al
```diff
@@ -173,18 +186,25 @@ codeunit 149005 "BCPT Line"
...
+        BCPTLogEntry."Orig. Operation" := copystr(Operation, 1, MaxStrLen(BCPTLogEntry."Orig. Operation"));
```
ChethanT:
```suggestion
        BCPTLogEntry."Orig. Operation" := CopyStr(Operation, 1, MaxStrLen(BCPTLogEntry."Orig. Operation"));
```

ChethanT (changes requested):

ChethanT (dismissed):

ChethanT (dismissed):

JesperSchulz:
We've got approval from @ChethanT, but the build is failing. Seems just to be some unused variables though, which should be easy to fix. One more round, and we can get this PR merged 😊

ChethanT (approved):

JesperSchulz (approved):

PR #500, opened 2024-01-22
Update issue template for bug creation

aholstrup1:
#### Summary 
Add a checkbox to verify that there is an intention to follow up the issue with a PR. If there isn't, redirect to other channels.
image

#### Work Item(s) 
[AB#497509](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/497509)

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -47,4 +47,15 @@ body:
       description: If needed add any other context about the problem here or include screenshots
       placeholder: Additional context, screenshots, log output, etc.
     validations:
-      required: true
+      required: false
+  - type: checkboxes
+    id: terms
+    attributes:
+      label: I intend on providing a fix for this issue myself
+      description: | 
+        If you plan to provide a fix for this issue, please check the box below. 
+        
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -47,4 +47,15 @@ body:
       description: If needed add any other context about the problem here or include screenshots
       placeholder: Additional context, screenshots, log output, etc.
     validations:
-      required: true
+      required: false
```
JesperSchulz, line 50: We should get this reviewed by a writer. I'll ping Brent :-)

brentholtorf: How about the following:

We change "Report and address a bug" header to "Fix a bug."

Label: I will provide a fix for a bug.
Description: To provide a fix for a bug, select this checkbox and then Get started. Thanks for contributing.

aholstrup1: @brentholtorf I updated the template with your suggestions. 

image

mazhelez (approved):

PR #503, opened 2024-01-22
[UriBuilder] Add function to add OData query parameters with preserved $ signs

encimita:
#### Summary
This PR tackles an issue with URL encoding of OData parameters.
Some servers, such as AppSource marketplace, don't interpret the encoded OData query parameter keys (`%24filter`) as equivalent to their unencoded form (`$filter`). Technically, this is by design given the encoding rules that Business Central uses, but this fix enables the caller of the AL code to decide whether to preserve the `$` signs or not when adding URL query parameters.

#### Work Item(s)
Fixes [AB#497036](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/497036)


MonicaPAhuja (approved):

src/System Application/Test/URI/src/UriBuilderQueryTest.Codeunit.al
```diff
@@ -314,18 +314,66 @@ codeunit 135072 "Uri Builder Query Test"
...
+        Assert.AreEqual('https://microsoft.com/' // Initial parameters are overwritten
+            + '?%C3%A8=%C3%A9AddedNotEncoded&%25C3%25A8=%25C3%25A9AddedEncoded' // Added parameters (the one that was already encoded should be double encoded)
             + '&moreGarbledStuff=%26%2F%5C%27%22%3F%3F%3F%2520' // Ensure escape and special characters are encoded
             + '&%EA%B3%A0%EC%96%91%EC%9D%B4=%3A30%26%EA%B3%A0%EC%96%91%EC%9D%B4', // Ensure character that use 3 bytes are also correctly encoded
             Uri.GetAbsoluteUri(), 'Unexpected URL.');
     end;
 
+    [Test]
+    [Scope('OnPrem')]
+    procedure TestExpectedODataEncoding_ODataParameters()
```
SBalslev, line 355: Can you add a test where you add a flag after adding a $ param?

src/System Application/App/URI/src/UriBuilderImpl.Codeunit.al
```diff
@@ -219,7 +229,10 @@ codeunit 3062 "Uri Builder Impl."
         foreach CurrentKey in KeysWithValueList.Keys() do begin
             KeysWithValueList.Get(CurrentKey, CurrentValues);
             foreach CurrentValue in CurrentValues do
-                FinalQuery += '&' + Uri.EscapeDataString(CurrentKey) + '=' + Uri.EscapeDataString(CurrentValue);
+                if UseODataEncoding then
```
grobyns: I'd get rid of this if my moving it down into the helper :-)

grobyns (dismissed):

src/System Application/Test/URI/src/UriBuilderQueryTest.Codeunit.al
```diff
@@ -314,18 +314,92 @@ codeunit 135072 "Uri Builder Query Test"
...
+        UriBuilder.AddODataQueryParameter('$expand', 'Products($filter=DiscontinuedDate eq null)');
+        UriBuilder.AddODataQueryParameter('moreGarbledStuff😊', '&/\''"*!???%20');
+        UriBuilder.AddQueryFlag('$newschemaversion');
+
+        // [Then] The resulting URI has encoded query parameters, except the $ sign in the parameter name
+        UriBuilder.GetUri(Uri);
+        Assert.AreEqual('https://microsoft.com/?%24top=33&%24skip=41' // Initial parameters
+            + '&%24filter=Name%20eq%20%27%26Contoso%27' // Filter
+            + '&%C3%A8=%C3%A9NotEncoded&%C3%A8=%C3%A9Encoded' // Other initial parameters
+            + '&%24expand=Products%28%24filter%3DDiscontinuedDate%20eq%20null%29' // $ is encoded in the value
```
AndreasMoth, line 399: Wouldn't you expect this to be &$expand? Just like in the previous test? After all you added "UriBuilder.AddODataQueryParameter('$expand', 'Products($filter=DiscontinuedDate eq null)');" the exact same way as above, hence I would expect the output to be the exact same.
I noticed the problem however is that the url is regenerated every time you add something new. If you add an OData Query parameter as the last, then ALL parameters will be formatted according to OData, if the last parameter is not like in this case, then all parameters are formatted as though they are not OData. This is because we convert the entire uri to a single string and then re-encode everything whenever you add a parameter.

AndreasMoth: I also noticed the comments are the same as in the previous test, even though the result is different.

encimita: You are right, the encoding will be the one of the last used `Add*Query*()` procedure called on the codeunit instance.

This is unfortunate, but it's a consequence of how the underlying dotnet works. `HttpUtility` returns to us the unencoded version of the query, so for example we lose information on whether the original parameter key was `%24expand` or `$expand`. At that point the best we can do is honour the encoding that the current call has requested. Alternatively, we would have to not use `HttpUtility.ParseQueryString` at all, and instead implement our own query string parser (which would be a much bigger work for the gained value).

Hence, in order not to revolutionize how the library works behind the scenes, I have just updated the comments on the facade codeunit. I will, however, fix the test comments (good catch!).

src/System Application/App/URI/src/UriBuilderImpl.Codeunit.al
encimita: Moved down!

grobyns (approved):

MonicaPAhuja (approved):

PR #508, opened 2024-01-23
Turn time token to uppercase before compare

JesperSchulz:
The case expects uppercase tokens. We must turn the input filter token into uppercase before we do the compare.

Fixes [AB#483241](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/483241)

### DIFF
--- a/src/System Application/App/Filter Tokens/src/FilterTokensImpl.Codeunit.al
+++ b/src/System Application/App/Filter Tokens/src/FilterTokensImpl.Codeunit.al
@@ -387,6 +387,7 @@ codeunit 58 "Filter Tokens Impl."
     local procedure ResolveTimeToken(var TimeFilter: Time; TimeToken: Text) Handled: Boolean
     begin
         TimeToken := DelChr(TimeToken); // Deletes all spaces in TimeToke
+        TimeToken := UpperCase(TimeToken); // Turn TimeToken into uppercase
         case TimeToken of
             NowTxt, 'NOW':
                 begin


JesperSchulz:
@NKarolak, took me a while to get to this, but here you go :-)

IhorHandziuk (approved):

NKarolak (approved):

src/System Application/App/Filter Tokens/src/FilterTokensImpl.Codeunit.al
```diff
@@ -387,6 +387,7 @@ codeunit 58 "Filter Tokens Impl."
     local procedure ResolveTimeToken(var TimeFilter: Time; TimeToken: Text) Handled: Boolean
     begin
         TimeToken := DelChr(TimeToken); // Deletes all spaces in TimeToke
```
NKarolak, line 389: If we are already here, we could also ...
```suggestion
        TimeToken := DelChr(TimeToken); // Deletes all spaces in TimeToken
```

JesperSchulz: Yeah, I saw that one... Usually I try to keep my PRs very clean and 100% related to the problem I'm trying to fix, but in this case I guess I could have made an exception ;-) Oh well...

src/System Application/App/Filter Tokens/src/FilterTokensImpl.Codeunit.al
```diff
@@ -387,6 +387,7 @@ codeunit 58 "Filter Tokens Impl."
...
+        TimeToken := UpperCase(TimeToken); // Turn TimeToken into uppercase
```
TKapitan, line 390: @JesperSchulz Well, it's Ok for the case statement, but I would probably prefer to pass the current value to the publisher (OnResolveTimeFilterToken).

I used lowercase in my blogpost in 2022 and this post had a lot of viewers :)) 

https://kepty.cz/2022/05/03/custom-filter-tokens/

TKapitan (changes requested):

src/System Application/App/Filter Tokens/src/FilterTokensImpl.Codeunit.al
TKapitan: ... and I see that we have also lowercase in two PTEs. We should probably convert the text to improve the stability, but this can break existing customizations.

TKapitan: And it's also strange that ResolveDateFilterToken has the publisher before the standard code with the handled pattern, but OnResolveTimeFilterToken is as else statement. 

For the date procedure, the event is before the conversion to uppercase.

TKapitan (changes requested):
Can break existing functionality.

mazhelez (approved):

src/System Application/App/Filter Tokens/src/FilterTokensImpl.Codeunit.al
JesperSchulz: OK, I'll fix the case only for now then.

JesperSchulz: Now it merged the PR :-P I'll create a follow up to address this right away.

PR #511, opened 2024-01-24
Address code review comments on ResolveTimeToken

JesperSchulz:
The case should not get changed in the event, as that might be breaking.

#### Work Item(s) 
Fixes [AB#483241](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/483241)


mazhelez (approved):

JesperSchulz:
Follow up PR for #508.

IhorHandziuk (approved):

JesperSchulz:
@TKapitan, I hope this change addresses your main concern in #508. I agree that we could invest more time in aligning the logic, but currently that doesn't have the highest of priorities.

darjoo (approved):

PR #516, opened 2024-01-25
#430 | Access to NewerThanDate in sent emails

### LINKED ISSUE #430: [Access Request] Sent Emails - NewerThanDate
### Describe the request

Hello everyone,
we use this BC function for opening sent e-mails from interaction log entries:

`Email.OpenSentEmails(Database::"Interaction Log Entry", Rec.SystemId);`

However, the sent emails are only displayed if they are less than one month old, because it is programmed in the OnOpenPage-trigger of the page "Sent Emails":

```
trigger OnOpenPage()
    begin
        NewerThanDate := CreateDateTime(CalcDate('<-30D>', Today()), Time());
        ...
    end;
```

We do not have access to the variable NewerThanDate, nor can we manipulate the variable using a parameter or a function.

Our ideas:

- Add a parameter for the datetime to the function Email.OpenSentEmails
- ... and/or set the flexible datetime yourself based on the creation date of the record provided to the function, so that the page is filtered to at least this datetime.
- ... and/or add an OpenSentEmails-procedure to the page "Sent Emails" and grant us access to the variable "NewerThanDate": changed to protected or via a set-function.

If more information is required, please let us know. 
Thank you very much in advance!

### Additional context

Emails are not intialy displayed if older than one month. We would like to get it more flexible.

GOB-Andrea:
Extended the access to the filter field "NewerThanDate" in the sent emails:
- new function with extended parameters for opening sent emails filtered to the source record and a specific date.
- removed the internal flag to set the date directly when opening the sent emails.
- "If" added so that the NewerThanDate is only overwritten with the "30 days filter" if it was previously empty.

Fixes #430  

Fixes [AB#497966](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/497966)

### DIFF
--- a/src/System Application/App/Email/src/Email/Sent/SentEmails.Page.al
+++ b/src/System Application/App/Email/src/Email/Sent/SentEmails.Page.al
@@ -213,6 +213,10 @@ page 8883 "Sent Emails"
         EmailConnector.ShowAccountInformation(Rec."Account Id");
     end;
 
+    /// <summary>
+    /// Set date filter for sent emails.
+    /// </summary>
+    /// <param name="NewDate">Earliest date to include sent emails from.</param>
     procedure SetNewerThan(NewDate: DateTime)
     begin
         NewerThanDate := NewDate;
@@ -223,7 +227,12 @@ page 8883 "Sent Emails"
         EmailAccountId := AccountId;
     end;
 
-    internal procedure SetRelatedRecord(TableID: Integer; SystemID: Guid)
+    /// <summary>
+    /// Set filter for related record on sent emails.
+    /// </summary>
+    /// <param name="TableID">The entity table.</param>
+    /// <param name="SystemID">A record to filter on.</param>
+    procedure SetRelatedRecord(TableID: Integer; SystemID: Guid)
     begin
         SourceTableID := TableID;
         SourceSystemID := SystemID;
--- a/src/System Application/App/Email/src/Email/Sent/SentEmails.Page.al
+++ b/src/System Application/App/Email/src/Email/Sent/SentEmails.Page.al
@@ -177,7 +177,9 @@ page 8883 "Sent Emails"
 
     trigger OnOpenPage()
     begin
-        NewerThanDate := CreateDateTime(CalcDate('<-30D>', Today()), Time());
+        if NewerThanDate = 0DT then
+            NewerThanDate := CreateDateTime(CalcDate('<-30D>', Today()), Time());
+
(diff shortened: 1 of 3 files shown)


GOB-Andrea:
@microsoft-github-policy-service agree [company="GOB Software & Systeme"]

GOB-Andrea:
@JesperSchulz Hi Jesper, I have created the PullRequest for issue #430. If I need to add anything or if I missed anything in this process, please don't hesitate to let me know. Thank you very much!

JesperSchulz:
> @JesperSchulz Hi Jesper, I have created the PullRequest for issue #430. If I need to add anything or if I missed anything in this process, please don't hesitate to let me know. Thank you very much!

That is fantastic, Andrea! I'm looking forward to look into your PR 😊 Please give me a few days!

src/System Application/App/Email/src/Email/Sent/SentEmails.Page.al
```diff
@@ -211,7 +213,7 @@ page 8883 "Sent Emails"
         EmailConnector.ShowAccountInformation(Rec."Account Id");
     end;
 
-    internal procedure SetNewerThan(NewDate: DateTime)
+    procedure SetNewerThan(NewDate: DateTime)
```
grobyns, line 220: Why do you need this method to be public?

grobyns: also, page 8888 "Sent Emails List Part" has these methods public and documented. I would think we need to align these two pages unless there is a good reason not to.

GOB-Andrea: @grobyns The page is public and can be called directly via .Run/RunModal. I wanted to create the option of directly setting the date filter in this process. Setting a filter on a page is not a sensitive functionality. Therefore, I would like to allow the access.

grobyns: but then why not also SetRelatedRecord? Why only SetNewerThanDate? 

As I've also mentioned, these methods are already public on the list part so I think we should align the pages so they have the same public methods (and with documentation)

GOB-Andrea: @grobyns Yes, I agree. I have aligned the function comments and set the procedure "SetRelatedRecord" to public. Object update is sent.

grobyns (approved):

JesperSchulz (approved):

PR #533, opened 2024-02-01
Add permissions test for No. Series

AndreasMoth:
The No. Series module should be able to use regardless of your permissions (but not the setup). Adding tests to that end.

Fixes [AB#498331](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/498331)

### DIFF
--- a/src/Business Foundation/App/NoSeries/src/Legacy/NoSeriesLinePurchase.Table.al
+++ b/src/Business Foundation/App/NoSeries/src/Legacy/NoSeriesLinePurchase.Table.al
@@ -22,6 +22,8 @@ table 12146 "No. Series Line Purchase"
     ObsoleteState = Pending;
     ObsoleteTag = '24.0';
 #endif
+    InherentEntitlements = r;
+    InherentPermissions = r;
 
     fields
     {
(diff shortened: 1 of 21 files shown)


grobyns (approved):

src/Business Foundation/App/NoSeries/src/Single/NoSeriesImpl.Codeunit.al
```diff
@@ -208,6 +212,7 @@ codeunit 304 "No. Series - Impl."
             Error(SeriesNotRelatedErr, DefaultNoSeriesCode, RelatedNoSeriesCode);
     end;
 
+    [InherentPermissions(PermissionObjectType::TableData, Database::"No. Series", 'r')]
```
stkillen: Should this inherent indirect read be set on table-level instead of function-level?

src/Business Foundation/App/NoSeries/src/Legacy/NoSeriesLinePurchase.Table.al
```diff
@@ -22,6 +22,8 @@ table 12146 "No. Series Line Purchase"
     ObsoleteState = Pending;
     ObsoleteTag = '24.0';
 #endif
+    InherentEntitlements = r;
```
stkillen: rX?

And these can also be removed from the permission sets

stkillen (approved):

grobyns (approved):

PR #535, opened 2024-02-02
remove NoSeriesMgt

grobyns:
#### Summary 
Removes NoSeriesMgt and fixes related issues. Introduces the No. Series - Setup facade codeunit
#### Work Item(s) 
Fixes [AB#498580](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/498580)


src/Business Foundation/App/NoSeries/src/Legacy/NoSeriesLineObsolete.TableExt.al
```diff
@@ -98,20 +98,20 @@ tableextension 309 NoSeriesLineObsolete extends "No. Series Line"
         exit(NoSeries.PeekNextNo(Rec, WorkDate()));
     end;
 
-    [Obsolete('This functionality has been removed and getting the number from a string is no longer part of No. Series.', '24.0')]
+    [Obsolete('This functionality has been made internal.', '24.0')]
```
NKarolak: As a developer getting this warning, I could not determine what to do with it. As I derive from the code, this procedure `ExtractNoFromCode` itself has not been made internal - it can be still called from outside. Same for the `GetFormattedNo` procedure.

You probably meant that this procedure will be removed later without public replacement?

grobyns: yes, the method will no longer be available. I'll update the message.

src/Business Foundation/App/NoSeries/src/Setup/NoSeries.Table.al
```diff
@@ -47,9 +48,10 @@ table 308 "No. Series"
 
             trigger OnValidate()
             var
-                NoSeriesMgt: Codeunit NoSeriesMgt;
+                NoSeriesSetupImpl: Codeunit "No. Series - Setup Impl.";
             begin
-                NoSeriesMgt.ValidateManualNos(Rec, xRec);
+                error('bar');
```
grobyns: ups ... LOL

src/Business Foundation/App/NoSeries/src/Legacy/NoSeriesManagement.Codeunit.al
```diff
@@ -518,6 +561,8 @@ codeunit 396 NoSeriesManagement
         ReplaceNoText(No, NewNo, 0, StartPos, EndPos);
     end;
 
+#if not CLEAN24
+    [Obsolete('The method was moved to No. Series Setup Impl. and is now internal.', '24.0')]
```
NKarolak: The Obsolete text does not match the content:
- The method is still available (was not moved yet).
- Within the procedure, I see no reference to the `No. Series Setup Impl.` yet.

src/Business Foundation/App/NoSeries/src/Legacy/NoSeriesManagement.Codeunit.al
```diff
@@ -931,11 +971,19 @@ codeunit 396 NoSeriesManagement
     begin
     end;
 
-    [IntegrationEvent(false, false)]
-    internal procedure OnBeforeUpdateNoSeriesLine(var NoSeriesLine: Record "No. Series Line"; NewNo: Code[20]; NewFieldName: Text[100]; var IsHandled: Boolean)
+#if not CLEAN24
+    [Obsolete('This is a temporary method for compatibility only. Please use the "No. Series" codeunit instead', '24.0')]
+    internal procedure RaiseObsoleteOnBeforeUpdateNoSeriesLine(var NoSeriesLine: Record "No. Series Line"; NewNo: Code[20]; NewFieldName: Text[100]; var IsHandled: Boolean)
```
NKarolak, line 976:
```suggestion
    internal procedure RaiseObsoleteOnBeforeUpdateNoSeriesLine(var NoSeriesLine: Record "No. Series Line"; NewNo: Code[20]; NewFieldName: Text[100]; var IsHandled: Boolean)
```
This procedure is internal and hence does not need to be obsoleted.

src/Business Foundation/App/NoSeries/src/Legacy/NoSeriesManagement.Codeunit.al
```diff
@@ -931,11 +971,19 @@ codeunit 396 NoSeriesManagement
...
     begin
+        OnBeforeUpdateNoSeriesLine(NoSeriesLine, NewNo, NewFieldName, IsHandled);
     end;
 
+    [Obsolete('This event is obsolete. Please use the extensibility options provided by the No. Series module.', '24.0')]
```
NKarolak, line 981: > "Please use the extensibility options provided by the No. Series module."

If the module provides a "twin event", then please name it directly.
Even if such an event does not exist, please point the developer at least to the relevant object(s).

src/Business Foundation/App/NoSeries/src/Legacy/NoSeriesObsolete.TableExt.al
```diff
@@ -54,23 +54,23 @@ tableextension 308 NoSeriesObsolete extends "No. Series"
 
 #if not CLEAN24
 #pragma warning disable AL0432
-    [Obsolete('The function has been moved to codeunit NoSeriesManagement', '24.0')]
+    [Obsolete('The function has been moved to codeunit "No. Series Setup Impl."', '24.0')]
```
NKarolak: General remark:
Sometimes you refer to "method", sometimes to "function".
The docs speak of "methods".
I personally prefer "procedure" for every method that was written in AL, but it would be helpful if there was some real Microsoft standard ;-)

src/Business Foundation/App/NoSeries/src/Legacy/NoSeriesObsolete.TableExt.al
```diff
@@ -85,18 +85,20 @@ tableextension 308 NoSeriesObsolete extends "No. Series"
...
+    [Obsolete('The event has been removed.', '24.0')]
+    // Symbol usage indicates no subscribers.
     [IntegrationEvent(false, false)]
     internal procedure OnBeforeValidateDefaultNos(var NoSeries: Record "No. Series"; var IsHandled: Boolean)
     begin
     end;
 
-    [Obsolete('The event has been moved to codeunit NoSeriesManagement', '24.0')]
+    [Obsolete('The event has been removed.', '24.0')]
+    // Symbol usage indicates no subscribers.
```
NKarolak: Please state whether there is a replacement. If there is, please name it.

src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
```diff
@@ -126,12 +130,233 @@ codeunit 305 "No. Series - Setup Impl."
...
+        IsHandled: Boolean;
+    begin
+        IsHandled := false;
+        NoSeries.OnBeforeValidateDefaultNos(NoSeries, IsHandled);
+        if not IsHandled then
+#pragma warning restore AL0432
+#else
+    begin
+#endif
+            if (NoSeries."Default Nos." = false) and (xRecNoSeries."Default Nos." <> NoSeries."Default Nos.") and (NoSeries."Manual Nos." = false) then
```
NKarolak, line 155:
```suggestion
            if (not NoSeries."Default Nos.") and (xRecNoSeries."Default Nos." <> NoSeries."Default Nos.") and (not NoSeries."Manual Nos.") then
```
Same for line 172.

NKarolak (changes requested):

src/Business Foundation/App/NoSeries/src/Legacy/NoSeriesManagement.Codeunit.al
grobyns: As with all obsoletions, the old method is still there because otherwise it would be breaking and also, you wouldn't get to see the obsolete warning. Now, we could change all these to state: The method will be moved, etc... but that's also not true. I could say the mehtod was copied and the old one will be removed. 
As for the second comment, I intend to keep the old code as unchanged as possible to reduce the risk of accidental breaks. 

Now, since i wrote this I added a facade and added the method to the facade so it is no longer internal...

src/Business Foundation/App/NoSeries/src/Legacy/NoSeriesManagement.Codeunit.al
grobyns: It does not need to be obsoleted but it can be so i'll keep it.

src/Business Foundation/App/NoSeries/src/Legacy/NoSeriesManagement.Codeunit.al
grobyns: There is no direct replacement. As I do not know why or how this event was used I also cannot recommend an alternative. Our data shows that only 1 app is using this event.

src/Business Foundation/App/NoSeries/src/Legacy/NoSeriesObsolete.TableExt.al
grobyns: I'll try and stick to method from now on since that what the docs use...

src/Business Foundation/App/NoSeries/src/Legacy/NoSeriesObsolete.TableExt.al
grobyns: there is no replacement. Our data shows there is no usage of these events. I asked the requestor of the events and they use it on-prem only. We are providing them with an alternative way of dealing with their requirement.

src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
grobyns: I agree with your comment but this code is obsolete now and I want to keep it as close to unchanged as I can to avoid accidental breaks.

src/Business Foundation/App/NoSeries/src/Legacy/NoSeriesObsolete.TableExt.al
```diff
@@ -82,21 +82,23 @@ tableextension 308 NoSeriesObsolete extends "No. Series"
         AllowGaps := NoSeriesSingle.MayProduceGaps();
     end;
 
-    [Obsolete('The function has been moved to codeunit NoSeriesManagement', '24.0')]
+    [Obsolete('The fmethod has been moved to codeunit NoSeriesManagement', '24.0')]
```
NKarolak:
```suggestion
    [Obsolete('The method has been moved to codeunit NoSeriesManagement', '24.0')]
```

grobyns: ups, good catch :-)

NKarolak (approved):

darjoo (approved):

stkillen (approved):

PredragMaricic (approved):

JesperSchulz (approved):

PR #539, opened 2024-02-05
Uptake BC Administrator role

stkillen:
#### Summary 
Uptake new BC Administrator role.

Add entitlements and plans for both internal and delegated BC Administrator.

#### Work Item(s) 
Fixes [#AB498639](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/498639/)

### DIFF
--- a/src/System Application/App/Azure AD Plan/src/PlanUpgrade.Codeunit.al
+++ b/src/System Application/App/Azure AD Plan/src/PlanUpgrade.Codeunit.al
@@ -30,6 +30,7 @@ codeunit 9057 "Plan Upgrade"
         AddMicrosoft365();
         AddEssentialAttach();
         AddD365Admin();
+        AddBCAdmin();
 
         AddDefaultPlanConfigurations();
     end;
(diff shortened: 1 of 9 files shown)


src/System Application/App/Entitlements/DelegatedBCAdminAgentPartner.Entitlement.al
```diff
@@ -0,0 +1,28 @@
...
+using System.Azure.Identity;
+using System.Environment.Configuration;
+using System.Email;
+using System.Apps;
+
+entitlement "Delegated BC Admin agent - Partner"
+{
+    Type = Role;
+    RoleType = Delegated;
+    Id = '00000000-0000-0000-0000-000000000010';
```
JesperSchulz, line 17: Just making sure this is the right Id?

JesperSchulz (approved):

src/System Application/App/Entitlements/DelegatedBCAdminAgentPartner.Entitlement.al
stkillen: Yes, this is the id for the delegated role.

src/System Application/App/Azure AD Plan/src/PlanUpgrade.Codeunit.al
```diff
@@ -207,6 +208,39 @@ codeunit 9057 "Plan Upgrade"
         UpgradeTag.SetUpgradeTag(PlanUpgradeTag.GetMicrosoft365UpgradeTag());
     end;
 
+    [NonDebuggable]
```
AndreasMoth, line 211: Any reason this is non-debuggable?

AndreasMoth (approved):

darjoo (approved):

src/System Application/App/Azure AD Plan/src/PlanUpgrade.Codeunit.al
stkillen: To be honest I can't see any reason in particular, but every other function in this codeunit that does similar upgrade is NonDebuggable

PR #543, opened 2024-02-06
Add actionable errors for errors in the new No. Series module

AndreasMoth:
In case that the user gets an error, we want to provide an easy way for the customer to fix those issues whenever possible. Hence adding some actionable questions to quickly resolve the issue up-front instead of having to look up docs on how to fix these.

Fixes [AB#476365](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/476365)


src/Business Foundation/App/NoSeries/src/Single/NoSeriesImpl.Codeunit.al
```diff
@@ -52,13 +55,31 @@ codeunit 304 "No. Series - Impl."
...
+    local procedure TestManualInternal(NoSeriesCode: Code[20]; ErrorText: Text)
     var
         NoSeries: Record "No. Series";
     begin
         NoSeries.Get(NoSeriesCode);
         if not NoSeries."Manual Nos." then
-            Error(ErrorText);
+            if HandleActionableManualNosError(NoSeries) then
+                if not NoSeries.Get(NoSeriesCode) and not NoSeries."Manual Nos." then
+                    Error(ErrorText);
```
NKarolak:
```suggestion
            if not HandleActionableManualNosError(NoSeries) then               
                Error(ErrorText);
```

See also 

As I understand, `HandleActionableManualNosError` returns `true` as soon as "Manual Nos." is already true or is confirmed by the user to set it as such.
I also don't see why we need a `NoSeries.Get` again in line 65 after it has been already fetched in line 62?

See also line 344.

src/Business Foundation/App/NoSeries/src/Single/NoSeriesImpl.Codeunit.al
```diff
@@ -52,13 +55,31 @@ codeunit 304 "No. Series - Impl."
...
+    end;
+
+    local procedure HandleActionableManualNosError(NoSeries: Record "No. Series"): Boolean
```
NKarolak:
```suggestion
    local procedure HandleActionableManualNosError(var NoSeries: Record "No. Series"): Boolean
```
There's a NoSeries.Modify(true) inside the procedure.

See also line 348.

NKarolak (changes requested):
This is not about an actionable error message, but about an additional confirmation dialog ;-)

src/Business Foundation/App/NoSeries/src/Single/NoSeriesImpl.Codeunit.al
AndreasMoth: Needed the Get since we want to confirm the Manual Nos. has been set, but I can set it as var instead

AndreasMoth:
> This is not about an actionable error message, but about an additional confirmation dialog ;-)

Yes, it was, a few iterations happened and now they are actionable error messages :)

NKarolak (approved):

darjoo (approved):

src/BusinessFoundationModules.code-workspace
```diff
@@ -2,15 +2,15 @@
 	"folders": [
 		{
 			"name": "No. Series",
-			"path": "Business Foundation/App/No Series"
+			"path": "Business Foundation\\App"
```
grobyns: well this is not correct... the name of the folder still says "No. Series" but now the path doesn't match.

also don't get why you're changing / to \\ ... / is nicer i think... 

Anywho... this workspace is BusinessFoundation_Modules_ so it is meant to show as
module1
module1testlib
module1test
module2
module2testlib
etc...

We might need a second workspace, BusinessFoundation.code-workspace which has
App
Test Library
Test

as you're doing here...

grobyns (changes requested):

grobyns (approved):
not happy about the workspace but fine...

JesperSchulz (approved):

PR #549, opened 2024-02-08
Language Id and Format Region override methods

Groenbech96:
#### Summary 
Allow partners and third party apps to control Language ID and Format Region when reports are called in static setting.
#### Work Item(s) 
Fixes [AB#497538](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/497538)


src/System Application/App/Language/src/LanguageImpl.Codeunit.al
```diff
@@ -36,8 +36,14 @@ codeunit 54 "Language Impl."
 
     procedure GetLanguageIdOrDefault(LanguageCode: Code[10]): Integer;
     var
+        Language: Codeunit Language;
         LanguageId: Integer;
+        Handled: Boolean;
     begin
+        Language.OnGetLanguageIdOrDefault(LanguageCode, LanguageId, Handled);
```
grobyns: Do we really need the 'handled' pattern here?

src/System Application/App/Language/src/LanguageImpl.Codeunit.al
```diff
@@ -36,8 +36,14 @@ codeunit 54 "Language Impl."
 
     procedure GetLanguageIdOrDefault(LanguageCode: Code[10]): Integer;
     var
+        Language: Codeunit Language;
         LanguageId: Integer;
+        Handled: Boolean;
```
grobyns: Should be IsHandled ... but better yes, we shouldn't have it at all

src/System Application/App/Language/src/LanguageImpl.Codeunit.al
grobyns: I don't think this is a good fix. These events run completely out of context. A subscriber will not know what impact they are having. 
Yes, when used properly (i.e. a manual subscriber, bound at the exact right time and unbound right after) might solve a particular issue but an automatic subscriber or a misplaced one will mess up a lot of things in unpredictable ways. 

The original issue is with some formatting on reports, and reports only. The reports switch language based on values in the records they deal with (on purpose). It seems that works a little too well in some specific cases. I would suggest looking into those specific cases. 

For example, if you run a report that takes the language or region from a customer record and you decide you don't like the language or region it is using, then why use a subscriber to the language method to override it? Why not change the value on the customer? 

There's only two ways to use these events: very targeted with manual subscribers, in this case the partner is running code and they can do all sorts of things OR, very generally, in which case you will change the behaviour for everything, everywhere since these events have no context. 

In short... I think we should look for a different solution to the original problem.

grobyns (changes requested):

src/System Application/App/Language/src/LanguageImpl.Codeunit.al
MODUSCarstenScholling: @grobyns @Groenbech96 

Our real problem is the fact that we do not know what is set in a report. You always talk about the "customer record", but the data is not only in the customer table, but also in contacts, setup tables, bank accounts and in multiple purchase and sales headers.

To make matters worse, anyone can now use constants in a report, data from third-party tables and much more, thereby overwriting the region with any data from everywhere. The caller has no control, because the api to overwrite these values is not available in Report.SaveAs().

In fact, I still expect an exported XML (data interchange format!) to contain decimals and date/time values in US English, which is to be corrected in the platform. However, manipulating the outcome from the application is also an important step forward.

These events are too general and, as written above, they lack context. The context may also be lost with Bind/UnbindSubscription().

Would you be open for another approach? Both Language and Language Impl. are already single instance Codeunits. What about the following methods?

- SetOverrideFormatRegion('en-US') -> Override all subsequent calls to GetFormatRegionOrDefault() until set back to empty or another value.
- SetOverrideNextFormatRegion('en-US') -> Override the next call to GetFormatRegionOrDefault(), automatically resets the override.

Same for Language.

So a developer can easily do this

```AL
Language.SetOverrideFormatRegion('en-US');
Report.SaveAs(...);
Language.SetOverrideFormatRegion(); // Overload without param. or Language.ResetOverrideFormatRegion()

// Handle proper XML output
```

What do you think? I know, setting overrides and not resetting it can cause massive issues, but filtering on FilterGroup 1 can do too...

grobyns: The problem with using this kind of override is that if an error happens, the incorrect value is set until the user closes the session and starts a new one. 
With manual binding, if an error happens, the unbinding should happen when the codeunit instance goes out of scope which is much more controllable. (i.e. don't use manual binding in a single instance cu and you should be fine).

I wonder if there is a way to combine some of these ideas such that these events are only raised when needed and subscribers are only called when needed and everything would automatically reset on errors to avoid unwanted after effects... it's still not as good as fixing the platform but would minimize issues...

grobyns: Thinking a bit more about the SetOverride... you imagine something like this i assume:

 ```
    procedure GetLanguageIdOrDefault(LanguageCode: Code[10]): Integer;
    var
        LanguageId: Integer;
    begin
        if LanguageIdOverride <> 0 then begin
            LanguageId := LanguageIdOverride;
            LanguageIdOverride := 0;
            exit(LanguageId);
        end;
        
        LanguageId := GetLanguageId(LanguageCode);

        if LanguageId = 0 then
            LanguageId := GlobalLanguage();

        exit(LanguageId);
    end;
```

same for FormatRegion. it would act exactly once and then reset. So you would never need to manually reset afterwards. If an error occurs after you set the override but before the call to GetLanguage then whatever is the next call would get it... This does seem to minimize any potential fallout so that's ok... not sure we can do much better as an app fix actually...

Comment (MODUSCarstenScholling) [reply src/System Application/App/Language/src/LanguageImpl.Codeunit.al]: Yes, exactly (for the idea with SetOverrideNextFormatRegion()). Let me check another option, not sure if this works. Gimme some minutes.

Comment (MODUSCarstenScholling) [reply src/System Application/App/Language/src/LanguageImpl.Codeunit.al]: While I could go with the shot and forget approach, there might be another option. But I have no clue if this is something you would do:

```AL
if not IsOnGetFormatRegionOrDefaultStaticallyBound() then begin
    OnGetFormatRegionOrDefaultManualBinding(FormatRegion);
end;
```

```AL
local procedure IsOnGetFormatRegionOrDefaultStaticallyBound(): Boolean
var
    EventSubscription: Record "Event Subscription";
begin
    EventSubscription.SetRange("Publisher Object Type", EventSubscription."Publisher Object Type"::Codeunit);
    EventSubscription.SetRange("Publisher Object ID", Codeunit::Language);
    EventSubscription.SetRange("Published Function", 'OnGetFormatRegionOrDefault');
    EventSubscription.SetRange("Subscriber Instance", 'Static-Automatic');

    exit(not EventSubscription.IsEmpty());
end;

internal procedure OnGetFormatRegionOrDefaultManualBinding(var FormatRegion: Text[80])
begin
end;
```

A bit tricky but would allow you to only allow manual even binding. Also, it could be initialized on the first call, changes would only occur if apps are changed.

Your decision, one time override or more complex checks for event bindings.

grobyns: I like the idea of checking for manual binding but not the implementation of it. This really should just be an attribute on the event publisher which states what kind of binding is allowed and the compiler should enforce it... 

I'll pass the idea on to the compiler team but for the app workaround I think we should go with the one-time override for now.

MODUSCarstenScholling: Sounds good to me, thank you!

nhsejth: > n fact, I still expect an exported XML (data interchange format!) to contain decimals and date/time values in US English, which is to be corrected in the platform. However, manipulating the outcome from the application is also an important step forward.

True. Now to the technical rationale behind the "weird/no-standard" formatting.  The xml format that is exported from reports is tightly coupled to the Word reporting story where it used to be the data source for application-based merge of Word templates. The column values are merged directly into the Word artifact with no further formatting. The next problem in xml emits is that many reports use a source expression that formats the value according to language and format and returns a string to the dataitem column (in particular an issue for date and time values where the AL formatting differs from the .Net formatting). If you want a pure Xml then the first requirement is data columns with no source expressions and an explicit definition of the raw xml output. Some of this can be added in a later platform update with extended report api's, but not something that can be rushed in.

MODUSCarstenScholling: I just wanted to have that mentioned here. If the platform in the future supports setting FormatRegion and Language for Report.SaveAs() programmatically like Report.Run() does (in the page, interactively), then no changes regarding the xml are needed.

nhsejth: Adding these two parameters in a parameter bag is included in current ideas. The report should be able to run without depending on triggers, request page xml (which is limited to page controls/filters) etc. to ensure a well-defined output without user or app interaction.

Groenbech96: Please review changes to PR. API now has 2 override methods.

src/System Application/App/Language/src/Language.Codeunit.al
```diff
@@ -67,6 +67,42 @@ codeunit 43 Language
         exit(LanguageImpl.GetLanguageId(LanguageCode));
     end;
 
+    /// <summary>  
+    /// Sets an override for the language ID.   
+    /// This function is used to override the default or user-selected language ID.   
+    /// Once set, the system will return this language ID for any subsequent calls to `GetLanguageIdOrDefault` method, until the override is cleared.
```
grobyns: This is not what was discussed. The override should reset after every call. If the override does not reset and an error occurs, the language will remain changed until the user starts a new session or hits another override.

grobyns (changes requested):

src/System Application/App/Language/src/LanguageImpl.Codeunit.al
MODUSCarstenScholling: Looks good to me, thank you.

src/System Application/App/Language/src/Language.Codeunit.al
MODUSCarstenScholling: Hmm, well. You are right.

Groenbech96: What if you have multiple calls to this GetLanguageId in a report? Then it will only override on the first, then remaining iterations will be printed in wrong language.

grobyns: as discussed, lets have an overload with and one without reset... and doc the error issue + add telemetry whenever override is used.

MODUSCarstenScholling:
It looks to me as if no further work is being done on this PR. This was to work around bugs related to a behavior change in the platform. This is an active case that I have open with Microsoft (2401220030002761) and we and our customers are waiting for a resolution.

When can we expect a solution and for which versions (23.x) is a backport planned?

nhsejth:
> It looks to me as if no further work is being done on this PR. This was to work around bugs related to a behavior change in the platform. This is an active case that I have open with Microsoft (2401220030002761) and we and our customers are waiting for a resolution.
> 
> When can we expect a solution and for which versions (23.x) is a backport planned?

The PR will be finalized soon, need the last test and review. The owner is oof this week and will handle it asap.

grobyns (approved):

haoranpb (dismissed):

haoranpb (approved):

grobyns (approved):

PR #550, opened 2024-02-08
Add GetCollectionType to Edit-In Excel API

DenLilleMand:
#### Summary 

The issue
We have received this issue: https://github.com/microsoft/ALAppExtensions/issues/21940 where the programmer wants to move all of the filters from one field to another. The only thing missing for this capability is that we should be able to dynamically detect if the filter is a AND or OR expression, this is not possible in the current implementation.

Expected behavior
We should expose the "Edit in Excel Filter Collection Type" property on the filter through the API, such that  clients of the API can dynamically detect what collection type the filter has.

#### Work Item(s) 
Fixes [AB#498510](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/498510/)


src/System Application/App/Edit in Excel/src/Filters/EditInExcelFieldFilter.Interface.al
```diff
@@ -27,6 +27,12 @@ interface "Edit in Excel Field Filter"
     /// <param name="FilterValue">The value which the field should be Equal to, Greater than etc.</param>
     procedure Get(Index: Integer; var EditinExcelFilterType: Enum "Edit in Excel Filter Type"; var FilterValue: Text)
 
+    /// <summary>
+    /// Get the filter collection type
+    /// </summary>
+    /// <returns>The collection type of the filter</returns>
+    procedure GetCollectionType(): Enum "Edit in Excel Filter Collection Type"
```
darjoo: Interfaces cannot be changed. Please create a v2 and obsolete this.

darjoo: Look at `Azure Key Vault` or ping me on teams if you need assistance, as I've had to do this twice now.

DenLilleMand:
Just putting this here, adding
```
    "preprocessorSymbols": [
        "CLEAN25"
    ]
``` to the `app.json` seems to compile fine.

**Comment by grobyns** [comment]
> Just putting this here, adding
> 
> ```
>     "preprocessorSymbols": [
>         "CLEAN25"
>     ]
> ```
> 
> to the `app.json` seems to compile fine.

which just goes to show: compilation is not enough :-P

File: src/System Application/App/Edit in Excel/src/Filters/EditInExcelFilters.Codeunit.al [hunk 2]
```diff
@@ -19,27 +19,58 @@ codeunit 1490 "Edit in Excel Filters"
...
 
+
+#if not CLEAN25
     /// <summary>
     /// Add the specified field using 'and' collection type.
     /// </summary>
     /// <param name="ODataFieldName">The OData name of the field referenced.</param>
     /// <param name="EditInExcelEdmType">The Edm type of the OData field as specified in the $metadata document.</param>
+#pragma warning disable AL0432
     procedure AddField(ODataFieldName: Text; EditInExcelEdmType: Enum "Edit in Excel Edm Type"): Interface "Edit in Excel Field Filter"
```
encimita, line 31: Mark this as obsolete as well with (pseudo): `[Obsolete("25.0", "Use AddFieldV2 instead")]`

encimita: same below

src/System Application/App/Edit in Excel/src/Filters/EditInExcelFieldFilter.Interface.al
```diff
@@ -11,13 +12,23 @@ namespace System.Integration.Excel;
 interface "Edit in Excel Field Filter"
 {
     Access = Public;
+#if not CLEAN25
+    ObsoleteState = Pending;
+    ObsoleteTag = '25.0';
+#else
+    ObsoleteState = Removed;
+    ObsoleteTag = '28.0';
```
encimita: Not sure we need this.

AFAIK the obsolete state removed is only for tables, because those persist in the SQL database even after obsoletion.

For interfaces, you should be able to just set the obsolete state to pending, tag to 25.0, and reason to your text. And instead of having the `#if` here, you can wrap the whole object in `#if not CLEAN25`, so that it goes away when we clean the 25 symbols.

darjoo: As Enrico mentioned, the else case is not needed. The interface will be removed in its entirety.

src/System Application/App/Edit in Excel/src/Filters/EditInExcelFldFilterImpl.Codeunit.al
```diff
@@ -52,6 +70,16 @@ codeunit 1492 "Edit in Excel Fld Filter Impl." implements "Edit in Excel Field F
         FilterCollectionNode.Collection.RemoveAt(Index);
     end;
 
+    procedure GetCollectionType(): Enum "Edit in Excel Filter Collection Type"
+    var
+        TempString: DotNet String;
+    begin
+        TempString := FilterCollectionNode.Operator();
```
darjoo, line 77: Is `FilterCollectionNode` guaranteed to be initialized when it gets here?

src/System Application/App/Edit in Excel/src/Filters/EditInExcelFilters.Codeunit.al
```diff
@@ -19,27 +19,58 @@ codeunit 1490 "Edit in Excel Filters"
...
+
+#if not CLEAN25
     /// <summary>
     /// Add the specified field with a specified collection type.
     /// </summary>
     /// <param name="ODataFieldName">The OData name of the field referenced.</param>
     /// <param name="EditInExcelFilterOperatorType">Specifies whether filters for this field have 'and' or 'or', such as Field1 = a|b|c (or operator) or Field1 = &lt;a&amp;>b&gt;amp;>c (and operator). Both operators for the same field is currently not supported.</param>
     /// <param name="EditInExcelEdmType">The Edm type of the OData field as specified in the $metadata document.</param>
+#pragma warning disable AL0432
     procedure AddField(ODataFieldName: Text; EditInExcelFilterOperatorType: Enum "Edit in Excel Filter Collection Type"; EditInExcelEdmType: Enum "Edit in Excel Edm Type"): Interface "Edit in Excel Field Filter"
```
darjoo, line 57: Add obsoletion tag. Helps to guarantee you have all `AddField` references removed, even though this is an internal method.

src/System Application/App/Edit in Excel/src/Filters/EditInExcelFldFilterImpl.Codeunit.al
DenLilleMand: @darjoo to get a reference to a code unit that implements this interface the `"Edit in Excel Filters Impl.":Get()` method is always called which calls `"EditinExcelFldFilterImpl.Initialize()"` that sets the `FilterCollectionNode`. Also similar methods such as `Count()` and `Remove()` would have had the same issue in that code unit.

src/System Application/App/Edit in Excel/src/Filters/EditInExcelFilters.Codeunit.al
```diff
@@ -48,17 +81,47 @@ codeunit 1490 "Edit in Excel Filters"
...
+        exit(EditinExcelFiltersImpl.Get(ODataFieldName));
+    end;
+#endif
+
+    /// <summary>
+    /// Get the field filters for the specified field.
+    /// </summary>
+    /// <param name="ODataFieldName">The OData name of the field referenced.</param>
+    /// <returns></returns>
+    procedure GetV2(ODataFieldName: Text): Interface "Edit in Excel Field Filter V2"
```
darjoo, line 124: Would it make more sense for this to be `Get(Text, var Interface)`? That way, in the future, we can obsolete this when needed and developers will not need to update anything for this except their variable that defines the interface.

src/System Application/App/Edit in Excel/src/Filters/EditInExcelFilters.Codeunit.al
```diff
@@ -19,27 +19,60 @@ codeunit 1490 "Edit in Excel Filters"
...
         exit(EditinExcelFiltersImpl.AddField(ODataFieldName, "Edit in Excel Filter Collection Type"::"and", EditInExcelEdmType));
     end;
+#endif
 
+    /// <summary>
+    /// Add the specified field using 'and' collection type.
+    /// </summary>
+    /// <param name="ODataFieldName">The OData name of the field referenced.</param>
+    /// <param name="EditInExcelEdmType">The Edm type of the OData field as specified in the $metadata document.</param>
+    procedure AddFieldV2(ODataFieldName: Text; EditInExcelEdmType: Enum "Edit in Excel Edm Type"): Interface "Edit in Excel Field Filter v2"
```
darjoo, line 43: Same thought as the other comment. To put the `interface` return as a var, making it simplier for developers in the future when it comes to obsoletion.

darjoo (approved):

src/System Application/App/Edit in Excel/src/Filters/EditInExcelFilters.Codeunit.al
DenLilleMand: Just putting this here: it is suppose to be a builder pattern, so returning by ref would ruin that design. Going to skip for now.

AndreasMoth, line 124: We have overloads in AL, so you can just keep this Get(ODataFieldName: Text): Interface "Edit in Excel Field Filter V2"

src/System Application/App/Edit in Excel/src/Filters/EditInExcelFilters.Codeunit.al
```diff
@@ -48,17 +81,47 @@ codeunit 1490 "Edit in Excel Filters"
...
 
+    /// <summary>
+    /// Add the specified field with an initial value using 'and' collection type.
+    /// This is mainly intended for fields that only filter on a single value.
+    /// </summary>
+    /// <param name="ODataFieldName">The OData name of the field referenced.</param>
+    /// <param name="EditInExcelFilterType">The filter type, such as Equal, Greater than etc.</param>
+    /// <param name="FilterValue">The value which the field should be Equal to, Greater than etc.</param>
+    /// <param name="EditInExcelEdmType">The Edm type of the OData field as specified in the $metadata document.</param>
+    procedure AddFieldV2(ODataFieldName: Text; EditInExcelFilterType: Enum "Edit in Excel Filter Type"; FilterValue: Text; EditInExcelEdmType: Enum "Edit in Excel Edm Type")
```
AndreasMoth, line 99: This function does the exact same as AddField, why not just keep AddField?

src/System Application/App/Edit in Excel/src/Filters/EditInExcelFilters.Codeunit.al
```diff
@@ -19,27 +19,60 @@ codeunit 1490 "Edit in Excel Filters"
...
     end;
+#endif
 
+    /// <summary>
+    /// Add the specified field with a specified collection type.
+    /// </summary>
+    /// <param name="ODataFieldName">The OData name of the field referenced.</param>
+    /// <param name="EditInExcelFilterOperatorType">Specifies whether filters for this field have 'and' or 'or', such as Field1 = a|b|c (or operator) or Field1 = &lt;a&amp;>b&gt;amp;>c (and operator). Both operators for the same field is currently not supported.</param>
+    /// <param name="EditInExcelEdmType">The Edm type of the OData field as specified in the $metadata document.</param>
+    procedure AddFieldV2(ODataFieldName: Text; EditInExcelFilterOperatorType: Enum "Edit in Excel Filter Collection Type"; EditInExcelEdmType: Enum "Edit in Excel Edm Type"): Interface "Edit in Excel Field Filter V2"
```
AndreasMoth, line 70: Just keep this AddField, use overloads. Adding V2 etc. to functions doesn't look very nice.. :/

src/System Application/App/Edit in Excel/src/Filters/EditInExcelFieldFilterv2.Interface.al
```diff
@@ -0,0 +1,47 @@
...
+interface "Edit in Excel Field Filter v2"
+{
+    Access = Public;
+
+    /// <summary>
+    /// Add a filter value
+    /// </summary>
+    /// <param name="EditInExcelFilterType">The filter type, such as Equal, Greater than.</param>
+    /// <param name="FilterValue">The value which the field should be Equal to, Greater than etc.</param>
+    procedure AddFilterValueV2(EditInExcelFilterType: Enum "Edit in Excel Filter Type"; FilterValue: Text): Interface "Edit in Excel Field Filter v2"
```
AndreasMoth, line 20: AddFilterValue

src/System Application/App/Edit in Excel/src/Filters/EditInExcelFilters.Codeunit.al
DenLilleMand: @AndreasMoth Yeah I wasn't sure what to do with it, in the beginning I left it as `AddField`, but then the builder pattern will mix `AddField` and `AddFieldV2`, I think it will just confuse everyone.

JesperSchulz (approved):

PR #567, opened 2024-02-12
Cache information for OnAddCommonCustomDimensions telemetry subscriber

darjoo:
#### Summary 
`AzureADGraph.GetTenantDetail(TenantInfo);` is called everytime and the information returned is the exact same.
Cache the information for each session that runs this.

#### Work Item(s) 
Fixes [AB#499449](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/499449)


JesperSchulz (approved):

src/System Application/App/Azure AD User Management/src/AzureADUserSubscribers.Codeunit.al
```diff
@@ -0,0 +1,67 @@
...
+    SingleInstance = true;
+
+    var
+        AzureADGraphUser: Codeunit "Azure AD Graph User";
+        AzureADGraph: Codeunit "Azure AD Graph";
+        AzureADPlan: Codeunit "Azure AD Plan";
+        Initialized, IsAzure, IsAdmin : Boolean;
+        CountryCode: Text;
+
+    local procedure GetInformation()
```
IhorHandziuk, line 26: We need to use the return value of TryGetCountryCode to avoid throwing an error. How about something like this:

```
    local procedure GetInformation()
    var
        PlanIds: Codeunit "Plan Ids";
        UserAccountHelper: DotNet NavUserAccountHelper;
    begin
        if Initialized then
            exit;
        Initialized := true;

        IsAzure := UserAccountHelper.IsAzure();
        if not IsAzure then
            exit;

        IsAdmin := AzureADGraphUser.IsUserDelegatedAdmin() or AzureADPlan.IsPlanAssignedToUser(PlanIds.GetGlobalAdminPlanId()) or AzureADPlan.IsPlanAssignedToUser(PlanIds.GetD365AdminPlanId());
        
        if TryGetCountryCode() then;
    end;
```

darjoo: Ah, very good catch. Let me fix that.

stkillen (approved):

JesperSchulz (approved):

PR #569, opened 2024-02-12
Add CredScan workflow

mazhelez:
#### Summary 
Add CredScan workflow

#### Work Item(s) 
Fixes AB#500399

### DIFF
--- a/.github/workflows/CredScan.yml
+++ b/.github/workflows/CredScan.yml
@@ -1,0 +1,24 @@
+name: CredScan
+
+on:
+  push:
+    branches: [ "main", "releases/*" ]
+  pull_request:
+    branches: [ "main", "releases/*" ]
+
+jobs:
+  CredScan:
+    runs-on: windows-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      - name: Credential Scanning
+        uses: microsoft/security-devops-action@v1.10.0
+        id: credscan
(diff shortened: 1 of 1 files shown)


.github/workflows/CredScan.yml
```diff
@@ -0,0 +1,24 @@
...
+
+on:
+  push:
+    branches: [ "main", "releases/*" ]
+  pull_request:
+    branches: [ "main", "releases/*" ]
+
+jobs:
+  CredScan:
+    runs-on: windows-latest
```
aholstrup1: Consider running on a linux runner if possible

aholstrup1 (approved):

aholstrup1 (approved):

.github/workflows/MSDO.yml
```diff
@@ -0,0 +1,31 @@
+name: MSDO
```
ankosiorek, line 1: What does MSDO stand for? Microsoft Security DevOps Analysis? maybe then MSDA? MSDOA

ankosiorek (approved):

.github/workflows/MSDO.yml
mazhelez: **M**icrosoft **S**ecurity **D**ev**O**ps

That I took from an example workflow and action usage. MSDO is the new cool kid on the block 😎

PR #573, opened 2024-02-13
Remove redundant internal event in the User Permissions module

IhorHandziuk:
#### Problem:
`NavUserAccountHelper.IsPermissionSetAssigned` doesn't work with buffered inserts, so when user records are inserted without a `Commit` (e. g. in tests), it throws the "User does not exist" exception. So far this problem has been solved with an internal event that skips the call alltogether in tests.

#### Solution:
Call `User.Get` before calling `NavUserAccountHelper.IsPermissionSetAssigned` to make sure buffered inserts are flushed.

#### Work Item(s) 
Fixes [AB#497548](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/497548)

### DIFF
--- a/src/System Application/App/User Permissions/src/UserPermissionsImpl.Codeunit.al
+++ b/src/System Application/App/User Permissions/src/UserPermissionsImpl.Codeunit.al
@@ -238,12 +238,11 @@ codeunit 153 "User Permissions Impl."
     var
         User: Record User;
         NavUserAccountHelper: DotNet NavUserAccountHelper;
-        Skip: Boolean;
     begin
         if HasUserPermissionSetDirectlyAssigned(UserSecurityId, Company, RoleId, ItemScope, AppId) then
             exit(true);
 
-        // NavUserAccountHelper doesn't work with buffered inserts.
+        // NavUserAccountHelper doesn't work with bulk (buffered) inserts.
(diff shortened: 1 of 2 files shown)


src/System Application/App/User Permissions/src/UserPermissionsImpl.Codeunit.al
```diff
@@ -236,14 +236,16 @@ codeunit 153 "User Permissions Impl."
...
+        User: Record User;
         NavUserAccountHelper: DotNet NavUserAccountHelper;
         Skip: Boolean;
     begin
         if HasUserPermissionSetDirectlyAssigned(UserSecurityId, Company, RoleId, ItemScope, AppId) then
             exit(true);
 
-        OnHasUserPermissionSetAssigned(Skip);
-        if Skip then
+        // NavUserAccountHelper doesn't work with buffered inserts.
```
NKarolak: Do you mean bulk inserts?
https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/administration/optimize-sql-bulk-inserts

JesperSchulz (approved):

darjoo (approved):

PR #592, opened 2024-02-19
Remove NoSeriesManagement (Clean tags)

AndreasMoth:
Adding CLEAN tags around NoSeriesManagement to remove it with cleanup of CLEAN24.

Fixes AB#471519


src/Business Foundation/App/NoSeries/src/Legacy/NoSeriesManagement.Codeunit.al
grobyns, line 1: you also need to clean up all the tags in the codeunit... :-S

JesperSchulz (approved):

PR #593, opened 2024-02-19
Enable premium users to access essential environments

stkillen:
#### Summary 
This PR enables mixing licenses (Basic, Essential, Premium). Restrictions on when these can be mixed will be introduced in Base app.

Obsolete CheckMixedPlans and remove errors. Instead add CheckMixedPlansExist which returns a boolean. Also introduce GetUserPlanExperience.

#### Work Item(s) 
Fixes [AB#459656](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/459656/)


src/System Application/App/Azure AD Plan/src/AzureADPlan.Codeunit.al
```diff
@@ -270,9 +283,35 @@ codeunit 9016 "Azure AD Plan"
...
+    end;
+#endif
+
+    /// <summary>
+    /// Checks whether the plan configuration mixes different plans.
+    /// Returns true if there are incompatible plans in the system. Only checks when conditions for plan check is met, otherwise returns false.
+    /// </summary>
+    /// <returns>Returns true if there are incompatible plans in the system.</returns>
+    [Scope('OnPrem')]
+    [NonDebuggable]
```
darjoo: NonDebuggable not required.

src/System Application/App/Azure AD Plan/src/AzureADPlan.Codeunit.al
```diff
@@ -270,9 +283,35 @@ codeunit 9016 "Azure AD Plan"
...
+    procedure CheckMixedPlansExist(): Boolean
+    begin
+        exit(AzureAdPlanImpl.CheckMixedPlans());
```
darjoo: I'd also update the internal function to use the new name

src/System Application/App/Azure AD Plan/src/AzureADPlanImpl.Codeunit.al
```diff
@@ -356,6 +340,31 @@ codeunit 9018 "Azure AD Plan Impl."
...
+    [NonDebuggable]
+    local procedure ShouldCheckMixedPlans(): Boolean
+    var
+        Company: Record Company;
+        EnvironmentInformation: Codeunit "Environment Information";
+    begin
+        if not EnvironmentInformation.IsSaaS() then
+            exit(false);
+
+        if not GuiAllowed() then
```
darjoo, line 346: Why should it not check if Gui is not allowed?

src/System Application/App/Azure AD Plan/src/UserPlanExperience.Enum.al
```diff
@@ -0,0 +1,46 @@
...
+    {
+        Caption = 'Premium';
+    }
+
+    /// <summary>
+    /// Basic plan.
+    /// </summary>
+    value(3; Basic)
+    {
+        Caption = 'Line';
```
AndreasMoth: Caption = 'Basic'

src/System Application/App/Azure AD Plan/src/AzureADPlanImpl.Codeunit.al
JesperSchulz: Indeed. This should work for API scenarios too.

src/System Application/App/Azure AD Plan/src/UserPlanExperience.Enum.al
```diff
@@ -0,0 +1,46 @@
...
+        Caption = 'Premium';
+    }
+
+    /// <summary>
+    /// Basic plan.
+    /// </summary>
+    value(3; Basic)
+    {
+        Caption = 'Basic';
+    }
```
JesperSchulz, line 37: Why not have all possible plans? Team Member etc. could come in handy too given time, could they not?

src/System Application/App/Azure AD Plan/src/AzureADPlanImpl.Codeunit.al
stkillen: We've never checked mixed plans when Gui is not allowed, so this would be a new enforcement.

Might be quite a lot of overhead to add to API scenarios?

darjoo: If you don't check it outside of GUIallowed sessions, then essentials can technically log into premium experience tier companies.

Fx, a user who is essentials. A JQ for for this user is then created for `Company 1` which is premium experience tier. They'd be able to login and do stuff in that background session.

src/System Application/App/Azure AD Plan/src/UserPlanExperience.Enum.al
stkillen: The purpose of the enum was to return only the information relevant for the mixed plans check, hence the name "User Plan Experience". Currently we're only checking mixed plans for Basic, Essentials and Premium licenses which is why these are the plans represented in the enum. Adding all plans / roles to this enum would add unnecessary overhead to the login flow. 

Furthermore, if we start considering all plans / roles, a single enum would not be sufficient to accurately represent the plans of a user, since one user can have more than one "plan".

So resolving this comment

src/System Application/App/Azure AD Plan/src/AzureADPlanImpl.Codeunit.al
stkillen: Resolving comment as changing the design here would require proper testing and insights into the performance impact of that change.

Safest option is to keep the same design and create a follow up work item to look into this if relevant.

src/System Application/App/Azure AD Plan/src/AzureADPlan.Codeunit.al
stkillen: Removed

darjoo (dismissed):

darjoo (approved):

JesperSchulz (approved):

PR #594, opened 2024-02-19
Add Legacy Action Bar setting

darjoo:
#### Summary 
Add a setting for Legacy Action Bar on the User Settings and My Settings pages for the user to customize.
Add platform subscriber to provide value of whether the setting has been enabled or not.

#### Work Item(s) 
Fixes [AB#499142](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/499142)


eikawata:
Looks good to me!

JesperSchulz (approved):

src/System Application/App/User Settings/src/UserSettingsImpl.Codeunit.al
```diff
@@ -368,6 +370,40 @@ codeunit 9175 "User Settings Impl."
         ApplicationUserSettings.Modify();
     end;
 
+    procedure LegacyActionBarEnabled(UserSecurityId: Guid): Boolean
```
AndreasMoth: The variable name UserSecurityId is the same as the global function name UserSecurityId(), this could be confusing.

AndreasMoth: We should probably have a rule to prevent that, just like we have for not naming variables the same as functions.

src/System Application/App/User Settings/src/UserSettingsImpl.Codeunit.al
```diff
@@ -368,6 +370,40 @@ codeunit 9175 "User Settings Impl."
...
+        GetAppSettings(UserSecurityId, ApplicationUserSettings);
+        ApplicationUserSettings."Legacy Action Bar" := false;
+        ApplicationUserSettings.Modify();
+    end;
+
+    procedure EnableLegacyActionBar(UserSecurityId: Guid)
+    var
+        ApplicationUserSettings: Record "Application User Settings";
+        CustomerExperienceSurvey: Codeunit "Customer Experience Survey";
+        FormsProId: Text;
```
AndreasMoth, line 394: FormsProcessId? or.. I'm not sure what Pro stands for, would be great to have written out.

src/System Application/App/User Settings/src/UserSettingsImpl.Codeunit.al
```diff
@@ -368,6 +370,40 @@ codeunit 9175 "User Settings Impl."
...
+        FormsProEligibilityId: Text;
+        IsEligible: Boolean;
+    begin
+        GetAppSettings(UserSecurityId, ApplicationUserSettings);
+        ApplicationUserSettings."Legacy Action Bar" := true;
```
AndreasMoth, line 399: We don't want to validate? That would allow partners to change this value with events (OnBeforeValidateField())

AndreasMoth: or we want them to hook in on global modify trigger?

src/System Application/App/User Settings/src/UserSettingsImpl.Codeunit.al
darjoo: I do not know what they mean either. This is old code that I was informed to take and place in directly to show the survey.

src/System Application/App/User Settings/src/UserSettings.Page.al
```diff
@@ -114,6 +114,14 @@ page 9204 "User Settings"
                         Caption = 'Teaching Tips';
                         ToolTip = 'Specifies whether to display short messages that inform, remind, or teach you about important fields and actions when you open a page.';
                     }
+                    field("Legacy Action Bar"; Rec."Legacy Action Bar")
+                    {
+                        ApplicationArea = All;
+                        Caption = 'Legacy Action Bar';
+#pragma warning disable AA0219
+                        ToolTip = 'Revert the action bar to the legacy terms and layout. For example, the action bar won''t be pinned on pages, the Home tab is named Process, and split buttons become multiple actions, which means more clicks. Also, your options for personalization will be limited.';
```
AndreasMoth, line 122: Specifies whether to revert...
This way you remove the warning disablement, especially when it's not needed.

src/System Application/App/User Settings/src/UserPersonalization.Page.al
```diff
@@ -116,6 +116,23 @@ page 9214 "User Personalization"
                             UserSettingsImpl.DisableTeachingTips(Rec."User SID");
                     end;
                 }
+                field("Legacy Action Bar"; LegacyActionBarEnabled)
+                {
+                    ApplicationArea = All;
+                    Caption = 'Legacy Action Bar';
+#pragma warning disable AA0219
+                    ToolTip = 'Revert the action bar to the legacy terms and layout. For example, the action bar won''t be pinned on pages, the Home tab is named Process, and split buttons become multiple actions, which means more clicks. Also, your options for personalization will be limited.';
```
AndreasMoth, line 124: Same comment as for the other page.

src/System Application/App/User Settings/src/UserSettingsImpl.Codeunit.al
darjoo: I don't call validate. So those events aren't called.
If the partner decides to go ahead and subscribe fx "OnBeforeInsert", "OnAfterInsert" and make it false, that'd be their decision.

src/System Application/App/User Settings/src/UserSettingsImpl.Codeunit.al
darjoo: Also why I didn't change the literal text names from 'modernactionbar'... to 'legacyactionbar'. I don't know how that breaks stuff.

src/System Application/App/User Settings/src/ApplicationUserSettings.Table.al
```diff
@@ -33,6 +33,15 @@ table 9222 "Application User Settings"
         {
             DataClassification = CustomerContent;
         }
+
+        /// <summary>
+        /// Specifies whether the legacy action bar is enabled.
+        /// </summary>
+        field(3; "Legacy Action Bar"; Boolean)
```
AndreasMoth, line 40: "Use Legacy Action Bar"? otherwise it more sounds like an enum where I can choose which Legacy Action Bar I want.

src/System Application/App/User Settings/src/UserSettingsImpl.Codeunit.al
darjoo: Standarized to the other methods, `UserSecurityID` instead of `UserSecurityId`. Not going in to change all methods to use `UID`.

src/System Application/App/User Settings/src/UserSettings.Page.al
darjoo: This was given by Jacob, sticking with it.

src/System Application/App/User Settings/src/UserPersonalization.Page.al
darjoo: This was given by Jacob, sticking with it.

src/System Application/App/User Settings/src/ApplicationUserSettings.Table.al
AndreasMoth: Could also be "Enable Legacy Action Bar", aligns better with the name of your page field.

darjoo: Keeping consistent with `Teaching Tips`. That doesn't specify what teaching tips either.

src/System Application/App/User Settings/src/UserPersonalization.Page.al
```diff
@@ -116,6 +116,23 @@ page 9214 "User Personalization"
                             UserSettingsImpl.DisableTeachingTips(Rec."User SID");
                     end;
                 }
+                field("Legacy Action Bar"; LegacyActionBarEnabled)
+                {
+                    ApplicationArea = All;
+                    Caption = 'Legacy Action Bar';
```
AndreasMoth, line 122: Caption = 'Enable Legacy Action Bar'

darjoo: Also given by Jacob. Not changing.
It initially was `Use Legacy Action Bar`

JesperSchulz (approved):

stkillen (approved):

PR #471, opened 2024-01-08
Uptake of SecretText

darjoo:
#### Summary 
Uptake of SecretText within the System Application.
Secrets, passwords, tokens etc are converted to use SecretText.

- Any functions that return secrets as a Text will be obsoleted with a replacement that returns SecretText. Uptake necessary.
- Any functions that return via var as Text will be obsoleted with a replacement that uses SecretText. Uptake necessary.
- Any events that use/return secret parameters will be obsoleted with a replacement that uses SecretText. Uptake necessary.
- Any functions that use secret Text parameters will just automatically be changed to SecretText. No change necessary.
- Allow debugging of Sys App

#### Work Item(s) 
Fixes [AB#463599](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/463599)

PR #603, opened 2024-02-20
Add back and add obsoletion to secrettext uptake functions

darjoo:
#### Summary 
Some functions were changed directly to `SecretText` without having obsoletion on the original function first.
This puts back those functions and adds obsoletion to them.

#### Work Item(s) 
Fixes [AB#463599](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/463599)

### DIFF
--- a/src/System Application/App/Azure Storage Services Authorization/src/AuthFormatHelper.Codeunit.al
+++ b/src/System Application/App/Azure Storage Services Authorization/src/AuthFormatHelper.Codeunit.al
@@ -41,6 +41,19 @@ codeunit 9060 "Auth. Format Helper"
         exit(DotNetDateTime.Parse(DateTimeAsXmlString).ToUniversalTime().ToString(FormatSpecifier));
     end;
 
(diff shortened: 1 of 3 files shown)


AndreasMoth (approved):

TKapitan:
@darjoo There is more procedures that needs to be reverted, for example this one https://github.com/microsoft/BCApps/pull/471/files#diff-6d1233c07fac23281b0926cb03abe7bb912f5a75892d59556b789a12b8ceddc3R64

darjoo:
@TKapitan No, that codeunit is internal and is fine to be changed directly.

darjoo:
On second look, all codeunits are internal, no reason to make this change. I will take another look through the original PR, but I every external should've been obsoleted with a new method. 
If no changes required, I will close this PR with no changes.

JesperSchulz (approved):

PR #607, opened 2024-02-21
No. Series: Add telemetry

AndreasMoth:
Adding telemetry to No. Series to determine if something fails and why.

Fixes [AB#496416](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/496416)

### DIFF
--- a/src/Business Foundation/App/NoSeries/src/Batch/NoSeriesBatchImpl.Codeunit.al
+++ b/src/Business Foundation/App/NoSeries/src/Batch/NoSeriesBatchImpl.Codeunit.al
@@ -17,6 +17,11 @@ codeunit 309 "No. Series - Batch Impl."
         SimulationMode: Boolean;
         CannotSaveNonExistingNoSeriesErr: Label 'Cannot save No. Series Line that does not exist: %1, %2', Comment = '%1 = No. Series Code, %2 = Line No.';
         CannotSaveWhileSimulatingNumbersErr: Label 'No. Series state cannot be saved while simulating numbers.';
+        NoSeriesBatchLbl: Label 'No. Series - Batch', Locked = true;
+        SimulationModeStartedTxt: Label 'No. Series simulation mode started.', Locked = true;
+        SavingSingleNoSeriesStateTxt: Label 'Saving single No. Series state.', Locked = true;
+        SavingAllNoSeriesStatesTxt: Label 'Saving all No. Series states.', Locked = true;
+        UpdatingNoSeriesLinesFromDbTxt: Label 'Updating No. Series lines from database.', Locked = true;
 
     procedure SetInitialState(TempNoSeriesLine: Record "No. Series Line" temporary)
     begin
@@ -124,6 +129,10 @@ codeunit 309 "No. Series - Batch Impl."
 
     procedure SetSimulationMode()
     begin
+        if SimulationMode then
+            exit;
+
+        Session.LogMessage('0000MI2', SimulationModeStartedTxt, Verbosity::Normal, DataClassification::SystemMetadata, TelemetryScope::ExtensionPublisher, 'Category', NoSeriesBatchLbl);
         SimulationMode := true;
     end;
 
@@ -133,6 +142,7 @@ codeunit 309 "No. Series - Batch Impl."
             Error(CannotSaveWhileSimulatingNumbersErr);
         if not TempGlobalNoSeriesLine.Get(TempNoSeriesLine."Series Code", TempNoSeriesLine."Line No.") then
             Error(CannotSaveNonExistingNoSeriesErr, TempNoSeriesLine."Series Code", TempNoSeriesLine."Line No.");
+        Session.LogMessage('0000MI3', SavingSingleNoSeriesStateTxt, Verbosity::Normal, DataClassification::SystemMetadata, TelemetryScope::ExtensionPublisher, 'Category', NoSeriesBatchLbl);
         UpdateNoSeriesLine(TempGlobalNoSeriesLine);
     end;
 
@@ -140,6 +150,7 @@ codeunit 309 "No. Series - Batch Impl."
     begin
         if SimulationMode then
             Error(CannotSaveWhileSimulatingNumbersErr);
+        Session.LogMessage('0000MI4', SavingAllNoSeriesStatesTxt, Verbosity::Normal, DataClassification::SystemMetadata, TelemetryScope::ExtensionPublisher, 'Category', NoSeriesBatchLbl);
         TempGlobalNoSeriesLine.Reset();
         if TempGlobalNoSeriesLine.FindSet() then
             repeat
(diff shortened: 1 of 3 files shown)


src/Business Foundation/App/NoSeries/src/Batch/NoSeriesBatchImpl.Codeunit.al
```diff
@@ -17,6 +17,11 @@ codeunit 309 "No. Series - Batch Impl."
         SimulationMode: Boolean;
         CannotSaveNonExistingNoSeriesErr: Label 'Cannot save No. Series Line that does not exist: %1, %2', Comment = '%1 = No. Series Code, %2 = Line No.';
         CannotSaveWhileSimulatingNumbersErr: Label 'No. Series state cannot be saved while simulating numbers.';
+        NoSeriesBatchLbl: Label 'No. Series - Batch', Locked = true;
+        SimulationModeStartedTxt: Label 'No. Series simulation mode started.', Locked = true;
+        SavingSingleNoSeriesStateTxt: Label 'Saving single No. Series state.', Locked = true;
+        SavingAllNoSeriesStatesTxt: Label 'Saving all No. Series states.', Locked = true;
+        UpdatingNoSeriesLinesFromDbTxt: Label 'Updating No. Series lines from database.', Locked = true;
```
NKarolak: Since all labels are locked, what about renaming them as `*Tok`?
(Applies to all other new ones in this PR as well.)

grobyns: 👍 Tok is meant to be used when the value of the label matches the name. As in: TokenTok: Label 'Token', Locked = true;
For these ones I guess we're close enough :-)

src/Business Foundation/App/NoSeries/src/Single/NoSeriesSequenceImpl.Codeunit.al
```diff
@@ -16,6 +16,8 @@ codeunit 307 "No. Series - Sequence Impl." implements "No. Series - Single"
 
     var
         NoOverFlowErr: Label 'Number series can only use up to 18 digit numbers. %1 has %2 digits.', Comment = '%1 is a string that also contains digits. %2 is a number.';
+        NoSeriesSequenceLbl: Label 'No. Series - Sequence', Locked = true;
+        UpdatingSequenceBasedOnTempValueTxt: Label 'Updating sequence based on temporary value.', Locked = true;
```
grobyns, line 20: see the Tok comment but... Why is this one Txt and all the others Lbl? :-P

src/Business Foundation/App/NoSeries/src/Single/NoSeriesStatelessImpl.Codeunit.al
```diff
@@ -17,6 +17,9 @@ codeunit 306 "No. Series - Stateless Impl." implements "No. Series - Single"
     var
         CannotAssignNumbersGreaterThanErr: Label 'You cannot assign numbers greater than %1 from the number series %2. No. assigned: %3', Comment = '%1=Last No.,%2=No. Series Code, %3=the new no.';
         WarnNoSeriesRunningOutMsg: Label 'The No. Series %1 is soon running out. The current number is %2 and the last allowed number of the sequence is %3.', Comment = '%1=No. Series code,%2=Current No.,%3=Last No. of the sequence';
+        NoSeriesLbl: Label 'No. Series', Locked = true;
+        NoSeriesOutOfRangeErr: Label 'The last number used is outside range.', Locked = true;
```
grobyns: Should also just be a Lbl ... (or Tok) ... it's not used in an error, it is just sent to telemetry...

grobyns (approved):

src/Business Foundation/App/NoSeries/src/Batch/NoSeriesBatchImpl.Codeunit.al
NKarolak: @grobyns 
Can you please explain the `Tok` rule? As I understood, the Label name should always reflect its content, at least more or less, not only for `Tok`. Furthermore, the [example in the docs](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/analyzers/codecop-aa0074) does the exact opposite of what you've just stated.
The community understood Tokens to be used this way: https://github.com/StefanMaron/BusinessCentral.LinterCop/wiki/LC0047 - wrong?

AndreasMoth: Just discussed with Nikola and Tok are for simple variables where the entire string is also part of the label, which I guess would work for the NoSeriesBatchTok. We are also looking into if we could add some other extension for these locked labels. I will go for Txt then, since 'No. Series simulation mode started.' is not really a token nor a label, it's a description of what is happening. I.E. a Text. Once we get this locked extension we can switch to that.

AndreasMoth: @NKarolak the docs looks to be out of date, I believe the example should be: GetTok: Label 'GET', Locked = true;
Will create a bug for that.

NKarolak: @AndreasMoth Thanks! While you are at it, could you also add some more explanations on when to use what, typically?

grobyns (approved):

darjoo (approved):

src/Business Foundation/App/NoSeries/src/Batch/NoSeriesBatchImpl.Codeunit.al
Arthurvdv: @AndreasMoth, bit of a side question here, but may I ask what would be your advice for the community on having a Code Analyzer for applying the right suffix on handeling labels?

As @NKarolak mentioned we've created two rules in our Community Code Analyzer; BusinessCentral.LinterCop

* [LC0046 - Label with suffix Tok must be locked.](https://github.com/StefanMaron/BusinessCentral.LinterCop/wiki/LC0046)
* [LC0047 - Locked Label must have a suffix Tok.](https://github.com/StefanMaron/BusinessCentral.LinterCop/wiki/LC0047)

With the information here I guess the LC0047 rule is invalid? Is it a good idea to adapt this to "the entire string is also part of the label", like the GetTok: Label 'GET' example?

AndreasMoth: @NKarolak @Arthurvdv I'm not going to try dictate any policies in a PR, but the way I go about it is, there are first some easy ones:
Err for Error messages "Error(NoLinesErr)"
Msg for Messages "Message(InvoicePostedMsg)"
Qst for questions in confirm dialogs "if Confirm(PostInvoiceQst) then"
Tok for any short tokens such as "GET", "PUT", "HTTPS" etc. where the name would then also align with the label itself, now these would generally always also be locked unless there's a very good reason.
Lbl and Txt I don't have a good explanation for.

I think as for these LC rules, I guess it depends on how partners wants to work with this. I can see these rules can help partners align their code better even if we don't have these.

I hope this answers all questions :). I do think however it's important to use what works for you as a partner.

PR #619, opened 2024-02-23
[Copilot] Add new events for when capabilities are activated/deactivated

darjoo:
#### Summary 
Provide events to be subscribed to for when capabilities are activated and deactivated.
Usage for this is primarily for setup/tear down, if required by a capability.

#### Work Item(s) 
Fixes [AB#502147](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/502147)

### DIFF
--- a/src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
+++ b/src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
@@ -21,6 +21,7 @@ codeunit 7774 "Copilot Capability Impl"
     Permissions = tabledata "Copilot Settings" = rimd;
 
     var
+        CopilotCapabilityCU: Codeunit "Copilot Capability";
         FeatureTelemetry: Codeunit "Feature Telemetry";
         CopilotCategoryLbl: Label 'Copilot', Locked = true;
         AzureOpenAiTxt: Label 'Azure OpenAI', Locked = true;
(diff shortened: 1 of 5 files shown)


src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
```diff
@@ -63,6 +64,7 @@ codeunit 7774 "Copilot Capability Impl"
 
         AddTelemetryDimensions(CopilotCapability, CallerModuleInfo.Id(), CustomDimensions);
         FeatureTelemetry.LogUsage('0000LDV', CopilotCategoryLbl, TelemetryRegisteredNewCopilotCapabilityLbl, CustomDimensions);
+        CopilotCapabilityCU.OnCapabilityActivated(CopilotCapability, CopilotSettings."App Id", CopilotSettings.Publisher);
```
msft-sam, line 67: it seems weird to do this as we will always trigger this event when registering any copilot capability, could we reconsider the use case?

what if the AOAI privacy notice is rejected? you could argue the capability is not active then for example, do we need to check all this additional logic here?

Additionally, it seems weird to do this via event since the caller will know all these things already (they can check `IsEnabled` and they are choosing to call `RegisterCapability`).

Imo we should aspire to, **at the most**, have the same lifecycle events of AL extensions (i.e. only have an activation event, definitely not a deactivation).
What happens in the case of an error? What if there was a commit in the subscriber and all other kinds of lovely data (in)consistencies?

darjoo:
Abandoning changes after discussion that this isn't the way we should do it. Especially for the use-cases we have at this time and future plans.

PR #643, opened 2024-02-26
[DRAFT] #606 Page Style As Enum

### LINKED ISSUE #606: [BC Idea]: Page Styles as Enum
### BC Idea Link

https://experience.dynamics.com/ideas/idea/?ideaid=95bf2064-f7cf-ee11-92bd-6045bdb614a3

### Description

Please add an Enum which represents the Page Style Expressions similar to the enum of the Cue Styles: enum 9701 "Cues And KPIs Style"

I have already added multiple times the same enum in different Apps to acchieve this but it would he helpful for us developer to have this in the system application.

The new enum should have the values described here:

https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/properties/devenv-style-property?wt.mc_id=d365bc_inproduct_alextension

Internal work item: [AB#502123](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/502123)

BernhardKloibmueller:
#### Summary 
adds an enum for the page styles with values described here:
https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/properties/devenv-style-property?wt.mc_id=d365bc_inproduct_alextension

#### Work Item(s) 
Fixes #606


src/System Application/App/Page Style/app.json
```diff
@@ -0,0 +1,31 @@
...
+        
+                           ],
+    "screenshots":  [
+
+                    ],
+    "platform":  "24.0.0.0",
+    "target":  "OnPrem",
+    "idRanges":  [
+                     {
+                         "from":  50000,
```
BernhardKloibmueller: I added it as a new app, I did not find any existing app where it was possible to add. Please provide the app where I should add or provide an idRange.

src/System Application/App/Page Style/PageStyle.al
```diff
@@ -0,0 +1,49 @@
+enum 50000 "Page Style"
```
ChrisBlankDe: I think the name "Page Style" is not perfekt. Its about Labels and Fields. Maybe "Control Style Type"  would be better.

src/System Application/App/Page Style/PageStyle.al
```diff
@@ -0,0 +1,49 @@
...
+    {
+        Caption = 'None', Comment = 'None';
+    }
+    value(1; Standard)
+    {
+        Caption = 'Standard', Comment = 'Standard';
+    }
+    value(2; StandardAccend)
+    {
+        Caption = 'StandardAccent', Comment = 'Blue';
```
ChrisBlankDe: Captions should be user friendly. So we should add some spaces. --> "Standard Accent"

BernhardKloibmueller: OK then we would need codeunits to translate the values.
If not implemented through a codeunit, I would need to lock the captions, as these vales should be constant.

The Intention would be an enum to be used for StyleExpr-Text variables in pages. Maybe you have a proper solution but I intended to use the enum directly.

I convert the PR to DRAFT because from this point there are too much questions from my point of view.
All imputs to improve are appreciated.

NKarolak: The `Comment` property is for translators of the `Caption` property; it is exported when you generate the g.xlf.
Since there is no Description property available for enum values, you could put it into comments.

src/System Application/App/Page Style/PageStyle.al
```diff
@@ -0,0 +1,49 @@
...
+
+    value(0; None)
+    {
+        Caption = 'None', Comment = 'None';
+    }
+    value(1; Standard)
+    {
+        Caption = 'Standard', Comment = 'Standard';
+    }
+    value(2; StandardAccend)
```
duichwer: There seems to be a typo.
I
```suggestion
    value(2; StandardAccent)
```

src/System Application/App/Page Style/PageStyle.al
JesperSchulz: I tend to agree, that the styling is for controls, rather than a page object. "Control Style" sounds good to me.

src/System Application/App/Page Style/PageStyle.al
JesperSchulz: Let's stick to the property values as defined here: https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/properties/devenv-style-property?wt.mc_id=d365bc_inproduct_alextension

src/System Application/App/Page Style/app.json
JesperSchulz: I'm also not sure we have a fitting module as of now. But I think a "Control Style" app is too narrow. Let's make a system enums app or something along those lines. The ID  range 1-10 would seem fitting for such an app.

I know of an enum or two which probably should move into this app, once created: e.g. https://github.com/microsoft/BusinessCentralApps/blob/main/App/Layers/W1/BaseApp/Licenses.Enum.al

src/System Application/App/Page Style/PageStyle.al
ChrisBlankDe: on second thought, you're right. but in that case we should lock the captions, right?

BernhardKloibmueller: Typo corrected and changed captions to locked

src/System Application/App/Page Style/PageStyle.al
BernhardKloibmueller: renamed to "Control Style"
set Namespace System.Visualization

src/System Application/App/Page Style/app.json
BernhardKloibmueller: Set the id-range from 1-10

src/System Application/App/Page Style/PageStyle.al
BernhardKloibmueller: type corrected

src/System Application/App/Page Style/PageStyle.al
BernhardKloibmueller: added the comments just like in the real options, so you have the explanation at hover. Thank you for pointing out

JesperSchulz:
@BernhardKloibmueller, please pause for a moment while we decide whether this should be a platform enum, rather than an app enum.

JesperSchulz:
Waiting for a conclusion. Ideally this would go into the platform, but I don't know what our capacity looks like. Waiting for compiler/platform to confirm.

JesperSchulz:
@BernhardKloibmueller, the compiler team would like to get some usage examples to better understand what this enum will be used for. Could you provide those please?

BernhardKloibmueller:
> @BernhardKloibmueller, the compiler team would like to get some usage examples to better understand what this enum will be used for. Could you provide those please?

The usage would be to use it in configurable setups for example. You want to set up Colors for the field style for some cases. Just like the Cue Setup. 

The other usage would be directly in code for multiple style options in pages. This is also used in Base App but with hardcoded strings at the moment.
Usage in Base Application is already wide spread with the string values for example in 
 - Table 981 "Payment Registration Buffer"
   - procedure GetDueDateStyle
   - procedure GetPmtDiscStyle
 - Table 36 "Sales Header"
   - procedure GetStatusStyleText
 - ...

JesperSchulz:
Great news: based on the above, the compiler team has agreed to add an option symbol to 2024 wave 2. We hence won't be needing app code!
@BernhardKloibmueller, is it OK if close this PR? I will also close the issue as completed when the platform has been updated.

BernhardKloibmueller:
@JesperSchulz

Yes then lets close this PR and we look forward to the next major. Thank you for handling this :)

PR #658, opened 2024-02-28
Add subscriber to system trigger OpenFeatureManagement

DenLilleMand:
#### Summary 
A link to the feature management page has been added to the system menu which calls a system trigger, this PR adds a subscriber to that event which opens the Feature Management page.

#### Work Item(s) 
Fixes [AB#502096](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/502096/)


src/System Application/App/Feature Key/src/FeatureManagementImpl.Codeunit.al
```diff
@@ -560,4 +560,17 @@ codeunit 2610 "Feature Management Impl."
...
+    var
+        FeatureManagementFacade: Codeunit "Feature Management Facade";
+        IsHandled: Boolean;
+        FeatureManagementPageID: Integer;
+    begin
+        FeatureManagementPageID := page::"Feature Management";
+        FeatureManagementFacade.OnBeforeOpenFeatureManagement(FeatureManagementPageID, IsHandled);
+        if not IsHandled then
+            PAGE.Run(FeatureManagementPageID);
+    end;
```
NKarolak: Just some formatting stuff, if you allow ;-)

```suggestion
    local procedure DefaultOpenFeatureManagement()
    var
        FeatureManagementFacade: Codeunit "Feature Management Facade";
        IsHandled: Boolean;
        FeatureManagementPageID: Integer;
    begin
        FeatureManagementPageID := Page::"Feature Management";
        FeatureManagementFacade.OnBeforeOpenFeatureManagement(FeatureManagementPageID, IsHandled);
        if not IsHandled then
            Page.Run(FeatureManagementPageID);
    end;
```

src/System Application/App/Feature Key/src/FeatureManagementImpl.Codeunit.al
```diff
@@ -560,4 +560,17 @@ codeunit 2610 "Feature Management Impl."
...
+    [EventSubscriber(ObjectType::Codeunit, Codeunit::"Feature Management Triggers", OpenFeatureManagement, '', false, false)]
+    local procedure DefaultOpenFeatureManagement()
+    var
+        FeatureManagementFacade: Codeunit "Feature Management Facade";
+        IsHandled: Boolean;
+        FeatureManagementPageID: Integer;
+    begin
+        FeatureManagementPageID := Page::"Feature Management";
+        FeatureManagementFacade.OnBeforeOpenFeatureManagement(FeatureManagementPageID, IsHandled);
+        if not IsHandled then
```
darjoo, line 573: If they don't handle it, we could end up opening a page they provided? 
We should do the following instead: 
1. They handle and open their own page
2. If they do not handle, we open our default page

darjoo (approved):

darjoo (commented):
Still trying to understand how to dismiss my own review on GH..

src/System Application/App/Feature Key/src/FeatureManagementImpl.Codeunit.al
DenLilleMand: @darjoo so you are just suggesting that I remove FeatureManagementPageID as an arg to OnBeforeOpenFeatureManagement ?

darjoo: It's fine to provide it, but pointless. So yes, I'd suggest to remove it.

DenLilleMand: @darjoo just FYI, I followed the same logic as `"Advanced Settings Ext."::OnBeforeOpenCompanySettings()` in BaseApp. In my opinion it seems like the right thing to do to give in the PageID in case they want to do something, put IsHandled := true and then open the page. It wouldn't make a difference right now, but if we add more logic than opening the page if it's not handled it might make sense in the future. What do you think about that?

grobyns: Why not 
```
if not IsHandled then
    Page.Run(Page::"Feature Management")
else
    Page.Run(FeatureManagementPageID);
```

? 

The event is triggered because you want to open a page. The method exists to open a page, why would it not open a page?

grobyns: also: blindly copying patterns from elsewhere is a surefire way to copy bugs, design mistakes and other issues. There might be good reasons why the code in "Advanced Settings Ext."::OnBeforeOpenCompanySettings() is the way it is but if you don't know why then best not to copy it...

DenLilleMand: @grobyns @darjoo I changed the code now, is this what you meant?

src/System Application/App/Feature Key/src/FeatureManagementImpl.Codeunit.al
```diff
@@ -560,4 +560,19 @@ codeunit 2610 "Feature Management Impl."
...
-}
+
+    [EventSubscriber(ObjectType::Codeunit, Codeunit::"Feature Management Triggers", OpenFeatureManagement, '', false, false)]
+    local procedure DefaultOpenFeatureManagement()
+    var
+        FeatureManagementFacade: Codeunit "Feature Management Facade";
+        IsHandled: Boolean;
+        FeatureManagementPageID: Integer;
+    begin
+        FeatureManagementPageID := Page::"Feature Management";
```
encimita, line 571: Not sure we actually need this line

encimita (approved):

src/System Application/App/Feature Key/src/FeatureManagementImpl.Codeunit.al
darjoo: Yes

src/System Application/App/Feature Key/src/FeatureManagementImpl.Codeunit.al
darjoo: Seems redundant since the subscriber would be overwriting the value. No harm either though.

darjoo (approved):

grobyns (approved):

PR #671, opened 2024-03-01
Properly handle paging (max. results and marker) for ListDirectory()

### LINKED ISSUE #509: [Bug]: Azure File Share ListDirectory not returning all files
### Describe the issue

The new Azure File Share functionality introduced in BC23.3 is a great new addition. Unfortunately, the ListDirectory function only returns the first 5000 files in the share and does not implement the Next Marker feature.

### Expected behavior

The API for the Azure File Share indicates that the ListDirectory will return only the first 5000 files in a shared folder, and it returns a NextMarker string in the response if there are further files to be listed. This NextMarker can be used on subsequent requests to retrieve the remaining files

https://learn.microsoft.com/en-us/rest/api/storageservices/list-directories-and-files

I would have expected this NextMarker to be part of the response from ListDirectory BC implementation. (It is present in the ListFileHandles function).

### Steps to reproduce

Create an Azure File Share, and a folder containing more than 5000 items.
Call the AFSFileClient.ListDirectory, and the returned file list contains only the first 5000 items.

        Authorization := StorageServiceAuthorization.UseReadySAS(AzureStorageSetup."Shared Access Signature");
        AFSFileClient.Initialize(AccountName, FileShare, Authorization);
        AFSOperationResponse := AFSFileClient.ListDirectory(FolderName, TempAFSDirectoryContent);

It is not possible to retrieve a list of the remaining files.

### Additional context

No additional context.

MODUSCarstenScholling:
#### Summary
Properly handle paging (max. results and marker) for ListDirectory(). This has been aligned with the existing implementation of ListFileHandles(). Furthermore, error returns are handled properly.

#### Work Item(s)
Fixes #509 

Fixes [AB#503159](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/503159)

### DIFF
--- a/src/System Application/App/Azure File Services API/src/AFSDirectoryContent.Table.al
+++ b/src/System Application/App/Azure File Services API/src/AFSDirectoryContent.Table.al
@@ -113,6 +113,11 @@ table 8950 "AFS Directory Content"
             DataClassification = SystemMetadata;
             Caption = 'URI', Locked = true;
         }
+        field(120; "Next Marker"; Text[2048])
+        {
+            DataClassification = SystemMetadata;
+            Caption = 'NextMarker';
+        }
     }
 
     keys
--- a/src/System Application/App/Azure File Services API/src/AFSDirectoryContent.Table.al
+++ b/src/System Application/App/Azure File Services API/src/AFSDirectoryContent.Table.al
@@ -115,7 +115,7 @@ table 8950 "AFS Directory Content"
         }
         field(120; "Next Marker"; Text[2048])
         {
-            DataClassification = SystemMetadata;
+            DataClassification = CustomerContent;
             Caption = 'NextMarker';
         }
     }
--- a/src/System Application/App/Azure File Services API/src/AFSFileClientImpl.Codeunit.al
+++ b/src/System Application/App/Azure File Services API/src/AFSFileClientImpl.Codeunit.al
@@ -186,8 +186,11 @@ codeunit 8951 "AFS File Client Impl."
         DirectoryURI := AFSHelperLibrary.GetDirectoryPathFromResponse(ResponseText);
 
         AFSHelperLibrary.DirectoryContentNodeListToTempRecord(DirectoryURI, DirectoryPath, NodeList, PreserveDirectoryContent, AFSDirectoryContent);
-        AFSDirectoryContent."Next Marker" := CopyStr(AFSHelperLibrary.GetNextMarkerFromResponse(ResponseText), 1, MaxStrLen(AFSDirectoryContent."Next Marker"));
-        AFSDirectoryContent.Modify();
+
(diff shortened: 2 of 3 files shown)


JesperSchulz (dismissed):

src/System Application/App/Azure File Services API/src/AFSDirectoryContent.Table.al
```diff
@@ -113,6 +113,11 @@ table 8950 "AFS Directory Content"
             DataClassification = SystemMetadata;
             Caption = 'URI', Locked = true;
         }
+        field(120; "Next Marker"; Text[2048])
+        {
+            DataClassification = SystemMetadata;
```
darjoo: What values can show up in "Next Marker"? Thinking if it's really `SystemMetadata` or `CustomerContent`

MODUSCarstenScholling: @darjoo Values like "'1!24!TWVkaXVtRGlyZWN0b3J5PzY-'". I assume an encoded position or something. Nothing really customer content related. Beside that, the content is only temporary.

darjoo: I double checked with our privacy champ, and it should be labeled `CustomerContent` as it is data provided by our customers. 
It should be labeled as `SystemMetadata` in Azure File Services, but it is treated as an external source of data from the customer coming into BC.

MODUSCarstenScholling: Updated. I would suggest reviewing Parent Directory, Etag, fields in the AFSHandle table and all tables in Blob Services module after this PR.

darjoo: Agreed, I've noted that with my privacy champ that we need to review, not just this table but others as well.

darjoo (approved):

JesperSchulz (approved):

src/System Application/App/Azure File Services API/src/AFSFileClientImpl.Codeunit.al
Pieterheinevdm, line 1: AFSDirectoryContent.Modify() raises an **error** if NodeList.Count = 0 then because in this case no AFSDirectoryContent was inserted

image

JesperSchulz: Could you create a new issue for that? We're not revisiting merged / closed PRs / issues.

Pieterheinevdm: https://github.com/microsoft/BCApps/issues/1409

PR #699, opened 2024-03-07
Temporarily disable AA0139

samyhaff:
#### Summary Set AA0139 as warning in the CodeCop ruleset, following the changes for reporting potential overflow in conversions in method invocation.

Related to [AB#481371](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/481371/).

### DIFF
--- a/src/rulesets/CodeCop.ruleset.json
+++ b/src/rulesets/CodeCop.ruleset.json
@@ -556,7 +556,8 @@
                   },
                   {
                       "id":  "AA0139",
-                      "action":  "Warning"
+                      "action":  "None",
+                      "justification": "This rule reports potential text overflows."
                   },
                   {
                       "id":  "AA0140",
--- a/src/rulesets/CodeCop.ruleset.json
+++ b/src/rulesets/CodeCop.ruleset.json
@@ -556,7 +556,7 @@
                   },
                   {
                       "id":  "AA0139",
-                      "action":  "Error"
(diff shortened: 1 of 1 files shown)


grobyns (changes requested):
We do not use Warning. Your choices are None or Error. If you set it to None, please note you will probably never re-enable it...

src/rulesets/CodeCop.ruleset.json
```diff
@@ -556,7 +556,8 @@
                   },
                   {
                       "id":  "AA0139",
-                      "action":  "Error"
+                      "action":  "None",
```
grobyns, line 559: I should also note:
Changing the rule here, in CodeCop.ruleset,json indicated the intent that the rule does not and will never apply to any MS code, anywhere, ever. 

IFF your intention was to 'temporarily' disable the rule then you should add the 'exception' to the ruleset.json file. 

The effect will be the same, the intent you indicate quite different.

Of course I am still of the opinion the rule should not be disabled at all, not even temporarily.

samyhaff: Since warnings are not used, then yes the intent is to temporarily disable it so that violations can be fixed.

AlexanderYakunin:
Why do we need to change this rule at all? The right and proven way is to fix issues first and setting Error helps to do that. If you set None, you will hide the issue instead of fixing it.

PR #714, opened 2024-03-09
OpenChangePasswordDialog new rules #328

### LINKED ISSUE #328: Event Request on Changing Password
**🕮 Describe the issue**
I would like to add an event that allows you to implement password policy on old and new password while changing your user passwords

**🔧 To Reproduce**

The current code only has only one integration event to modify password minimum length only

**📣 Expected behavior**

The current code works fine, this is an additional feature. You can then check that users don't reenter their old passwords asthe new password

**Screenshots**
image

**Additional context**

Drakonian:
#### Summary 

If OldPassword is passed, we check that the new password should match the OldPassword
Also in this case we check that the old password matches the entered old password
New test

#### Work Item(s) 
Fixes #328 

Fixes [AB#506715](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/506715)


src/System Application/App/Password/src/PasswordDialog.Page.al
```diff
@@ -27,6 +27,12 @@ page 9810 "Password Dialog"
                 ExtendedDatatype = Masked;
                 ToolTip = 'Specifies the current password, before the user defines a new one.';
                 Visible = ShowOldPassword;
+                trigger OnValidate()
+                begin
+                    if OldPasswordToCompare <> '' then
```
grobyns: All logic should be moved the impl codeunit

src/System Application/App/Password/src/PasswordDialog.Page.al
```diff
@@ -85,12 +95,16 @@ page 9810 "Password Dialog"
...
+        PasswordSameAsOldErr: Label 'The new password cannot be the same as the old password.';
+        OldPasswordMismatchErr: Label 'The old password does not match the entered password.';
         [NonDebuggable]
         PasswordValue: Text;
         [NonDebuggable]
         ConfirmPasswordValue: Text;
         [NonDebuggable]
         OldPasswordValue: Text;
+        [NonDebuggable]
+        OldPasswordToCompare: Text;
```
grobyns: why are these not SecretText?

Drakonian: @grobyns
Actually the reason is that on the page the user cannot interact directly with SecretText, variables with type text are used for input. Therefore, it was more convenient to compare text with text at once. 

For the same reason, I decided to place the logic of checks directly on the page.
Anyway, I'll fix what you asked for )

Drakonian:
PR is updated

src/System Application/App/Password/src/PasswordDialogImpl.Codeunit.al
```diff
@@ -99,15 +99,15 @@ codeunit 9811 "Password Dialog Impl."
     [NonDebuggable]
     procedure ValidateOldPasswordMatch(OldPasswordToCompare: SecretText; OldPasswordEntered: SecretText)
     begin
-        if OldPasswordToCompare.Unwrap() <> '' then
+        if not OldPasswordToCompare.IsEmpty() then
```
Drakonian: It is better to use the built-in IsEmpty() function to check for emptiness

JesperSchulz:
Build is failing. Please update PR and let's do another run 😊

Drakonian:
Updated

src/System Application/App/Password/src/PasswordDialogImpl.Codeunit.al
```diff
@@ -92,5 +95,21 @@ codeunit 9811 "Password Dialog Impl."
...
+    begin
+        if not OldPasswordToCompare.IsEmpty() then
+            if OldPasswordToCompare.Unwrap() <> OldPasswordEntered.Unwrap() then
+                Error(OldPasswordMismatchErr);
+    end;
+
+    [NonDebuggable]
+    procedure ValidateNewPasswordUniqueness(OldPasswordToCompare: SecretText; NewPassword: SecretText)
+    begin
+        if not OldPasswordToCompare.IsEmpty() then
```
JesperSchulz: You could reduce nesting here by by doing an "if isempty then exit". Just a suggestion.

JesperSchulz (approved):

src/System Application/App/Password/src/PasswordDialogImpl.Codeunit.al
Drakonian: I agree. I often write like that myself, don't know why I haven't done the same here :)

JesperSchulz (dismissed):

src/System Application/App/Password/src/PasswordDialog.Page.al
```diff
@@ -27,6 +27,10 @@ page 9810 "Password Dialog"
                 ExtendedDatatype = Masked;
                 ToolTip = 'Specifies the current password, before the user defines a new one.';
                 Visible = ShowOldPassword;
+                trigger OnValidate()
+                begin
+                    PasswordDialogImpl.ValidateOldPasswordMatch(OldPasswordToCompare, OldPasswordValue);
```
darjoo: Shouldn't we also do the same check `if RequiresPasswordValidation then...` here?

Drakonian: As far as I understood from the discussion of the problem it was decided that this rule should always be active. Besides, `RequiresPasswordValidation` is now used only to check the strength of the password and the user has the option not to enable it. 

Based on all this information, it seems to me that I don't need to be bound to `RequiresPasswordValidation`

JesperSchulz:
I'll do some functional testing of those changes early next week 🙂 If all goes well, we merge!

src/System Application/Test/Password/src/PasswordDialogTest.Codeunit.al
```diff
@@ -243,14 +269,29 @@ codeunit 135033 "Password Dialog Test"
...
+        Assert.IsTrue(PasswordDialog.OldPassword.Visible(), 'Old Password Field should be visible.');
+        Assert.IsTrue(PasswordDialog.ConfirmPassword.Visible(), 'Confirm Password Field should be visible.');
         PasswordDialog.OldPassword.SetValue(InValidPasswordTxt);
         PasswordDialog.Password.SetValue(ValidPasswordTxt);
         PasswordDialog.ConfirmPassword.SetValue(ValidPasswordTxt);
         PasswordDialog.OK().Invoke();
     end;
 
+    [ModalPageHandler]
+    procedure ChangePasswordDialogWitldOldPasswordModalPageHandler(var PasswordDialog: TestPage "Password Dialog");
```
JesperSchulz: There seems to be some typo in the procedure name?
...WitldOld...

src/System Application/Test/Password/src/PasswordDialogTest.Codeunit.al
```diff
@@ -221,6 +224,29 @@ codeunit 135033 "Password Dialog Test"
...
     end;
 
+    [Test]
+    [HandlerFunctions('ChangePasswordDialogWitldOldPasswordModalPageHandler')]
+    procedure OpenChangePasswordDialogWithOldPasswordTest();
+    var
+        Password: SecretText;
+        OldPassword: SecretText;
+    begin
+        // [SCENARIO] Open Password dialog in change password mode. Old password
```
JesperSchulz: Old password?

src/System Application/Test/Password/src/PasswordDialogTest.Codeunit.al
Drakonian: Should be WithOldPassword, fixed

src/System Application/Test/Password/src/PasswordDialogTest.Codeunit.al
Drakonian: Typo removed, thanks

JesperSchulz (dismissed):
Looks good to me. Manual tests conducted successfully.

Drakonian:
Something's broken, the build's been going on for 3 hours.
image

JesperSchulz:
> Something's broken, the build's been going on for 3 hours. image

@mazhelez / @aholstrup1, you're on!

src/System Application/App/Password/src/PasswordDialogImpl.Codeunit.al
```diff
@@ -73,6 +75,7 @@ codeunit 9811 "Password Dialog Impl."
         PasswordDialog: Page "Password Dialog";
     begin
         PasswordDialog.EnableChangePassword();
+        PasswordDialog.SetOldPasswordToCompareSecretValue(OldPassword);
```
AndreasMoth: The OldPassword parameter is a bit confusing now.. So as an input it's the old password that should be compared to, as an output it's the old password the user typed. Now these are not necessarily equal, if the OldPassword input was empty, then the OldPassword output is not necessarily empty. Of course if OldPassword input has a value, then OldPassword output are equal.

src/System Application/App/Password/src/PasswordDialogManagement.Codeunit.al
```diff
@@ -116,7 +116,7 @@ codeunit 9810 "Password Dialog Management"
     /// Opens a dialog for the user to change a password and returns the old and new typed passwords if there is no validation error,
     /// otherwise an empty text are returned.
     /// </summary>
-    /// <param name="OldPassword">Out parameter, the old password user typed on the dialog.</param>
+    /// <param name="OldPassword">Out parameter, the old password to compare with the password user typed on the dialog.</param>
```
AndreasMoth: We should specify OldPassword is also an input parameter. From this I expect the function itself determines the old password, compares and gives me the old password in case I want to do some kind of comparison.

src/System Application/App/Password/src/PasswordDialogImpl.Codeunit.al
Drakonian: Exactly, they can have differences and it's all because the original function had a very strange design where the OldPassword was just what the user entered in the dialog box, all the checks were kind of out of brackets. This in turn created confusion about how to use the function at all. 

https://github.com/microsoft/BCApps/issues/328#issuecomment-1983618268

To support backwards compatibility, when passing a blank to OldPassword (and this function was very often used this way) it will work as it did before the change. But only when OldPassword is filled in will the function actually perform a comparison. In addition, checking that the old password does not match the new one makes sense only when the old password has been sent. Otherwise there is nothing to check.

So the logic is as follows
Input OldPassword = '' works as before, we just return the entered password, no checks.
Input OldPassword <> '' if we do pass OldPassword it means that we assume a specific OldPassword and if it doesn't match the password the user fills in, an error is inevitable and necessary. Only in this case there is any chance to compare if OldPassword doesn't match NewPassword.

I'd say more confusing was how the function worked previously

src/System Application/App/Password/src/PasswordDialogManagement.Codeunit.al
Drakonian: I agree now this is input and output parameter

Drakonian: It probably makes no sense to return OldPassword at all, since we introduce checks and standards at the PasswordDialogManagment level.

Perhaps the best way would be to mark the old method as obsolete in future releases and create a new one (but would have to give a new name, number of parameters the same).

or we keep as it is (but specify param as In & Out)

@AndreasMoth what do you think?

AndreasMoth: Yes, that makes sense, to at least remove "var" from the OldPassword (through obsoletion, remember CLEAN tags).
You can also argue if we should even return the password, why does the caller need to get that password? But for now, just remove the "var" from OldPassword. We could also rename OldPassword to CurrentPassword while Password should be NewPassword (after all, you give the current password and get out the new password - renaming is not super important though).

Drakonian:
@AndreasMoth @JesperSchulz 

I've updated the code and updated the variable names.

I would like to point out that this strange design when OldPassword was returned was only needed for OnPrem report 9810 “Change Password”. 
The point is that it's quite an unusual situation when we need to return the entered OldPassword and at the same time we can't find out which password is actually correct. This information is not available and its validation is done in Database.ChangeUserPassword(Text, Text)
https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/methods-auto/database/database-changeuserpassword-method

My original code supported in/out at the same time, which essentially supported the Change Password report scenario. In order to keep this support in the report the old implementation is used instead of the new one.

Let me know what you think about it.

AndreasMoth (approved):

src/System Application/App/Password/src/PasswordDialogManagement.Codeunit.al
```diff
@@ -77,6 +77,18 @@ codeunit 9810 "Password Dialog Management"
...
 #pragma warning restore AL0432
     end;
+
+    /// <summary>
+    /// Opens a dialog for the user to change a password and returns the old and new typed passwords if there is no validation error,
+    /// otherwise an empty text are returned.
+    /// </summary>
+    /// <param name="OldPassword">Out parameter, the old password user typed on the dialog.</param>
+    /// <param name="Password">Out parameter, the new password user typed on the dialog.</param>
+    [Obsolete('Replaced by OpenPasswordChangeDialog without out OldPassword param', '24.0')]
```
JesperSchulz:
```suggestion
    [Obsolete('Replaced by OpenPasswordChangeDialog without out OldPassword param', '25.0')]
```

JesperSchulz (approved):

JesperSchulz:
@Drakonian, test compilation now seems to fail. Will you investigate?

Drakonian:
@JesperSchulz

Yes of course, the problem is that Test Password is a separate app and it cannot access internal implementation from System Application. The strangest thing is that I didn't get this error in testing locally, probably the build system in the project is still not perfect.

I will now update the test that refers to internal codeunit.

grobyns:
> @JesperSchulz
> 
> Yes of course, the problem is that Test Password is a separate app and it cannot access internal implementation from System Application. The strangest thing is that I didn't get this error in testing locally, probably the build system in the project is still not perfect.
> 
> I will now update the test that refers to internal codeunit.

FYI: tests should not have direct access to internals. If some internals access is needed, then it should be through a test library app which encapsulates the internal and exposes some public method allowing what you need. If you can avoid using internals completely, that's great, if not, please add methods to a test library :-)

Drakonian:
@JesperSchulz 
Ready to check

JesperSchulz (dismissed):

Drakonian:
@JesperSchulz broken builder?

image

id looks correct

image

Drakonian:
image

JesperSchulz:
@Drakonian, no, there's still a permission set with ID 135033. Now that you've changed the app.json, that ID is outside of the ID range - hence the compilation error. Add the range from 135033 to 135033, and that issue should be gone.

Drakonian:
> @Drakonian, no, there's still a permission set with ID 135033. Now that you've changed the app.json, that ID is outside of the ID range - hence the compilation error. Add the range from 135033 to 135033, and that issue should be gone.

Corrected. 
Do you have any idea why this is compiling locally? 😮

JesperSchulz:
> > @Drakonian, no, there's still a permission set with ID 135033. Now that you've changed the app.json, that ID is outside of the ID range - hence the compilation error. Add the range from 135033 to 135033, and that issue should be gone.
> 
> Corrected to old ID range. Do you have any idea why this is compiling locally? 😮

You are probably only compiling the entire System Application in one go, and not the modules individually - is my guess 🙂

Drakonian:
> > > @Drakonian, no, there's still a permission set with ID 135033. Now that you've changed the app.json, that ID is outside of the ID range - hence the compilation error. Add the range from 135033 to 135033, and that issue should be gone.
> > 
> > 
> > Corrected to old ID range. Do you have any idea why this is compiling locally? 😮
> 
> You are probably only compiling the entire System Application in one go, and not the modules individually - is my guess 🙂

I feel a little confused about having to run the CI process several times and it finds new errors. It would be cool to run it locally somehow 😞 
I hope you're not wasting too much time on this

new correction ready, should be last 😈

JesperSchulz:
No worries. I just click a button 🙂 It's perfectly normal that you need to submit several updates before builds succeed. Of course you can (and should) enable all code analyzers with the correct rules, but things like compiling the modules individually or running CLEAN builds, is too much to ask. We all here in the BC team use CI to verify those things.

JesperSchulz (approved):

AndreasMoth (approved):

PR #735, opened 2024-03-12
Marketing Text: JA/Japanese: Remove lower boundary check as this blocks languages where spaces are…

henrikfrovst:
#### Summary 
Remove lower boundary check as this blocks languages where spaces are not used.

#### Work Item(s)
Fixes # [AB#506451](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/506451)

### DIFF
--- a/src/System Application/App/Entity Text/src/EntityTextImpl.Codeunit.al
+++ b/src/System Application/App/Entity Text/src/EntityTextImpl.Codeunit.al
@@ -359,8 +359,6 @@ codeunit 2012 "Entity Text Impl."
                 begin
                     SplitCompletion := Completion.Split(EncodedNewlineTok + EncodedNewlineTok);
                     FormatValid := SplitCompletion.Count() = 2; // a tagline + paragraph must contain an empty line
-                    if FormatValid then
-                        FormatValid := SplitCompletion.Get(2).Split(' ').Count() >= MinParagraphWords; // the paragraph must be more than MinParagraphWords words
                 end;
             TextFormat::Paragraph:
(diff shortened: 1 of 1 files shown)


src/System Application/App/Entity Text/src/EntityTextImpl.Codeunit.al
```diff
@@ -359,8 +359,6 @@ codeunit 2012 "Entity Text Impl."
                 begin
                     SplitCompletion := Completion.Split(EncodedNewlineTok + EncodedNewlineTok);
                     FormatValid := SplitCompletion.Count() = 2; // a tagline + paragraph must contain an empty line
-                    if FormatValid then
```
msft-sam, line 362: Should we keep the check for English?

henrikfrovst: Not sure why it was there in the first place, other than ensuring some text? but perhaps we can change it to go for characters rather than words?

msft-sam: I guess characters could work - it was mostly there to ensure we don't give a very short paragraph of text

qasimikram: In my opinion we should have a rule in prompt how long a description we need rather then having such condition in AL. Shorter result is still better then no result IMO :)

qasimikram (approved):
Look good to me.

JesperSchulz (approved):

PR #785, opened 2024-03-19
[24.0] Obsolete legacy BusinessChart addin

stkillen:
#### Summary 
The control addin "Microsoft.Dynamics.Nav.Client.BusinessChart" was removed from the App DB in version 24.0. Any usages of this will fail or new tenants.

Introduce this addin as Obsolete = Pending to ensure that the removal from the App DB is non-breaking.

#### Work Item(s) 
Fixes [AB#507953](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/507953)

### DIFF
--- a/src/System Application/App/ControlAddIns/src/MicrosoftDynamicsNavClientBusinessChart.ControlAddin.al
+++ b/src/System Application/App/ControlAddIns/src/MicrosoftDynamicsNavClientBusinessChart.ControlAddin.al
@@ -1,0 +1,60 @@
+#if not CLEAN24
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System;
+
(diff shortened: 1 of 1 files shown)


darjoo (approved):

src/System Application/App/ControlAddIns/src/MicrosoftDynamicsNavClientBusinessChart.ControlAddin.al
```diff
@@ -0,0 +1,60 @@
+#if not CLEAN24
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System;
```
stkillen: Ideally should not be within a namespace, but globally available until removed.

```suggestion
```

AndreasMoth (approved):

darjoo (approved):

JesperSchulz (approved):

PR #794, opened 2024-03-20
Implement different fallback mechanism when determining baselines for main branch

mazhelez:
Automation for updating app baselines on `main` branch should always look for baselines version in `bcinsider` storage account.

[AB#502086](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/502086)

### DIFF
--- a/build/scripts/EnlistmentHelperFunctions.psm1
+++ b/build/scripts/EnlistmentHelperFunctions.psm1
@@ -206,7 +206,15 @@ function Get-PackageLatestVersion() {
                 }
             }
 
-            return Get-LatestBCArtifactVersion -minimumVersion $minimumVersion
+            $currentBranch = Get-CurrentBranch
+            $storageAccountOrder = @("bcartifacts", "bcinsider")
+            if($currentBranch -eq "main") {
+                # Always use bcartifacts for main branch
+                $storageAccountOrder = @("bcinsider")
+            }
+
+
+            return Get-LatestBCArtifactVersion -minimumVersion $minimumVersion -storageAccountOrder $storageAccountOrder
         }
         default {
             throw "Unknown package source: $($package.Source)"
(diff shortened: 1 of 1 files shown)


build/scripts/EnlistmentHelperFunctions.psm1
```diff
@@ -206,7 +206,15 @@ function Get-PackageLatestVersion() {
                 }
             }
 
-            return Get-LatestBCArtifactVersion -minimumVersion $minimumVersion
+            $currentBranch = Get-CurrentBranch
+            $storageAccountOrder = @("bcartifacts", "bcinsider")
+            if($currentBranch -eq "main") {
```
aholstrup1, line 211: Shouldn't we always use bcinsider for .x branches well?

aholstrup1 (dismissed):

build/scripts/EnlistmentHelperFunctions.psm1
mazhelez: No, actually, most of the time we should use _bcartifacts_ for baselines for .x branches. Only around major branching, when the .0 branch isn't shipped yet, we'll need to fallback to _bcinsider_.

build/scripts/EnlistmentHelperFunctions.psm1
```diff
@@ -206,7 +206,15 @@ function Get-PackageLatestVersion() {
...
+                # Always use bcartifacts for main branch
```
mazhelez:
```suggestion
                # Always use bcinsider for baselines for the main branch
```

build/scripts/EnlistmentHelperFunctions.psm1
aholstrup1: Ah yes you're right.. I was thinking of artifact versions

aholstrup1 (approved):

jonasblunck (approved):

PR #825, opened 2024-03-25
Updated devcontainer experience

aholstrup1:
#### Summary 
This PR would add a script to set up a dev container for BCApps. The script uses containerhelper to set up a standard container. Once a container is set up, all the apps are moved to dev scope and the AL projects in the repo get a .vscode folder generated with a launch.json and settings.json

#### Work Item(s) 
Fixes [AB#521408](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/521408)


build/scripts/DevEnv/ALDev.psm1
```diff
@@ -0,0 +1,219 @@
...
+
+.PARAMETER LaunchSettings
+A hash table containing the settings that should be set in the default launch configuration.
+
+.PARAMETER ProjectSettings
+A hash table containing the settings that should be set in the project settings.
+#>
+function Configure-ALProject(
+    [Parameter(Mandatory = $true)]
+    [string]$ProjectFolder,
```
mazhelez, line 115:
```suggestion
    [string] $ProjectFolder,
```

build/scripts/DevEnv/NewDevContainer.psm1
```diff
@@ -0,0 +1,242 @@
...
+    .DESCRIPTION
+    Checks if the specified database exists on the SQL server
+    .PARAMETER DatabaseName
+    The name of the database to check
+    .PARAMETER DatabaseServer
+    The hostname of the SQL server
+    .OUTPUTS
+    Returns true if database exists otherwise false
+#>
+function Test-NavDatabase()
```
mazhelez: Maybe skip the `Nav` part. Or make it `BC`?

build/scripts/DevEnv/NewDevContainer.psm1
```diff
@@ -0,0 +1,242 @@
...
+
+    UPDATE [$DatabaseName].[dbo].[NAV App Installed App]
+    SET [Version Major] = $Major, [Version Minor] = $Minor, [Version Build] = 0, [Version Revision] = 0
+    WHERE Name = '$Name' and Publisher = '$Publisher';
+
+    UPDATE [$DatabaseName].[dbo].[`$ndo`$navappschematracking]
+    SET [version] = '$Major.$Minor.0.0', [baselineversion] = '$Major.$Minor.0.0'
+    WHERE [name] = '$Name' and [publisher] = '$Publisher';
+"@
+
```
mazhelez, line 118:
```suggestion
```

build/scripts/DevEnv/NewDevContainer.psm1
```diff
@@ -0,0 +1,242 @@
...
+        [string]$Minor,
+        [Parameter(Mandatory = $false)]
+        [string]$Publisher = 'Microsoft',
+        [Parameter(Mandatory = $false)]
+        [string]$DatabaseServer = '.'
+    )
+
+    if (!$Major)
+    {
+        throw "The `$env:BUILDVERSION is 0. Check if the environment has been configured correctly."
```
mazhelez: This message seems off.

build/scripts/DevEnv/NewDevContainer.psm1
```diff
@@ -0,0 +1,242 @@
...
+    .PARAMETER Major
+    The major version number
+    .PARAMETER Minor
+    The minor version number
+    .PARAMETER Publisher
+    The publisher of the extension
+    .PARAMETER DatabaseServer
+    The hostname of the SQL server
+#>
+function Set-ExtensionVersion()
```
mazhelez: Can we call it "app" and not "extension"?

build/scripts/DevEnv/NewDevContainer.psm1
```diff
@@ -0,0 +1,242 @@
...
+        } finally {
+            Write-Host "Starting server instance $($server.ServerInstance)" -ForegroundColor Green
+            Start-NAVServerInstance -ServerInstance $server.ServerInstance
+        }
+
+
+    } -argumentList $NewDevContainerModule,$RepoVersion -usePwsh $false
+}
+
+Export-ModuleMember -Function *-*
```
mazhelez: Better list functions one by one.

mazhelez (commented):
Looks great!

Left minor comments here and there.

build/scripts/DevEnv/NewDevContainer.ps1
```diff
@@ -0,0 +1,34 @@
+[CmdletBinding()]
+param(
+    [Parameter(Mandatory = $false)]
+    [string] $ContainerName = "BC-$(Get-Date -Format 'yyyyMMdd')",
+    [Parameter(Mandatory = $false)]
+    [ValidateSet('Windows', 'UserPassword')]
+    [string] $Authentification = "UserPassword"
```
mazhelez: "Authentication"?

build/scripts/DevEnv/NewDevEnv.ps1
```diff
@@ -1,110 +1,83 @@
...
-        The password to use for the container.
-    .Parameter projectPaths
-        The paths of the AL projects to build. May contain wildcards.
-    .Parameter workspacePath
-        The path of the workspace to build. The workspace file must be in JSON format.
-    .Parameter alGoProject
-        The path of the AL-Go project to build.
-    .Parameter packageCacheFolder
-        The folder to store the built artifacts.
+    .SYNOPSIS
```
mazhelez, line 2: The help words are not case-sensitive, but we better align the PS doc across the repo. I see 'Synopsis' and 'SYNOPSIS' all around: https://github.com/search?q=repo%3Amicrosoft%2FBCApps+.Synopsis&type=code

Copilot generates 'Synopsis' for me 😄

build/scripts/DevEnv/NewDevEnv.ps1
```diff
@@ -1,110 +1,83 @@
...
-    [Parameter(Mandatory = $true, ParameterSetName = 'ProjectPaths')]
-    [string[]] $projectPaths,
-
-    [Parameter(Mandatory = $true, ParameterSetName = 'WorkspacePath')]
-    [string] $workspacePath,
-
-    [Parameter(Mandatory = $true, ParameterSetName = 'ALGoProject')]
-    [string] $alGoProject,
-
+    [ValidateSet('Windows', 'UserPassword')]
```
mazhelez: There was something about "Windows" authentication that didn't work as expected. Freddy might remember what it was.

build/scripts/DevEnv/NewDevEnv.ps1
```diff
@@ -1,110 +1,83 @@
...
+    .DESCRIPTION
+    This script creates a new development environment with a new container, sets up vscode for development against the container, and optionally builds and publishes apps.
+    .PARAMETER ContainerName
+    The name of the container to create. The default value is "BC-$(Get-Date -Format 'yyyyMMdd')".
+    .PARAMETER Authentication
+    The authentication type to use when creating the container. The default value is "UserPassword".
+    .PARAMETER SkipVsCodeSetup
+    If specified, vscode will not be set up for development against the container.
+    .PARAMETER ProjectPaths
+    The paths to the AL projects to build and publish. This parameter is mutually exclusive with the WorkspacePath and AlGoProject parameters.
```
mazhelez, line 13: This doesn't look enforced, and it looks like it's allowed to enter all three. Not sure what could be the side effect of that.

build/scripts/DevEnv/NewDevEnv.psm1
```diff
@@ -221,19 +129,20 @@ function BuildApp {
...
+        Write-Host "App $appFile already exists in $appOutputFolder. Removing..."
         $appFile = (Join-Path $appOutputFolder $appFile -Resolve)
-    } else {
-        # Create compiler folder on demand
-        if(-not $compilerFolder.Value) {
-            Write-Host "Creating compiler folder..." -ForegroundColor Yellow
-            $compilerFolder.Value = CreateCompilerFolder -packageCacheFolder $packageCacheFolder
-            Write-Host "Compiler folder: $($compilerFolder.Value)" -ForegroundColor Yellow
-        }
+        Remove-Item -Path $appFile -Force
```
mazhelez: Why?
Wouldn't this mean that apps could be built multiple times unnecessarily?

mazhelez (commented):
Intermediate review. I'll try to circle back later on.
See comments below.

One thing I noticed is that building apps and creating a container is not done in parallel now. Why not?

build/scripts/DevEnv/NewDevEnv.ps1
aholstrup1: It generates SYNOPSIS for me 😄 Probably why we've been a bit inconsistent with that. Let's address that in a different PR though

build/scripts/DevEnv/NewDevEnv.ps1
aholstrup1: I haven't tested it but I noticed that the Resolve-ProjectPaths helper function you wrote allows for all three and just appends the results together 🤷‍♂️ I can test out if it works

build/scripts/DevEnv/NewDevEnv.ps1
aholstrup1: Yeah there is :) I'll talk to Freddy to see if he can help out. Otherwise, we can start with only UserPassword

build/scripts/DevEnv/NewDevEnv.psm1
aholstrup1: You're right.. I changed it because the existing implementation didn't recompile e.g. the System App even if I wanted it to be rebuilt. I noticed there is a $rebuild variable in the existing implementation but it's never set anywhere. Perhaps I can leverage that. 

```
if((Test-Path (Join-Path $appOutputFolder $appFile)) -and (-not $rebuild)) {
        Write-Host "App $appFile already exists in $appOutputFolder. Removing..."
        $appFile = (Join-Path $appOutputFolder $appFile -Resolve)
        Remove-Item -Path $appFile -Force
    }
```

I just want to make sure we recompile e.g. the System App if I run `.\build\scripts\DevEnv\NewDevEnv.ps1 -ProjectPaths '.\src\System Application\App'` even if it was already previously compiled.

build/scripts/DevEnv/NewDevEnv.ps1
mazhelez: As long as there are no duplicate projects, it's fine to allow all three parameters.

build/scripts/DevEnv/NewDevEnv.psm1
mazhelez: What if `.\build\scripts\DevEnv\NewDevEnv.ps1 -ProjectPaths '.\src\System Application\*'` fails on building the last app. Do you expect re-running the script to re-compile all apps?
Same goes for re-publishing the apps.

aholstrup1: Changed this behaviour now such that we'll try to use the cached apps if they exist. If you don't want to used the cached apps you should add the -RebuildApps flag to NewDevEnv.ps1

build/scripts/DevEnv/NewDevEnv.ps1
aholstrup1: Let's look at this in another iteration :)

build/scripts/DevEnv/NewDevEnv.ps1
```diff
@@ -91,20 +79,18 @@ if($createContainerJob) {
...
 
-Write-Host "Publishing apps took $($publishingAppsStats.TotalSeconds) seconds"
+# Set up vscode for development against the container (i.e. set up launch.json and settings.json)
+if (-not $SkipVsCodeSetup)
+{
+    Write-Host "Configuring vscode for development against the container..."
+    Configure-ALProjectsInPath -ContainerName $ContainerName -Authentication $Authentication
+}
 
+Write-Host "Development environment setup completed successfully" -ForegroundColor Green
```
mazhelez: If the apps failed publishing, then was the dev env setup really successful?

Consider altering the message in case of partial failure. Interesting idea could be to report 3 different statuses: on container creation, on building apps and on publishing apps.

E.g.:
```
✅Container BCApps-123 created successfully
✅Apps app1 and app2 built successfully
❌Failed to publish the following apps:...
```
or
```
❌Container BCApps-123 failed to be created.
✅Apps app1 and app2 built successfully
🟡Skipped publishing apps as container was not created
```

aholstrup1: Made something like that (ish) 😄 Not as nice but it should still give an accurate representation of what the script did

mazhelez (approved):

build/scripts/DevEnv/ALDev.psm1
```diff
@@ -0,0 +1,225 @@
...
+        {
+            $CustomLaunchSettings[$propertyName] = $DefaultLaunchSettings[$propertyName]
+        }
+    }
+    return $CustomLaunchSettings
+}
+
+function GetDefaultSettings([switch]$LaunchJson, [switch]$SettingsJson)
+{
+    if ($LaunchJson)
```
ankosiorek, line 42: There is also an option that both switch would be provided, not sure if that's breaking or sth ;)

ankosiorek (approved):

build/scripts/DevEnv/ALDev.psm1
aholstrup1: Yeah that's true. Wouldn't really make sense though 😄 We only use this internally so for now I think it's fine

PR #827, opened 2024-03-25
Add Support to specifiy ODataQuery Parameters in Microsoft Graph Client

### LINKED ISSUE #802: [BC Idea]: Improve Microsoft Graph Module with OData Query Parameters
### BC Idea Link

https://experience.dynamics.com/ideas/idea/?ideaid=e0452d86-70e7-ee11-a73e-002248504629

### Description

We cannot specify ODataQuery Parameter as optional paramter.

This should be added to work appropriate with microsoft graph endpoints.

Internal work item: [AB#522447](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/522447)

pri-kise:
#### Summary
This PR adds support so specify OData Query Parameters via the Microsoft Graph Client

#### Work Item(s)
Fixes #802

Fixes [AB#522447](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/522447)

Fixes [AB#527604](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/527604)

### DIFF
--- a/src/System Application/App/MicrosoftGraph/src/GraphOptionalParameters.Codeunit.al
+++ b/src/System Application/App/MicrosoftGraph/src/GraphOptionalParameters.Codeunit.al
@@ -28,30 +28,13 @@ codeunit 9353 "Graph Optional Parameters"
     end;
 
     /// <summary>
-    /// Sets the value for 'If-None-Match' HttpHeader for a request.
+    /// Sets the value for a HttpHeader for a request.
     /// </summary>
-    /// <param name="Value">Text value specifying the HttpHeader value</param>
-    procedure SetIfNoneMatchRequestHeader("Value": Text)
+    /// <param name="GraphRequestHeader">The Request Header</param>
+    /// <param name="HeaderValue">Text value specifying the HttpHeader value</param>
+    procedure SetRequestHeader(GraphRequestHeader: Enum "Graph Request Header"; HeaderValue: Text)
     begin
-        GraphOptionalParametersImpl.SetIfNoneMatchRequestHeader("Value");
-    end;
-
-    /// <summary>
-    /// Sets the value for 'Prefer' HttpHeader for a request.
-    /// </summary>
-    /// <param name="Value">Text value specifying the HttpHeader value</param>
-    procedure SetPreferRequestHeader("Value": Text)
-    begin
-        GraphOptionalParametersImpl.SetPreferRequestHeader("Value");
-    end;
-
-    /// <summary>
-    /// Sets the value for 'ConsistencyLevel' HttpHeader for a request.
-    /// </summary>
-    /// <param name="Value">Text value specifying the HttpHeader value</param>
-    procedure SetConsistencyLevelRequestHeader("Value": Text)
-    begin
-        GraphOptionalParametersImpl.SetConsistencyLevelRequestHeader("Value");
+        GraphOptionalParametersImpl.SetRequestHeader(GraphRequestHeader, HeaderValue);
     end;
 
     internal procedure GetRequestHeaders(): Dictionary of [Text, Text]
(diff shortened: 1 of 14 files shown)


src/System Application/App/MicrosoftGraph/src/GraphOptionalParameters.Codeunit.al
```diff
@@ -72,23 +69,120 @@ codeunit 9353 "Graph Optional Parameters"
...
 
-    var
-        QueryParameters: Dictionary of [Text, Text];
-        RequestHeaders: Dictionary of [Text, Text];
+    /// <summary>
+    /// Sets the value for the OData Query Parameter '$filter'.
+    /// see: https://learn.microsoft.com/en-us/graph/query-parameters?tabs=http#filter-parameter
+    /// </summary>
+    /// <param name="Value">Text value specifying the query parameter</param>
+    procedure SetODataQueryParameterFilter("Value": Text)
```
AndreasMoth: Looking at all these new functions, I'm thinking it would be simpler to have an enum with Query Parameter options and then all these new functions simply turn into SetOdataQueryParameter(ODataQueryParameter: Enum "OData Query Parameter"; Value: Text). This way it's very easy to add another option, instead of having to add multiple functions here. It's also easier to see have an overview of what you can do in this codeunit. You can set query parameter, you can set request header etc. Then if you are interested in setting query parameter, you can see which are available in the enum. We may even be able to make this enum Extensible, which means Partners won't have to go add new functions every time they need a new ODataQueryParameter.

AndreasMoth: TBH, I feel we should do the same for request header, simply have SetRequestHeader(RequestHeaderType: Enum "Request Header Type"; Value: Text);

pri-kise: I wanted to make it as easy as possible by providing specific procedures for query parameters that only accept an integer value.
By making it an enum the partner would have to apply proper formatting to the value.

What I like on your suggestion is that it might be better for extensibility.

Maybe we can get a 2nd opinion on this topic before I start to refactor everrything.

JesperSchulz: ...which brings us back to the discussion if this isn't a fundamental building block, which should provide all capabilities out of the box, rather than being easily extensible. When it comes to system application modules, I pick "as easy as possible" over "extensibility" any day. But I can see the advantage of both solutions here.
@pri-kise, have you only implemented a small subset of query parameters so far? Are there many more to come?

pri-kise: @JesperSchulz i added all the OData query parameters that are mentioned in the following list:
https://learn.microsoft.com/en-us/graph/query-parameters?tabs=http#odata-system-query-options

pri-kise: @devgar since you provided a recent PR for this module what would you prefer?
@darjoo you've reviewed my original PR. What's your opinion?

devgar: I feel I am no one to give my opinion, but I agree that the more flexibility is better, but this implementation seems complete enough for the MS Graph spec.

JesperSchulz: @devgar, here on GitHub we value everyone's opinion! Please don't ever hold back sharing yours with us 🙂

AndreasMoth: As I see, it would be easy with an enum, you just have two Set functions SetRequestHeader() and SetODataQueryParameter(). As you code, you know whether you want a request header or a query parameter. Once you select it, you then get a list of only the valid headers or only the valid query parameters you can add.
Regarding the Integer, it's actually still stored as a text and provided to the web request as text itself.
So you can simply format it yourself, of course you can give it a string instead and mess it up, but you can do that with any other value as well if you provide an invalid value.

pri-kise: I updated the code with new procedures and enums

pri-kise: @AndreasMoth could you take another review after my changes? :)

AndreasMoth (dismissed):

JesperSchulz (dismissed):
Let us get this one on the road! 🥳

src/System Application/App/MicrosoftGraph/src/GraphOptionalParametersImpl.Codeunit.al
```diff
@@ -0,0 +1,78 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+namespace System.Integration.Graph;
+
+using System.Integration.Graph;
```
JesperSchulz:
```suggestion
namespace System.Integration.Graph;

```

JesperSchulz (approved):

AndreasMoth (approved):

pri-kise:
@JesperSchulz can this maybe be backported to 24.x ?

JesperSchulz:
> @JesperSchulz can this maybe be backported to 24.x ?

Of course. I'll create a PR.

JesperSchulz:
PR created: #1042

PR #826, opened 2024-03-25
Move to generated keys in RSA tests

Groenbech96:
#### Summary 
Add methods to generate XML keys. Use new method to generate keys needed in tests.

#### Work Item(s) 
Fixes [AB#500563](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/500563)

PR #835, opened 2024-03-26
[releases/24.0] Move to generated keys in RSA tests

Groenbech96:
This pull request backports #826 to releases/24.0

Fixes AB#[**Insert Work Item Number Here**]


darjoo:
Does this need to be fixed into 24.0? Isn't 24.x sufficient?

PR #842, opened 2024-03-27
Features/484133 drag and drop attach multiple files (email attachment part)

gggdttt:
#### Summary 
Part of the slice work - support drag and drop multiple file in Email attachment
#### Work Item(s) 
* Also refactor some code 
*  support drag and drop multiple file in Email attachment
* No test for now because cannot init `fileUpload` in AL now, need to wait for the testability from Develop Tool Team and then add tests
*


src/System Application/App/Email/src/Message/EmailAttachments.Page.al
```diff
@@ -53,6 +54,28 @@ page 8889 "Email Attachments"
...
+                Visible = IsEmailEditable;
+
+                trigger OnAction(files: List of [FileUpload])
+                var
+                    EmailEditor: Codeunit "Email Editor";
+                    TempInStream: InStream;
+                    SingleFile: FileUpload;
+                begin
+                    foreach SingleFile in files do begin
+                        SingleFile.CreateInStream(TempInStream, TextEncoding::UTF8);
```
gggdttt, line 73: I am suggested to use UTF8 as default encoding type. 

We have had issues with partners moving their customers from one Windows machine to a different one, and they had different CodePages, which is the MSDos encoding, and that is causing issue.

I am asking if client can support to choose Encoding in the filepicker. If they can do, we can allow user to choose Encoding here.

src/System Application/App/Email/src/Email/Outbox/EmailEditor.Codeunit.al
```diff
@@ -162,18 +162,24 @@ codeunit 8906 "Email Editor"
...
         UploadIntoStream('', '', '', FileName, Instream);
+        UploadAttachmentFromStream(EmailMessageImpl, FileName, Instream);
+    end;
+
+    internal procedure UploadAttachmentFromStream(EmailMessageImpl: Codeunit "Email Message Impl."; FileName: Text; Instream: InStream)
+    var
+        AttachmentName, ContentType : Text[250];
+        AttachmentSize: Integer;
+    begin
         if FileName = '' then
```
haoranpb: You can move this out of the procedure

haoranpb: i.e. 

```
if FileName <> '' then
    UploadAttachmentFromStream(EmailMessageImpl, FileName, Instream);
```

src/System Application/App/Email/src/Message/EmailAttachments.Page.al
```diff
@@ -53,6 +54,28 @@ page 8889 "Email Attachments"
     {
         area(Processing)
         {
+            fileuploadaction(UploadMultiple)
+            {
+                ApplicationArea = All;
+                Image = Attach;
+                Caption = 'Add files';
```
haoranpb, line 61: Assume you discussed this with UI team?

src/System Application/App/Email/src/Email/Outbox/EmailEditor.Codeunit.al
```diff
@@ -162,18 +162,24 @@ codeunit 8906 "Email Editor"
...
             exit;
 
         AttachmentName := CopyStr(FileName, 1, 250);
         ContentType := EmailMessageImpl.GetContentTypeFromFilename(FileName);
-        AttachamentSize := EmailMessageImpl.AddAttachmentInternal(AttachmentName, ContentType, Instream);
+        AttachmentSize := EmailMessageImpl.AddAttachmentInternal(AttachmentName, ContentType, Instream);
```
haoranpb, line 180: I'm actually curious if this Telemetry is ever used...

src/System Application/App/Email/src/Message/EmailAttachments.Page.al
```diff
@@ -53,6 +54,28 @@ page 8889 "Email Attachments"
...
+                begin
+                    foreach SingleFile in files do begin
+                        SingleFile.CreateInStream(TempInStream, TextEncoding::UTF8);
+                        EmailEditor.UploadAttachmentFromStream(EmailMessageImpl, SingleFile.FileName, TempInStream);
+                    end;
+                    UpdateDeleteActionEnablement();
+                end;
+            }
 
             action(Upload)
```
haoranpb, line 81: Obsolete this action if not used anymore, don't forget the CLEAN tag.

haoranpb (changes requested):

src/System Application/App/Email/src/Message/EmailAttachments.Page.al
gggdttt: Checked with client team : choose UTF-8.

src/System Application/App/Email/src/Message/EmailAttachments.Page.al
gggdttt: Yes, I have checked with Jacob.

src/System Application/App/Email/src/Message/EmailAttachments.Page.al
```diff
@@ -53,15 +54,40 @@ page 8889 "Email Attachments"
...
+                    end;
+                    UpdateDeleteActionEnablement();
+                end;
+            }
+#if not CLEAN27
             action(Upload)
             {
+                ObsoleteState = Pending;
+                ObsoleteReason = 'Action Upload is replaced by action UploadMultiple. Action Upload will be removed in version 27.0';
+                ObsoleteTag = '27.0';
```
pri-kise: The ObsoleteTag should be `25.0`.

src/System Application/App/Email/src/Email/Outbox/EmailEditor.Codeunit.al
gggdttt: Yes, keep it for now

src/System Application/Test/Email/src/EmailTest.Codeunit.al
```diff
@@ -300,7 +300,6 @@ codeunit 134685 "Email Test"
         Assert.IsFalse(EmailEditor.BccField.Editable(), 'Bcc field was editable');
         Assert.IsFalse(EmailEditor.SubjectField.Editable(), 'Subject field was editable');
         Assert.IsFalse(EmailEditor.BodyField.Editable(), 'Body field was editable');
-        Assert.IsFalse(EmailEditor.Attachments.Upload.Visible(), 'Upload Action is visible.');
```
haoranpb, line 303: Minor: wrap them with `#if not CLEAN25` instead, the functionality will still exist for a while, we can keep the tests for the same period of time

haoranpb (approved):

src/System Application/App/Email/src/Email/Outbox/EmailEditor.Codeunit.al
```diff
@@ -162,18 +162,22 @@ codeunit 8906 "Email Editor"
     var
         FileName: Text;
         Instream: InStream;
-        AttachmentName, ContentType : Text[250];
-        AttachamentSize: Integer;
     begin
         UploadIntoStream('', '', '', FileName, Instream);
-        if FileName = '' then
```
AndreasMoth, line 169: Feel free to won't fix since it's only a single exit, but I think these exits makes the code easier to understand and especially add more exceptions to. Thinking of reasons to exit instead of reasons to execute.

src/System Application/App/Email/src/Email/Outbox/EmailEditor.Codeunit.al
```diff
@@ -162,18 +162,22 @@ codeunit 8906 "Email Editor"
...
-            exit;
+        if FileName <> '' then
+            UploadAttachmentFromStream(EmailMessageImpl, FileName, Instream);
+    end;
 
+    internal procedure UploadAttachmentFromStream(EmailMessageImpl: Codeunit "Email Message Impl."; FileName: Text; Instream: InStream)
```
AndreasMoth: Remove internal. No need to mark procedure as internal since the object already is. Hence this is redundant.

AndreasMoth: Just call it UploadAttachment, the FromStream comes from our legacy days when we could not do overrides, hence we had to come up with unique names. Now that we can do overrides, we should use them.

AndreasMoth: I believe this function should also be public on the facade, so partners can call this as well.

src/System Application/App/Email/src/Message/EmailAttachments.Page.al
```diff
@@ -53,15 +54,41 @@ page 8889 "Email Attachments"
...
+                Scope = Page;
+                Visible = IsEmailEditable;
 
+                trigger OnAction(files: List of [FileUpload])
+                var
+                    EmailEditor: Codeunit "Email Editor";
+                    TempInStream: InStream;
+                    SingleFile: FileUpload;
+                begin
+                    foreach SingleFile in files do begin
```
AndreasMoth, line 72: This should be moved to the codeunit, would make sense partners can also call EmailEditor.UploadAttachment(files)

gggdttt: Unfortunately now we have bug: FileUpload cannot be used in Codeunit. Will refactor this to a codeunit once this bug is solved

src/System Application/App/Email/src/Email/Outbox/EmailEditor.Codeunit.al
haoranpb: Personally, I prefer this way. So, folks don't have to scroll up to the top of the file to figure out if they defined a `public` or `internal` produce.

haoranpb (approved):

darjoo (approved):

JesperSchulz (approved):

PR #855, opened 2024-04-02
Add TriggerMovedTableSchemaCheck

grobyns:
#### Summary 
Adds TriggerMovedTableSchemaCheck

#### Work Item(s) 
Fixes [AB#523755](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/523755)

### DIFF
--- a/src/Business Foundation/App/NoSeries/src/Upgrade/NoSeriesInstaller.Codeunit.al
+++ b/src/Business Foundation/App/NoSeries/src/Upgrade/NoSeriesInstaller.Codeunit.al
@@ -48,13 +48,13 @@ codeunit 329 "No. Series Installer"
 #endif
     begin
 #pragma warning disable AA0175
-        NoSeries.FindFirst();
-        NoSeriesLine.FindFirst();
-        NoSeriesRelationship.FindFirst();
-        NoSeriesTenant.FindFirst();
+        if NoSeries.FindFirst() then;
+        if NoSeriesLine.FindFirst() then;
+        if NoSeriesRelationship.FindFirst() then;
+        if NoSeriesTenant.FindFirst() then;
 #if not CLEAN24
-        NoSeriesLineSales.FindFirst();
-        NoSeriesLinePurchase.FindFirst();
+        if NoSeriesLineSales.FindFirst() then;
+        if NoSeriesLinePurchase.FindFirst() then;
 #endif
 #pragma warning restore AA0175
     end;
--- a/src/Business Foundation/App/NoSeries/src/Upgrade/NoSeriesInstaller.Codeunit.al
+++ b/src/Business Foundation/App/NoSeries/src/Upgrade/NoSeriesInstaller.Codeunit.al
@@ -15,6 +15,7 @@ codeunit 329 "No. Series Installer"
 
     trigger OnInstallAppPerCompany()
     begin
+        TriggerMovedTableSchemaCheck();
         SetupNoSeriesImplementation();
     end;
 
@@ -33,5 +34,28 @@ codeunit 329 "No. Series Installer"
         NoSeriesLine.ModifyAll(Implementation, "No. Series Implementation"::Normal, false);
 
         UpgradeTag.SetUpgradeTag(NoSeriesUpgradeTags.GetImplementationUpgradeTag());
+    end;
+
+    local procedure TriggerMovedTableSchemaCheck()
+    var
+        NoSeries: Record "No. Series";
+        NoSeriesLine: Record "No. Series Line";
+        NoSeriesRelationship: Record "No. Series Relationship";
+        NoSeriesTenant: Record "No. Series Tenant";
+#if not CLEAN24
+        NoSeriesLineSales: Record "No. Series Line Sales";
+        NoSeriesLinePurchase: Record "No. Series Line Purchase";
+#endif
+    begin
+#pragma warning disable AA0175
+        NoSeries.FindFirst();
+        NoSeriesLine.FindFirst();
+        NoSeriesRelationship.FindFirst();
+        NoSeriesTenant.FindFirst();
+#if not CLEAN24
+        NoSeriesLineSales.FindFirst();
+        NoSeriesLinePurchase.FindFirst();
+#endif
+#pragma warning restore AA0175
     end;
 }
--- a/src/Business Foundation/App/NoSeries/src/Upgrade/NoSeriesInstaller.Codeunit.al
+++ b/src/Business Foundation/App/NoSeries/src/Upgrade/NoSeriesInstaller.Codeunit.al
@@ -15,7 +15,7 @@ codeunit 329 "No. Series Installer"
 
     trigger OnInstallAppPerCompany()
     begin
-        TriggerMovedTableSchemaCheck();
+        TriggerMovedTableSchemaSanityCheck();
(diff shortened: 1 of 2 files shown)


jehelles (dismissed):

src/Business Foundation/App/NoSeries/src/Upgrade/NoSeriesInstaller.Codeunit.al
```diff
@@ -34,4 +35,27 @@ codeunit 329 "No. Series Installer"
...
+        NoSeriesLine: Record "No. Series Line";
+        NoSeriesRelationship: Record "No. Series Relationship";
+        NoSeriesTenant: Record "No. Series Tenant";
+#if not CLEAN24
+        NoSeriesLineSales: Record "No. Series Line Sales";
+        NoSeriesLinePurchase: Record "No. Series Line Purchase";
+#endif
+    begin
+#pragma warning disable AA0175
+        if NoSeries.FindFirst() then;
```
nikolakukrika, line 63: Why do we need this code? First iteration was throwing an error and this iteration is simply reading a row?
Why don't we provide a proper error message except the error of not finding first?
Should we emit telemetry if we cannot find it?

src/Business Foundation/App/NoSeries/src/Upgrade/NoSeriesInstaller.Codeunit.al
```diff
@@ -34,4 +35,27 @@ codeunit 329 "No. Series Installer"
 
         UpgradeTag.SetUpgradeTag(NoSeriesUpgradeTags.GetImplementationUpgradeTag());
     end;
+
+    local procedure TriggerMovedTableSchemaCheck()
```
nikolakukrika: The name should be updated that this code is using OnInstall trigger and explain more about the usage of this method.

nikolakukrika (changes requested):

src/Business Foundation/App/NoSeries/src/Upgrade/NoSeriesInstaller.Codeunit.al
grobyns: it's a local in an install codeunit... how is that not clear enough?

src/Business Foundation/App/NoSeries/src/Upgrade/NoSeriesInstaller.Codeunit.al
grobyns: it was requested by server as an extra check to load schema metadata just to make sure... the if is there in case the table is empty, which is fine. There's nothing to be done if the code passes. If it fails then it would have failed to sync schema metadata and code execution would stop anyways...

nikolakukrika (dismissed):
If you are implementing this consider:
Moving the check to both OnInstall and Upgrade, you may miss the check otherwise
Name it SchemaSanityCheck or something, add comments why we have this confusing code
Add a method to validate after upgrade is done to check that the data is moved 
Consider emitting telemetry and setting up the alerts 
Have the methods only in v24.x so we track the upgrade, no need to have this in master going forward

src/Business Foundation/App/NoSeries/src/Upgrade/NoSeriesInstaller.Codeunit.al
haoranpb: Wondering: If we don't have this check, and the sync schema metadata failed, what would happen?

I assume the upgrade will fail anyway?

grobyns: Server runtime requested the change as an extra sanity check. The if .. then is there so code doesn't fail when there is no data (the data itself is not relevant). There is no need for an error or telemetry. If the code runs all is good, if it fails it would fail due to a schema bug on the server which would stop all .al code execution anyway.

src/Business Foundation/App/NoSeries/src/Upgrade/NoSeriesInstaller.Codeunit.al
grobyns: added to upgrade too

grobyns:
> If you are implementing this consider: Moving the check to both OnInstall and Upgrade, you may miss the check otherwise Name it SchemaSanityCheck or something, add comments why we have this confusing code Add a method to validate after upgrade is done to check that the data is moved Consider emitting telemetry and setting up the alerts Have the methods only in v24.x so we track the upgrade, no need to have this in master going forward

The check is now on install and on upgrade. Added Sanity to the method name. Added some explanation in a doc trigger.
If we want to remove the code after it has been backported (HF'd) to 24.0 we can but process dictates we do the change in main, then 24.x then 24.0.

haoranpb (approved):

src/Business Foundation/App/NoSeries/src/Upgrade/NoSeriesInstaller.Codeunit.al
```diff
@@ -34,4 +35,41 @@ codeunit 329 "No. Series Installer"
 
         UpgradeTag.SetUpgradeTag(NoSeriesUpgradeTags.GetImplementationUpgradeTag());
     end;
+
+    /// <summary>
+    /// This method is used to ensure that the runtime metadata matches the schema for moved tables.
+    /// </summary>
+    /// <remarks>
+    /// The if .. then statements ensure the code does not fail when the tables are empty. The presence of data is not important, the FindFirst will trigger a schema check.
+    /// Should this code fail it would indicate a bug in the server code. The code is not expected to fail.
```
AndreasMoth, line 44: I don't think application code should validate platform code. Platform should validate their own code.. otherwise I guess we need to do this every time we move a table?

grobyns: @jehelles ?

darjoo: I have to second this. I understand the server is asking for a second check, but is this a one off? Or is this for every time we move tables? 
If it's one off fine, shouldn't it already have been tested by them? And if it's every time, then clearly something is missing in platform instead.

jehelles: @AndreasMoth 
@grobyns 
@darjoo 
This is a one-off, for the first time, we don't expect any issues, and of course it's tested already (as part of the existing upgrade test, as well as manual testing).

jehelles: You are free to remove it in a month or two.

haoranpb: In that case, should we wrap it in a CLEAN tag?

grobyns: @haoranpb there is no need. Once the change is in 24.0 we can delete the code from main immediately.

darjoo (approved):

JesperSchulz (approved):

PR #903, opened 2024-04-09
Only table with type Normal is supported in Data Classification Worksheet and Data Privacy Utilities

haoranpb:
#### Summary 

Add the public procedure `IsSupportedTable`: returns True if the table has type Normal and not obsoleted.

Will need to uptake in NAV for "Data Privacy Utilities" after this is merged.

#### Work Item(s) 
Fixes AB#523733

### DIFF
--- a/src/System Application/App/Data Classification/src/DataClassificationMgt.Codeunit.al
+++ b/src/System Application/App/Data Classification/src/DataClassificationMgt.Codeunit.al
@@ -186,6 +186,16 @@ codeunit 1750 "Data Classification Mgt."
     end;
 
     /// <summary>
+    /// Check if the table is supported in the Data Privacy Utilities and Data Classification Worksheet.
+    /// </summary>
+    /// <param name="TableNo"></param>
+    /// <returns>True if the table has type Normal and not obsoleted</returns>
+    procedure IsSupportedTable(TableNo: Integer): Boolean
+    begin
+        exit(DataClassificationMgtImpl.IsSupportedTable(TableNo));
+    end;
+
+    /// <summary>
     /// Raises an event that allows subscribers to insert Data Privacy Entities in the DataPrivacyEntities record.
     /// Throws an error when it is not called with a temporary record.
     /// </summary>
--- a/src/System Application/App/Data Classification/src/DataClassificationMgtImpl.Codeunit.al
+++ b/src/System Application/App/Data Classification/src/DataClassificationMgtImpl.Codeunit.al
@@ -40,12 +40,14 @@ codeunit 1753 "Data Classification Mgt. Impl."
     var
         DataSensitivity: Record "Data Sensitivity";
     begin
-        DataSensitivity.Init();
-        DataSensitivity."Company Name" := CopyStr(CompanyName(), 1, MaxStrLen(DataSensitivity."Company Name"));
-        DataSensitivity."Table No" := TableNo;
-        DataSensitivity."Field No" := FieldNo;
-        DataSensitivity."Data Sensitivity" := DataSensitivityOption;
-        DataSensitivity.Insert();
+        if IsSupportedTable(TableNo) then begin
+            DataSensitivity.Init();
+            DataSensitivity."Company Name" := CopyStr(CompanyName(), 1, MaxStrLen(DataSensitivity."Company Name"));
+            DataSensitivity."Table No" := TableNo;
+            DataSensitivity."Field No" := FieldNo;
+            DataSensitivity."Data Sensitivity" := DataSensitivityOption;
+            DataSensitivity.Insert();
+        end;
     end;
 
     procedure SetSensitivities(var DataSensitivity: Record "Data Sensitivity"; Sensitivity: Option)
@@ -396,6 +398,19 @@ codeunit 1753 "Data Classification Mgt. Impl."
         DataSensitivity.FilterGroup(2);
         DataSensitivity.SetRange("Table No", TableNo);
         Page.RunModal(Page::"Data Classification Worksheet", DataSensitivity);
+    end;
+
+    procedure IsSupportedTable(TableNo: Integer): Boolean
+    var
+        TableMetadata: Record "Table Metadata";
+    begin
+        if TableMetadata.Get(TableNo) then
+            if (TableMetadata.ObsoleteState = TableMetadata.ObsoleteState::Removed) or (TableMetadata.TableType <> TableMetadata.TableType::Normal) then
+                exit(false)
+            else
+                exit(true);
+
+        exit(false);
     end;
 }
 


src/System Application/Test/Data Classification/src/DataClassificationMgtTests.Codeunit.al
```diff
@@ -145,8 +142,8 @@ codeunit 135150 "Data Classification Mgt. Tests"
         DataPrivacyEntities.DeleteAll();
         DataSensitivity.DeleteAll();
 
-        // [GIVEN] A fictive table number and an entry in the Data Privacy Entities corresponding to this table number
-        TableNo := 50001;
+        // [GIVEN] A random table number and an entry in the Data Privacy Entities corresponding to this table number
+        TableNo := Database::"Data Sensitivity";
```
grobyns, line 148: Isn't the point of using 50001 as TableNo that it does not exist in the database?

src/System Application/Test/Data Classification/src/DataClassificationMgtTests.Codeunit.al
```diff
@@ -104,7 +101,7 @@ codeunit 135150 "Data Classification Mgt. Tests"
         DataSensitivity.DeleteAll();
 
         // [GIVEN] A table number, a field number and a data sensitivity
-        TableNo := 18;
+        TableNo := Database::"Data Sensitivity";
```
grobyns, line 106: why the change? before it was using the Customer table, now Data Sensitivity... what's the diff?

src/System Application/Test/Data Classification/src/DataClassificationMgtTests.Codeunit.al
```diff
@@ -549,7 +544,7 @@ codeunit 135150 "Data Classification Mgt. Tests"
         DataSensitivity.DeleteAll();
 
         // [GIVEN] A table number
-        TableNo := 1180;
+        TableNo := Database::"Data Sensitivity";
```
grobyns, line 549: I somewhat get changing 18 or 27 or 1180 to some table(s) that do exist, if that then doesn't influence the behaviour of the test... but shouldn't the replacements be as unique? it seems this PR is changing (almost) everything to Data Sensitivity... Does it matter that all these tests used different values before and now all use the same?

src/System Application/Test/Data Classification/src/DataClassificationMgtTests.Codeunit.al
```diff
@@ -872,8 +860,24 @@ codeunit 135150 "Data Classification Mgt. Tests"
...
+        TableMetaData: Record "Table Metadata";
+        Field: Record Field;
+        DataClassificationMgtImpl: Codeunit "Data Classification Mgt. Impl.";
+    begin
+        FieldCount := 0;
+        if TableMetaData.FindSet() then
+            repeat
+                if DataClassificationMgt.IsSupportedTable(TableMetaData.ID) then begin
+                    Field.SetRange(TableNo, TableMetaData.ID);
+                    DataClassificationMgtImpl.GetEnabledSensitiveFields(Field);
```
grobyns: use of internals .. tsk tsk... should use a test library...

src/System Application/Test/Data Classification/src/DataClassificationMgtTests.Codeunit.al
haoranpb: Resolving so we use the comment below for the same discussion.

src/System Application/Test/Data Classification/src/DataClassificationMgtTests.Codeunit.al
haoranpb: The change to `TableNo` is related to the change in `DataClassificationMgtImpl.Codeunit.al`:

I added the logic `procedure IsSupportedTable`. If the table doesn't exit, it will return `false`. Table 18, 27... do not exist in the context of the test, so they will not be inserted into the `Data Sensitivity` table, then causing the tests to fail.

The reason behind the change is: `Data Privacy Utilities` (using configuration package) [only support TableType = Normal](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_git/NAV?path=/App/Layers/W1/BaseApp/System/RapidStart/ConfigValidateManagement.Codeunit.al&version=GBmaster&line=949&lineEnd=949&lineStartColumn=21&lineEndColumn=41&lineStyle=plain&_a=contents), so we will align this with `Data Classification Worksheet`.

src/System Application/Test/Data Classification/src/DataClassificationMgtTests.Codeunit.al
haoranpb: Good catch will move it after I fix all the tests...

src/System Application/Test/Data Classification/src/DataClassificationMgtTests.Codeunit.al
grobyns: just out of curiosity, what happens when i install an app with a table, classify it, then uninstall the app? The data remains in our system, the classification should too... in case the app is ever re-installed... or something else...

src/System Application/Test/Data Classification/src/DataClassificationMgtTests.Codeunit.al
haoranpb: Reading through the test, don't think the Table ID matters in this test scenario.

src/System Application/Test/Data Classification/src/DataClassificationMgtTests.Codeunit.al
haoranpb: Good question.

The Data Classification will stay in our system. We do not delete data in `Data Sensitivity` table. Unless it's unclassified, (this is done via `SyncAllFields`, it can be triggered by an action on the `Data Privacy Utilities` page, it's called Add New Fields or something)

```
/// <summary>
/// Synchronizes the Data Sensitivity table with the Field table. It inserts new values in the Data Sensitivity table for the
/// fields that the Field table contains and the Data Sensitivity table does not and it deletes the unclassified fields from
/// the Data Sensitivity table that the Field table does not contain.
/// </summary>
procedure SyncAllFields()
```

haoranpb: Good question.

The Data Classification will stay in our system. We do not delete data in `Data Sensitivity` table. Unless it's unclassified, (this is done via `SyncAllFields`, it can be triggered by an action on the `Data Privacy Utilities` page, it's called Add New Fields or something)

```
/// <summary>
/// Synchronizes the Data Sensitivity table with the Field table. It inserts new values in the Data Sensitivity table for the
/// fields that the Field table contains and the Data Sensitivity table does not and it deletes the unclassified fields from
/// the Data Sensitivity table that the Field table does not contain.
/// </summary>
procedure SyncAllFields()
```

ikoletic (approved):

darjoo (approved):

gggdttt (approved):

grobyns (approved):

PR #945, opened 2024-04-16
Fix return value for ProcessFunctionCall

haoranpb:
#### Summary 

Fix return value for procedure `ProcessFunctionCall`. Also explicit return value false

#### Work Item(s) 
Fixes [AB#525473](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/525473)

### DIFF
--- a/src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
+++ b/src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
@@ -378,6 +378,7 @@ codeunit 7772 "Azure OpenAI Impl"
         end;
     end;
 
+    /// <returns>True if Function Call Response is set</returns>
     local procedure ProcessFunctionCall(Functions: JsonArray; var ChatMessages: Codeunit "AOAI Chat Messages"; var AOAIFunctionResponse: Codeunit "AOAI Function Response"): Boolean
     var
         Function: JsonObject;
@@ -417,12 +418,14 @@ codeunit 7772 "Azure OpenAI Impl"
             if TryExecuteFunction(AOAIFunction, Arguments, FunctionResult) then begin
                 AOAIFunctionResponse.SetFunctionCallingResponse(true, true, AOAIFunction.GetName(), FunctionResult, '', '');
                 exit(true);
-            end else
-                AOAIFunctionResponse.SetFunctionCallingResponse(true, false, AOAIFunction.GetName(), FunctionResult, GetLastErrorText(), GetLastErrorCallStack())
-        else
+            end else begin
+                AOAIFunctionResponse.SetFunctionCallingResponse(true, false, AOAIFunction.GetName(), FunctionResult, GetLastErrorText(), GetLastErrorCallStack());
+                exit(true);
+            end
+        else begin
             AOAIFunctionResponse.SetFunctionCallingResponse(true, false, FunctionName, FunctionResult, StrSubstNo(FunctionCallingFunctionNotFoundErr, FunctionName), '');
-
-        exit(false);
+            exit(true);
+        end;
     end;
 
     [TryFunction]
--- a/src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
+++ b/src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
@@ -388,38 +388,41 @@ codeunit 7772 "Azure OpenAI Impl"
         FunctionResult: Variant;
     begin
         if Functions.Count = 0 then
-            exit;
+            exit(false);
 
         Functions.Get(0, Token);
         Function := Token.AsObject();
 
         if Function.Get('type', Token) then begin
             if Token.AsValue().AsText() <> 'function' then
-                exit;
+                exit(false);
         end else
-            exit;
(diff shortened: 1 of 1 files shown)


ChethanT (approved):

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
```diff
@@ -388,38 +388,41 @@ codeunit 7772 "Azure OpenAI Impl"
...
-            else
+            if TryExecuteFunction(AOAIFunction, Arguments, FunctionResult) then begin
+                AOAIFunctionResponse.SetFunctionCallingResponse(true, true, AOAIFunction.GetName(), FunctionResult, '', '');
+                exit(true);
+            end else
                 AOAIFunctionResponse.SetFunctionCallingResponse(true, false, AOAIFunction.GetName(), FunctionResult, GetLastErrorText(), GetLastErrorCallStack())
         else
             AOAIFunctionResponse.SetFunctionCallingResponse(true, false, FunctionName, FunctionResult, StrSubstNo(FunctionCallingFunctionNotFoundErr, FunctionName), '');
+
+        exit(false);
```
msft-sam: Does this mean we lose the error information in the response? (should it be true?)

haoranpb: Good catch

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
```diff
@@ -388,38 +389,43 @@ codeunit 7772 "Azure OpenAI Impl"
...
             FunctionName := Token.AsValue().AsText()
         else
-            exit;
+            exit(false);
 
         if Function.Get('arguments', Token) then
             // Arguments are stored as a string in the JSON
             Arguments.ReadFrom(Token.AsValue().AsText());
 
         if ChatMessages.GetFunctionTool(FunctionName, AOAIFunction) then
```
ChethanT, line 416: Isn't this same as 
image

haoranpb: Yes, but feels bit wrong to use a catch all to return `true`...

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
```diff
@@ -378,6 +378,7 @@ codeunit 7772 "Azure OpenAI Impl"
         end;
     end;
 
+    /// <returns>True if Function Call Response is set</returns>
```
grobyns: please remove this comment/doc. let's not start a bad trend of half baked doc sections... either you document the full thing or not... :-)

haoranpb: Gone.

I think Darrick is planning to re-structure this bit future, so not gonna do the full document now.

grobyns (approved):

Drakonian:
I guess it's all about the availability of Microsoft services at the moment?
image

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
```diff
@@ -388,38 +388,43 @@ codeunit 7772 "Azure OpenAI Impl"
...
+            if TryExecuteFunction(AOAIFunction, Arguments, FunctionResult) then begin
+                AOAIFunctionResponse.SetFunctionCallingResponse(true, true, AOAIFunction.GetName(), FunctionResult, '', '');
+                exit(true);
+            end else begin
+                AOAIFunctionResponse.SetFunctionCallingResponse(true, false, AOAIFunction.GetName(), FunctionResult, GetLastErrorText(), GetLastErrorCallStack());
+                exit(true);
+            end
+        else begin
             AOAIFunctionResponse.SetFunctionCallingResponse(true, false, FunctionName, FunctionResult, StrSubstNo(FunctionCallingFunctionNotFoundErr, FunctionName), '');
+            exit(true);
```
AndreasMoth, line 426: From what I can see we always return true, so shouldn't we just return true at the end of the function? Otherwise adding another case statement you could potentially not return true (reproducing the issue you just fixed). At least I think it makes sense to have a global exit value of true, even if you want to handle these individually..

haoranpb: The return value of the procedure should be determined by whether the `Response` is set or not.

And by default, that value is not set, which correspond to return value default to false.

(Will also tag @darjoo here as I think the purpose of return value is not obvious, we might want to re-structure)

msft-sam: Agree it is not obvious, had to read the calling code to figure out the intent :)
While you could simplify to a single `exit(true)` I agree that setting it after the response does help with understanding what the purpose of it is.

Yes it would be better to refactor this to be a little more clear - but we can take that separate to this bug fix.

PR #975, opened 2024-04-23
No error on CodeCop rule 0248

esbenk:
New code cop rule is coming, but is causing too many diagnostics right now. 

Related to
[AB#480821](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/480821): [AL] [Idea] Codeunit self-reference (this keyword)

### DIFF
--- a/src/rulesets/CodeCop.ruleset.json
+++ b/src/rulesets/CodeCop.ruleset.json
@@ -994,7 +994,7 @@
                   {
                       "id":  "AA0248",
                       "action":  "None",
-                      "justification": "New rule for missing this. qualification on members. Turned off globally."
+                      "justification": "Rule for missing 'this.' qualification on members. Not an error, rule turned off globally."
                   },
                   {
                       "id":  "AA0249",
--- a/src/rulesets/CodeCop.ruleset.json
+++ b/src/rulesets/CodeCop.ruleset.json
@@ -993,7 +993,8 @@
                   },
                   {
                       "id":  "AA0248",
-                      "action":  "Error"
+                      "action":  "None",
+                      "justification": "New rule for missing this. qualification on members. Turned off globally."
                   },
                   {
                       "id":  "AA0249",


grobyns (approved):

src/rulesets/CodeCop.ruleset.json
```diff
@@ -993,7 +993,8 @@
                   },
                   {
                       "id":  "AA0248",
-                      "action":  "Error"
+                      "action":  "None",
+                      "justification": "New rule for missing this. qualification on members. Turned off globally."
```
AndreasMoth: perhaps "This rule has too many false positives. It will soon be replaced with another rule."

esbenk: They will not be false positives and there will not be another rule, but you (app-team) must decide if you want to use this. qualification. We need a diagnostic to trigger the CodeFixer for it.

AndreasMoth: Ok, then I misunderstood the justification for this. "New rule missing this" which new rule, missing what? "qualitifation on members", not sure what this means, so something about members doesn't work, but I don't understand what.

AndreasMoth: Please update the justification to explain the problem. Also the description of this PR says "New code cop rule is coming" but you say "there will not be another rule", these are contradictory, no?

esbenk: Adding support a this keyword. You can optionally use it to qualify access to e.g. global vars/methods with this.MyGlobalVar := 42.

grobyns:
```suggestion
                      "justification": "Rule for missing 'this.' qualification on members. Not an error, rule turned off globally."
```

maybe? @AndreasMoth

AndreasMoth (approved):

grobyns (approved):

Issue #298, opened 2023-11-02
[Extensibility Bug]: Provide a possibility to set the Telemetry Logger Interface explicit instead of relying on a event subscriber on publisher


pri-kise:
### Describe the issue

I'd like to have the possibility to explicit the the "Telemetry Logger" Interface, when using the Telemetry or FeatureTelemetry Codeunits.

Currently we need a subscriber in any of our apps (with identical publisher).
And since we don't know which app is installed we might subscribe multiple times to this event publisher, which results in an additional telemetry event, that I want to avoid.

### Expected behavior

it should be possible to explicit set the interface via a Setter Method on the Public Facade Codeunits.

or in other words:

it should be possible to set an TelemetryLogger Interface to skip the following lines 

```al
        TelemetryLoggers.SetCurrentPublisher(Publisher);
        TelemetryLoggers.OnRegisterTelemetryLogger();

        if TelemetryLoggers.GetTelemetryLogger(TelemetryLogger) then
```
https://github.com/microsoft/BCApps/blob/636c354bc6d2b27ba6527fd193ac494af7f8ebfa/src/System%20Application/App/Telemetry/src/Logging/TelemetryImpl.Codeunit.al#L41C1-L41C1

### Steps to reproduce

Create two apps with the same publisher and subscribe to the the TelemetryLogger Event.

### Additional context

I already talked to some people of microsoft at Directions EMEA 2023 and they suggested to open an issue for this topic to discuss here fruther with @IhorHandziuk who created this module.

IhorHandziuk:
Hi @pri-kise, I think adding the ability to provide the telemetry logger implementation on a variable instance is not the best approach, as it will add more lines of code and potentially more confusion to something that is suppoused to be very siple to use.

Normally we recommend that partners extract the implementation of the telemetry logger in a separate app, and then add it as a dependency to the existing apps (see this [Viva Engage post](https://www.yammer.com/dynamicsnavdev/#/Threads/show?threadId=2411336245280768)).
If this is not a viable option for some reason, we can also just remove the warning about having multiple telemetry logger implementations altogether (and assume that the extension developer is aware that a random imlementation will be picked).

pri-kise:
I will discuss with my other collegaues internally and will close this issue until we have evaluated the best approach.

JohnnyUndercover:
Why can't more than one interface be saved in the Telemetry Logger codeunit? (List of [interface] would be nice ;) )

[...]

PR #995, opened 2024-04-26
Enable Telemetry for ISV

haoranpb:
### Summary

The main goal of the effort is to bring Telemetry to ISVs when they are utilizing functionalities from Microsoft. So they can get insights into usage/adoption/quality etc.

We will only uptake this capability for the `AI SDK` for the moment and collect feedback before continue moving forward.

There are two main concerns while implementing:
1. No breaking. The changes should not bring any breaking changes.
2. Performance. `Telemetry` is wildly used, we should not impact performance much while bringing value.

Imagine a callstack like: (Note, "Publisher B" is usually Micosoft)

```
Foo App - "Publisher A"
Bar App - "Publisher A"
SDK App - "Publisher B" <- Telemetry.LogMessage
System App - Microsoft
```

#### How Telemetry module works currently
1. `Telemetry.LogMessage` is called from `Publisher B`
3. `CurrentPublisher` is set to `Publisher B`
4. `OnRegisterTelemetryLogger` event is fired, all event subscribers will try to register their `Telemetry Logger`
5. But only one `Telemetry Logger` from `Publisher B` will be saved and take effect

#### How Telemetry module might work in future (simplified)

1. `Telemetry.LogMessage` is called from `Publisher B`
2. `CallStackPublishers` is set to `["Publisher B", "Publisher A"]`
3. `OnRegisterTelemetryLogger` event is fired, all event subscribers will try to register their `Telemetry Logger`
4. `Telemetry Logger` from `["Publisher B", "Publisher A"]` will be saved and take effect

### Work Item(s)
Fixes [AB#492500](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/492500)


src/System Application/App/Telemetry/src/Logging/TelemetryImpl.Codeunit.al
```diff
@@ -33,16 +33,31 @@ codeunit 8712 "Telemetry Impl."
...
+            // When Scope is All, all publishers get a copy of Telemetry (should be 3rd-party).
+            RegisteredTelemetryLoggers := TelemetryLoggers.GetRegisteredTelemetryLoggers();
+            foreach TelemetryLogger in RegisteredTelemetryLoggers do
+                // we set the scope to ExtensionPublisher for 3rd-party loggers to avoid multiple logging into environment telemetry.
+                // todo: need to filter out first-party loggers from RegisteredTelemetryLoggers
+                TelemetryLogger.LogMessage(EventId, Message, Verbosity, DataClassification, TelemetryScope::ExtensionPublisher, CustomDimensions);
+
+            if TelemetryLoggers.GetFirstPartyTelemetryLogger(MicrosoftTelemetryLogger) then
+                // Then we use first-party logger to log telemetry to All only once.
+                MicrosoftTelemetryLogger.LogMessage(EventId, Message, Verbosity, DataClassification, TelemetryScope::All, CustomDimensions);
```
IhorHandziuk: This will make sure that we always get a copy of 3rd party telemetry, even though we never want to have it.

This prompted me think that maybe we should add more guards to make sure no one receives telemetry they never wanted. I think we should add some opt-in mechanism (maybe a new parameter in the `OnRegisterTelemetryLogger` event) for IsVs to start receiving telemetry from other IsVs, with the possibility to only allow certain publishers.
This way we would be able to:
- Not receive telemetry from any 3rd party IsVs ourselves
- Allow IsVs to only receive events coming from Microsoft apps
- Allow IsVs to receive events from _all_ other IsVs, if that's what they want

Another point to discuss is whether it's okay to reuse the `TelemetryScope::All` value for sharing telemetry between IsVs. Maybe it would be best to add a new value to this enum to indicate that telemetry is shared both with the customer (environment telemetry) _and_ with other IsVs?

haoranpb: Agree, need to better handle who get what Telemetry.

Also thought about if we want to add one more value to the `Enum::TelemetryScope`, guess that depends on whether we want to support:
> Allow IsVs to receive events from all other IsVs, if that's what they want

src/System Application/App/Telemetry/src/Logging/TelemetryImpl.Codeunit.al
```diff
@@ -33,16 +33,31 @@ codeunit 8712 "Telemetry Impl."
     procedure LogMessageInternal(EventId: Text; Message: Text; Verbosity: Verbosity; DataClassification: DataClassification; TelemetryScope: TelemetryScope; CustomDimensions: Dictionary of [Text, Text]; CallerCustomDimensions: Dictionary of [Text, Text]; Publisher: Text)
     var
         TelemetryLoggers: Codeunit "Telemetry Loggers";
-        TelemetryLogger: Interface "Telemetry Logger";
+        TelemetryLogger, MicrosoftTelemetryLogger : Interface "Telemetry Logger";
+        RegisteredTelemetryLoggers: List of [Interface "Telemetry Logger"];
```
pri-kise: Does this mean that we have Dictitionary of [Guid, Interface], too?

@JohnnyUndercover suggested to save the most explicit implementation for the publisher.
https://github.com/microsoft/BCApps/issues/298#issuecomment-1892249123

haoranpb: Hi @pri-kise , thanks for taking a look at this PR!

> Does this mean that we have Dictitionary of [Guid, Interface], too?

No, we are still experimenting a few things and trying to find the best way to move forward. as you can see, this PR still couldn't pass build.

> save the most explicit implementation for the publisher.

We have taken this into consideration during out internal discussion, will share more information in this PR once we have more conclusion.

---

For now, this draft PR will remain proof of concept. Feel free to share more related feedback/suggestions here since #298 is closed.

src/System Application/App/Telemetry/src/Logging/TelemetryImpl.Codeunit.al
```diff
@@ -34,15 +34,33 @@ codeunit 8712 "Telemetry Impl."
...
-        if TelemetryLoggers.GetTelemetryLogger(TelemetryLogger) then
-            TelemetryLogger.LogMessage(EventId, Message, Verbosity, DataClassification, TelemetryScope, CustomDimensions);
+        case TelemetryScope of
+            // When Scope is ExtensionPublisher: only the current publisher gets a copy of Telemetry.
+            TelemetryScope::ExtensionPublisher:
+                if TelemetryLoggers.GetTelemetryLoggerFromCurrentPublisher(TelemetryLogger) then
+                    TelemetryLogger.LogMessage(EventId, Message, Verbosity, DataClassification, TelemetryScope::ExtensionPublisher, CustomDimensions);
+
+            // When Scope is All: 1. the current publisher gets a copy of Telemetry, 2. the environment gets a copy of Telemetry, 3. all registered loggers on the callstack get a copy of Telemetry.
+            TelemetryScope::All:
```
haoranpb: Echoing second part of the [comment from Ihor](https://github.com/microsoft/BCApps/pull/995#discussion_r1580705354). The `TelemetryScope::All` is getting confusing here.

I'd like to avoid any breaking change in this regard, so maybe something like:

`TelemetryScope::ExtensionPublisher` => Current Publisher

`TelemetryScope::All` => Current Publisher & Environment

`TelemetryScope::AllWithISV` => Current Publisher & Environment & all registered loggers on the callstack.

---

This way, no behavior changes unless developers explicitly opt-in the new enum value.

src/System Application/App/Telemetry/src/Logging/TelemetryImpl.Codeunit.al
```diff
@@ -34,15 +34,33 @@ codeunit 8712 "Telemetry Impl."
     var
         TelemetryLoggers: Codeunit "Telemetry Loggers";
         TelemetryLogger: Interface "Telemetry Logger";
+        RelevantTelemetryLoggers: List of [Interface "Telemetry Logger"];
     begin
         AddCustomDimensionsFromSubscribers(CustomDimensions, Publisher);
         AddCustomDimensionsSafely(CustomDimensions, CallerCustomDimensions);
 
         TelemetryLoggers.SetCurrentPublisher(Publisher);
```
haoranpb: Small performance improvement can be made here:
If we add something like `TelemetryLoggers.SetCurrentTelemetryScope`, then we don't need to collect other extension's loggers when scope is set to `ExtensionPublisher`.

src/System Application/App/Telemetry/src/Logging/TelemetryLoggersImpl.Codeunit.al
```diff
@@ -23,24 +25,40 @@ codeunit 8709 "Telemetry Loggers Impl."
...
         if Publisher = CurrentPublisher then
-            if not IsLoggerFound then begin
-                CurrentTelemetryLogger := TelemetryLogger;
-                IsLoggerFound := true;
-            end else
-                Session.LogMessage('0000G7J', StrSubstNo(MultipleTelemetryLoggersFoundErr, Publisher), Verbosity::Warning, DataClassification::SystemMetadata, TelemetryScope::All, 'Category', TelemetryLibraryCategoryTxt)
+            if not IsLoggerFromCurrentPublisherFound then begin
+                TelemetryLoggerFromCurrentPublisher := TelemetryLogger;
+                IsLoggerFromCurrentPublisherFound := true;
+            end;
```
haoranpb: This logic may not be most performant. It will run on very `Register` (could be multiple times per `LogMessage`), but it only save time once per `LogMessage`

src/System Application/App/Telemetry/src/Logging/TelemetryImpl.Codeunit.al
haoranpb: After some thought, given that we are only sending Telemetry to publishers on the callstack, think the first point is less relevant now.

If "Publisher A" decides to utilize an SDK App from "Publisher B", then they should receive Telemetry from the "SDK App" (at least by default, maybe we could consider an opt-out mechanism).

Will track the second point in the comment below

src/System Application/App/Telemetry/src/Logging/TelemetryImpl.Codeunit.al
haoranpb: An update now that the PR is getting ready:
We will not have `Dictitionary of [Guid, Interface]` as part of this effort. However, I can't see why not if there is a strong case for it.

And after some thoughts and discussions, we will keep the current design for one implementation per publisher for now. One main reason pushes me into making this design is performance.

The recommendation for partners will remain the same: Extract the implementation of the telemetry logger into a separate app, and then add it as a dependency to other existing apps.

src/System Application/App/Telemetry/src/Logging/TelemetryImpl.Codeunit.al
haoranpb: There are a few reasons for the extra value:
1. Flexibility. Have the option to send Telemetry to only 1. current publisher and 2. Environment if folks want to
2. Performance. It takes a bit more time to send Telemetry to ISVs on the callstack, so it would be more performant to ask everyone to opt-in this capability
3. App Insights Storage: There will be more Telemetry for partner's App Insights. if we re-use `TelemetryScope::All`

However, none of the reasons above are a deal breaker. It might be complicated to add an extra value to TelemetryScope:
1. Flexibility. `Session.LogMessage` can still be used to achieve flexibility
2. Performance. We will go through all Telemetry module usage in Microsoft.
3. App Insights Storage. This can be tricky, hard to estimate the impact, effort in step 2 might help.

haoranpb: Adding the `Enum "AL Telemetry Scope"` to mitigate this.

Keeping it internal so we can collect feedback and keep the flexibility to change

src/System Application/App/Telemetry/src/Logging/ALTelemetryScope.Enum.al
```diff
@@ -0,0 +1,27 @@
...
+
+    value(0; ExtensionPublisher)
+    {
+        Caption = 'Extension Publisher';
+        // map to TelemetryScope::ExtensionPublisher
+    }
+    value(1; Environment)
+    {
+        Caption = 'Environment';
+        // map to TelemetryScope::All
```
pri-kise: I think that you should use the summary xml comment for such explanations instead of a simple comment here.
This would align with other parts of the system application

haoranpb: Good catch, fixed

src/System Application/App/AI/src/Copilot/CopilotTelemetry.Codeunit.al
```diff
@@ -1,47 +0,0 @@
-// ------------------------------------------------------------------------------------------------
```
darjoo: Do not delete. Called by platform

haoranpb: Done, reverted those changes

darjoo (approved):

Issue #685, opened 2024-03-04
[Bug]: When using Azure Shared Key to load files to Azure File Share you are presented with an Authentication error


PJohnstonSysco:
### Describe the issue

When you try to load a file to an Azure File Share while connected using an Azure Share Key you are presented with an authentication error. You do not get the same error when using the List File function or Download Files

### Expected behavior

The file should be loaded to the Azure File Share without error

### Steps to reproduce

Use an Azure Shared Key when authenticating with the New Azure File module and try to upload a file to the File Share

### I will provide a fix for a bug

- [X] I will provide a fix for a bug

JesperSchulz:
Sounds like a bug! Of course approved 😊

Drakonian:
@PJohnstonSysco If you don't have time, I can do the fix

PJohnstonSysco:
@Drakonian I have got caught up with other things. If you have time that would be great

PR #1006, opened 2024-04-29
Fix error in AFS shared key authorization (#961)

### LINKED ISSUE #961: [Bug]: Error on uploading file on Azure File Share
### Describe the issue

When you try to upload a file to an Azure File Share (i tried both PutFileUI and PutFileStream) an authentication error occurs: "403 Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.".
I can read, download and delete the files, so I'm pretty sure that the authentication is done correctly using the Shared Key.
The module is not well documented so I'm not sure if something is missing in my procedure.

### Expected behavior

the file is uploaded correctly on the file share

### Steps to reproduce

Initialize the "AFS File Client" codeunit authenticating with shared key and try to call the PutFileUI method.
The file will be requested to the user but the response of the operation will not be successful.

### Additional context

_No response_

Krajzys:
#### Summary 

This is a PR, that should fix the error that occured when creating or sending file to Azure File Share with AFS Library.
I also created a simple test to ensure it doesn't happen again.

The issue was with two things:
1. Sorting of the headers - In BC the order of the headers in the string to sign was different than in the documentation, which led azure to calculate a different hash for the authorization. I added sorting of the headers in the same way as implemented in ABS module, which should ensure correct order.
2. The header x-ms-content-length required when using CreateFile call, was added to the content headers, which led to it not being correctly added to the string to sign. I changed it to be added to request headers.

#### Work Item(s) 
Fixes #961 #685 

Fixes [AB#532987](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/532987)

### DIFF
--- a/src/System Application/App/Azure File Services API/src/AFSOperationPayloadImpl.Codeunit.al
+++ b/src/System Application/App/Azure File Services API/src/AFSOperationPayloadImpl.Codeunit.al
@@ -6,6 +6,7 @@
 namespace System.Azure.Storage.Files;
 
 using System.Azure.Storage;
+using System;
 
 codeunit 8952 "AFS Operation Payload Impl."
 {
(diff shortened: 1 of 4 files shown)


JesperSchulz (approved):

gggdttt (approved):

darjoo (approved):

pkoutsogilas:
@Krajzys: Nice fix, thanks!
@JesperSchulz: Is this going to be backported to 24.2?

JesperSchulz:
> @Krajzys: Nice fix, thanks! @JesperSchulz: Is this going to be backported to 24.2?

Sure, I can do that 😊

f4n0:
@JesperSchulz Is there a hotfix for BC24 while we wait an actual fix?

PR #1029, opened 2024-05-01
Add ability to delete orphaned email messages

AndreasMoth:
This PR provides the ability to clean up orphaned email messages you may have created without an email outbox or sent email referencing them.
This is not a normal situation to end up in and only occurs if you simply create the message without linking.
It's not the most optimized solution, it's simply there to allow cleaning up incorrectly created messages.

Fixes [AB#524544](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/524544)

### DIFF
--- a/src/System Application/App/Email/src/Message/EmailMessage.Codeunit.al
+++ b/src/System Application/App/Email/src/Message/EmailMessage.Codeunit.al
@@ -75,7 +75,7 @@ codeunit 8904 "Email Message"
     /// Deletes messages that does not have a reference from either the email outbox nor sent email.
     /// This functionality is only needed if email messages have been created without any email outbox or sent email referencing it, otherwise they will be cleaned up automatically.
     /// </summary>
-    /// <param name="StartMessageId">The email message id to start from.</param>
+    /// <param name="StartMessageId">The email message id to start from. Using empty guid will start from the beginning.</param>
     /// <param name="MessagesToIterate">Number of email messages to loop over.</param>
     /// <returns>The next email message id to be checked. Returns empty guid if there are no more messages.</returns>
     procedure DeleteOrphanedMessages(StartMessageId: Guid; MessagesToIterate: Integer) NextMessageId: Guid
--- a/src/System Application/App/Email/src/Message/EmailMessage.Codeunit.al
+++ b/src/System Application/App/Email/src/Message/EmailMessage.Codeunit.al
@@ -72,6 +72,18 @@ codeunit 8904 "Email Message"
     end;
 
     /// <summary>
+    /// Deletes messages that does not have a reference from either the email outbox nor sent email.
+    /// This functionality is only needed if email messages have been created without any email outbox or sent email referencing it, otherwise they will be cleaned up automatically.
+    /// </summary>
+    /// <param name="StartMessageId">The email message id to start from.</param>
+    /// <param name="MessagesToIterate">Number of email messages to loop over.</param>
+    /// <returns>The next email message id to be checked. Returns empty guid if there are no more messages.</returns>
+    procedure DeleteOrphanedMessages(StartMessageId: Guid; MessagesToIterate: Integer) NextMessageId: Guid
+    begin
+        exit(EmailMessageImpl.DeleteOrphanedMessages(StartMessageId, MessagesToIterate));
+    end;
+
+    /// <summary>
     /// Gets the body of the email message.
     /// </summary>
     /// <returns>The body of the email.</returns>
--- a/src/System Application/App/Email/src/Message/EmailMessageImpl.Codeunit.al
+++ b/src/System Application/App/Email/src/Message/EmailMessageImpl.Codeunit.al
@@ -621,20 +621,20 @@ codeunit 8905 "Email Message Impl."
         exit(EmailMessage.Id);
     end;
 
-    local procedure DeleteIfOrphaned(var EmailMessage: Record "Email Message"): Boolean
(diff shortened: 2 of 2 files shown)


src/System Application/App/Email/src/Message/EmailMessage.Codeunit.al
```diff
@@ -71,6 +71,18 @@ codeunit 8904 "Email Message"
         exit(EmailMessageImpl.Get(MessageId));
     end;
 
+    /// <summary>
+    /// Deletes messages that does not have a reference from either the email outbox nor sent email.
+    /// This functionality is only needed if email messages have been created without any email outbox or sent email referencing it, otherwise they will be cleaned up automatically.
+    /// </summary>
+    /// <param name="StartMessageId">The email message id to start from.</param>
```
stkillen: Should document that StartMessageId can be null for first request

src/System Application/App/Email/src/Message/EmailMessageImpl.Codeunit.al
```diff
@@ -599,6 +599,43 @@ codeunit 8905 "Email Message Impl."
...
     end;
 
+    procedure DeleteOrphanedMessages(StartMessageId: Guid; MessagesToIterate: Integer) NextMessageId: Guid
+    var
+        EmailMessage: Record "Email Message";
+        EmptyGuid: Guid;
+        MessageNo: Integer;
+    begin
+        EmailMessage.SetLoadFields(Id);
+        EmailMessage.SetFilter(Id, '>=%1', StartMessageId);
```
stkillen, line 610: Larger or equal than a guid value? Is this intentional?

darjoo: This wouldn't be right. We use `CreateGuid()` for the Id, so it doesn't guarantee that it's going to be in sequence.

darjoo (commented):
Can we instead create a retention policy that has an additional validation, only if orphan, will it be deleted.

src/System Application/App/Email/src/Message/EmailMessageImpl.Codeunit.al
```diff
@@ -599,6 +599,43 @@ codeunit 8905 "Email Message Impl."
...
+        if not EmailMessage.FindSet() then
```
darjoo, line 611: This would lock all rows after that Id. Customer can run into locking issues, if they have a ton which need to be deleted but are also trying to send emails where processes are running outside of business hours.

src/System Application/App/Email/src/Message/EmailMessageImpl.Codeunit.al
AndreasMoth: Yes, you start with zero guid and work your way up. I know we use CreateGuid but this is for cleaning up your ophaned messages. As Darrick says, we use CreateGuid() so they are not in sequence. That's why we need to start from zero guid and work our way through, we can't start from some clever place as they are all over.

src/System Application/App/Email/src/Message/EmailMessageImpl.Codeunit.al
AndreasMoth: It only locks if there's a locktable. Furthermore this is exactly why partner can choose how many messages to iterate over, then commit etc. This way we also avoid having IcMs for iterating too many or if we iterate too few and it takes too long. They are in control.

darjoo: The findset will lock all the rows it finds with `updlock`. That's the default. 
I'd suggest we put readisolation to readcommited and only in the delete, we set a filter to that id and delete. Then we won't lock anything besides the one we're deleting. 
The code will run slower, but we eliminate the possibility of getting a lock issue if they iterate too many.

AndreasMoth: https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/methods-auto/record/record-findset-boolean-method
"Set this parameter to true if you want to modify any records in the set; otherwise, set the parameter to false. If you set this parameter to true, then the records will be read with Updlock. If you set this parameter to false, then you can still modify the records in the set, but these updates will not be performed optimally. The default value is false."

So no lock is set (since default value is false), with that being said, we can still change readisolation to only read committed, since I think by default it reads uncommitted. Will add that.

src/System Application/App/Email/src/Message/EmailMessageImpl.Codeunit.al
```diff
@@ -599,6 +599,44 @@ codeunit 8905 "Email Message Impl."
...
+    begin
+        EmailOutbox.SetRange("Message Id", EmailMessage.Id);
+        if not EmailOutbox.IsEmpty() then
+            exit(false);
+
+        SentEmail.SetRange("Message Id", EmailMessage.Id);
+        if not SentEmail.IsEmpty() then
+            exit(false);
+
+        exit(EmailMessage.Delete());
```
darjoo: If the delete fails, we continue on to the next. Might be misleading to the dev that we actually deleted everything they wanted.
May not be necessary to fix as this is still an edge case we're dealing with.

darjoo (dismissed):

darjoo (approved):

stkillen (approved):

JesperSchulz (approved):

PR #1048, opened 2024-05-03
[24.x] Add ability to delete orphaned email messages (#1029)

AndreasMoth:
This PR provides the ability to clean up orphaned email messages you may have created without an email outbox or sent email referencing them. This is not a normal situation to end up in and only occurs if you simply create the message without linking.
It's not the most optimized solution, it's simply there to allow cleaning up incorrectly created messages.

Fixes [AB#533515](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/533515)

### DIFF
--- a/src/System Application/App/Email/src/Message/EmailMessage.Codeunit.al
+++ b/src/System Application/App/Email/src/Message/EmailMessage.Codeunit.al
@@ -72,6 +72,18 @@ codeunit 8904 "Email Message"
     end;
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


haoranpb:
It might be a late message here, since it's a backport, but consider adding tests.

It will ensure the functionality actually works, and also provide an example for partners to get started.

JesperSchulz (approved):

PR #1062, opened 2024-05-06
 Add ShowOnly procedure to ABSOptionalParameters Codeunit

### LINKED ISSUE #894: [BC Idea]: Allow listing files/directories only in azure blob storage
### BC Idea Link

https://experience.dynamics.com/ideas/idea/?ideaid=d783420e-c2f5-ee11-a73d-6045bd808754

### Description

Add showonly procedure to ABSOptionalParameters to allow filtering for deleted/files/directories (implement as enum). This Parameter could be used when calling ListBlobs.

Internal work item: [AB#524631](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/524631)

ChrisBlankDe:
#### Summary
Allow listing files/directories only in azure blob storage.

#### Work Item(s) 
Fixes #894 

Fixes [AB#524631](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/524631)

### DIFF
--- a/src/System Application/App/Azure Blob Services API/src/Helper/ABSOptionalParameters.Codeunit.al
+++ b/src/System Application/App/Azure Blob Services API/src/Helper/ABSOptionalParameters.Codeunit.al
@@ -403,6 +403,16 @@ codeunit 9047 "ABS Optional Parameters"
         SetParameter('include', "Value");
     end;
 
+    /// <summary>
+    /// Specifies the datasets to be returned in the response.
+    /// </summary>
+    /// see: https://learn.microsoft.com/rest/api/storageservices/list-blobs?tabs=microsoft-entra-id#uri-parameters
+    /// <param name="Value">The dataset to include in text</param>
+    procedure ShowOnly("Value": Text)
+    begin
+        SetParameter('showonly', "Value");
+    end;
+
     local procedure SetParameter(Header: Text; HeaderValue: Text)
     begin
         Parameters.Remove(Header);


JesperSchulz (approved):

src/System Application/App/Azure Blob Services API/src/Helper/ABSOptionalParameters.Codeunit.al
```diff
@@ -403,6 +403,16 @@ codeunit 9047 "ABS Optional Parameters"
         SetParameter('include', "Value");
     end;
 
+    /// <summary>
+    /// Specifies the datasets to be returned in the response.
```
mazhelez, line 407: Doc is off,

ChrisBlankDe: i don't understand your comment. can you please repeat this in more detail?

JesperSchulz: Your code documentation summary states "Specifies the datasets to be returned in the response.", just as the parameter description says "The dataset to include in text". Looks like a copy/paste from above. But I actually think it's correct? Please confirm.

JesperSchulz (changes requested):

JesperSchulz (approved):
Reading the documentation once more, I stand by my approval :-)

https://learn.microsoft.com/en-us/rest/api/storageservices/list-blobs?tabs=microsoft-entra-id#uri-parameters

src/System Application/App/Azure Blob Services API/src/Helper/ABSOptionalParameters.Codeunit.al
ChrisBlankDe: Confirm. No copy from above. Copy/Paste from msft learn: https://learn.microsoft.com/en-us/rest/api/storageservices/list-blobs?tabs=microsoft-entra-id#uri-parameters
I did not copy the listing of the possible values since we probably won't notice immediately if something changes there.
every developer who works with the azure blob looks there anyway.

mazhelez: My bad then.

mazhelez (approved):

PR #1076, opened 2024-05-07
ExtensionDeploymentStatus new Actions

### LINKED ISSUE #977: [BC Idea]: Action in Page 2508 to Extension Management
### BC Idea Link

https://experience.dynamics.com/ideas/idea/?ideaid=7a60b923-4f01-ef11-a73c-0022484ddea9

### Description

When uploading a PTE we have to refresh the "Extension Installation Status" page, which the is the main page in the browser. When we want to navigate back to the "installed extensions" page you have to search again.

the idea is a simple navigation button to open the page "2500 "Extension management"

Internal work item: [AB#527490](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/527490)

JulianTillmann:
#### Summary
Added new actions to page "Extensions Deployment Status"

- Refresh: for refreshing the status of the deployment
- Upload Extension: to re-upload an extension if the first try failed for some reason
- Extension Management: to go back to the extension management page

Unfortunately it's not possible to test this Onprem and there does not seem to be a whole lot of test automation for this which I could re-use. Hopefully you can qualitity check this manually.

#### Work Item(s)
Fixes #977

Fixes [AB#527490](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/527490)

### DIFF
--- a/src/System Application/App/Extension Management/src/ExtensionDeploymentStatus.Page.al
+++ b/src/System Application/App/Extension Management/src/ExtensionDeploymentStatus.Page.al
@@ -111,7 +111,8 @@ page 2508 "Extension Deployment Status"
                     NavAppTenantOperationTable: Record "NAV App Tenant Operation";
                     ExtensionOperationImpl: Codeunit "Extension Operation Impl";
                 begin
-                    NavAppTenantOperationTable.CopyFilters(Rec);
+                    NavAppTenantOperationTable.Copy(Rec);
+                    NavAppTenantOperationTable.SetFilter(Status, '%1|%2|%3', NavAppTenantOperationTable.Status::Unknown, NavAppTenantOperationTable.Status::NotFound, NavAppTenantOperationTable.Status::InProgress);
                     if NavAppTenantOperationTable.FindSet() then
                         repeat
                             ExtensionOperationImpl.RefreshStatus(NavAppTenantOperationTable."Operation ID");
@@ -120,6 +121,7 @@ page 2508 "Extension Deployment Status"
             }
             action("Upload Extension")
             {
+                ApplicationArea = All;
                 Caption = 'Upload Extension';
                 Image = Import;
                 RunObject = page "Upload And Deploy Extension";
--- a/src/System Application/App/Extension Management/src/ExtensionDeploymentStatus.Page.al
+++ b/src/System Application/App/Extension Management/src/ExtensionDeploymentStatus.Page.al
@@ -115,7 +115,7 @@ page 2508 "Extension Deployment Status"
                     if NavAppTenantOperationTable.FindSet() then
                         repeat
                             ExtensionOperationImpl.RefreshStatus(NavAppTenantOperationTable."Operation ID");
-                        until NavAppTenantOperationTable.Next = 0;
+                        until NavAppTenantOperationTable.Next() = 0;
                 end;
             }
             action("Upload Extension")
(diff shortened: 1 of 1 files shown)


JesperSchulz:
Hi @JulianTillmann,

you've got a few easy to fix codecop violations. Could you look into those to get green ticks in the CI step? Then we'll process the PR 🙂

src/System Application/App/Extension Management/src/ExtensionDeploymentStatus.Page.al
```diff
@@ -99,6 +100,49 @@ page 2508 "Extension Deployment Status"
...
+                    ExtensionOperationImpl: Codeunit "Extension Operation Impl";
+                begin
+                    NavAppTenantOperationTable.CopyFilters(Rec);
+                    if NavAppTenantOperationTable.FindSet() then
+                        repeat
+                            ExtensionOperationImpl.RefreshStatus(NavAppTenantOperationTable."Operation ID");
+                        until NavAppTenantOperationTable.Next() = 0;
+                end;
+            }
+            action("Upload Extension")
```
EmilDamsbo, line 122: Everywhere else, we are adding the `ApplicationArea = All`
While it is not required, let's be consistent in this and add it here too.

EmilDamsbo (changes requested):

src/System Application/App/Extension Management/src/ExtensionDeploymentStatus.Page.al
```diff
@@ -99,6 +100,49 @@ page 2508 "Extension Deployment Status"
...
+                ApplicationArea = All;
+                ToolTip = 'Refresh the deployment details.';
+                Image = Refresh;
+
+                trigger OnAction()
+                var
+                    NavAppTenantOperationTable: Record "NAV App Tenant Operation";
+                    ExtensionOperationImpl: Codeunit "Extension Operation Impl";
+                begin
+                    NavAppTenantOperationTable.CopyFilters(Rec);
```
BazookaMusic: I think this doesn't copy marked records (rows you select with cntrl)

This one should (although it also copies the value of the record)
https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/methods-auto/record/record-copy-method

I

BazookaMusic: Without marked records, all the rows in view will be refreshed even if you select some

BazookaMusic: I might be wrong though and copy filters does copy the marked records, it should be straightforward to test by selecting rows and calling the action

src/System Application/App/Extension Management/src/ExtensionDeploymentStatus.Page.al
```diff
@@ -99,6 +100,49 @@ page 2508 "Extension Deployment Status"
...
+                    if NavAppTenantOperationTable.FindSet() then
+                        repeat
+                            ExtensionOperationImpl.RefreshStatus(NavAppTenantOperationTable."Operation ID");
```
thloke, line 118: We only need to refresh status if the status is `Unknown`, `NotFound`, or `InProgress`. Once it's `Failed` or `Completed` it's not going to change anymore. This should cut down the number of operations that actually need refreshing. Might be easier still to just set a filter on `NavAppTenantOperationTable` before `FindSet` to get the records that aren't in `Failed` or `Completed` state.

EmilDamsbo (approved):

thloke (approved):

PR #1099, opened 2024-05-13
add locktable to no series stateless impl

grobyns:
#### Summary 
Two or more sessions cannot work simultaneously and get new numbers from number series when using the stateteless impl. because it doesn't lock the no. series line. This PR adds a locktable.

#### Work Item(s) 
Fixes [AB#534262](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/534262)

### DIFF
--- a/src/Business Foundation/App/NoSeries/src/Single/NoSeriesStatelessImpl.Codeunit.al
+++ b/src/Business Foundation/App/NoSeries/src/Single/NoSeriesStatelessImpl.Codeunit.al
@@ -28,6 +28,11 @@ codeunit 306 "No. Series - Stateless Impl." implements "No. Series - Single"
 
     procedure GetNextNo(var NoSeriesLine: Record "No. Series Line"; UsageDate: Date; HideErrorsAndWarnings: Boolean): Code[20]
     begin
+        if not NoSeriesLine.IsTemporary() then begin
+            NoSeriesLine.LockTable();
+            NoSeriesLine.Find();
+        end;
+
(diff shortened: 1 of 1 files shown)


AndreasMoth (approved):

JesperSchulz (approved):

src/Business Foundation/App/NoSeries/src/Single/NoSeriesStatelessImpl.Codeunit.al
```diff
@@ -28,6 +28,11 @@ codeunit 306 "No. Series - Stateless Impl." implements "No. Series - Single"
 
     procedure GetNextNo(var NoSeriesLine: Record "No. Series Line"; UsageDate: Date; HideErrorsAndWarnings: Boolean): Code[20]
     begin
+        if not NoSeriesLine.IsTemporary() then begin
+            NoSeriesLine.LockTable();
```
BardurKnudsen, line 32: Fine, but I would go for a more modern approach and write NoSeriesLine.ReadIsolation(IsolationLevel::UpdLock).

Also, I think this part goes together with the 'if ModifySeries then ....' section in GetNextNoInternal, so either the update is moved into this function or the lock+read is moved to GetNextNoInternal.

BardurKnudsen: The same locking is also missing in NoSeriesSequenceImpl.

PR #1225, opened 2024-05-31
Incremented Version number to 24.3

mazhelez:
Incremented Version number to 24.3
AB#537084


darjoo (approved):

aholstrup1 (commented):
We also have a repoversion in .github/AL-Go-Settings.json that needs to be updated

PR #1243, opened 2024-06-03
fix bug related to daily no. series lines

grobyns:
#### Summary 
The No.  Series module tries to use the No Series Line given before looking for a new line to use. This has the effect that as long as the usagedate is newer the given line would be used even though a closer match exists. 
This fix makes it so the given line is only used if the starting dates from the given line and the new line match. If the dates don't match, the new line is used. 

#### Work Item(s) 
Fixes [AB#537258](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/537258)

### DIFF
--- a/src/Business Foundation/App/NoSeries/src/Single/NoSeriesImpl.Codeunit.al
+++ b/src/Business Foundation/App/NoSeries/src/Single/NoSeriesImpl.Codeunit.al
@@ -150,6 +150,7 @@ codeunit 304 "No. Series - Impl."
     procedure GetNoSeriesLine(var NoSeriesLine: Record "No. Series Line"; NoSeriesCode: Code[20]; UsageDate: Date; HideErrorsAndWarnings: Boolean): Boolean
     var
         NoSeriesRec: Record "No. Series";
+        NoSeriesLine2: Record "No. Series Line";
         NoSeries: Codeunit "No. Series";
         NoSeriesErrorsImpl: Codeunit "No. Series - Errors Impl.";
 #if not CLEAN24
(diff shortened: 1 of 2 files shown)


src/Business Foundation/App/NoSeries/src/Single/NoSeriesImpl.Codeunit.al
```diff
@@ -163,12 +164,21 @@ codeunit 304 "No. Series - Impl."
             UsageDate := WorkDate();
 
         // Find the No. Series Line closest to the usage date
-        NoSeriesLine.Reset();
-        NoSeriesLine.SetCurrentKey("Series Code", "Starting Date");
-        NoSeriesLine.SetRange("Series Code", NoSeriesCode);
-        NoSeriesLine.SetRange("Starting Date", 0D, UsageDate);
-        NoSeriesLine.SetRange(Open, true);
-        if (NoSeriesLine."Line No." <> 0) and (NoSeriesLine."Series Code" = NoSeriesCode) then begin
+        NoSeriesLine2.Reset();
```
AndreasMoth, line 167: NoSeriesLine may be temporary and have lines that have been closed (but are not closed in the database which this looks up). If it's temporary, we should be looking up in NoSeriesLine (or a reference to the same data at least)

grobyns: As you can see a bit below, NoSeriesLine2 is only used to check whether the line we would find matches in start date to the line given. All other code still acts on NoSeriesLine ..

WaelAbuSeada (approved):

PR #1249, opened 2024-06-03
[Edit-In Excel] Remove Invalid Filter errors when tfield does not exist on page + is not a key

DenLilleMand:
#### Summary 
We have 2 issues with invalid filters:

(1) People will try to filter on fields that are not keys and is not exposed on the page.
(2) People will try to filter on table fields that does not have the same external odata name as its corresponding page field (Platform just externalize the table field name right now)

This PR fixes issue (1).

#### Work Item(s) 
Fixes
[AB#537326](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/537326/)


src/System Application/App/Edit in Excel/src/Filters/EditInExcelFiltersImpl.Codeunit.al
```diff
@@ -161,6 +166,18 @@ codeunit 1491 "Edit in Excel Filters Impl."
             exit(false);
         end;
 
+        ExternalTableFieldName := NameJsonToken.AsValue().AsText();
+
+        PageControlField.SetRange(PageNo, PageNumber); // Does this method work also for fields added through Extensions ?
```
JesperSchulz: Do you intend to address this question before check in?

src/System Application/App/Edit in Excel/src/Filters/EditInExcelFiltersImpl.Codeunit.al
```diff
@@ -143,12 +146,14 @@ codeunit 1491 "Edit in Excel Filters Impl."
     end;
 
     local procedure ReadNodeValues(JsonFilterObject: JsonObject; ODataJsonPayload: JsonObject; var ODataFieldName: Text;
```
JesperSchulz: This formatting / indentation looks odd. Is that intended?

src/System Application/App/Edit in Excel/src/Filters/EditInExcelFiltersImpl.Codeunit.al
```diff
@@ -120,11 +122,12 @@ codeunit 1491 "Edit in Excel Filters Impl."
                             ChildNodesJsonToken.AsObject(),
                             ODataJsonPayload,
                             FieldFilterGroupingOperator,
-                            "Edit in Excel Filter Collection Type".FromInteger(ExcelFilterNodeType.AsInteger()));
+                            "Edit in Excel Filter Collection Type".FromInteger(ExcelFilterNodeType.AsInteger()),
```
JesperSchulz, line 126: Why the line break here?

JesperSchulz (changes requested):
Some minor comments. Take a look 🙂

src/System Application/App/Edit in Excel/src/EditinExcelImpl.Codeunit.al
```diff
@@ -43,6 +43,7 @@ codeunit 1482 "Edit in Excel Impl."
         CreateEndpointForObjectTxt: Label 'Creating endpoint for %1 %2.', Locked = true;
         EditInExcelHandledTxt: Label 'Edit in excel has been handled.', Locked = true;
         EditInExcelOnlySupportPageWebServicesTxt: Label 'Edit in Excel only support web services created from pages.', Locked = true;
+        EditInExcelInvalidFilterErr: Label 'We had to remove the filters applied to the following fields, because they are not available in the Office Add-In. Please notice that as a result of this that the number of rows you exported to Excel could vary.\n %1', Locked = true;
```
encimita: Since we use this in a `Message()`, should it be translated and hence not locked?

(Also, smal typo "notice that ... that").

encimita: Oh, one more thing: the new line in the AL `Message()` function is not `\n` but rather `\`

See:
https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/methods-auto/dialog/dialog-message-method#:~:text=Use%20a%20backslash%20(%5C)%20to%20start%20a%20new%20line

src/System Application/App/Edit in Excel/src/Filters/EditInExcelFiltersImpl.Codeunit.al
```diff
@@ -185,6 +202,24 @@ codeunit 1491 "Edit in Excel Filters Impl."
...
+    begin
+        if PageMetadata.Get(PageNumber) then begin
+            FieldMetadata.SetRange(FieldMetadata.TableNo, PageMetadata.SourceTable);
+            FieldMetadata.SetRange(FieldMetadata.ExternalName, ExternalTableFieldName);
+            if FieldMetadata.FindFirst() then begin
+                KeyRec.SetRange(TableNo, PageMetadata.SourceTable);
+                KeyRec.SetRange("Key", FieldMetadata.FieldName);
+                exit(not KeyRec.IsEmpty())
+            end;
+        end;
```
encimita, line 232: Keep the `SetRange` consistent, e.g.
```diff
+FieldMetadata.SetRange(TableNo, PageMetadata.SourceTable);
-FieldMetadata.SetRange(FieldMetadata.TableNo, PageMetadata.SourceTable);
```

encimita (approved):
Looks good to me apart from a couple of minor comments and Jesper's comments

src/System Application/App/Edit in Excel/src/EditinExcelImpl.Codeunit.al
StefanSosic: Missing comment for %1 in newly added Label EditInExcelInvalidFilterErr

src/System Application/App/Edit in Excel/src/EditinExcelImpl.Codeunit.al
```diff
@@ -865,13 +867,32 @@ codeunit 1482 "Edit in Excel Impl."
...
     end;
-}
+
+    local procedure FormatFilterErrors(FilterErrors: Dictionary of [Text, Boolean]): Text
+    var
+        ConcatenatedErrors: Text;
+        ErrorText: Text;
+        IsFirst: Boolean;
+    begin
+        ConcatenatedErrors := '';
```
StefanSosic: Local variable in local procedure, why "clear" it?

src/System Application/App/Edit in Excel/src/EditinExcelImpl.Codeunit.al
```diff
@@ -865,13 +867,32 @@ codeunit 1482 "Edit in Excel Impl."
...
+        IsFirst := true;
+
+        foreach ErrorText in FilterErrors.Keys() do begin
```
StefanSosic: I don't like the pattern used here.
Take a look how it already done in Base App for example:
image

So, adding, than removing trailing ','

src/System Application/App/Edit in Excel/src/Filters/EditInExcelFiltersImpl.Codeunit.al
```diff
@@ -185,6 +202,24 @@ codeunit 1491 "Edit in Excel Filters Impl."
         exit(true);
     end;
 
+    local procedure IsKey(PageNumber: Integer; ExternalTableFieldName: Text): Boolean
+    var
+        PageMetadata: Record "Page Metadata";
```
StefanSosic: Sort the variables

src/System Application/App/Edit in Excel/src/Filters/EditInExcelFiltersImpl.Codeunit.al
```diff
@@ -185,6 +202,24 @@ codeunit 1491 "Edit in Excel Filters Impl."
...
+        FieldMetadata: Record "Field";
+        KeyRec: Record "Key";
+    begin
+        if PageMetadata.Get(PageNumber) then begin
```
StefanSosic, line 223: Un-necessary identation, I would be here more exclusive. 
if not PageMetadata.Get than exit(false), than you don't surround by begin and end and code looks nicer

StefanSosic (changes requested):

src/System Application/App/Edit in Excel/src/Filters/EditInExcelFiltersImpl.Codeunit.al
DenLilleMand: @JesperSchulz Just a continuation of what we already had, do you want everything on one line or?

encimita: I think this might be an issue with how Github wraps the line, and it looks like there is a line break in the middle of the enum name.
image

But you can see that the line number does not increase, and if you go to the three dots on the top right of the file and click View File, indentation/newlines actually look fine IMHO :)

src/System Application/App/Edit in Excel/src/EditinExcelImpl.Codeunit.al
```diff
@@ -43,6 +43,7 @@ codeunit 1482 "Edit in Excel Impl."
         CreateEndpointForObjectTxt: Label 'Creating endpoint for %1 %2.', Locked = true;
         EditInExcelHandledTxt: Label 'Edit in excel has been handled.', Locked = true;
         EditInExcelOnlySupportPageWebServicesTxt: Label 'Edit in Excel only support web services created from pages.', Locked = true;
+        EditInExcelInvalidFilterErr: Label 'We had to remove the filters applied to the following fields, because they are not available in the Office Add-In. Please notice that as a result of this the number of rows you exported to Excel could vary.\ %1', Locked = true;
```
encimita: (I see the label is still `Locked` and hence will not be translated, is that on purpose?)

src/System Application/App/Edit in Excel/src/EditinExcelImpl.Codeunit.al
```diff
@@ -865,13 +867,26 @@ codeunit 1482 "Edit in Excel Impl."
...
-        EditinExcelFilters.ReadFromJsonFilters(Filter, Payload, TenantWebService."Object ID");
+        EditinExcelFilters.ReadFromJsonFilters(Filter, Payload, TenantWebService."Object ID", FilterErrors);
         EditinExcel.OnEditInExcelWithFilters(ServiceName, EditinExcelFilters, SearchString, Handled);
         if Handled then begin
             Session.LogMessage('0000IG8', EditInExcelHandledTxt, Verbosity::Normal, DataClassification::SystemMetadata, TelemetryScope::ExtensionPublisher, 'Category', EditInExcelTelemetryCategoryTxt);
             exit;
         end;
-
+        if FilterErrors.Count() > 0 then
+            Message(StrSubstNo(EditInExcelInvalidFilterErr, FormatFilterErrors(FilterErrors)));
```
encimita: Minor: `Message` has an overload where you can specify the values to replace in the message without having to use `Strsubstno`. Functionality is the same, but it gains a bit in readability.

```suggestion
            Message(EditInExcelInvalidFilterErr, FormatFilterErrors(FilterErrors));
```

src/System Application/App/Edit in Excel/src/Filters/EditInExcelFiltersImpl.Codeunit.al
```diff
@@ -161,6 +169,26 @@ codeunit 1491 "Edit in Excel Filters Impl."
...
+
+        FieldAlName := AlNameJsonToken.AsValue().AsText();
+
+        PageControlField.SetRange(PageNo, PageNumber);
+        PageControlField.SetRange(ControlName, FieldAlName);
+
+        if PageControlField.IsEmpty() then
+            if not IsKey(PageNumber, ExternalTableFieldName) then begin
+                Session.LogMessage('0000I5U', FieldNotOnThePageTxt, Verbosity::Normal, DataClassification::SystemMetadata, TelemetryScope::ExtensionPublisher, 'Category', EditInExcelTelemetryCategoryTxt);
+                if not FilterErrors.ContainsKey(ExternalTableFieldName) then
```
encimita: Is this a typo that we check if it contains `ExternalTableFieldName` but then we add `FieldAlName`?

encimita (approved):

src/System Application/App/Edit in Excel/src/Filters/EditInExcelFiltersImpl.Codeunit.al
DenLilleMand: @encimita no it's on purpose, because the keys in this dictionary is eventually used in the UI to tell the user what filters we removed, I thought that there was a bigger chance they would enjoy the AL field name instead of the external field name. I guess that the field caption would be ideal actually.

encimita: Ok fair, but then shouldn't we check if the `FieldAlName` is in the dictionary and then add `FieldAlName`?

Instead of checking if `ExternalTableFieldName` is in the dictionary and then we add something else (`FieldAlName`)?

DenLilleMand: @encimita Yeah for sure, huge bug, ty!

src/System Application/App/Edit in Excel/src/EditinExcelImpl.Codeunit.al
DenLilleMand: @encimita No, I think I want it translated

StefanSosic: @DenLilleMand than there should be comment for %1, after removing Locked

DenLilleMand: @StefanSosic Yeah I added that, the compiler complained w/o it.

JesperSchulz (approved):

StefanSosic (approved):

encimita (approved):

PR #1260, opened 2024-06-04
[Copilot] Add authorization interface for managed resources

encimita:
#### Summary 
Adds a new function to `AzureOpenAI.Codeunit.al` that allows partners approved for the private preview program to use the Microsoft managed AI resources.

Reworked some of the logic in `AOAIAuthorization.Codeunit.al` to more clearly distinguish between the three authorization cases that we have today (first party, managed, self-managed), and created an enum to store that information.

Polish logic to disallow background calls to Azure OpenAI. The previous logic wanted to express that calls in the background are allowed if it's the partner's own AOAI resource (the partner decides in that case), but not if it's using Microsoft managed resources.

_Note: the logic at the moment does not verify that the partner AOAI deployment exists and is valid in the case of Managed resources. This is because we are currently in private preview, so the selected partners are already allowlisted. We will add the verification logic when we move to public preview._

#### Work Item(s) 
Fixes [AB#537361](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/537361)
Fixes [AB#527684](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/527684)

### DIFF
--- a/src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
+++ b/src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
@@ -22,35 +22,61 @@ codeunit 7767 "AOAI Authorization"
         Deployment: Text;
         [NonDebuggable]
         ApiKey: SecretText;
+        [NonDebuggable]
+        ManagedResourceDeployment: Text;
+        ResourceUtilization: Enum "AOAI Resource Utilization";
 
     [NonDebuggable]
     procedure IsConfigured(CallerModule: ModuleInfo): Boolean
     var
-        AzureOpenAiImpl: Codeunit "Azure OpenAI Impl";
+        AzureOpenAIImpl: Codeunit "Azure OpenAI Impl";
         CurrentModule: ModuleInfo;
         ALCopilotFunctions: DotNet ALCopilotFunctions;
     begin
         NavApp.GetCurrentModuleInfo(CurrentModule);
 
-        if Deployment = '' then
-            exit(false);
+        case ResourceUtilization of
+            Enum::"AOAI Resource Utilization"::FirstParty:
+                exit((Deployment <> '') and ALCopilotFunctions.IsPlatformAuthorizationConfigured(CallerModule.Publisher(), CurrentModule.Publisher()));
+            Enum::"AOAI Resource Utilization"::SelfManaged:
+                exit((Deployment <> '') and (Endpoint <> '') and (not ApiKey.IsEmpty()));
+            Enum::"AOAI Resource Utilization"::MicrosoftManaged:
+                exit((Deployment <> '') and (Endpoint <> '') and (not ApiKey.IsEmpty()) and (ManagedResourceDeployment <> '') and AzureOpenAIImpl.IsTenantAllowlistedForFirstPartyCopilotCalls());
+        end;
 
-        if (Endpoint = '') and ApiKey.IsEmpty() then
-            exit(AzureOpenAiImpl.IsTenantAllowlistedForFirstPartyCopilotCalls()
-                or ALCopilotFunctions.IsPlatformAuthorizationConfigured(CallerModule.Publisher(), CurrentModule.Publisher()));
-
-        if (Endpoint = '') or ApiKey.IsEmpty() then
-            exit(false);
-
-        exit(true);
+        exit(false);
     end;
 
     [NonDebuggable]
-    procedure SetAuthorization(NewEndpoint: Text; NewDeployment: Text; NewApiKey: SecretText)
+    procedure SetMicrosoftManagedAuthorization(NewEndpoint: Text; NewDeployment: Text; NewApiKey: SecretText; NewManagedResourceDeployment: Text)
     begin
+        ClearVariables();
+
+        ResourceUtilization := Enum::"AOAI Resource Utilization"::MicrosoftManaged;
         Endpoint := NewEndpoint;
         Deployment := NewDeployment;
         ApiKey := NewApiKey;
+        ManagedResourceDeployment := NewManagedResourceDeployment;
+    end;
+
+    [NonDebuggable]
+    procedure SetSelfManagedAuthorization(NewEndpoint: Text; NewDeployment: Text; NewApiKey: SecretText)
+    begin
+        ClearVariables();
+
+        ResourceUtilization := Enum::"AOAI Resource Utilization"::SelfManaged;
+        Endpoint := NewEndpoint;
+        Deployment := NewDeployment;
+        ApiKey := NewApiKey;
+    end;
+
+    [NonDebuggable]
+    procedure SetFirstPartyAuthorization(NewDeployment: Text)
+    begin
+        ClearVariables();
+
+        ResourceUtilization := Enum::"AOAI Resource Utilization"::FirstParty;
+        Deployment := NewDeployment;
     end;
 
     [NonDebuggable]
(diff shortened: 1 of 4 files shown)


src/System Application/App/AI/src/Azure OpenAI/AOAIResourceUtilization.Enum.al
```diff
@@ -0,0 +1,36 @@
...
+/// </summary>
+enum 7771 "AOAI Resource Utilization"
+{
+    Access = Internal;
+    Extensible = false;
+
+    /// <summary>
+    /// The first party utilization (only available for Microsoft first party apps).
+    /// </summary>
+    value(0; FirstParty)
```
darjoo: Use "First Party", "Microsoft Managed" etc instead of concatenating the names without space.

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAI.Codeunit.al
```diff
@@ -70,6 +70,25 @@ codeunit 7771 "Azure OpenAI"
         exit(AzureOpenAIImpl.IsInitialized(CopilotCapability, ModelType, CallerModuleInfo));
     end;
 
+    /// <summary>
+    /// NOTE: This function is currently only available to selected partners.
```
darjoo: I'd rather have this note in remarks. Though I understand why you put it here. 
If you have no strong preference, I'd move it :D

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAI.Codeunit.al
```diff
@@ -70,6 +70,25 @@ codeunit 7771 "Azure OpenAI"
...
+    /// <param name="ModelType">The model type to set authorization for.</param>
+    /// <param name="Endpoint">The endpoint to use to verify access to Azure OpenAI. This is used only for verification, not for actual Azure OpenAI calls.</param>
+    /// <param name="Deployment">The deployment to use to verify access to Azure OpenAI. This is used only for verification, not for actual Azure OpenAI calls.</param>
+    /// <param name="ApiKey">The API key to use  to verify access to Azure OpenAI. This is used only for verification, not for actual Azure OpenAI calls.</param>
+    /// <param name="ManagedResourceDeployment">The managed deployment to use for the model type.</param>
+    /// <remarks>Endpoint would look like: https://resource-name.openai.azure.com/
+    /// Deployment would look like: gpt-35-turbo-16k
+    /// </remarks>
+    [NonDebuggable]
+    procedure SetManagedResourceAuthorization(ModelType: Enum "AOAI Model Type"; Endpoint: Text; Deployment: Text; ApiKey: SecretText; ManagedResourceDeployment: Text)
```
darjoo, line 87: Managed resource shouldn't need endpoint and apikey. Also there should only be one deployment parameter.

EvgenijKorovin (approved):

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -32,27 +35,50 @@ codeunit 7767 "AOAI Authorization"
...
-        if (Endpoint = '') and ApiKey.IsEmpty() then
-            exit(AzureOpenAiImpl.IsTenantAllowlistedForFirstPartyCopilotCalls()
-                or ALCopilotFunctions.IsPlatformAuthorizationConfigured(CallerModule.Publisher(), CurrentModule.Publisher()));
+        exit(false);
+    end;
 
-        if (Endpoint = '') or ApiKey.IsEmpty() then
-            exit(false);
+    [NonDebuggable]
+    procedure SetMicrosoftManagedAuthorization(NewEndpoint: Text; NewDeployment: Text; NewApiKey: SecretText; NewManagedResourceDeployment: Text)
```
darjoo, line 51: Related to the facade API comment, no reason for endpoint and key right? Also only one deployment parameter.

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
```diff
@@ -479,9 +501,17 @@ codeunit 7772 "Azure OpenAI Impl"
         ALCopilotFunctions: DotNet ALCopilotFunctions;
         ALCopilotOperationResponse: DotNet ALCopilotOperationResponse;
         Error: Text;
+        EmptySecretText: SecretText;
     begin
         ClearLastError();
-        ALCopilotAuthorization := ALCopilotAuthorization.Create(AOAIAuthorization.GetEndpoint(), AOAIAuthorization.GetDeployment(), AOAIAuthorization.GetApiKey());
+        case AOAIAuthorization.GetResourceUtilization() of
+            Enum::"AOAI Resource Utilization"::MicrosoftManaged:
+                ALCopilotAuthorization := ALCopilotAuthorization.Create(EmptySecretText, AOAIAuthorization.GetManagedResourceDeployment(), EmptySecretText);
```
darjoo, line 509: Fine for now, but I'd prefer if we add additional method on platform. `CreateMicrosoftManaged` forexample.

darjoo: For clearer API calls and to get rid of the `EmptySecretText` variable

MonicaPAhuja (approved):

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
encimita: Agree with this. We will need to do it anyway when we move this to public preview, so deferring to that time.

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAI.Codeunit.al
encimita: No strong preference! Just wanted it to be prominent in VSCode, but remarks also works :)

src/System Application/App/AI/src/Azure OpenAI/AOAIResourceUtilization.Enum.al
encimita: On it :)

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
encimita: See other comment!

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAI.Codeunit.al
encimita: We need to be able to check that the partner has a valid active AOAI deployment for them to be able to use managed resources. We will not do the check for now (private preview) but we have to do it for public preview. It seems the easiest way is to have them specify still their API key, deployment name and endpoint, and we can verify it's a valid one. So this way the final interface is already the one in this PR, which requires these values.

MonicaPAhuja (approved):

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAI.Codeunit.al
darjoo: Lets obsolete it in the future and get rid of this once the AOAI requirement goes away.

darjoo (approved):

PR #1267, opened 2024-06-06
NoSeriesManagement - add var to handle obsolete No Series

bussik:
On the previous versions of BC there was an event:
image

But now there is new event:
image
And var is missing despite of that the function in body has a var.

### DIFF
--- a/src/Business Foundation/App/NoSeries/src/Legacy/NoSeriesManagement.Codeunit.al
+++ b/src/Business Foundation/App/NoSeries/src/Legacy/NoSeriesManagement.Codeunit.al
@@ -869,7 +869,7 @@ codeunit 396 NoSeriesManagement
     end;
 
     [Obsolete('This is a temporary method for compatibility only. Please use the "No. Series" codeunit instead', '24.0')]
-    internal procedure RaiseObsoleteOnAfterGetNextNo3(NoSeriesLine: Record "No. Series Line"; ModifySeries: Boolean)
+    internal procedure RaiseObsoleteOnAfterGetNextNo3(var NoSeriesLine: Record "No. Series Line"; ModifySeries: Boolean)
     begin
         OnAfterGetNextNo3(NoSeriesLine, ModifySeries);
     end;


AndreasMoth:
@bussik can you explain a bit more of what problem you are trying to solve and why you need this event? We have the new interface as well that you can implement, please also let us know why that interface does not work.

bussik:
> @bussik can you explain a bit more of what problem you are trying to solve and why you need this event? We have the new interface as well that you can implement, please also let us know why that interface does not work.

There are some partners solutions that are still on the old No. Series, and there is a need to have possible for change No. Series number.

StefanSosic:
> > @bussik can you explain a bit more of what problem you are trying to solve and why you need this event? We have the new interface as well that you can implement, please also let us know why that interface does not work.
> 
> @AndreasMoth There are some partners solutions that are still on the old No. Series, and there is a need to have possibility for change No. Series number. And it is strange that before there is possibility to do it, but in new version not.

Hi @bussik, 
make sure first that Issue #1268 get approved. Then move to implementing the changes and submitting PR for review, so that we don't review something which might not be Approved.

JesperSchulz:
@AndreasMoth, please review.

bussik:
> @AndreasMoth, please review.

Hello, when can we expect the bugfix? It will crash No. Series after upgrade to BC24.

grobyns:
Hi @bussik,
I'm don't think this fix will accomplish what you want. The code where this method is called looks like this:
```
        Result := NoSeriesSingle.GetNextNo(NoSeriesLine, UsageDate, HideErrorsAndWarnings);
        NoSeriesManagement.RaiseObsoleteOnAfterGetNextNo3(NoSeriesLine, true);
        exit(Result);
```
So var or no var, the NoSeriesLine passed to the event can't affect the result at this point. This was done on purpose as you should not be able to ask for a number from one no series only to have an event subscriber give you a number from another. 

I'd be happy to understand the scenario you're facing better so we can come up with a solution together. Unfortunately I don't see how this PR fixes the issue you describe... 
br/Gert

bussik:
> Hi @bussik, I'm don't think this fix will accomplish what you want. The code where this method is called looks like this:
> 
> ```
>         Result := NoSeriesSingle.GetNextNo(NoSeriesLine, UsageDate, HideErrorsAndWarnings);
>         NoSeriesManagement.RaiseObsoleteOnAfterGetNextNo3(NoSeriesLine, true);
>         exit(Result);
> ```
> 
> So var or no var, the NoSeriesLine passed to the event can't affect the result at this point. This was done on purpose as you should not be able to ask for a number from one no series only to have an event subscriber give you a number from another.
> 
> I'd be happy to understand the scenario you're facing better so we can come up with a solution together. Unfortunately I don't see how this PR fixes the issue you describe... br/Gert

Here it is explained more:
https://github.com/microsoft/ALAppExtensions/issues/26721#issuecomment-2188565599

grobyns:
I replied again here: https://github.com/microsoft/ALAppExtensions/issues/26721#issuecomment-2188565599

in addition to adding a var to the parameter, the code where it is called would also need a fix.

PR #1275, opened 2024-06-07
Model Upgrade and Function Calling added

qasimikram:
#### Summary 

- Model upgraded to GPT 4o
- Function calling added
- Feature improvements

#### Work Item(s) 
Fixes [AB#521281](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/521281)


src/System Application/App/Entity Text/src/EntityTextImpl.Codeunit.al
```diff
@@ -388,76 +303,100 @@ codeunit 2012 "Entity Text Impl."
...
-        NewLineChar := 10;
-
         NavApp.GetCurrentModuleInfo(EntityTextModuleInfo);
         if (not (Endpoint = '')) and (not (Deployment = ''))
         then
             AzureOpenAI.SetAuthorization(Enum::"AOAI Model Type"::"Chat Completions", Endpoint, Deployment, ApiKey)
         else
             if (not IsNullGuid(CallerModuleInfo.Id())) and (CallerModuleInfo.Publisher() = EntityTextModuleInfo.Publisher()) then
-                AzureOpenAI.SetAuthorization(Enum::"AOAI Model Type"::"Chat Completions", AOAIDeployments.GetGPT35TurboLatest())
+                AzureOpenAI.SetAuthorization(Enum::"AOAI Model Type"::"Chat Completions", AOAIDeployments.GetGPT4Preview()) //ToDo: Change to the correct deployment
```
msft-sam: Feel free to remove the todo - will log a bug once we can switch back to latest 😄

src/System Application/App/Entity Text/src/EntityTextImpl.Codeunit.al
```diff
@@ -388,76 +303,100 @@ codeunit 2012 "Entity Text Impl."
...
         AzureOpenAI.SetCopilotCapability(Enum::"Copilot Capability"::"Entity Text");
 
         AOAICompletionParams.SetMaxTokens(2500);
         AOAICompletionParams.SetTemperature(0.7);
+
+        AOAIChatMessages.AddTool(MagicFunction);
+
+        GenerateProdMktAdFunction.SetTextFormat(TextFormat);
+        AOAIChatMessages.AddTool(GenerateProdMktAdFunction);
+        AOAIChatMessages.SetToolChoice('auto');
```
msft-sam, line 362: I know there is the magic function as well - but perhaps we can have a single function and then set the tool choice to that function to ensure the call happens?

qasimikram: If I force the model to use just one function i.e. generate marketing text, then for all the harmful input the model will try to generate some marketing text. There is another option which make the model to always return a tools call i.e. 'Required', this option is not available in AOAI yet, but once available I plan to use that.

darjoo (approved):

src/System Application/App/Entity Text/src/EntityTextPrompts.Codeunit.al
```diff
@@ -0,0 +1,148 @@
...
+using System.Telemetry;
+
+codeunit 2016 "Entity Text Prompts"
+{
+    Access = Internal;
+    InherentEntitlements = X;
+    InherentPermissions = X;
+
+    [NonDebuggable]
+    internal procedure GetAzureKeyVaultSecret(var SecretValue: Text; SecretName: Text)
```
ChethanT, line 17: Consider using SecretText as the datatype for SecretValue param

ChethanT (approved):

PR #1331, opened 2024-06-13
Change Temp Current Sequence No to BigInteger

haoranpb:
#### Summary 

Replace `field(15; "Temp Current Sequence No."; Integer)` with `BigInteger` type.

Directly updating the field with new capability from Platform.

#### Work Item(s) 
Fixes [AB#535564](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/535564)


src/Business Foundation/App/NoSeries/src/Setup/NoSeriesLine.Table.al
```diff
@@ -131,6 +131,16 @@ table 309 "No. Series Line"
...
             DataClassification = SystemMetadata;
             Access = Internal;
+            ObsoleteReason = 'This field is replaced by field 16 - "Temp. Current Sequence No.".';
+            ObsoleteTag = '24.0';
+            ObsoleteState = Removed;
+        }
+        field(16; "Temp. Current Sequence No."; BigInteger)
+        {
+            Caption = 'Temporary Sequence Number';
+            Description = 'The field should only have value during runtime';
```
AndreasMoth: Maybe "This field holds the next sequence number during batching for sequences. It is only used during runtime and no values will be stored in the database.

AndreasMoth (dismissed):

src/Business Foundation/App/NoSeries/src/Setup/NoSeriesLine.Table.al
haoranpb:
```suggestion
            Description = 'This field holds the next sequence number during batching for sequences. It is only used during runtime and no values will be stored in the database.';
```

grobyns (approved):

src/Business Foundation/App/NoSeries/src/Setup/NoSeriesLine.Table.al
```diff
@@ -131,6 +131,18 @@ table 309 "No. Series Line"
             Caption = 'Temporary Sequence Number';
             DataClassification = SystemMetadata;
             Access = Internal;
+            ObsoleteReason = 'This field is replaced by field 16 - "Temp. Current Sequence No.".';
+#pragma warning disable AS0115
+            ObsoleteState = Removed;
```
StefanSosic: Is it better to be first pending and then with pragma also Removed in another version?

AndreasMoth (dismissed):

src/Business Foundation/App/NoSeries/src/Setup/NoSeriesLine.Table.al
AndreasMoth: I believe this should be fine, it's an internal field that should not be used by anyone else.

AndreasMoth: This is no longer relevant as we now have capability of upgrading integer field to BigInteger.

AndreasMoth (dismissed):

StefanSosic (approved):

darjoo (approved):

PR #1344, opened 2024-06-17
New pages for scheduled performance profiles

BazookaMusic:
#### Summary
This PR adds new pages and a supporting codeunit to setup and configure scheduled performance profiles. The idea of the feature is to allow an admin or a user to setup a schedule for when performance profiling should happen. The schedule contains details like the user, session type and when profiling should happen. Then, during user activity, if these details match, the session will be profiled.

Fixes [AB#538509](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/538509)


kalberes (approved):

src/System Application/App/Performance Profiler/src/ActivityType.Enum.al
```diff
@@ -0,0 +1,38 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+namespace System.Tooling;
+
+/// <summary>
+/// The supported activity types for creating scheduled based sampling profiles
+/// </summary>
+enum 1932 "Activity Type"
```
nikolakukrika: We should get a bit more specific name so we do not get clashes. 
Performance Profiler Activity Type?

nikolakukrika: Perf. Profiler Activity Type

src/System Application/App/Performance Profiler/src/PerfProfilerScheduleCard.Page.al
```diff
@@ -0,0 +1,216 @@
...
+                    ApplicationArea = All;
+                    Caption = 'Enabled';
+                    ToolTip = 'Specifies whether the schedule is enabled.';
+                    AboutText = 'Specifies whether the schedule is enabled.';
+                }
+                field("Start Time"; Rec."Starting Date-Time")
+                {
+                    ApplicationArea = All;
+                    Caption = 'Start Time';
+                    ToolTip = 'Specifies the time the schedule will start.';
```
nikolakukrika: Is it will start or it was started?
The same question for will end vs ended?

src/System Application/App/Performance Profiler/src/PerformanceProfiles.Page.al
```diff
@@ -0,0 +1,208 @@
...
+            {
+                ApplicationArea = All;
+                Image = Download;
+                Enabled = Rec."Activity ID" <> '';
+                Caption = 'Download';
+                ToolTip = 'Download the performance profile file.';
+
+                trigger OnAction()
+                var
+                    SampPerfProfilerImplCodeunit: Codeunit "Sampling Perf. Profiler Impl.";
```
grobyns:
```suggestion
                    SamplingPerfProfilerImpl: Codeunit "Sampling Perf. Profiler Impl.";
```

grobyns: no need to add the type to the variable name

src/System Application/App/Performance Profiler/src/ActivityType.Enum.al
nikolakukrika: Small question, is this activity type or session type?

src/System Application/App/Performance Profiler/src/PerformanceProfileHelper.Codeunit.al
```diff
@@ -0,0 +1,55 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+namespace System.Tooling;
+
+using System.Security.User;
+
+codeunit 1934 "Performance Profile Helper"
```
grobyns: names like management, helper, misc should be avoided. These becomes buckets of random stuff. We want single responsibility codeunits and, when you have those, the name should reflect what is in it... Maybe "Activity Type Impl." for the first bit?

src/System Application/App/Performance Profiler/src/PerformanceProfilerInstaller.Codeunit.al
```diff
@@ -0,0 +1,66 @@
...
+    InherentPermissions = X;
+    InherentEntitlements = X;
+    Permissions = tabledata Field = r;
+
+    trigger OnInstallAppPerCompany()
+    begin
+        AddRetentionPolicyAllowedTables(false);
+    end;
+
+    procedure AddRetentionPolicyAllowedTables(ForceUpdate: Boolean)
```
nikolakukrika, line 26: you want to call the same code from the upgrade. The apps are probably not installed in SaaS, I'm not sure if we are raising this trigger.

src/System Application/App/Performance Profiler/src/PerformanceProfileHelper.Codeunit.al
grobyns: just noticed this is called from the Scheduled Perf Profiler Impl. so why not just keep the code there?

src/System Application/App/Performance Profiler/src/PerformanceProfilerInstaller.Codeunit.al
```diff
@@ -0,0 +1,66 @@
...
+
+using System.Environment.Configuration;
+using System.DataAdministration;
+using System.PerformanceProfile;
+using System.Reflection;
+using System.Upgrade;
+
+codeunit 1933 "Performance Profiler Installer"
+{
+    Subtype = Install;
```
nikolakukrika, line 15: Check if the brand new deployment has this setup, the deployment code is tricky. See the comment below regarding upgrade.

src/System Application/App/Performance Profiler/src/PerformanceProfiles.Page.al
```diff
@@ -0,0 +1,208 @@
...
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Tooling;
+
+using System.PerformanceProfile;
+
+/// <summary>
+/// List for performance profiles generated by profile schedules
+/// </summary>
+page 1931 "Performance Profiles"
```
nikolakukrika: Minor - Performance Profile List sounds better

src/System Application/App/Performance Profiler/src/PerformanceProfiles.Page.al
```diff
@@ -0,0 +1,208 @@
...
+            {
+            }
+
+            actionref(Refresh; RefreshPage)
+            {
+            }
+
+            actionref(DownloadProfile; Download)
+            {
+
```
nikolakukrika: Minor, blank line

kalberes (dismissed):

src/System Application/Test/Performance Profiler/src/ScheduledPerfProfilingTest.Codeunit.al
```diff
@@ -0,0 +1,217 @@
...
+
+    local procedure SetupClientType(var PerformanceProfileScheduler: Record "Performance Profile Scheduler"; ClientType: Option; var ActivityType: Enum "Activity Type")
+    begin
+        PerformanceProfileScheduler."Client Type" := ClientType;
+        ScheduledPerfProfiler.MapRecordToActivityType(PerformanceProfileScheduler, ActivityType);
+    end;
+
+    local procedure AddTwoUsers(TempUser: Record User temporary)
+    var
+        i: Integer;
```
nikolakukrika: Minor - Capital I

nikolakukrika (approved):

src/System Application/App/Performance Profiler/src/ActivityType.Enum.al
kalberes: Depends on the context. From a user's perspective it is an Activity. From the server's perspective is hijacking an NST session type.

src/System Application/App/Performance Profiler/src/PerfProfilerScheduleCard.Page.al
kalberes: It could actually be both :). Probably it would be better without specifying any "tense".

nikolakukrika: I find it bad that it is both, it should either be after the fact - we update when it was started or after the start action is invoked.

If we update the expected time after it started, then we should get a new field. 
UI control could be calculated. Alternative is to get a new field, scheduled start time and started time.

See if you want to fix, if not, wontfix this comment.

kalberes: All depends on when teh NST will add/modify a record. It is normally a will if you add ahead of time, or now (you can ask what is now of course). But you can also start in the past. I think we should not overcomplicate things. You want to schedule a session starting at a given datetime and ending at a given datetime. That is the simplest IMO.

Put it in another way. It is a wontfix for now

src/System Application/App/Performance Profiler/src/PerformanceProfileHelper.Codeunit.al
kalberes: @grobyns It is called multiple places not only Scheduled Perf Profiler Impl. And the word "helper" is used a few times in the system app as an object suffix. I do not understand what is the problem with a generic helper **internal** utility object (emphasis on **internal**)? I get the idea of purism of single responsibility code units. But than we could add quite a few. But I am also thinking on how the NST works and maybe the impact on the metadata sub-system when creating an abundance of objects.

But that said we will do it

src/System Application/App/Performance Profiler/src/ActivityType.Enum.al
kalberes: I think the latter makes sense.

src/System Application/App/Performance Profiler/src/PerformanceProfiles.Page.al
kalberes: Agree

src/System Application/App/Performance Profiler/src/PerformanceProfileHelper.Codeunit.al
BazookaMusic: Moved the code and renamed the codeunit

src/System Application/App/Performance Profiler/src/PerformanceProfiles.Page.al
BazookaMusic: Fixed

kalberes (approved):

src/System Application/App/Performance Profiler/src/PerfProfActivityMapper.Codeunit.al
```diff
@@ -0,0 +1,37 @@
...
+codeunit 1934 "Perf. Prof. Activity Mapper"
+{
+    Access = Internal;
+    InherentEntitlements = X;
+    InherentPermissions = X;
+
+
+    procedure MapActivityTypeToClientType(var ClientTpe: Option ,,"Web Service",,,Background,,"Web Client",,,,; ActivityType: Enum "Perf. Profile Activity Type")
+    begin
+        if (ActivityType = ActivityType::WebClient) then
```
StefanSosic: Can you refactor this to case condition, much cleaner

src/System Application/App/Performance Profiler/src/PerfProfActivityMapper.Codeunit.al
```diff
@@ -0,0 +1,37 @@
...
+            if (ActivityType = ActivityType::Background) then
+                ClientTpe := ClientTpe::Background
+            else
+                if (ActivityType = ActivityType::WebAPIClient) then
+                    ClientTpe := ClientTpe::"Web Service";
+    end;
+
+    procedure MapClientTypeToActivityType(ClientTpe: Option ,,"Web Service",,,Background,,"Web Client",,,,; var ActivityType: Enum "Perf. Profile Activity Type")
+    begin
+        if (ClientTpe = ClientTpe::Background) then
```
StefanSosic: Can you refactor this to case condition, much cleaner

src/System Application/App/Performance Profiler/src/ScheduledPerfProfilerImpl.Codeunit.al
```diff
@@ -0,0 +1,178 @@
...
+
+    procedure ValidateThreshold(var PerformanceProfileScheduler: Record "Performance Profile Scheduler")
+    begin
+        if (PerformanceProfileScheduler."Profile Creation Threshold" <= 0) then
+            PerformanceProfileScheduler.Validate("Profile Creation Threshold", 500);
+    end;
+
+    local procedure Intersects(First: record "Performance Profile Scheduler"; Second: Record "Performance Profile Scheduler"): Boolean
+    var
+        startInterval1: DateTime;
```
StefanSosic: Upper case for first char of variables

src/System Application/Test/Performance Profiler/src/ScheduledPerfProfilingTest.Codeunit.al
```diff
@@ -0,0 +1,217 @@
...
+        Any: Codeunit Any;
+        Assert: Codeunit "Library Assert";
+        ScheduledPerfProfiler: Codeunit "Scheduled Perf. Profiler";
+        ProfileStartingDateLessThenEndingDateErr: Label 'The performance profile starting date must be set before the ending date.';
+        ProfileHasAlreadyBeenScheduledErr: Label 'Only one performance profile session can be scheduled for a given activity type for a given user for a given period.';
+
+    [Test]
+    procedure TestInitializedData()
+    var
+        TempPerformanceProfileScheduler: Record "Performance Profile Scheduler" temporary;
```
StefanSosic, line 31: // [SCENARIO] missing for all tests

src/System Application/Test/Performance Profiler/src/ScheduledPerfProfilingTest.Codeunit.al
```diff
@@ -0,0 +1,217 @@
...
+    end;
+
+    [Test]
+    procedure TestMapRecordToActivityType()
+    var
+        TempPerformanceProfileScheduler: Record "Performance Profile Scheduler" temporary;
+        ExpectedActivityTypeMsg: Label 'Expected %1 actvity type. Actual type %2:', Locked = true;
+        ActivityType: Enum "Activity Type";
+    begin
+        // [WHEN] a web client session type is used
```
StefanSosic: In all tests you start with [WHEN] where it makes more sense to [GIVEN].

Please go to: https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-extension-advanced-example-test

Read more regards SCENARIO Tag and GIVEN-WHEN-THEN Tags

image

Structure of all tests should be adjusted

src/System Application/Test/Performance Profiler/src/ScheduledPerfProfilingTest.Codeunit.al
```diff
@@ -0,0 +1,217 @@
...
+    procedure TestUserFilter()
+    var
+        TempPerformanceProfileScheduler: Record "Performance Profile Scheduler" temporary;
+        TempUser: Record User temporary;
+        ActivityType: Enum "Activity Type";
+    begin
+        // [WHEN] we have a user that is just a default user 
+        this.AddTwoUsers(TempUser);
+        if not TempUser.FindSet() then
+            exit;
```
StefanSosic: This means you try to add AddTwoUsers users, and than if not TempUser.FindSet(), test will be still success, even though it didn't created users and it exited early

StefanSosic: I would use only FindSet without if condition, so it throws an error if anything. If kind of un-necessary. Since in AddTwoUsers if it doesn't insert it will error, so here it must have Records, if not defenetly you want test to fail.

StefanSosic (changes requested):

src/System Application/App/Performance Profiler/src/PerfProfActivityMapper.Codeunit.al
kalberes: Why is it cleaner? It is a very simple if.

src/System Application/Test/Performance Profiler/src/ScheduledPerfProfilingTest.Codeunit.al
nikolakukrika: Lets not use if, we should fail the test if the temp user does not exist.

nikolakukrika (approved):

src/System Application/App/Performance Profiler/src/PerfProfActivityMapper.Codeunit.al
StefanSosic: Its nasted if, case provides better structure of code in this case

src/System Application/App/Performance Profiler/src/PerformanceProfilerInstaller.Codeunit.al
BazookaMusic: We are following the existing pattern, like in EmailInstaller.Codeunit.al. This is the system application, so it will be installed in SaaS because base app depends on it.

src/System Application/App/Performance Profiler/src/PerformanceProfilerInstaller.Codeunit.al
BazookaMusic: See answer above, we are using an existing pattern.

src/System Application/App/Performance Profiler/src/PerfProfActivityMapper.Codeunit.al
BazookaMusic: Done

src/System Application/App/Performance Profiler/src/PerfProfActivityMapper.Codeunit.al
BazookaMusic: Done

src/System Application/Test/Performance Profiler/src/ScheduledPerfProfilingTest.Codeunit.al
BazookaMusic: Fixed

src/System Application/Test/Performance Profiler/src/ScheduledPerfProfilingTest.Codeunit.al
BazookaMusic: Added a scenario for all

src/System Application/App/Performance Profiler/src/ScheduledPerfProfilerImpl.Codeunit.al
BazookaMusic: Fixed

src/System Application/Test/Performance Profiler/src/ScheduledPerfProfilingTest.Codeunit.al
BazookaMusic: Fixed

src/System Application/Test/Performance Profiler/src/ScheduledPerfProfilingTest.Codeunit.al
StefanSosic: @BazookaMusic After the most recent commit, still adjustments needed. It's better that there are GIVEN tags now, but WHEN is missing. Also, be aware that it could be multiple GIVEN tags.

kalberes: @StefanSosic 
Code can also be read and understood. And it is always better then any intent written in English.
Spoken languages cannot abstract things in a way that code does. Different semantics. That said we will add back the WHEN-s.

StefanSosic: @kalberes Answer is wrong in many ways. I don't agree with this statement at all. 

**First**, the initial point is to structure it by tags, not to put all in [WHEN], and that's it, it's more correct than not to use any tags. Also, using [WHEN] where it was [GIVEN], it's just the wrong tag, if you don't care about tags better not use it, than just use it on wrong way. 
**Second**, a rough description is fine. You can read more about it on https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-extension-advanced-example-test 
Or take a look at other tests which are existing though app. It should not explain what code does, but explain what test does, what is given, what are you doing and what are you verifying.
**Third**, if all tests follow this guideline why wouldn't this one? Consistency. 
**Fourth**, it would be the same as saying you don't need documentation for anything, just look up in code, and yet Docs exist.

It's excuse for not wanting to follow simple guideline, which is consistent on all tests. Is it something that takes too much time? It can take 5 mins to adjust this test Codeunit. I am more for approach if you do it, than do it as it should be done, on right way. Don't throws just random tags in your test procedures and than give excuses for that.

StefanSosic: @JesperSchulz @nikolakukrika 
Could you guys take a look at this test structure? Do we just throw any tags around in test procedures?

kalberes (dismissed):

kalberes (dismissed):

src/System Application/Test/Performance Profiler/src/ScheduledPerfProfilingTest.Codeunit.al
BazookaMusic: The tags were fixed yesterday before the discussion, so there's no longer "GIVEN" tags in place of [WHEN]. On the one hand, this is not the place to change this convention because all other tests follow it. 

However, to Kalman's point, this convention is not ideal for maintainability. Comments are not in the type system and tend to get outdated as code ages. They can even be misleading as code is moved around.

Ideally we could have a unit testing library that expresses this flow, however the lack of anonymous functions probably makes this hard at the moment.

src/System Application/Test/Performance Profiler/src/ScheduledPerfProfilingTest.Codeunit.al
BazookaMusic: @StefanSosic Btw kudos for this catch and thanks for the review, the test was not correct.

src/System Application/App/Performance Profiler/permissions/PerformanceProfilerObjects.PermissionSet.al
```diff
@@ -5,11 +5,19 @@
...
     Assignable = false;
 
     Permissions = codeunit "Sampling Performance Profiler" = X,
-                  page "Performance Profiler" = X;
+                  codeunit "Scheduled Perf. Profiler" = X,
+                  page "Performance Profiler" = X,
+                  page "Perf. Profiler Schedules List" = X,
+                  page "Perf. Profiler Schedule Card" = X,
+                  page "Performance Profile List" = X,
+                  tabledata "Performance Profile Scheduler" = Rimd,
```
darjoo: The tabledata should be in another permissionset. view and edit. Edit includes view and view includes this.

src/System Application/App/Permissions/SystemTablesBasic.PermissionSet.al
```diff
@@ -65,5 +66,7 @@ permissionset 66 "System Tables - Basic"
                   tabledata "Style Sheet" = RIMD,
                   tabledata "System Object" = Rimd,
                   tabledata "Tenant Profile Page Metadata" = Rimd,
-                  tabledata "Token Cache" = Rimd;
+                  tabledata "Token Cache" = Rimd,
+                  tabledata "Performance Profile Scheduler" = Rimd,
```
darjoo: These would then go away and the new permission set edit should be included instead

src/System Application/App/Performance Profiler/src/PerfProfActivityMapper.Codeunit.al
```diff
@@ -0,0 +1,37 @@
...
+namespace System.Tooling;
+
+codeunit 1934 "Perf. Prof. Activity Mapper"
+{
+    Access = Internal;
+    InherentEntitlements = X;
+    InherentPermissions = X;
+
+
+    procedure MapActivityTypeToClientType(var ClientTpe: Option ,,"Web Service",,,Background,,"Web Client",,,,; ActivityType: Enum "Perf. Profile Activity Type")
```
darjoo: ClientType typo

src/System Application/App/Performance Profiler/src/PerfProfActivityMapper.Codeunit.al
```diff
@@ -0,0 +1,37 @@
...
+            ActivityType::WebClient:
+                ClientTpe := ClientTpe::"Web Client";
+            ActivityType::Background:
+                ClientTpe := ClientTpe::Background;
+            ActivityType::WebAPIClient:
+                ClientTpe := ClientTpe::"Web Service";
+        end;
+    end;
+
+    procedure MapClientTypeToActivityType(ClientTpe: Option ,,"Web Service",,,Background,,"Web Client",,,,; var ActivityType: Enum "Perf. Profile Activity Type")
```
darjoo: ClientType typo

src/System Application/App/Performance Profiler/src/PerfProfActivityMapper.Codeunit.al
```diff
@@ -0,0 +1,37 @@
...
+// ------------------------------------------------------------------------------------------------
+namespace System.Tooling;
+
+codeunit 1934 "Perf. Prof. Activity Mapper"
+{
+    Access = Internal;
+    InherentEntitlements = X;
+    InherentPermissions = X;
+
+
```
darjoo:
```suggestion
```

src/System Application/App/Performance Profiler/src/PerfProfileActivityType.enum.al
```diff
@@ -0,0 +1,38 @@
...
+/// </summary>
+enum 1932 "Perf. Profile Activity Type"
+{
+    Access = Public;
+    Extensible = false;
+
+    /// <summary>
+    /// The web client activity type
+    /// </summary>
+    value(0; WebClient)
```
darjoo: Can we separate these words with a space? "Web Client" and same for the api client "Web API Client"

src/System Application/App/Performance Profiler/src/PerfProfilerSchedulesList.Page.al
```diff
@@ -0,0 +1,140 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Tooling;
+
+using System.PerformanceProfile;
+
+
```
darjoo:
```suggestion
```

src/System Application/App/Performance Profiler/src/ScheduledPerfProfiler.Codeunit.al
```diff
@@ -0,0 +1,127 @@
...
+    /// Maps a session type to an activity type.
+    /// </summary>
+    /// <param name="PerformanceProfileScheduler">The "Performance Profile Scheduler" record </param>
+    ///<returns>The user name if found, else empty</returns>
+    procedure MapRecordToUserName(PerformanceProfileScheduler: Record "Performance Profile Scheduler"): Text
+    begin
+        exit(ScheduledPerfProfilerImpl.MapRecordToUserName(PerformanceProfileScheduler));
+    end;
+
+
```
darjoo:
```suggestion
```

src/System Application/App/Performance Profiler/src/ScheduledPerfProfilerImpl.Codeunit.al
```diff
@@ -0,0 +1,199 @@
...
+        RecordRef.SetTable(PerformanceProfileScheduler);
+    end;
+
+    procedure FilterUsers(var RecordRef: RecordRef; SecurityID: Guid; ForceFilterToUser: Boolean)
+    var
+        UserPermissions: Codeunit "User Permissions";
+        FilterView: Text;
+        FilterTextTxt: Label 'where("User ID"=filter(''%1''))', locked = true;
+
+    begin
```
darjoo, line 56:
```suggestion
    begin
```

BazookaMusic: Done

src/System Application/App/Performance Profiler/src/ScheduledPerfProfiler.Codeunit.al
BazookaMusic: Done

src/System Application/App/Performance Profiler/src/PerfProfilerSchedulesList.Page.al
BazookaMusic: Done

src/System Application/App/Performance Profiler/src/PerfProfileActivityType.enum.al
BazookaMusic: Done

src/System Application/App/Performance Profiler/src/PerfProfActivityMapper.Codeunit.al
BazookaMusic: Done

src/System Application/App/Performance Profiler/src/PerfProfActivityMapper.Codeunit.al
BazookaMusic: Done

src/System Application/App/Permissions/SystemTablesBasic.PermissionSet.al
BazookaMusic: I added the two new permission sets

darjoo (approved):

src/System Application/App/Performance Profiler/src/ScheduledPerfProfiler.Codeunit.al
```diff
@@ -0,0 +1,126 @@
...
+    /// </summary>
+    /// <param name="PerformanceProfileScheduler">The "Performance Profile Scheduler" record </param>
+    ///<returns>The user name if found, else empty</returns>
+    procedure MapRecordToUserName(PerformanceProfileScheduler: Record "Performance Profile Scheduler"): Text
+    begin
+        exit(ScheduledPerfProfilerImpl.MapRecordToUserName(PerformanceProfileScheduler));
+    end;
+
+    /// <summary>
+    /// Initalizes the fields for the "Performance Profile Scheduler" receord
```
JesperSchulz:
```suggestion
    /// Initalizes the fields for the "Performance Profile Scheduler" record
```

src/System Application/App/Performance Profiler/src/ScheduledPerfProfiler.Codeunit.al
```diff
@@ -0,0 +1,126 @@
...
+    /// </summary>
+    /// <param name="PerformanceProfileScheduler">The record</param>
+    /// <param name="SecurityID">The security ID to filter the opening of the page</param>
+    procedure FilterUsers(var PerformanceProfileScheduler: Record "Performance Profile Scheduler"; SecurityID: Guid)
+    begin
+        ScheduledPerfProfilerImpl.FilterUsers(PerformanceProfileScheduler, SecurityID, false);
+    end;
+
+    /// <summary>
+    /// Filters the record with the users that allowed to view the Performance Profile Scheduler page
```
JesperSchulz:
```suggestion
    /// Filters the record with the users that are allowed to view the Performance Profile Scheduler page
```

src/System Application/App/Performance Profiler/src/ScheduledPerfProfiler.Codeunit.al
```diff
@@ -0,0 +1,126 @@
...
+    /// </summary>
+    /// <param name="PerformanceProfileScheduler">The "Performance Profile Scheduler" record </param>
+    /// <param name="ActivityType">The activity enum type</param>
+    procedure MapRecordToActivityType(PerformanceProfileScheduler: Record "Performance Profile Scheduler"; var ActivityType: Enum "Perf. Profile Activity Type")
+    begin
+        ScheduledPerfProfilerImpl.MapRecordToActivityType(PerformanceProfileScheduler, ActivityType);
+    end;
+
+    /// <summary>
+    /// Filters the record with the users that allowed to view the Performance Profile Scheduler page
```
JesperSchulz:
```suggestion
    /// Filters the record with the users that are allowed to view the Performance Profile Scheduler page
```

JesperSchulz (dismissed):

JesperSchulz (approved):

darjoo (approved):

PR #1361, opened 2024-06-18
Manually calculate Field Type in OnAfterGetCurrRecord

jehelles:
Manually calculate Field Type in OnAfterGetCurrRecord, as it might not be calculated if the field is invisible.

Needed for [AB#537292](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/537292)


BardurKnudsen (commented):
Just add 'Rec.' as the automatic comments suggest.

grobyns (approved):

BardurKnudsen (approved):

PR #1369, opened 2024-06-19
Escape special characters in ABS Blob Storage blob filenames

### LINKED ISSUE #1088: [Bug]: ABS Operation fails when blob contains # in file name
### Describe the issue

I'll describe this issue based on the GetBlobAsStream procedure of the ABS Blob Client Codeunit as I haven't tested with other operations yet.

The blob name isn't escaped in the operation so the response returns a 404. In case the ABS Operation Response return parameter isn't used and instead only the InStream is used further in the code, the InStream contains 215 Bytes of data.

GetBlobAsStream calls the SetBlobName(BlobName) procedure of the "ABS Operation Payload" codeunit which in turn just assigns the parameter to its global variable "BlobName". In the procedure PrepareRequestMsg of the "ABS Web Request Helper" Codeunit the Request Uri is set via the ConstructUri procedure of the "ABS Operation Payload" Codeunit, which calls the ConstructUri procedure in the "ABS URI Helper" Codeunit.

ConstructUri in the URI Helper codeunit calls a procedure called "AppendBlobIfNecessary" that does.. just that:
```
if not ConstructedUrl.EndsWith('/') then
            ConstructedUrl += '/';
        ConstructedUrl += BlobName;
```
If the file name of the blob contains a # this won't get escaped which leads to the HttpClient.Send() request not finding the specified blob. I assume in itself not escaping the # is correct, as it's used to jump to anchor points in a website, but I don't see those being applicable in a storage container.

I could provide a fix for this, as the easiest solution in my opinion would be to either call EscapeDataString() from the Uri Codeunit when SetBlobName() is called, or when the Blob name is appended in AppendBlobIfNecessary. This also doesn't hurt when other special characters are present in the file name (tested with !$# characters, when escaping them, the blob is still found, to be tested if that works with other operation types as well).

Appreciate any opinions on how to best fix this.

### Expected behavior

The # in the blob name gets escaped properly when the request uri is constructed.

### Steps to reproduce

- create a storage account
- create a container
- upload a file containing a # in the file name
- use the GetBlobAsStream() procedure in the "ABS Blob Client" to read this blob into an InStream
- look at the error message in the "ABS Operation Response" Codeunit, which should show a 404 in this case

code snippet to reproduce:
```
action(ABSOperation)
{
    Caption = 'ABS Operation';
    ToolTip = 'Tries to get a blob with the special character # in the file name from Azure Blob Storage.';
    ApplicationArea = All;
    Image = TestFile;
    trigger OnAction()
    var
        ABSClient: Codeunit "ABS Blob Client";
        ABSStorageAuth: Codeunit "Storage Service Authorization";
        ABSOperationResponse: Codeunit "ABS Operation Response";
        IABSServiceAuth: Interface "Storage Service Authorization";
        InStr: InStream;
    begin
        IABSServiceAuth := ABSStorageAuth.CreateSharedKey('<insertSharedKey>');
        ABSClient.Initialize('<insertStorageAccountName>', '<insertContainerName>', IABSServiceAuth);
        ABSOperationResponse := ABSClient.GetBlobAsStream('<insertFile#Name.pdf>', InStr);
        if not ABSOperationResponse.IsSuccessful() then
            Message(ABSOperationResponse.GetError());
    end;
}
```

### Additional context

_No response_

ptrk-tr:
#### Summary 
Escapes special characters in blob file names in the ABS Blob Storage module so that operations on these files work as intended.
#### Work Item(s) 
Fixes #1088

Fixes [AB#539027](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/539027)

### DIFF
--- a/src/System Application/App/Azure Blob Services API/src/ABSOperationPayload.Codeunit.al
+++ b/src/System Application/App/Azure Blob Services API/src/ABSOperationPayload.Codeunit.al
@@ -6,6 +6,7 @@
 namespace System.Azure.Storage;
 
 using System;
+using System.Utilities;
 
 codeunit 9042 "ABS Operation Payload"
 {
@@ -98,8 +99,10 @@ codeunit 9042 "ABS Operation Payload"
 
     [NonDebuggable]
     procedure SetBlobName("Blob": Text);
+    var
+        Uri: Codeunit Uri;
     begin
-        BlobName := "Blob";
+        BlobName := Uri.EscapeDataString("Blob");
     end;
 
     [NonDebuggable]


darjoo (approved):

rvanbekkum:
This change turned out to be a breaking change. The behaviour of the ABS Blob Client changed, now altering the input paths.
We already had an `Uri.EscapeDataString` call in our solution, so due to this change the ABS Blob Client module ended up double encoding everything.

Could there have been anyway we could have detected this? Even unit tests wouldn't be able to discover this.

I think a warning (like the ones we get for obsoletions) would have worked for us to detect this breaking change in behaviour. An obsoletion would not be appropriate though (just like changing `TableType` of a table object to `Temporary` is also not really an obsoletion, but the obsoletion properties are used nonetheless).

@JesperSchulz Is it something to consider for the AL language to support warnings that are not obsoletions? Or do you have any other ideas on how this could be approached?
Goal is to get informed/warned about possible breaking changes in behaviour that could affect us.

PR #1442, opened 2024-07-01
Create format validation for phone nos.

### LINKED ISSUE #1493: System Application part of BusinessCentralApps #767: Implement phone no. validation to 'Alternative Address' and 'Contact Alt. Address'
### Describe the issue

See https://github.com/microsoft/BusinessCentralApps/issues/767.

DonHendriksen:
#### Summary 
Provides methods to validate that entered data is in the correct format, such as phone numbers.

#### Work Item(s) 
Fixes #1493


pri-kise:
Some comments on this small PR.
1. I kind of like the idea to have a format validation codeunit in the System Application. Since there can be other data type that we want to check in the future I'd suggest to rename the codeunit to a more generic name. (e.g. `Codeunit Format Validation` or `Codeunit Check Format`) 
2. Although this is codeunit does not contain much code you should follow the facade pattern.
    Every procedure in the public codeunit should contain a documentation header.
3. Can you maybe add a 2nd public procedure that accepts an PhoneNo Text as Input and simple returns a boolean when the format is valid
4. You cannot depend on BaseApp Objects on tests for the system application
    If you want to test this logic in the System Application then you maybe need to create a new Test Table that contains a Phone No Text field for this purpose.

DonHendriksen:
@pri-kise, I processed the first 3 remarks. Implemented the facade pattern. Please have a first look.

Some questions:
1. I need a work item. I cannot link it to the existing one #767. Do we need to create a new issue for it?
2. I created a new app for Format Validation. Can you provide me, for production and test app, the app id and the object number range? If possible check dependencies and brief/description. Any suggestions are welcome.
3. What do you think of the public procedures in the facade codeunit?
4. For testing purposes I can create a new app in the Test Library. A simple FormatValidation table which currently only contains a phone no. field. And this table can be used to test the facade codeunit. Is this approach okay?

src/System Application/App/Format Validation/src/FormatValidationImpl.Codeunit.al
```diff
@@ -0,0 +1,44 @@
...
+
+        exit(false);
+    end;
+
+    procedure ThrowErrorIfPhoneNoContainsCharacters(PhoneNo: Text[30]; SourceRecordId: RecordId; SourceFieldId: Integer)
+    var
+        RecRef: RecordRef;
+        FieldRef: FieldRef;
+    begin
+        if not ValidateIfPhoneNoContainsCharacters(PhoneNo: Text[30];
```
StefanSosic: this is weird formatting here

DonHendriksen: Excuse me, how that happened I don't know. But is resolved. Please also have a look at my questions.

src/System Application/App/Format Validation/src/FormatValidationImpl.Codeunit.al
```diff
@@ -0,0 +1,44 @@
...
+    end;
+
+    procedure ThrowErrorIfPhoneNoContainsCharacters(PhoneNo: Text[30]; SourceRecordId: RecordId; SourceFieldId: Integer)
+    var
+        SourceRecordId: RecordId;
+        RecRef: RecordRef;
+        FieldRef: FieldRef;
+        SourceFieldId: Integer;
+    begin
+        if not ValidateIfPhoneNoContainsCharacters(PhoneNo: Text[30]) then
```
pri-kise: This won't compile.
It should be:
```suggestion
        if not ValidateIfPhoneNoContainsCharacters(PhoneNo) then
```

DonHendriksen: Solved. I did not test and did not compile yet because there are to many open links. I want to develop close to final before I want to take the next steps...

pri-kise:
Regaring your questions:

>I need a work item. I cannot link it to the existing one https://github.com/microsoft/BCApps/pull/767. Do we need to create a new issue for it?

It would be the easiest if you you create an issue on this repository. @JesperSchulz  can then approve it.

> I created a new app for Format Validation. Can you provide me, for production and test app, the app id and the object number range? If possible check dependencies and brief/description. Any suggestions are welcome.

For the appId. -> Simple take a new Guid
Number Ranges are always a littel bit tricky. Searching fro free Ids isn't that easy.. Maybe @JesperSchulz  can help out on this.

> What do you think of the public procedures in the facade codeunit?

lgtm

>For testing purposes I can create a new app in the Test Library. A simple FormatValidation table which currently only contains a phone no. field. And this table can be used to test the facade codeunit. Is this approach okay?

I think this would be the correct approach.
New Test Library with a simple table.
I think this is a valid approach.  Similiar was done for the translation module. https://github.com/microsoft/BCApps/blob/main/src/System%20Application/Test%20Library/Translation/src/TranslationTestTable.Table.al
I would suggest to use an Integer as primary key (similiar to the translation module)

src/System Application/Test/Format Validation/src/FormatValidationTests.Codeunit.al
```diff
@@ -0,0 +1,94 @@
...
+    begin
+        // [SCENARIO] Expect no error message when validating a phone on. which does not contain characters.
+        Initialize();
+        CreateFormatValidationTestData(LibraryUtility.GenerateRandomNumericText(MaxStrLen(FormatValidationTestTable."Phone No.")), FormatValidationTestTable);
+
+        FormatValidation.ThrowErrorIfPhoneNoContainsCharacters(FormatValidationTestTable."Phone No.", Database::"Format Validation Test Table", FormatValidationTestTable.FieldNo("Phone No."));
+    end;
+
+    [Test]
+    procedure CreateFormatValidationDataWithInvalidPhoneNo_ValidatePhoneNo_ExpectNoError()
```
pri-kise: A small Copy Past Issue?
```suggestion
    procedure CreateFormatValidationDataWithInvalidPhoneNo_ValidatePhoneNo_ExpectError()
```

pri-kise:
In general this PR is now in a good shape for review by someone of Microsoft. 👍

pri-kise (approved):

DonHendriksen:
@pri-kise @StefanSosic What are the next steps? I want to keep this PR rolling.

I still need object nos before I can build and publish the changes. I currently cannot test the changes in this PR. And if this one is approved I still have the PR in the BusinessCentralApps repo pending, to call these new validations from Contact Alt. Address.

src/System Application/App/Format Validation/src/FormatValidation.Codeunit.al
```diff
@@ -0,0 +1,38 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Utilities;
+
+/// <summary>
+/// Codeunit that exposes format validation functionality.
+/// </summary>
+codeunit 5070 "Format Validation"
```
JesperSchulz: Please use ID 348.

src/System Application/App/Format Validation/src/FormatValidationImpl.Codeunit.al
```diff
@@ -0,0 +1,41 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Utilities;
+
+using System;
+
+codeunit 5071 "Format Validation Impl."
```
JesperSchulz: Please use ID 349.

src/System Application/Test Library/FormatValidation/src/FormatValidationTestTable.table.al
```diff
@@ -0,0 +1,33 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+// namespace ?
+
+table 130443 "Format Validation Test Table"
```
JesperSchulz: Please use ID 130449.

src/System Application/Test/Format Validation/src/FormatValidationTests.Codeunit.al
```diff
@@ -0,0 +1,94 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Test.Utilities;
+
+codeunit 139194 "Format Validation Tests"
```
JesperSchulz: Please use ID 130449.

src/System Application/Test Library/FormatValidation/src/FormatValidationTestTable.table.al
```diff
@@ -0,0 +1,33 @@
...
+{
+    Caption = 'Format Validation Test Table';
+    DataClassification = SystemMetadata;
+
+    fields
```
JesperSchulz, line 14: Is it really necessary to introduce this table just to test the validation of a phone no? I'd argue it's not.

JesperSchulz (changes requested):
Let's address the ID question first. Generally the design guidelines are now followed and this is a proper module, but I'm not sure I'm a fan of the two public procedures. I will set aside time on Monday to take a closer look.

JesperSchulz:
I've created an issue for you and have linked everything up alright. After code review on Monday we'll settle on a final design and then we'll get this one in before the week is over. I hope that works for you!

src/System Application/Test/Format Validation/src/FormatValidationTests.Codeunit.al
DonHendriksen: This object number is also assigned to the test table.

src/System Application/Test Library/FormatValidation/src/FormatValidationTestTable.table.al
DonHendriksen: It is not necessary. In the unit test codeunit it is also possible to assign a value without using a table. Please discuss it internally because this was the suggestion from your colleagues.

JesperSchulz: I think the suggestion came from @pri-kise, who is one of our most active GitHub contributors and a valued MVP, but he is not a colleague :-) Be that as it may, I value everyone's input on equal terms. Let me think about this for a minute.

src/System Application/App/Format Validation/app.json
```diff
@@ -0,0 +1,39 @@
...
+            "name": "Format Validation Test",
+            "publisher": "Microsoft"
+        }
+    ],
+    "screenshots": [
+    ],
+    "platform": "25.0.0.0",
+    "idRanges": [
+        {
+            "from": 710,
```
StefanSosic: Object IDs used now are: 348 and 349
Update idRanges

src/System Application/Test Library/FormatValidation/app.json
```diff
@@ -0,0 +1,31 @@
...
+                            "id": "fd3d0b8e-61fa-4f87-b2fc-c4a1f3e53a63",
+                            "name": "Format Validation",
+                            "publisher": "Microsoft",
+                            "version": "25.0.0.0"
+                        }
+    ],
+    "screenshots": [
+    ],
+    "platform": "25.0.0.0",
+    "idRanges": [
```
StefanSosic, line 24: update idRanges here also
using 130449

src/System Application/Test/Format Validation/app.json
```diff
@@ -0,0 +1,44 @@
...
+                            "name":  "Library Assert",
+                            "publisher":  "Microsoft",
+                            "version":  "25.0.0.0"
+                        }
+                     ],
+    "screenshots":  [
+
+                    ],
+    "platform":  "25.0.0.0",
+    "idRanges":  [
```
StefanSosic, line 37: update idRanges

src/System Application/App/Format Validation/src/FormatValidation.Codeunit.al
```diff
@@ -0,0 +1,38 @@
...
+/// </summary>
+codeunit 348 "Format Validation"
+{
+    Access = Public;
+
+    /// <summary>
+    /// Validates if a phone no. contains characters.
+    /// </summary>
+    /// <param name="PhoneNo">The phone no. to validate</param>
+    procedure ValidateIfPhoneNoContainsCharacters(PhoneNo: Text[30]): Boolean
```
nikolakukrika, line 19: This not a good design. We cannot add methods like this in a Format Validation codeunit. It needs to be more generic as it should handle more than Phone number and it should contain more functionality. If we want a module, it should validate e.g. BankAccount numbers (IBAN, US format...), Emails, home address... We need at least 3 examples and a generic interface.

src/System Application/App/Format Validation/src/FormatValidation.Codeunit.al
```diff
@@ -0,0 +1,38 @@
...
+        exit(FormatValidationImpl.ValidateIfPhoneNoContainsCharacters(PhoneNo));
+    end;
+
+    /// <summary>
+    /// Validates and throws an error if a phone no. contains characters.
+    /// </summary>
+    /// <param name="PhoneNo">The phone no. to validate</param>
+    /// <param name="SourceRecordId">The error message is a fielderror which needs the source record.</param>
+    /// <param name="SourceFieldId">The error message is a fielderror which needs the source field.</param>
+    procedure ThrowErrorIfPhoneNoContainsCharacters(PhoneNo: Text[30]; SourceRecordId: RecordId; SourceFieldId: Integer)
```
nikolakukrika, line 32: It is unclear when the Validate should be called and when ThrowError should be called. I would expect Validate to handle both cases.

src/System Application/Test Library/FormatValidation/src/FormatValidationTestTable.table.al
StefanSosic: I wouldn't use here a test table just for this and send sql queries.
If we would keep it, I would than think about using it as Temp table. 

Anyway, not really for it, from the tests written it's just un-necessary to fill the table then to test. Its a simple test, wouldn't overcomplicate this.

src/System Application/App/Format Validation/src/FormatValidation.Codeunit.al
StefanSosic: I was here thinking the same, and if it was the point to skip erroring and just have validation, would introduce a parameter to handle it.

DonHendriksen: The error message implemented is a fielderror, which is the standard check. So we need a table so we can unit test the result as fielderror.
It is on request to implement the procedure to check and additionally a procedure which throws the field error. So the dev choices are on purpose.

src/System Application/Test Library/FormatValidation/src/FormatValidationTestTable.table.al
DonHendriksen: Changed it to temp table.

src/System Application/App/Format Validation/src/FormatValidation.Codeunit.al
DonHendriksen: @nikolakukrika, I requested an additional check on field phone no. for table contact alt. address similar to the format validation for contact phone no. I understand that the module should be extended in time, but that's out of scope. The module is created, and it up to everyone who wants to contribute to extend it. My boundaries are reached.

src/System Application/App/Format Validation/app.json
```diff
@@ -0,0 +1,39 @@
...
+            "id": "7e3b999e-1182-45d2-8b82-d5127ddba9b2",
+            "name": "DotNet Aliases",
+            "publisher": "Microsoft",
+            "version": "25.0.0.0"
+        }
+    ],
+    "internalsVisibleTo": [
+        {
+            "id": "a7ab976b-78a1-451b-8dda-313a17b50441",
+            "name": "Format Validation Test",
```
grobyns, line 24: internals should only be made visible to test libraries, if at all

src/System Application/Test/Format Validation/src/FormatValidationTests.Codeunit.al
JesperSchulz: Codeunits and tables can have the same ID.

src/System Application/App/Format Validation/src/FormatValidation.Codeunit.al
JesperSchulz: @DonHendriksen, this small PR has led to rather intense discussions internally. Some love the idea of having a format validation module, others don't. Some think your PR is fine, some strongly disagree.

The biggest push back is on the fact, that your module has too narrow a scope, which might lead to us not getting the design right. It's not that all kinds of format validations must be implemented now, but they should be considered to nail down a future proof design. Sure, we could go for the exact thing you need right here, right now, but then it probably should not be a system application module, as those typically are supposed to have a broader applicability.

Let's be more concrete: In essence, your function validates that a phone no. doesn't contain letters. Why is that specific to a phone no? Why not make a general validation, that a string doesn't contain letters? This might come in handy other places!

But what if someone actually wants to allow to store a phone no. like 1-800-CALLMENOW, which is seen in some countries. How does one now override the behavior? There is no interface where one could change the implementation. There are no events either - which we'd also ideally would like to avoid, now that we're designing a new module. Sure, that also doesn't exist today, but does that mean a new module shouldn't support that?

I was much in favor of a format validation module, but maybe building it around a single format validation with hardly any requirements to other formats doesn't work. It also seems like we're asked to wildly over-engineer, what initially was a simple module.

I'm still between chairs, but I'm tempted to go back on the module and return to your original implementation in the BaseApp - maybe even simplifying that one even more.

pri-kise: I think the worst what could happen is that this PR is cancelled.
It would be better if the critics provide help in making this PR to something really helpful.
If we need an interface for this -> then help in providing a concept for this. If I knew a good example for extensibility or an "Inteface-Handled" pattern then we could uptake this pattern here, but nothing comes to my mind.
What is a generic interface...?

To be honest I never understood why the mail validation mas done in the mail account codeunit.
https://github.com/microsoft/BCApps/blob/main/src/System%20Application/App/Email/src/Account/EmailAccount.Codeunit.al
We could move this a generic Format Validation Module.

DonHendriksen: @JesperSchulz, I appreciate your response and clarification. And I fully agree on a module which is based on a good foundation. But what is the foundation, what are the restrictions?
My 'complain' is that there is no beginning and no end of this story. Since the PR is created we go from left to right and back.
Who can give the final conclusion on this topic. Extending the PR to something really useful, fine. Going back and develop the changes in base app, also fine.
Please let me know...

DonHendriksen: Hi @JesperSchulz, it is quit around here. Can we discuss what will be the final solution?

StefanSosic: @DonHendriksen Jesper was last two weeks on vacation and he will be one more week on vacation, that's why it's quiet. He will probably get back to you upon returning when he catches up all what was happening :)

DonHendriksen: @JesperSchulz, I guess all of us enjoyed a holiday. Now we are all back I really like to proceed with this topic. How can we come to a final conclusion for the phone no. check?

The fact is that there are clear restrictions for Dutch phone nos., but these cannot be applied as a default check because other countries have different restrictions for phone nos. So it should be something country specific.
Or, as you considered, going back to where I started with.

Can you, or can other contributors, draw a final conclusion on this?

Thanks and regards,
Don

JesperSchulz: @DonHendriksen, I agree. Time to reach a conclusion. Let's go with your original solution, and I will work on a format module incl. refactoring when time allows. We need to have some more design discussions before we embark on this adventure. I still think a formatting module like the one you started here would be a very valuable addition to the product, but it requires an initial investment I cannot set aside time for right now. This work won't be lost - we'll simply park it for the time being.

JesperSchulz:
We'll go with the original solution, as we would need more than a single scenario to build a good validation module.
As a spin-off of this PR we've renewed discussions to which extend we should have a type system in BC, where one can specify the type on data in certain fields, which would give certain special actions and validation. E.g. on could in metadata specify, that a field is a phone no, and then number validation would occur, just as a "call now" action would render next to the field. Or addresses would get a "show on map" action and have auto complete capabilities. That would be a much larger effort, but probably the right thing.

For this case, we'll go back to simple validation logic. I'll wrap up the other PR now.

PR #1444, opened 2024-07-01
Scheduled Profiler - Add event subscriber to link help page with schedules + profiling status on the schedules page

BazookaMusic:
#### Summary
This PR adds two features:
1) It adds an event subscriber to link the new help page link with the profiler schedules page
2) It adds a new group to show the scheduled profiling status on the schedules page. This new group contains information about if profiling is active for the current session and shows the active schedule.

#### Work Item(s)
Fixes [AB#501774](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/501774)

### DIFF
--- a/src/System Application/App/Performance Profiler/src/PerfProfilerSchedulesList.Page.al
+++ b/src/System Application/App/Performance Profiler/src/PerfProfilerSchedulesList.Page.al
@@ -28,35 +28,35 @@ page 1933 "Perf. Profiler Schedules List"
         {
             group("Profiling Status")
             {
-                Caption = 'Profiling Status';
                 AboutTitle = 'Profiling Status';
-                AboutText = 'See if profiling is enabled for the current user session.';
+                AboutText = 'Specifies if profiling is enabled for the current user session.';
+                Caption = 'Profiling Status';
 
                 field("Profiling Enabled"; IsProfilingEnabled)
                 {
+                    AboutText = 'Specifies if profiling is enabled for the current user session.';
                     Caption = 'Profiling Enabled';
-                    ToolTip = 'Shows if profiling is enabled for the current user session.';
-                    AboutText = 'Shows if profiling is enabled for the current user session.';
                     Editable = false;
+                    ToolTip = 'Specifies if profiling is enabled for the current user session.';
                 }
 
                 field("Active Schedule ID"; ActiveScheduleId)
                 {
-                    Caption = 'Active Schedule ID';
-                    ToolTip = 'Shows the ID of the active schedule.';
                     AboutText = 'The ID of the active schedule.';
+                    Caption = 'Active Schedule ID';
                     Editable = false;
                     DrillDown = true;
+                    ToolTip = 'Specifies the ID of the active schedule.';
 
                     trigger OnDrillDown()
                     var
                         PerformanceProfileScheduler: Record "Performance Profile Scheduler";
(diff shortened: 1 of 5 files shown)


BazookaMusic:
Needs a platform uptake

src/System Application/App/Performance Profiler/src/PerfProfilerSchedulesList.Page.al
```diff
@@ -26,8 +26,50 @@ page 1933 "Perf. Profiler Schedules List"
...
+            group("Profiling Status")
             {
+                Caption = 'Profiling Status';
+                AboutTitle = 'Profiling Status';
+                AboutText = 'See if profiling is enabled for the current user session.';
+
+                field("Profiling Enabled"; IsProfilingEnabled)
+                {
+                    Caption = 'Profiling Enabled';
+                    ToolTip = 'Shows if profiling is enabled for the current user session.';
```
NKarolak: Tooltips should begin with "Specifies".
https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/user-assistance#guidelines-for-tooltip-text
(check all other new tooltips as well)

src/System Application/App/Performance Profiler/src/PerfProfilerSchedulesList.Page.al
```diff
@@ -26,8 +26,50 @@ page 1933 "Perf. Profiler Schedules List"
...
+                    trigger OnDrillDown()
+                    var
+                        PerformanceProfileScheduler: Record "Performance Profile Scheduler";
+                        ScheduleCardPage: Page "Perf. Profiler Schedule Card";
+                    begin
+                        if (IsNullGuid(ActiveScheduleId)) then
+                            exit;
+
+                        if (not PerformanceProfileScheduler.Get(ActiveScheduleId)) then
+                            exit;
```
NKarolak:
```suggestion
                        if IsNullGuid(ActiveScheduleId) then
                            exit;

                        if not PerformanceProfileScheduler.Get(ActiveScheduleId) then
                            exit;
```

NKarolak (changes requested):
Nit: I prefer to have all properties alphabetically sorted. That way, it's easier to compare what is set and what is not between other fields or objects.

src/System Application/App/Performance Profiler/src/PerfProfilerSchedulesList.Page.al
BazookaMusic: Done

src/System Application/App/Performance Profiler/src/PerfProfilerSchedulesList.Page.al
BazookaMusic: Fixed in all the tooltips I could find

BazookaMusic:
> Nit: I prefer to have all properties alphabetically sorted. That way, it's easier to compare what is set and what is not between other fields or objects.

I see the utility of this but I'd rather sort the properties I'd like to compare rather than spend all the time to sort properties manually for all fields/objects etc.

NKarolak (approved):

onbuyuka (approved):

onbuyuka (approved):

StefanSosic (approved):

PR #1519, opened 2024-07-09
Fix empty No Series code in GetNoSeriesLine

Groenbech96:
#### Summary 

Issue happens when we have a combination of No Series that does not have a No Series Line. 
And also have an No Series Line with empty "Series Code".

When No Series dont have a No Series Line, it is empty initialized. 
Then the code to GetNoSeries will do a findlast that will find that No Series Line with empty code. And then because logic says it found the line, it will try and do a Get later, which then fails.

#### Work Item(s) 
Fixes [AB#540058](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/540058)

### DIFF
--- a/src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
+++ b/src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
@@ -116,8 +116,10 @@ codeunit 305 "No. Series - Setup Impl."
             NoSeriesLine.SetRange("Series Code", NoSeriesRec.Code);
         end;
 
-        if not NoSeriesLine.FindFirst() then
+        if not NoSeriesLine.FindFirst() then begin
             NoSeriesLine.Init();
+            NoSeriesLine."Series Code" := NoSeriesRec.Code;
+        end;
 
         if ResetForDrillDown then begin
             NoSeriesLine.SetRange("Starting Date");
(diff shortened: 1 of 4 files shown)


src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
```diff
@@ -116,8 +116,10 @@ codeunit 305 "No. Series - Setup Impl."
             NoSeriesLine.SetRange("Series Code", NoSeriesRec.Code);
         end;
 
-        if not NoSeriesLine.FindFirst() then
+        if not NoSeriesLine.FindFirst() then begin
             NoSeriesLine.Init();
+            NoSeriesLine."Series Code" := NoSeriesRec.Code;
```
haoranpb, line 121: Can we move the modification out of the procedure? `SetNoSeriesCurrentLineFilters`

Groenbech96: Then we should move the init out.

src/Business Foundation/Test/NoSeries/src/NoSeriesTests.Codeunit.al
```diff
@@ -908,6 +909,54 @@ codeunit 134530 "No. Series Tests"
         LibraryAssert.AreEqual('7', NoSeries.GetNextNo(TempNoSeriesLine, WorkDate()), 'GetNextNo with temporary record returned wrong value');
     end;
 
+    [Test]
+    procedure TestNoSeriesEmptyCodeInLine()
```
grobyns, line 913: it would be good to add the same tests on the old/legacy code to ensure we get the same behaviour as before where possible...

grobyns (dismissed):

src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
Groenbech96: Also there is no modify. We just make sure to init correctly, by setting Pk when init.

grobyns (approved):

t-prda (approved):

PR #1587, opened 2024-07-16
Scheduled Profiler - Fix value update deadlock with different validation strategy + misc fixes

BazookaMusic:
#### Summary
Right now, it's only possible to create or set start/end dates of a schedule for the next week. The reason why is because due to how we made the validation triggers, the start time and end time triggers block each other from ever completing when one gets an incorrect value.

The solution is to keep one validate trigger on the "End Time" and also validate that the dates are correct before saving the values to the database. This way, there's no way the triggers can block each other.

Furthermore, this fixes a few minor issues from internal testing, like field order, descriptions and an incorrect reference for the retention policy.

#### Work Item
Fixes [AB#501774](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/501774)

### DIFF
--- a/src/System Application/App/Performance Profiler/src/PerfProfilerScheduleCard.Page.al
+++ b/src/System Application/App/Performance Profiler/src/PerfProfilerScheduleCard.Page.al
@@ -55,6 +55,11 @@ page 1932 "Perf. Profiler Schedule Card"
                     Caption = 'Start Time';
                     ToolTip = 'Specifies the start time of the schedule.';
                     AboutText = 'The start time of the schedule.';
+
+                    trigger OnValidate()
+                    begin
+                        ScheduledPerfProfiler.ValidatePerformanceProfileSchedulerDatesRelation(Rec);
+                    end;
                 }
                 field("End Time"; Rec."Ending Date-Time")
                 {
@@ -65,7 +70,7 @@ page 1932 "Perf. Profiler Schedule Card"
 
                     trigger OnValidate()
                     begin
-                        ScheduledPerfProfiler.ValidatePerformanceProfileSchedulerDates(Rec, MaxRetentionPeriod);
+                        ScheduledPerfProfiler.ValidatePerformanceProfileSchedulerDatesRelation(Rec);
                         ScheduledPerfProfiler.ValidatePerformanceProfileEndTime(Rec);
                     end;
                 }
--- a/src/System Application/App/Performance Profiler/src/PerfProfilerScheduleCard.Page.al
+++ b/src/System Application/App/Performance Profiler/src/PerfProfilerScheduleCard.Page.al
@@ -203,12 +203,12 @@ page 1932 "Perf. Profiler Schedule Card"
     trigger OnModifyRecord(): Boolean
     begin
         ScheduledPerfProfiler.MapActivityTypeToRecord(Rec, Activity);
-        ValidateRecord();
+        this.ValidateRecord();
     end;
 
     trigger OnInsertRecord(BelowxRec: Boolean): Boolean
     begin
-        ValidateRecord();
+        this.ValidateRecord();
     end;
 
     var
--- a/src/System Application/App/Performance Profiler/src/PerfProfilerScheduleCard.Page.al
+++ b/src/System Application/App/Performance Profiler/src/PerfProfilerScheduleCard.Page.al
@@ -93,6 +93,7 @@ page 1932 "Perf. Profiler Schedule Card"
                     TableRelation = User."User Security ID";
                     Lookup = true;
                 }
+
                 field(Activity; Activity)
                 {
                     ApplicationArea = All;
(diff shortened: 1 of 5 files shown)


StefanSosic (approved):

src/System Application/App/Performance Profiler/src/PerfProfilerScheduleCard.Page.al
```diff
@@ -203,11 +199,17 @@ page 1932 "Perf. Profiler Schedule Card"
     trigger OnModifyRecord(): Boolean
     begin
         ScheduledPerfProfiler.MapActivityTypeToRecord(Rec, Activity);
-        ScheduledPerfProfiler.ValidatePerformanceProfileSchedulerRecord(Rec, Activity);
+        this.ValidateRecord();
```
NKarolak, line 210:
```suggestion
        ValidateRecord();
```
Not sure about your current styleguide, but IMHO `this` should be used either everywhere (e.g., also for all global variable calls) or nowhere within an object.
Same in line 207.

src/System Application/App/Performance Profiler/src/PerfProfilerScheduleCard.Page.al
```diff
@@ -203,11 +199,17 @@ page 1932 "Perf. Profiler Schedule Card"
...
     end;
 
     trigger OnInsertRecord(BelowxRec: Boolean): Boolean
     begin
+        this.ValidateRecord();
+    end;
+
+    local procedure ValidateRecord()
```
NKarolak: Styleguide? Usually local procedures are placed at the bottom of the file, under the variables section.

BazookaMusic: I will move them and I guess this is the pattern in other files, but I couldn't find any reference of this in a styleguide. Do you happen to have a link?

NKarolak: No, just plain observation :-)

src/System Application/App/Performance Profiler/src/PerfProfilerScheduleCard.Page.al
NKarolak: I've learned only now that you plan to convert the whole System App to `this` ([source](https://x.com/KarolakNatalie/status/1813513871241642277)). However, I would expect it to happen separately, in a separate PR across all System App objects.

src/System Application/App/Performance Profiler/src/PerfProfilerScheduleCard.Page.al
```diff
@@ -198,24 +197,32 @@ page 1932 "Perf. Profiler Schedule Card"
...
         ScheduledPerfProfiler.MapActivityTypeToRecord(Rec, Activity);
         RetentionPeriod := ScheduledPerfProfiler.GetRetentionPeriod();
+        ProfileCreationThreshold := Rec."Profile Creation Threshold";
     end;
 
     trigger OnModifyRecord(): Boolean
     begin
         ScheduledPerfProfiler.MapActivityTypeToRecord(Rec, Activity);
-        ScheduledPerfProfiler.ValidatePerformanceProfileSchedulerRecord(Rec, Activity);
+        ValidateRecord();
```
kalberes: this.

src/System Application/App/Performance Profiler/src/PerfProfilerScheduleCard.Page.al
```diff
@@ -144,16 +141,18 @@ page 1932 "Perf. Profiler Schedule Card"
                         end;
                     end;
                 }
-                field("Profile Creation Threshold"; Rec."Profile Creation Threshold")
+                field("Activity Duration Threshold"; ProfileCreationThreshold)
```
kalberes, line 148: "Activity Duration Threshold in ms"

src/System Application/App/Performance Profiler/src/PerfProfilerSchedulesList.Page.al
```diff
@@ -33,21 +33,13 @@ page 1933 "Perf. Profiler Schedules List"
                 AboutText = 'Specifies if profiling is enabled for the current user session.';
                 Caption = 'Profiling Status';
 
-                field("Profiling Enabled"; IsProfilingEnabled)
+                field("Active Schedule ID"; ActiveScheduleDisplayName())
```
kalberes: this.

kalberes (approved):

src/System Application/App/Performance Profiler/src/PerfProfilerScheduleCard.Page.al
BazookaMusic: The caption has the (ms)

onbuyuka (approved):

src/System Application/App/Performance Profiler/src/PerfProfilerScheduleCard.Page.al
BazookaMusic: It's fine to fix in new code I think

kalberes (approved):

src/System Application/App/Performance Profiler/src/PerfProfilerScheduleCard.Page.al
```diff
@@ -97,6 +93,7 @@ page 1932 "Perf. Profiler Schedule Card"
                     TableRelation = User."User Security ID";
                     Lookup = true;
                 }
+
```
NKarolak:
```suggestion
```

NKarolak (changes requested):

kalberes (dismissed):

onbuyuka (dismissed):

darjoo (approved):

onbuyuka (dismissed):

darjoo (approved):

kalberes (approved):

PR #1625, opened 2024-07-23
Make Edit-In Excel use mysettings.language instead of mysettings.regi…

DenLilleMand:
#### Summary 
Make Edit-In Excel use mysettings.language instead of mysettings.region as add-in language

#### Work Item(s) 
Fixes #[AB#538298](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/538298/)

### DIFF
--- a/src/System Application/App/Edit in Excel/src/EditInExcelWorkbookImpl.Codeunit.al
+++ b/src/System Application/App/Edit in Excel/src/EditInExcelWorkbookImpl.Codeunit.al
@@ -163,7 +163,7 @@ codeunit 1489 "Edit in Excel Workbook Impl."
         ConnectionInfo.HostName := HostName;
 
         DataEntityExportInfo.Connection := ConnectionInfo;
-        DataEntityExportInfo.Language := LanguageIDToCultureName(WindowsLanguage);
+        DataEntityExportInfo.Language := LanguageIDToCultureName(UserSettingsLanguageID());
         DataEntityExportInfo.EnableDesign := true;
         DataEntityExportInfo.RefreshOnOpen := true;
         DataEntityExportInfo.DateCreated := CurrentDateTime();
@@ -231,6 +231,35 @@ codeunit 1489 "Edit in Excel Workbook Impl."
         exit(99);
     end;
 
+    local procedure UserSettingsLanguageID(): Integer
+    var
+        DefaultLanguageID: Integer;
+        UserSessionSettings: SessionSettings;
+    begin
+        if TryUserSettingsLanguageID() then begin
+            UserSessionSettings.Init();
+            if UserSessionSettings.LanguageId() <> 0 then
+                exit(UserSessionSettings.LanguageId());
+        end;
+
+        if WindowsLanguage() <> 0 then
+            DefaultLanguageID := WindowsLanguage()
+        else
+            DefaultLanguageID := 1033; // English - United States
+
+        exit(DefaultLanguageID);
+    end;
+
+    [TryFunction]
+    local procedure TryUserSettingsLanguageID()
+    var
+        UserSessionSettings: SessionSettings;
+    begin
+        UserSessionSettings.Init();
+        if UserSessionSettings.LanguageId() = 0 then
+            error('Unable to get the SessionSettings.LanguageId()')
+    end;
+
     local procedure LanguageIDToCultureName(LanguageID: Integer): Text
     var
         CultureInfo: DotNet CultureInfo;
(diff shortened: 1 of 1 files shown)


src/System Application/App/Edit in Excel/src/EditInExcelWorkbookImpl.Codeunit.al
```diff
@@ -231,12 +231,39 @@ codeunit 1489 "Edit in Excel Workbook Impl."
         exit(99);
     end;
 
+    local procedure UserSettingsLanguageID(): Integer
+    var
+        DefaultLanguageID: Integer;
+        UserSessionSettings: SessionSettings;
+    begin
+        if TryUserSettingsLanguageID() then begin
```
StefanMaron: Why using a `TryFunction` here? You just end up with duplicate code.
If the user does not have a default language you will default with english anyway.

src/System Application/App/Edit in Excel/src/EditInExcelWorkbookImpl.Codeunit.al
```diff
@@ -231,12 +231,39 @@ codeunit 1489 "Edit in Excel Workbook Impl."
...
+            UserSessionSettings.Init();
+            exit(UserSessionSettings.LanguageId());
+        end;
+
+        if WindowsLanguage() <> 0 then
+            DefaultLanguageID := WindowsLanguage()
+        else
+            DefaultLanguageID := 1033; // English - United States
```
StefanMaron: Instead of hardcoding english, `GlobalLanguage()` might be better

src/System Application/App/Edit in Excel/src/EditInExcelWorkbookImpl.Codeunit.al
```diff
@@ -231,12 +231,39 @@ codeunit 1489 "Edit in Excel Workbook Impl."
...
+    var
+        DefaultLanguageID: Integer;
+        UserSessionSettings: SessionSettings;
+    begin
+        if TryUserSettingsLanguageID() then begin
+            UserSessionSettings.Init();
+            exit(UserSessionSettings.LanguageId());
+        end;
+
+        if WindowsLanguage() <> 0 then
```
StefanMaron: You can always assign DefaultLanugageID with WindowsLanuguage() and then only overwrite it if it equals `0`.
That way you can avoid to call WindowsLanuguage() twice.

StefanMaron (changes requested):

src/System Application/App/Edit in Excel/src/EditInExcelWorkbookImpl.Codeunit.al
```diff
@@ -231,12 +231,39 @@ codeunit 1489 "Edit in Excel Workbook Impl."
...
+    end;
+
+    [TryFunction]
+    local procedure TryUserSettingsLanguageID()
+    var
+        UserSessionSettings: SessionSettings;
+    begin
+        UserSessionSettings.Init();
+        if UserSessionSettings.LanguageId() = 0 then
+            error('Unable to get the SessionSettings.LanguageId()')
```
StefanSosic: Error starts with upper E

StefanSosic: Should be Label to be translatable

src/System Application/App/Edit in Excel/src/EditInExcelWorkbookImpl.Codeunit.al
StefanSosic: I would agree here + don't agree with structure, as Stefan said its duplicate.
Different approaches, calling procedure with var parameter than exit with LanguageId for exmaple, something like this:
if TryUserSettingsLanguageID(LanguageID) then 
exit(LanguageID);

StefanSosic (changes requested):

src/System Application/App/Edit in Excel/src/EditInExcelWorkbookImpl.Codeunit.al
DenLilleMand: @StefanMaron I've changed the implementation now to just use the `GlobalLanguage()` and removed everything else, thanks for the advice

StefanMaron:
Lol, thats even better

encimita (approved):

StefanMaron (approved):

darjoo (approved):

PR #1660, opened 2024-07-29
[Copilot] Support multiple tool calls

msft-sam:
#### Summary
Currently the Copilot developer tools will only process and invoke the first tool call provided in the tool calls response.
However, as the tool call response is an array, there could be multiple tools returned.

This updates the handling logic of the tool calls to process all the function responses. Additionally, it is now possible to specify the invocation behavior, such that the functions are invoked, not invoked, or invoked and returned to the LLM until no further tool calls are returned.

#### Work Item(s) 
Fixes [AB#542604](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/542604)

### DIFF
--- a/src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
+++ b/src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
@@ -421,7 +421,7 @@ codeunit 7772 "Azure OpenAI Impl"
             ChatMessages.AddToolCalls(CompletionToken.AsArray());
 
             if not ProcessToolCalls(CompletionToken.AsArray(), ChatMessages, AOAIOperationResponse) then begin
-                AOAIFunctionResponse.SetFunctionCallingResponse(true, Enum::"AOAI Function Response Status"::FunctionInvalid, '', '', '', '', '');
+                AOAIFunctionResponse.SetFunctionCallingResponse(true, Enum::"AOAI Function Response Status"::"Function Invalid", '', '', '', '', '');
                 AOAIOperationResponse.AddFunctionResponse(AOAIFunctionResponse);
             end;
 
@@ -430,7 +430,7 @@ codeunit 7772 "Azure OpenAI Impl"
                 if not AOAIFunctionResponse.IsSuccess() then
                     FeatureTelemetry.LogError('0000MTB', CopilotCapabilityImpl.GetAzureOpenAICategory(), StrSubstNo(TelemetryFunctionCallingFailedErr, AOAIFunctionResponse.GetFunctionName()), AOAIFunctionResponse.GetError(), AOAIFunctionResponse.GetErrorCallstack(), Enum::"AL Telemetry Scope"::All, CustomDimensions);
 
-            if ChatMessages.GetToolInvokePreference() in [Enum::"AOAI Tool Invoke Preference"::InvokeToolsOnly, Enum::"AOAI Tool Invoke Preference"::Automatic] then
+            if ChatMessages.GetToolInvokePreference() in [Enum::"AOAI Tool Invoke Preference"::"Invoke Tools Only", Enum::"AOAI Tool Invoke Preference"::Automatic] then
                 AOAIOperationResponse.AppendFunctionResponsesToChatMessages(ChatMessages);
 
             Telemetry.LogMessage('0000MFH', TelemetryChatCompletionToolCallLbl, Verbosity::Normal, DataClassification::OrganizationIdentifiableInformation, Enum::"AL Telemetry Scope"::All, CustomDimensions);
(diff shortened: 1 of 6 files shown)


src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
```diff
@@ -473,15 +488,24 @@ codeunit 7772 "Azure OpenAI Impl"
...
+                    AOAIFunctionResponse.SetFunctionCallingResponse(true, true, AOAIFunction.GetName(), FunctionId, FunctionResult, '', '');
+                    AOAIOperationResponse.AddFunctionResponse(AOAIFunctionResponse);
+                    exit(true);
+                end else begin
+                    AOAIFunctionResponse.SetFunctionCallingResponse(true, false, AOAIFunction.GetName(), FunctionId, FunctionResult, GetLastErrorText(), GetLastErrorCallStack());
+                    AOAIOperationResponse.AddFunctionResponse(AOAIFunctionResponse);
+                    exit(true);
+                end
+            else begin
                 AOAIFunctionResponse.SetFunctionCallingResponse(true, true, AOAIFunction.GetName(), FunctionId, FunctionResult, '', '');
```
msft-sam: In this case the `AOAI Tool Invoke Preference` must be set to `Manual`.
I wonder if it makes more sense for the `NewFunctionCallSuccess` parameter to be set to `false` (second param), although this would cause the `IsSuccess` checks to return false later on.

I can see an argument both ways - it did "successfully" not call the function, as requested, so it shouldn't be treated as an error condition, but also, the function was not invoked - so what should `IsSuccess` return on the `AOAIFunctionResponse`? (I guess what we decide there will help determine what we set here)

@qasimikram do you have any opinions here?

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIToolInvokePreference.Enum.al
```diff
@@ -0,0 +1,40 @@
...
+        Caption = 'Manual', Locked = true;
+    }
+
+    /// <summary>
+    /// Automatically invoke the tool calls, and send them back to the LLM until no more tool calls are returned before returning to the caller.
+    /// </summary>
+    value(2; Automatic)
+    {
+        Caption = 'Automatic', Locked = true;
+    }
```
NKarolak: Why won't the values be translated?
```suggestion
    value(0; InvokeToolsOnly)
    {
        Caption = 'Invoke Tools Only';
    }

    /// <summary>
    /// Require manually invocation of the tool calls (i.e. the Copilot toolkit will not invoke the tools).
    /// Does not append the tool results to the chat history.
    /// </summary>
    value(1; Manual)
    {
        Caption = 'Manual';
    }

    /// <summary>
    /// Automatically invoke the tool calls, and send them back to the LLM until no more tool calls are returned before returning to the caller.
    /// </summary>
    value(2; Automatic)
    {
        Caption = 'Automatic';
    }
```

msft-sam: No strong reason either way, also cleaned up the comments a bit

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
darjoo: I haven't gone through the whole PR, but my thought here would be.
`AOAIOperationResponse.IsSuccess` = `true`
`AOAIOperationResponse.IsFunctionCall` = `true`
`AOAIFunctionResponse.IsInvoked` = `false`
`AOAIFunctionResponse.IsSuccess` = `false`
We should have a distinction between a successful function call and if we invoked the function call. If we try and mix the two distinctions into one, it would be misleading for the majority of developers.

darjoo (dismissed):

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
msft-sam: Will keep the operation response as is (like above), for the function responses I'll add another enum to indicate the state better since `IsSuccess` is kinda undefined until the function has been invoked (and will save us from adding more bools as time goes on).

Will have an enum consisting of these states:
- `NotInvoked` - if it is a Manual operation
- `InvokeSuccess` - if the function was invoked and completed successfully
- `InvokeFail` - if the function was invoked and failed in some way

msft-sam: I also included all the other failure modes we have today, so should be much easier to see what went wrong with the function call

src/System Application/App/AI/src/Azure OpenAI/Operation Response/AOAIFunctionResponseStatus.Enum.al
```diff
@@ -0,0 +1,54 @@
...
+/// </summary>
+enum 7777 "AOAI Function Response Status"
+{
+    Access = Public;
+    Extensible = false;
+
+    /// <summary>
+    /// The function has not been invoked.
+    /// </summary>
+    value(0; NotInvoked)
```
darjoo:
```suggestion
    value(0; "Not Invoked")
```

Add spaces to the names

src/System Application/App/AI/src/Azure OpenAI/Operation Response/AOAIOperationResponse.Codeunit.al
```diff
@@ -14,11 +14,18 @@ codeunit 7770 "AOAI Operation Response"
...
+        LastAOAIFunctionResponse: Codeunit "AOAI Function Response";
+#endif
+        AOAIFunctionResponses: List of [Codeunit "AOAI Function Response"];
         StatusCode: Integer;
         Success: Boolean;
         Result: Text;
         Error: Text;
+        IncorrectRoleErr: Label 'The last chat message must have a role of assistant.';
+        IncorrectToolCallsErr: Label 'The last chat message does not contain any tool calls to respond to.';
+        IncorrectToolCountsErr: Label 'The provided number of tool calls to not match the number of function calls. There may be unsupported tool call types.';
```
qasimikram: Typo 
```suggestion
        IncorrectToolCountsErr: Label 'The provided number of tool calls do not match the number of function calls. There may be unsupported tool call types.';
```

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIChatMessages.Codeunit.al
```diff
@@ -69,6 +69,12 @@ codeunit 7763 "AOAI Chat Messages"
         AOAIChatMessagesImpl.AddAssistantMessage(NewMessage);
     end;
```
qasimikram, line 71: Function summary and param details missing.

qasimikram (dismissed):
Few minor things, apart from that looks good.

qasimikram (approved):

PR #1692, opened 2024-07-31
[Copilot] Set Allow Data Movement as default

DenLilleMand:
#### Summary 

We are changing Copilot features to be on by default even when moving data across geographical boundaries, making it an opt-out instead of opt-in for the regions where OpenAI infrastructure is not hosted.

#### Work Item(s) 
Fixes [AB#537341](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/537341/)


src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
```diff
@@ -135,19 +133,9 @@ codeunit 7772 "Azure OpenAI Impl"
 
                     exit(false);
                 end;
-            else begin
+            else
                 // Privacy notice not set, we will not cross geo-boundries
```
darjoo: Remove comment, this is now invalid

darjoo (approved):
Are the docs being updated? https://learn.microsoft.com/en-us/dynamics365/business-central/enable-ai

Please sync with Brent to get those updated as that will no longer be true in v25 with this change.

EvgenijKorovin (approved):

darjoo (approved):

PR #1693, opened 2024-07-31
Add Source Control details to Extension Settings

blrobl:
#### Work Item(s) 
Fixes [AB#543124](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/543124)

### DIFF
--- a/src/System Application/App/Extension Management/src/ExtensionSettings.Page.al
+++ b/src/System Application/App/Extension Management/src/ExtensionSettings.Page.al
@@ -98,9 +98,9 @@ page 2511 "Extension Settings"
                 }
             }
 
-            group("Source control details")
+            group("Source Control details")
             {
-                Caption = 'The source control details for the extension';
+                Caption = 'Source Control details';
                 Visible = IsSourceInformationAvailable;
 
                 field(RepositoryUrl; RepositoryUrl)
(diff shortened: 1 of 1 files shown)


src/System Application/App/Extension Management/src/ExtensionSettings.Page.al
```diff
@@ -97,6 +97,25 @@ page 2511 "Extension Settings"
                     ToolTip = 'Specifies if the publisher allows a symbol package download will contain the source code and all other resources that were part of the extension package.';
                 }
             }
+
+            group("Source Control details")
+            {
+                Caption = 'Source Control details';
```
StefanSosic: To me looks off with lowercase 'details'
All other good, however you decide

StefanSosic (approved):

src/System Application/App/Extension Management/src/ExtensionSettings.Page.al
NKarolak: I believe that this mix does not comply with the current styleguide:
https://learn.microsoft.com/en-us/style-guide/capitalization

Either `Source control details` or `Source Control Details`; same for the field group name.

krupybalu (approved):

SBalslev (approved):

PR #1723, opened 2024-08-05
Exposing all visible and available record fields of a given page from Page Summary module.

MonicaPAhuja:
#### Summary 
Exposing all visible and available record fields of a given page from Page Summary module.
Exposing tooltip property for the fields

#### Work Item(s) 
Fixes 
[AB#539642](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/539642)


src/System Application/App/Page Summary Provider/src/PageSummaryProviderImpl.Codeunit.al
```diff
@@ -43,7 +42,10 @@ codeunit 2717 "Page Summary Provider Impl."
             AddErrorMessage(ResultJsonObject, InvalidBookmarkErrorCodeTok, InvalidBookmarkErrorMessageTxt);
             exit(Format(ResultJsonObject)); // Bookmark is invalid, so returning the information we actually have about the page
         end;
-        exit(GetFieldsSummary(PageId, RecId, Bookmark, ResultJsonObject));
+
+        GetFieldsSummary(PageId, RecId, Bookmark, ResultJsonObject);
```
msft-sam: Should we check for an error message if the fields summary failed? (Not sure how it would combine with record fields if the summary failed)

src/System Application/App/Page Summary Provider/src/PageSummaryProviderImpl.Codeunit.al
```diff
@@ -117,16 +122,47 @@ codeunit 2717 "Page Summary Provider Impl."
...
-        if not TryGetPageSummaryFields(PageId, RecId, Bookmark, ResultJsonObject) then begin
+        if not TryGetPageSummaryFields(PageId, RecId, Bookmark, ResultJsonObject) then
             AddErrorMessage(ResultJsonObject, FailedGetSummaryFieldsCodeTok, GetLastErrorText());
-            exit(Format(ResultJsonObject));
+    end;
+
+    local procedure GetRecordFields(PageId: Integer; Bookmark: Text; var ResultJsonObject: JsonObject)
+    begin
+        // Get all visible and available table fields that back the controls that are visible on the page
+        if TryGetAvailableRecordFieldsData(PageId, Bookmark, ResultJsonObject) then;
```
encimita: Who do we expect to throw an `Error`? Should we add a `Session.LogMessage`/warning here?

encimita: (or `AddErrorMessage`)

src/System Application/App/Page Summary Provider/src/PageSummaryProvider.Codeunit.al
```diff
@@ -26,12 +26,25 @@ codeunit 2718 "Page Summary Provider"
     ///   "pageType":"Card",
     ///   "summaryType":"Brick",
```
encimita, line 27: It seems like what we are returning now is no longer a summarytype Brick? We are now including way more than just the brick fields.

Not sure what we use the summarytype for at all, but consider whether it makes sense to differentiate the brick case from the all-visible-fields case

encimita (dismissed):

darjoo (dismissed):
Enrico & Sam have good comments, please take a look.

src/System Application/App/Page Summary Provider/src/PageSummaryProviderImpl.Codeunit.al
```diff
@@ -117,16 +122,47 @@ codeunit 2717 "Page Summary Provider Impl."
...
+    end;
+
+    [TryFunction]
+    local procedure TryGetAvailableRecordFieldsData(PageId: Integer; Bookmark: Text; var ResultJsonObject: JsonObject)
+    var
+        GenericList: DotNet GenericList1;
```
DenLilleMand, line 158: Instead of `GenericList` what about `AvailableTableFields` or similar?
If we agree that `GenericList` should have a name that shows what it contains instead of its type, then there are a lot of `GenericList` variables in this module that should be renamed.

src/System Application/App/Page Summary Provider/src/PageSummaryProvider.Codeunit.al
MonicaPAhuja: As discussed, all visible fields belong in separate array

MonicaPAhuja: summary type brick is still correct as the summarised fields contain the brick. And We don't read that summaryType anywhere anyway. TIS just looks at the page type to decide what to do

src/System Application/App/Page Summary Provider/src/PageSummaryProviderImpl.Codeunit.al
MonicaPAhuja: Fixed

src/System Application/App/Page Summary Provider/src/PageSummaryProviderImpl.Codeunit.al
MonicaPAhuja: Fixed

src/System Application/App/Page Summary Provider/src/PageSummaryProviderImpl.Codeunit.al
```diff
@@ -116,17 +133,50 @@ codeunit 2717 "Page Summary Provider Impl."
...
+        if not TryGetPageSummaryFields(PageId, RecId, Bookmark, ResultJsonObject) then
             AddErrorMessage(ResultJsonObject, FailedGetSummaryFieldsCodeTok, GetLastErrorText());
-            exit(Format(ResultJsonObject));
+    end;
+
+    local procedure GetRecordFields(PageId: Integer; Bookmark: Text; var ResultJsonObject: JsonObject)
+    begin
+        // Get all visible and available table fields that back the controls that are visible on the page
+        if TryGetAvailableRecordFieldsData(PageId, Bookmark, ResultJsonObject) then
+            Session.LogMessage('', StrSubstNo(GetRecordFieldsFailureTelemetryTxt, PageId), Verbosity::Error, DataClassification::SystemMetadata, TelemetryScope::ExtensionPublisher, 'Category', PageSummaryCategoryLbl);
```
encimita: tag

src/System Application/App/Page Summary Provider/src/PageSummaryProviderImpl.Codeunit.al
```diff
@@ -106,7 +113,17 @@ codeunit 2717 "Page Summary Provider Impl."
...
 
-    procedure GetFieldsSummary(PageId: Integer; RecId: RecordId; Bookmark: Text; var ResultJsonObject: JsonObject): Text
+    local procedure AddFields(PageId: Integer; RecId: RecordId; Bookmark: Text; var ResultJsonObject: JsonObject)
+    begin
+        // Add Summary fields
+        GetFieldsSummary(PageId, RecId, Bookmark, ResultJsonObject);
+
+        // Add Record fields
+        if not ResultJsonObject.Contains('error') then
+            GetRecordFields(PageId, Bookmark, ResultJsonObject);
```
DenLilleMand, line 125: To me it looks like `GetFieldsSummary` and `GetRecordFields` do the same thing more or less, the only difference being that one of them is `pfields` and another is `tfields`,
Wouldn't it be more suitable to call it  `GetRecordFieldsSummary` then ?

DenLilleMand (approved):

src/System Application/App/Page Summary Provider/src/PageSummaryProviderImpl.Codeunit.al
```diff
@@ -43,7 +42,11 @@ codeunit 2717 "Page Summary Provider Impl."
             AddErrorMessage(ResultJsonObject, InvalidBookmarkErrorCodeTok, InvalidBookmarkErrorMessageTxt);
             exit(Format(ResultJsonObject)); // Bookmark is invalid, so returning the information we actually have about the page
         end;
-        exit(GetFieldsSummary(PageId, RecId, Bookmark, ResultJsonObject));
+
+        // Add header
```
encimita:
```suggestion
        // Add summary fields and record fields
```

src/System Application/App/Page Summary Provider/src/PageSummaryProviderImpl.Codeunit.al
```diff
@@ -78,7 +81,11 @@ codeunit 2717 "Page Summary Provider Impl."
         Bookmark := Format(RecId, 0, 10); // 10 = Format RecordId into string
         if Bookmark = '' then
             exit(Format(ResultJsonObject));
-        exit(GetFieldsSummary(PageId, RecId, Bookmark, ResultJsonObject));
+
+        // Add header
```
encimita:
```suggestion
        // Add summary fields and record fields
```

src/System Application/App/Page Summary Provider/src/PageSummaryProviderImpl.Codeunit.al
msft-sam: This comment doesn't seem correct?

src/System Application/App/Page Summary Provider/src/PageSummaryProviderImpl.Codeunit.al
msft-sam: and here

src/System Application/App/Page Summary Provider/src/PageSummaryProviderImpl.Codeunit.al
msft-sam: Fields summary is a "summary" of the fields (i.e. it could be the brick definition, or some custom AL could provide its own summary), record fields are the backing record fields that are visible on the page - think the naming here is good (look at the example response on the `PageSummaryProvider` codeunit to get an idea of what the resulting JSON looks like)

src/System Application/App/Page Summary Provider/src/PageSummaryProviderImpl.Codeunit.al
msft-sam: The usual convention is to name it the dotnet type, imo it would be better if the dotnet object would return an AL list - maybe one day in the future 😄

MonicaPAhuja: The usual convention is to name it the dotnet type, so will keep it like that

encimita (approved):

darjoo (approved):

PR #1743, opened 2024-08-06
Allow using latest BCArtifact instead of hardcoded version

aholstrup1:
#### Summary 

* Allow the "artifact" setting to use latest available version instead of a hardcoded version 
* Run stability jobs on all official branches 

#### Work Item(s) 
Fixes [AB#543528](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/543528)

### DIFF
--- a/.github/actions/RunAutomation/SubmitStabilityJob/run.ps1
+++ b/.github/actions/RunAutomation/SubmitStabilityJob/run.ps1
@@ -6,14 +6,14 @@ param (
 $repository = $runParameters.Repository
 $targetBranch = $runParameters.TargetBranch
 $workflowName = " CI/CD"
-$workflowRunTime = Get-Date
+$workflowRunTime = Get-Date -AsUTC
 
 Write-Host "Running the workflow '$workflowName' on branch $targetBranch"
 gh workflow run --repo $repository --ref $targetBranch $workflowName
 
 # Get the workflow run URL to display in the message
 
-while((Get-Date) -lt $workflowRunTime.AddMinutes(1)) {
+while((Get-Date -AsUTC) -lt $workflowRunTime.AddMinutes(1)) {
     Start-Sleep -Seconds 5 # wait for 5 seconds for the workflow to start
     $workflowRun = gh run list --branch $targetBranch --event workflow_dispatch --workflow $workflowName --repo $repository --json createdAt,url --limit 1 | ConvertFrom-Json
 
--- a/.github/actions/RunAutomation/SubmitStabilityJob/run.ps1
+++ b/.github/actions/RunAutomation/SubmitStabilityJob/run.ps1
@@ -5,7 +5,7 @@ param (
 
 $repository = $runParameters.Repository
 $targetBranch = $runParameters.TargetBranch
-$workflowName = " Update AL-Go System Files"
+$workflowName = " CI/CD"
 $workflowRunTime = Get-Date
 
 Write-Host "Running the workflow '$workflowName' on branch $targetBranch"
--- a/.github/actions/RunAutomation/UpdateALGo/run.ps1
+++ b/.github/actions/RunAutomation/UpdateALGo/run.ps1
@@ -6,14 +6,14 @@ param (
 $repository = $runParameters.Repository
 $targetBranch = $runParameters.TargetBranch
 $workflowName = " Update AL-Go System Files"
-$workflowRunTime = Get-Date
+$workflowRunTime = Get-Date -AsUTC
 
 Write-Host "Running the workflow '$workflowName' on branch $targetBranch"
 gh workflow run --repo $repository --ref $targetBranch $workflowName
 
 # Get the workflow run URL to display in the message
 
-while((Get-Date) -lt $workflowRunTime.AddMinutes(1)) {
+while((Get-Date -AsUTC) -lt $workflowRunTime.AddMinutes(1)) {
     Start-Sleep -Seconds 5 # wait for 5 seconds for the workflow to start
     $workflowRun = gh run list --branch $targetBranch --event workflow_dispatch --workflow $workflowName --repo $repository --json createdAt,url --limit 1 | ConvertFrom-Json
 
--- a/.github/workflows/SubmitStabilityJobs.yaml
+++ b/.github/workflows/SubmitStabilityJobs.yaml
@@ -19,7 +19,7 @@ jobs:
         id: getOfficialBranches
         uses: microsoft/BCApps/.github/actions/GetGitBranches@main
         with:
-          include: "['main', 'releases/*']"
+          include: "['releases/*']"
 
   SubmitStabilityJobs:
     name: "[${{ matrix.branch }}] Submit Stability Job"
--- a/.github/workflows/SubmitStabilityJobs.yaml
+++ b/.github/workflows/SubmitStabilityJobs.yaml
@@ -9,8 +9,8 @@ permissions: read-all
 
 jobs:
   GetBranches:
-    if: github.repository_owner == 'microsoft'
     name: Get Official Branches
+    if: github.repository_owner == 'microsoft'
     runs-on: ubuntu-latest
     outputs:
       officialBranches: ${{ steps.getOfficialBranches.outputs.branchesJson }}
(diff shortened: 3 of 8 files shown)


.github/workflows/SubmitStabilityJobs.yaml
```diff
@@ -0,0 +1,41 @@
...
+    if: github.repository_owner == 'microsoft'
+    runs-on: ubuntu-latest
+    outputs:
+      officialBranches: ${{ steps.getOfficialBranches.outputs.branchesJson }}
+    steps:
+      - name: Get Official Branches
+        id: getOfficialBranches
+        uses: microsoft/BCApps/.github/actions/GetGitBranches@main
+        with:
+          include: "['main', 'releases/*']"
```
mazhelez: `main` could be omitted here.

.github/actions/RunAutomation/SubmitStabilityJob/run.ps1
```diff
@@ -0,0 +1,35 @@
+param (
+    [Parameter(Mandatory=$true)]
+    $runParameters
+)
+
+$repository = $runParameters.Repository
+$targetBranch = $runParameters.TargetBranch
+$workflowName = " CI/CD"
+$workflowRunTime = Get-Date
```
mazhelez:
```suggestion
$workflowRunTime = Get-Date -AsUTC
```

I think the `$workflowRun.createdAt` down is in universal time. Please double-check.

.github/actions/RunAutomation/UpdateALGo/run.ps1
```diff
@@ -5,11 +5,10 @@ param (
 
 $repository = $runParameters.Repository
 $targetBranch = $runParameters.TargetBranch
-
-Write-Host "Running the workflow Update AL-Go System Files on branch $targetBranch"
-
 $workflowName = " Update AL-Go System Files"
 $workflowRunTime = Get-Date
```
mazhelez:
```suggestion
$workflowRunTime = Get-Date -AsUTC
```

build/scripts/EnlistmentHelperFunctions.psm1
```diff
@@ -315,6 +315,15 @@ function Update-BCArtifactVersion {
 
     Write-Host "Latest BCArtifact URL: $latestArtifactUrl"
 
+    if ($currentArtifactUrl -notlike "https*") {
```
mazhelez, line 323: No need for `$latestArtifactUrl = Get-LatestBCArtifactUrl -minimumVersion $currentVersion` then (line 314).

build/scripts/EnlistmentHelperFunctions.psm1
```diff
@@ -315,6 +315,15 @@ function Update-BCArtifactVersion {
...
+        Write-Host "Using latest BCArtifact version for the current repo version: $currentVersion"
+        if ($latestArtifactUrl -contains "bcartifacts") {
+            $latestArtifactUrl = "bcartifacts/Sandbox/$currentVersion/base/latest"
+        } else {
+            $latestArtifactUrl = "bcinsider/Sandbox/$currentVersion/base/latest"
```
mazhelez: Shouldn't there be a check to verify this version exists?

E.g. If we bump the version to 24.5, but there is still no 24.5 artifact yet? Should happen around branching, when we change the repo version (and all app versions), but only change the artifact version once there is an artifact with that version.

build/scripts/EnlistmentHelperFunctions.psm1
aholstrup1: Why not? 🤔 We still want to use the latestArtifactUrl when currentArtifactUrl matches "https*" i.e. in main. 

And in release branches I still use it to see if we should use '**bcinsider**/Sandbox/$currentVersion/base/latest' or '**bcartifacts**/Sandbox/$currentVersion/base/latest'. That way we should get a PR generated for releases/25.5 when we start pushing 25.5 artifacts to bcartifacts.

build/scripts/EnlistmentHelperFunctions.psm1
aholstrup1: We run `Get-LatestBCArtifactUrl -minimumVersion $currentVersion` earlier in the function. This will fail if there are no 24.5 artifacts.

.github/actions/RunAutomation/SubmitStabilityJob/run.ps1
aholstrup1: As far as I can see, that is the case yes.

build/scripts/EnlistmentHelperFunctions.psm1
mazhelez: Oh, yeah, no I understand better what's happening here.

build/scripts/EnlistmentHelperFunctions.psm1
mazhelez: Yes, you are right.

So, this new piece of code just converts a full URL to e.g. `bcinsider/Sandbox/24.5/latest`
Does it then make sense to be part of `Get-LatestBCArtifactUrl`. E.g.
`Get-LatestBCArtifactUrl -minimumVersion 24.4 -asURL` would return the full URL and ``Get-LatestBCArtifactUrl -minimumVersion 24.4` would return... the not full URL 😄 
You got the idea, right?

Also, instead of checking `$currentArtifactUrl -notlike "https*"`, why not check the branch name? That way we won't have to edit the artifact property in 6 release branches and we won't have to remember to change it when branching major versions ((the automation will do all that).

mazhelez (approved):
Left some suggestions. It looks okay, but I think it could be optimized (see comments).

build/scripts/EnlistmentHelperFunctions.psm1
aholstrup1: > So, this new piece of code just converts a full URL to e.g. bcinsider/Sandbox/24.5/latest
Does it then make sense to be part of Get-LatestBCArtifactUrl. E.g.
Get-LatestBCArtifactUrl -minimumVersion 24.4 -asURL would return the full URL and ``Get-LatestBCArtifactUrl -minimumVersion 24.4` would return... the not full URL 😄
You got the idea, right?

Yeeah I suppose. Except the default behavior of Get-LatestBCArtifact**Url** should probably be to return... a URL 😄 But I can add a `-AsPattern` switch (or something like that) to change this.

> Also, instead of checking $currentArtifactUrl -notlike "https*", why not check the branch name? That way we won't have to edit the artifact property in 6 release branches and we won't have to remember to change it when branching major versions ((the automation will do all that).

My idea was to have the automation respect the whatever we decide the "artifact" setting should be (either URL or pattern). Around the time we branch for a new major version we might want to keep the artifact URLs in the new .x branch for a little bit so we know we're using the right artifacts. There have been cases in the past where we picked up the wrong artifacts. 

As you said, the downside is that I'll have to go update 6 branches once this PR is merged and longer term, we'll have to remember to change it when we branch for a new major version. However, since our automation will try to update the artifact version every day, we will be reminded that we forgot to change it. And once we decide to change it, we can just edit the automated PRs. Shouldn't take long

mazhelez: Isn't this exactly why we need the automation to try set the artifact setting based on the branch?

Say we branch 25.x, `artifact` is full URL and then the automation tries to change it to a pattern (the PR will be the reminder).
Same for the cases when we might need to revert to using full URL. The next day the automation will try to set a pattern and the PR build on the automated PR will tell us if it's good to go.

aholstrup1: > Say we branch 25.x, artifact is full URL and then the automation tries to change it to a pattern (the PR will be the reminder).

That's not ideal if we want to keep getting the full URLs for a few days after we've done branching for 25.x e.g. to make sure we didn't produce an artifact with a higher version out of master before doing the version update to 26.0 in master. Might be nice that the first bcartifact update in 25.x just updates the URL so we know we're picking up a bcartifact produced from 25.x 🤷‍♂️

StefanSosic (approved):

build/scripts/EnlistmentHelperFunctions.psm1
mazhelez: > make sure we didn't produce an artifact with a higher version out of master before doing the version update to 26.0 in master.

That's part of the branching process and shouldn't affect this automation.

I say YOLO, what's the worst that can happen? We built against an older platform and will eventually surface somewhere. Anyway, it is up to you.

mazhelez (approved):

mazhelez (approved):

PR #1812, opened 2024-08-14
Scheduled Profiler - Bugfixes from testing

BazookaMusic:
#### Summary
Some bugs were identified on further testing of the feature. These are:
1) While non-admin users cannot see other users' profiles and schedules, they can create them as insert wasn't guarded on the page.
2) The description is not correctly validated to be non-empty
3) On the performance profiles page, we incorrectly include "(ms)" in the caption although the duration type is formatted with a textual form of the duration.

#### Work Item(s)
Fixes [AB#544034](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/544034)

### DIFF
--- a/src/System Application/App/Performance Profiler/src/PerfProfilerScheduleCard.Page.al
+++ b/src/System Application/App/Performance Profiler/src/PerfProfilerScheduleCard.Page.al
@@ -80,7 +80,12 @@ page 1932 "Perf. Profiler Schedule Card"
                     Caption = 'Description';
                     ToolTip = 'Specifies the description of the schedule.';
                     AboutText = 'The description of the schedule.';
-                    NotBlank = true;
+
+                    trigger OnValidate()
+                    begin
+                        if Rec.Description = '' then
+                            Error(NotEmptyDescriptionErr);
+                    end;
                 }
             }
 
@@ -228,6 +233,7 @@ page 1932 "Perf. Profiler Schedule Card"
         MaxRetentionPeriod: Duration;
         NoRetentionPolicySetupErr: Label 'No retention policy setup found for the performance profiles table.';
         CreateRetentionPolicySetupTxt: Label 'Create a retention policy setup';
+        NotEmptyDescriptionErr: Label 'Description must be filled in.';
 
     local procedure ValidateRecord()
     begin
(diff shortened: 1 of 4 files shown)


src/System Application/App/Performance Profiler/src/PerfProfilerScheduleCard.Page.al
```diff
@@ -223,10 +233,17 @@ page 1932 "Perf. Profiler Schedule Card"
         MaxRetentionPeriod: Duration;
         NoRetentionPolicySetupErr: Label 'No retention policy setup found for the performance profiles table.';
         CreateRetentionPolicySetupTxt: Label 'Create a retention policy setup';
+        NotEmptyDescriptionErr: Label 'The description must be filled in.';
```
darjoo: The variable name dis different from what the label is. It sounds like the description is not empty. When the label is a message saying that is must be filled in.
I would rename to something like `DescriptionNotFilledErr` or `EmptyDescriptionErr`.

src/System Application/App/Performance Profiler/src/ScheduledPerfProfilerImpl.Codeunit.al
```diff
@@ -220,4 +231,5 @@ codeunit 1932 "Scheduled Perf. Profiler Impl."
         ProfileCannotBeInThePastErr: Label 'A schedule cannot be set to run in the past.';
         ScheduleDurationCannotExceedRetentionPeriodErr: Label 'The performance profile schedule duration cannot exceed the retention period.';
         ScheduleEndTimeCannotBeEmptyErr: Label 'The performance profile schedule must have an end time.';
+        CannotCreateSchedulesForOtherUsersErr: Label 'Only user admins can create schedules for other users.';
```
darjoo: Sounds weird. `Only admin users...` maybe? Check with the design team, Jacob, as they should review labels.

darjoo: Additionally, is the text actually correct? It doesn't exactly require an admin user does it? A user could be given `SECURITY` (Which is what `CanManageUsersOnTenant` checks for).

darjoo (approved):

src/System Application/App/Performance Profiler/src/ScheduledPerfProfilerImpl.Codeunit.al
BazookaMusic: Fixed

src/System Application/App/Performance Profiler/src/PerfProfilerScheduleCard.Page.al
BazookaMusic: Fixed

JesperSchulz (approved):

PR #1835, opened 2024-08-19
[Copilot] New IsCapabilityRegisteredSaaS api

darjoo:
#### Summary 
New api `IsCapabilityRegisteredSaaS` will check if a capability has been registered and specifically if it is on `SaaS` environment.

#### Work Item(s) 
Fixes [AB#544933](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/544933)

### DIFF
--- a/src/System Application/App/AI/src/Copilot/CopilotCapability.Codeunit.al
+++ b/src/System Application/App/AI/src/Copilot/CopilotCapability.Codeunit.al
@@ -92,7 +92,21 @@ codeunit 7773 "Copilot Capability"
     /// <remarks>Capabilities are tied to the module registering it. Checking for a capability will check if the enum and app id of the module exists.</remarks>
     procedure IsCapabilityRegistered(CopilotCapability: Enum "Copilot Capability"; AppId: Guid): Boolean
     begin
-        exit(CopilotCapabilityImpl.IsCapabilityRegistered(CopilotCapability, AppId));
+        exit(CopilotCapabilityImpl.IsCapabilityRegistered(CopilotCapability, AppId, false));
+    end;
+
+    /// <summary>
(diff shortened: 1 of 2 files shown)


StefanSosic (approved):

mikebcMSFT:
Are there downsides to also checking for SaaS rather than cleaner separation of concern and focusing only on registration check?

For example, if our registration system changes in the future (note: no plans to do so), then apps will lose the SaaS check.

src/System Application/App/AI/src/Copilot/CopilotCapability.Codeunit.al
```diff
@@ -92,7 +92,21 @@ codeunit 7773 "Copilot Capability"
...
+        exit(CopilotCapabilityImpl.IsCapabilityRegistered(CopilotCapability, AppId, false));
+    end;
+
+    /// <summary>
+    /// Check if your capability has been registered and the environment is on SaaS.
+    /// </summary>
+    /// <param name="CopilotCapability">The capability.</param>
+    /// <returns>True if the capability has been registered.</returns>
+    /// <remarks>Capabilities are tied to the module registering it. Checking for a capability will check if the enum and app id of your module exists.</remarks>
+    procedure IsCapabilityRegisteredSaaS(CopilotCapability: Enum "Copilot Capability"): Boolean
```
msft-sam, line 104: Similar to Mike's comment, the purpose of this procedure is to determine if your copilot action should be visible. The actual implementation logic isn't relevant.

Should we rename it to `IsCapabilityActionVisible`, or something along those lines? Then if the logic changes, we don't need to obsolete.

PR #1854, opened 2024-08-21
Page Summary provider - Add ability to exclude media, media set and blobs from page summary provider

nikolakukrika:
#### Work Item(s) 
Fixes [AB#540386](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/540386)

### DIFF
--- a/src/System Application/App/Page Summary Provider/permissions/PageSummaryProviderObj.PermissionSet.al
+++ b/src/System Application/App/Page Summary Provider/permissions/PageSummaryProviderObj.PermissionSet.al
@@ -11,6 +11,7 @@ permissionset 2716 "Page Summary Provider - Obj."
     Assignable = false;
 
     Permissions = table "Page Summary Settings" = X,
+                  table "Page Summary Parameters" = X,
                   codeunit "Page Summary Provider" = X,
                   codeunit "Page Summary Settings" = X,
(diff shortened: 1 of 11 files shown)


WaelAbuSeada (approved):

src/System Application/App/Page Summary Provider/permissions/PageSummaryProviderObj.PermissionSet.al
```diff
@@ -11,6 +11,7 @@ permissionset 2716 "Page Summary Provider - Obj."
     Assignable = false;
 
     Permissions = table "Page Summary Settings" = X,
+                  table "Page Summary Parameters" = X,
```
JesperSchulz, line 14: Temporary tables don't need permissions.

JesperSchulz (approved):

src/System Application/App/Page Summary Provider/permissions/PageSummaryProviderObj.PermissionSet.al
nikolakukrika: As talked offline, we will keep it to align with other objects. We do not have inherent permissions unfortunately on this module. Sam or Monica can fix it once they are refactoring this system.

PR #1869, opened 2024-08-26
[AI Test Tool] Add support for multiple iterations of the same codeunit

stkillen:
#### Summary 
Allow multiple iterations of the same codeunit in AI Test Tool

#### Work Item(s) 
Fixes [AB#543079](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/543079)

### DIFF
--- a/src/Tools/AI Test Toolkit/src/TestSuite/AITALTestSuiteMgt.Codeunit.al
+++ b/src/Tools/AI Test Toolkit/src/TestSuite/AITALTestSuiteMgt.Codeunit.al
@@ -74,17 +74,21 @@ codeunit 149037 "AIT AL Test Suite Mgt"
         TempTestMethodLine: Record "Test Method Line" temporary;
         ALTestSuite: Record "AL Test Suite";
         TestInputsManagement: Codeunit "Test Inputs Management";
+        Iteration: Integer;
     begin
         ALTestSuite := GetOrCreateALTestSuite(AITTestMethodLine);
 
-        TempTestMethodLine."Line Type" := TempTestMethodLine."Line Type"::Codeunit;
-        TempTestMethodLine."Test Codeunit" := AITTestMethodLine."Codeunit ID";
-        TempTestMethodLine."Test Suite" := AITTestMethodLine."AL Test Suite";
-        TempTestMethodLine."Data Input Group Code" := TestInput."Test Input Group Code";
-        TempTestMethodLine."Data Input" := TestInput.Code;
-        TempTestMethodLine.Insert();
+        for Iteration := 0 to AITTestMethodLine.Iterations - 1 do begin
+            TempTestMethodLine."Line Type" := TempTestMethodLine."Line Type"::Codeunit;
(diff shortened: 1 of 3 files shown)


src/Tools/AI Test Toolkit/src/TestSuite/AITALTestSuiteMgt.Codeunit.al
```diff
@@ -74,17 +74,21 @@ codeunit 149037 "AIT AL Test Suite Mgt"
...
     begin
         ALTestSuite := GetOrCreateALTestSuite(AITTestMethodLine);
 
-        TempTestMethodLine."Line Type" := TempTestMethodLine."Line Type"::Codeunit;
-        TempTestMethodLine."Test Codeunit" := AITTestMethodLine."Codeunit ID";
-        TempTestMethodLine."Test Suite" := AITTestMethodLine."AL Test Suite";
-        TempTestMethodLine."Data Input Group Code" := TestInput."Test Input Group Code";
-        TempTestMethodLine."Data Input" := TestInput.Code;
-        TempTestMethodLine.Insert();
+        for Iteration := 0 to AITTestMethodLine.Iterations - 1 do begin
```
nikolakukrika, line 81: This implementation is problematic as we may end up inserting too many lines and it is not the best.

We need to implement automation of the Run method. User would invoke run method many times to create different iterations with history. The implementation should be to loop at the action itself each time mimicking like the user has invoked the action again e.g. 20 times for 20 iterations. This would be the best design because this is the flow that we are implementing. 

There should be a dialog to make the task stop.

src/Tools/AI Test Toolkit/src/TestSuite/AITTestMethodLine.Table.al
```diff
@@ -166,6 +166,18 @@ table 149032 "AIT Test Method Line"
             FieldClass = FlowField;
             CalcFormula = count("AIT Log Entry" where("Test Suite Code" = field("Test Suite Code"), "Test Method Line No." = field("Line No."), Version = field("Base Version Filter")));
         }
+        field(32; Iterations; Integer)
```
nikolakukrika, line 169: IMO Iteration should be on the suite level. We can also have it on the line, but then we need to implement the loop at the run of the line. That could be confusing as we do not have a way of running a single test. 
Line should be a single test, many iterations - suite

nikolakukrika (changes requested):

src/Tools/AI Test Toolkit/src/TestSuite/AITALTestSuiteMgt.Codeunit.al
stkillen: Then we should also add functionality to allow easy access to aggregated results across multiple versions. Otherwise, it's going to be difficult to get an average accuracy number, which is the main purpose of allowing multiple iterations.

stkillen:
Closing PR based on above feedback. Changing to per-suite level. See https://github.com/microsoft/BCApps/pull/2108

PR #1871, opened 2024-08-26
Disable Suggest No. Series features for now.

JesperSchulz:
#### Summary 
Disable the "Suggest No. Series" feature until we've got all necessary sign-offs and approvals.

#### Work Item(s) 
Fixes [AB#545893](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/545893)

### DIFF
--- a/src/Business Foundation/App/NoSeriesCopilot/src/Register/NoSeriesCopilotInstall.Codeunit.al
+++ b/src/Business Foundation/App/NoSeriesCopilot/src/Register/NoSeriesCopilotInstall.Codeunit.al
@@ -13,10 +13,11 @@ codeunit 290 "No. Series Copilot Install"
     InherentPermissions = X;
     InherentEntitlements = X;
 
-    trigger OnInstallAppPerDatabase()
-    var
-        NoSeriesCopilotRegister: Codeunit "No. Series Copilot Register";
-    begin
-        NoSeriesCopilotRegister.RegisterCapability();
-    end;
+    // TODO: Implement this trigger when all green lights are given.
+    // trigger OnInstallAppPerDatabase()
+    // var
+    //     NoSeriesCopilotRegister: Codeunit "No. Series Copilot Register";
+    // begin
+    //     NoSeriesCopilotRegister.RegisterCapability();
+    // end;
 }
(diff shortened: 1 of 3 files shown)


darjoo (approved):
Sad, but approved.

darjoo:
@JesperSchulz Can you comment out the `RegisterCapability` function? I noticed that it is actually also subscribing to an event in the Copilot & AI Capabilities page, that will install it too :D

darjoo (changes requested):
See previous comment

WaelAbuSeada (dismissed):

nikolakukrika (dismissed):

darjoo (approved):

nikolakukrika (approved):

src/Business Foundation/App/NoSeriesCopilot/src/Register/NoSeriesCopilotInstall.Codeunit.al
```diff
@@ -13,10 +13,11 @@ codeunit 290 "No. Series Copilot Install"
     InherentPermissions = X;
```
AndreasMoth, line 13: Quick comment for when this is re-enabled. This should be run on upgrade, not install. We have another install codeunit for NoSeries, but it's backed up by an upgrade and it's only because the first install of BusinessFoundation should be an install (even that is not always completely true). After that, this will be an upgrade.

src/Business Foundation/App/NoSeriesCopilot/src/Register/NoSeriesCopilotRegister.Codeunit.al
```diff
@@ -15,11 +15,12 @@ codeunit 327 "No. Series Copilot Register"
...
-    local procedure HandleOnRegisterCopilotCapability()
-    begin
-        RegisterCapability();
-    end;
+    // TODO: Implement this trigger when all green lights are given.
+    // [EventSubscriber(ObjectType::Page, Page::"Copilot AI Capabilities", 'OnRegisterCopilotCapability', '', false, false)]
+    // local procedure HandleOnRegisterCopilotCapability()
+    // begin
+    //     RegisterCapability();
+    // end;
```
AndreasMoth, line 23: I noticed there's a TODO for the label LearnMoreUrlTxt, is that something we should fix? Not sure how that made it into the repo. At least we should put it into the work item for re-enabling this again.

JesperSchulz: Will do!

src/Business Foundation/App/NoSeriesCopilot/src/Register/NoSeriesCopilotInstall.Codeunit.al
JesperSchulz: Got will. Will address when re-enabling. Thanks 🙏

PR #1887, opened 2024-08-27
[Email] Add retention policy for email inbox

darjoo:
#### Summary 
Add retention policy for Email Inbox. The length being 2 days retention.

#### Work Item(s) 
Fixes [AB#539754](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/539754)

### DIFF
--- a/src/System Application/App/Email/src/EmailInstaller.Codeunit.al
+++ b/src/System Application/App/Email/src/EmailInstaller.Codeunit.al
@@ -89,8 +89,7 @@ codeunit 1596 "Email Installer"
         RetentionPolicySetup.Validate("Apply to all records", true);
         RetentionPolicySetup.Validate("Retention Period", RetentionPolicySetupCU.FindOrCreateRetentionPeriod("Retention Period Enum"::"2 Days"));
         RetentionPolicySetup.Validate(Enabled, true);
-        if not RetentionPolicySetup.Insert(true) then
-            RetentionPolicySetup.Modify(true);
+        RetentionPolicySetup.Insert(true);
 
         if IsInitialSetup then
             UpgradeTag.SetUpgradeTag(GetEmailInboxPolicyAddedToAllowedListUpgradeTag());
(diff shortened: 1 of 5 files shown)


src/System Application/App/Email/src/EmailInstaller.Codeunit.al
```diff
@@ -46,21 +48,73 @@ codeunit 1596 "Email Installer"
...
+
+        RetentionPolicySetup.SetRange("Table Id", Database::"Email Inbox");
+        if not RetentionPolicySetup.IsEmpty() then
+            exit;
+
+        RetentionPolicySetup.Validate("Table Id", Database::"Email Inbox");
+        RetentionPolicySetup.Validate("Apply to all records", true);
+        RetentionPolicySetup.Validate("Retention Period", RetentionPolicySetupCU.FindOrCreateRetentionPeriod("Retention Period Enum"::"2 Days"));
+        RetentionPolicySetup.Validate(Enabled, true);
+        RetentionPolicySetup.Insert(true);
```
stkillen, line 95: Wouldn't this fail on `ForceUpdate = true` if this is already inserted? It also seems it's only called with ForceUpdate false. Is this parameter needed?

darjoo: No, it wouldn't fail. There's a check above that the record doesn't exist. It'd exit if it exists.

stkillen (dismissed):

src/System Application/App/Retention Policy/src/Retention Period/RetentionPeriodEnum.Enum.al
```diff
@@ -29,6 +29,14 @@ enum 3900 "Retention Period Enum" implements "Retention Period"
         Caption = 'Custom';
     }
     /// <Summary>
+    /// The "2 Days" value results in a retention period where records that are older than 2 days are deleted.
+    /// </Summary>
+    value(10; "2 Days")
```
grobyns: do we need this? Why not use a custom date calc thing?

JesperSchulz (dismissed):
I find two days extremely short, but ok :-)

grobyns (approved):

darjoo:
> I find two days extremely short, but ok :-)

They can always disable or lengthen if needed.

PR #1899, opened 2024-08-28
[Copilot] GetCopilotCapabilityStatus subscriber not exiting for other capabilities

darjoo:
#### Summary 
The event is used by platform only, but now that we have other apps that will be called from platform, this subscriber needs to only handle the capabilities it knows of.

#### Work Item(s) 
Fixes [AB#546657](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/546657)


src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
```diff
@@ -277,6 +277,10 @@ codeunit 7774 "Copilot Capability Impl"
     begin
         CopilotCapability := Enum::"Copilot Capability".FromInteger(Capability);
 
+        if (CopilotCapability <> Enum::"Copilot Capability"::Chat) or
```
gggdttt: Not sure if we should throw an error here.

gggdttt: Feel free to  resolve it if there is reason we cannot throw error from here

gggdttt (approved):

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
darjoo: We should not be throwing an error. The subscriber should not handle this if it is not of these types

Aleyenda (dismissed):

Aleyenda (approved):

PR #1927, opened 2024-09-02
[AI Test Tool] Improve compare page

stkillen:
### Summary 
This PR adds some improvements to the compare page, specifically:
1. Introduce new page that shows entire history of runs
1. Add capability to view by tag or version
1. Allow filtering results to specific test lines
1. Improve the tool tips, captions and variable names. Delete unused fields and pages
1. Introduce new page for viewing test input/output and comparing
1. Add missing objects to permission sets

#### Run history page, view by version:
image

#### Run history page, view by tag:
image

#### Run history page, filter history to specific lines:
image

#### Compare test data
image

#### Work Item(s) 
Fixes [AB#541857](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/541857/)


src/Tools/AI Test Toolkit/src/Logs/AITLogEntries.Page.al
```diff
@@ -280,34 +280,13 @@ page 149033 "AIT Log Entries"
         ShowSensitiveData: Boolean;
 
     trigger OnAfterGetRecord()
+    var
+        AITLogEntryCU: Codeunit "AIT Log Entry";
```
nikolakukrika: Good rule is no abbreviations. CU for me is See you :D
Maybe AITLogEntryCodeunit?

src/Tools/AI Test Toolkit/src/Logs/AITLogEntry.Codeunit.al
```diff
@@ -5,21 +5,163 @@
...
         AITLogEntry.Run();
     end;
+
+    procedure UpdateRunHistory(Code: Code[100]; LineNo: Integer; AITViewBy: Enum "AIT Run History - View By"; var TempAITRunHistory: Record "AIT Run History" temporary)
+    var
+        AITRunHistory: Record "AIT Run History";
+        SeenTags: List of [Text[20]];
+    begin
+        TempAITRunHistory.DeleteAll();
+        AITRunHistory.SetRange("Test Suite Code", Code);
```
t-prda, line 51: Are we missing any filters before AITRunHistory.FindSet()?
If not, then can we do the AITRunHistory.FindSet() once and do the check for ViewBy inside the repeat?

src/Tools/AI Test Toolkit/src/Logs/AITLogEntries.Page.al
```diff
@@ -87,8 +87,11 @@ page 149033 "AIT Log Entries"
                     ToolTip = 'Specifies the test input of the test.';
 
                     trigger OnDrillDown()
+                    var
+                        AITTestData: Page "AIT Test Data Compare";
                     begin
-                        Message(Rec.GetInputBlob());
+                        AITTestData.SetRecord(Rec);
+                        AITTestData.Run();
```
nikolakukrika: Why not a static invocation - Page.Run(Page::"AIT Test Data Compare", Rec)?

src/Tools/AI Test Toolkit/src/Logs/AITLogEntry.Codeunit.al
```diff
@@ -5,21 +5,163 @@
...
+                        TempAITRunHistory.Insert();
+                    end;
+                    SeenTags.Add(AITRunHistory.Tag);
+                until AITRunHistory.Next() = 0;
+
+        if (LineNo <> 0) then
+            TempAITRunHistory.SetRange("Line No. Filter", LineNo)
+    end;
+
+    procedure LookupTestMethodLine(TestSuiteCode: Code[100]; var LineNoFilter: Text; var LineNo: Integer)
```
nikolakukrika: Maybe this should go to the different codeunit, like test suite management? Do we have anything similar there?

src/Tools/AI Test Toolkit/src/Logs/AITLogEntry.Codeunit.al
```diff
@@ -5,21 +5,163 @@
...
+            exit;
+
+        AITTestMethodLines.GetRecord(AITTestMethodLine);
+
+        AITTestMethodLine.CalcFields("Codeunit Name");
+        LineNoFilter := StrSubstNo(LineNoFilterLbl, AITTestMethodLine."Codeunit ID", AITTestMethodLine."Codeunit Name", AITTestMethodLine."Input Dataset");
+        LineNo := AITTestMethodLine."Line No.";
+    end;
+
+    procedure UpdateTestInput(TestInput: Text; TestInputView: Enum "AIT Test Input - View"): Text
```
nikolakukrika: Same comment for the methods below, it seems like it should go to the different object.

src/Tools/AI Test Toolkit/src/Logs/AITTestDataCompare.Page.al
```diff
@@ -0,0 +1,94 @@
...
+                                UpdateTestData();
+                            end;
+                        }
+
+                        part("Data Output"; "AIT Test Data")
+                        {
+                        }
+                    }
+
+
```
nikolakukrika, line 71: many blank lines

src/Tools/AI Test Toolkit/src/Logs/AITTestDataCompare.Page.al
```diff
@@ -0,0 +1,94 @@
...
+                            trigger OnValidate()
+                            begin
+                                UpdateTestData();
+                            end;
+                        }
+
+                        part("Data Output"; "AIT Test Data")
+                        {
+                        }
```
nikolakukrika, line 68: Consider adding caption

src/Tools/AI Test Toolkit/src/Logs/AITLogEntries.Page.al
```diff
@@ -241,7 +247,20 @@ page 149033 "AIT Log Entries"
...
+                Image = CompareCOA;
+                ToolTip = 'View Test Data.';
 
+                trigger OnAction()
+                var
+                    AITTestData: Page "AIT Test Data Compare";
+                begin
+                    AITTestData.SetRecord(Rec);
+                    AITTestData.Run();
+                end;
```
Groenbech96, line 254: RunTrigger? is that possible

nikolakukrika (approved):

src/Tools/AI Test Toolkit/src/Logs/AITLogEntry.Codeunit.al
```diff
@@ -5,21 +5,163 @@
...
     Access = Internal;
 
+    var
+        LineNoFilterLbl: Label 'Codeunit %1 "%2" (Input: %3)', Locked = true;
+
     procedure DrillDownFailedAITLogEntries(AITSuiteCode: Code[100]; LineNo: Integer; VersionNo: Integer)
     var
         AITLogEntries: Record "AIT Log Entry";
+    begin
+        AITLogEntries.SetRange(Version, VersionNo);
```
Groenbech96, line 16: Add index for this.

src/Tools/AI Test Toolkit/src/Logs/AITLogEntry.Codeunit.al
```diff
@@ -5,21 +5,163 @@
...
+        DrillDownFailedAITLogEntries(AITLogEntries, AITSuiteCode, LineNo);
+    end;
+
+    procedure DrillDownFailedAITLogEntries(AITSuiteCode: Code[100]; LineNo: Integer; Tag: Text[20])
+    var
+        AITLogEntries: Record "AIT Log Entry";
+    begin
+        AITLogEntries.SetRange(Tag, Tag);
```
Groenbech96, line 24: Index

src/Tools/AI Test Toolkit/src/Logs/AITLogEntry.Codeunit.al
```diff
@@ -5,21 +5,163 @@
...
+        DrillDownFailedAITLogEntries(AITLogEntries, AITSuiteCode, LineNo);
+    end;
```
Groenbech96, line 26: Why not just set filter where you call this function? instead of having two new functions that does nothing

src/Tools/AI Test Toolkit/src/Logs/AITLogEntry.Codeunit.al
```diff
@@ -5,21 +5,163 @@
...
+                repeat
+                    if not SeenTags.Contains(AITRunHistory.Tag) then begin
+                        TempAITRunHistory.TransferFields(AITRunHistory);
+                        TempAITRunHistory.Insert();
+                    end;
+                    SeenTags.Add(AITRunHistory.Tag);
+                until AITRunHistory.Next() = 0;
+
+        if (LineNo <> 0) then
+            TempAITRunHistory.SetRange("Line No. Filter", LineNo)
```
Groenbech96: This part is confusing. Rest of code seems to populate the temp record. 
Maybe move out of function to make this function easier to understand.

Groenbech96: Maybe call it GetRunHistory instead.

src/Tools/AI Test Toolkit/src/Logs/AITLogEntry.Codeunit.al
```diff
@@ -5,21 +5,163 @@
...
+    var
+        AITTestMethodLine: Record "AIT Test Method Line";
+        AITTestMethodLines: Page "AIT Test Method Lines Lookup";
+    begin
+        AITTestMethodLine.SetRange("Test Suite Code", TestSuiteCode);
+
+        AITTestMethodLines.SetTableView(AITTestMethodLine);
+        AITTestMethodLines.LookupMode(true);
+
+        if not (AITTestMethodLines.RunModal() = Action::LookupOK) then
```
Groenbech96: <> intead of a negated =

src/Tools/AI Test Toolkit/src/Logs/AITRunHistory.Page.al
```diff
@@ -0,0 +1,222 @@
...
+    end;
+
+    local procedure UpdateRunHistory()
+    var
+        AITLogEntry: Codeunit "AIT Log Entry";
+    begin
+        if not ApplyLineFilter then
+            LineNo := 0;
+
+        AITLogEntry.UpdateRunHistory(TestSuiteCode, LineNo, ViewBy, Rec);
```
Groenbech96: Probably just do filter here after Get or before, what with the LineNo. Then remove it as param.

src/Tools/AI Test Toolkit/src/Logs/AITLogEntry.Codeunit.al
```diff
@@ -5,21 +5,163 @@
...
+        else
+            TestData.Initialize(TestDataText);
+    end;
+
+    local procedure GetTestDataElement(ElementName: Text; TestData: Codeunit "Test Input Json"): Text
+    var
+        ElementTestDataJson: Codeunit "Test Input Json";
+        ElementExists: Boolean;
+    begin
+        ElementTestDataJson := TestData.ElementExists('turns', ElementExists);
```
Groenbech96: This function is totally opposite once expectation. It saus "ElementExists" so i would think it returns a bool. 
but no, it returns data? and also takes a bool param.

src/Tools/AI Test Toolkit/src/Logs/AITLogEntry.Codeunit.al
```diff
@@ -5,21 +5,163 @@
...
+            TestInputView::Question:
+                exit(GetTestDataElement('question', TestData));
+            TestInputView::Context:
+                exit(GetTestDataElement('context', TestData));
+            TestInputView::"Test Setup":
+                exit(GetTestDataElement('test_setup', TestData));
+            TestInputView::"Ground Truth":
+                exit(GetTestDataElement('ground_truth', TestData));
+            TestInputView::"Expected Data":
+                exit(GetTestDataElement('expected_data', TestData));
```
Groenbech96: Good candiate for using an interface instead. The seems like something that should be an Enum, then we can add to it as requirements expand.

src/Tools/AI Test Toolkit/src/Logs/AITRunHistory.Page.al
```diff
@@ -0,0 +1,222 @@
...
+                        ToolTip = 'Specifies the version.';
+                    }
+                    field(Tag; Rec.Tag)
+                    {
+                        Caption = 'Tag';
+                        ToolTip = 'Specifies the tag of the version.';
+                    }
+                    field("No. of Tests - By Version"; Rec."No. of Tests Executed")
+                    {
+                        Visible = ViewBy = ViewBy::Version;
```
Groenbech96, line 98: Instead of trying to do this ViewBy stuff, why not just use filters normally? 
Then you dont need the table to contain both things either? And people can use BC platform filters as normal.

src/Tools/AI Test Toolkit/src/Logs/AITRunHistory.Page.al
```diff
@@ -0,0 +1,222 @@
...
+                        Editable = false;
+                        Caption = 'No. of Tests Failed';
+                        ToolTip = 'Specifies the number of tests that failed in the version.';
+                        Style = Unfavorable;
+
+                        trigger OnDrillDown()
+                        var
+                            AITLogEntryCU: Codeunit "AIT Log Entry";
+                        begin
+                            AITLogEntryCU.DrillDownFailedAITLogEntries(Rec."Test Suite Code", Rec."Line No. Filter", Rec.Version);
```
Groenbech96:
```suggestion
                            Rec.SetRange(Version, Rec.Version)
                            AITLogEntryCU.DrillDownFailedAITLogEntries(Rec."Test Suite Code", Rec."Line No. Filter");
```

Groenbech96: This is where this function was used. Why not just set range here.

Groenbech96 (commented):
Left some inital comments. Lets go over them first, then i will continue review the rest.

Groenbech96 (changes requested):
Left some inital comments. Lets go over them first, then i will continue review the rest.

src/Tools/AI Test Toolkit/src/Logs/AITLogEntry.Codeunit.al
stkillen: It's specifically related to viewing test data, which is stored in AIT Log Entry. It's not really relevant in another context. I think this is more accurate placement of the function than placing it in a generic management codeunit that already has way to wide of a responsibility?

src/Tools/AI Test Toolkit/src/Logs/AITLogEntry.Codeunit.al
stkillen: It's specifically related to viewing test data, which is stored in AIT Log Entry. It's not really relevant in another context. I think this is more accurate placement of the function than placing it in a generic management codeunit that already has way to wide of a responsibility?

src/Tools/AI Test Toolkit/src/Logs/AITLogEntry.Codeunit.al
stkillen: It is already an enum. I think implementing an interface for this scenario seems a little over-engineered. Should I then have 7 implementation codeunits with a single 1-line function? Or what is your suggestion here?

src/Tools/AI Test Toolkit/src/Logs/AITLogEntry.Codeunit.al
stkillen: The bool param is passed by var and set to true/false depending on whether the element exist. Then the data is also returned if it exist. I didn't implement this functionality, simply using existing functionality in data-driven framework.

src/Tools/AI Test Toolkit/src/Logs/AITRunHistory.Page.al
stkillen: Why even have function in the first place? We could just do everything in-line here?

I like that it's clean on the page and code is in the codeunit. Generally we also want to remove code from pages and have it in codeunits. It's more of a philosophical discussion about when you should create a function...
If I had to SetRange on 5 fields, would you agree that it's right to move it to a separate function in a codeunit so we avoid having code on the page? If not, wow many lines of code do you need before you should have a separate function?

I think it makes sense to have these overloads, to be able to filter on a specific version or tag when opening the page. From an API pov it also makes sense to have them.

src/Tools/AI Test Toolkit/src/Logs/AITRunHistory.Page.al
stkillen: Again, I think we should centralize code in codeunits when possible.

src/Tools/AI Test Toolkit/src/Logs/AITRunHistory.Page.al
stkillen: You can't do that with normal filters.. It's aggregated data based on either version or tag. That's the whole reason I had to do that

src/Tools/AI Test Toolkit/src/Logs/AITLogEntry.Codeunit.al
stkillen: This sets the flowfilter on the record. I think it's ok to populate and set a filter in one function? 

Renamed to GetRunHistory

src/Tools/AI Test Toolkit/src/Logs/AITRunHistory.Page.al
stkillen: The number of records that we need to display is different based on the view. If it's version, then we need to get the aggregated numbers for each version, but you can have multiple versions with the same tag. 

So, resolving this as it's not possible to achieve the desired functionality with just normal filters.

src/Tools/AI Test Toolkit/src/Logs/AITLogEntry.Codeunit.al
stkillen: Resolving as it's already an enum and implementing an interface and 7 implementations seems over-engineered for this purpose. Please reopen if you disagree.

src/Tools/AI Test Toolkit/src/Logs/AITRunHistory.Page.al
stkillen: Resolving as I think it's worth abstracting this in it's own overloaded function.

src/Tools/AI Test Toolkit/src/Logs/AITLogEntry.Codeunit.al
stkillen: They don't do nothing. I think it's nice having this in it's own separate function and not having to do part of the filtering happen on the page and the other part in the codeunit. This makes it clearer what's happening and what's being filtered on imo.

src/Tools/AI Test Toolkit/src/Logs/AITLogEntry.Codeunit.al
stkillen: Added

src/Tools/AI Test Toolkit/src/Logs/AITLogEntry.Codeunit.al
stkillen: Added

src/Tools/AI Test Toolkit/src/Logs/AITLogEntry.Codeunit.al
stkillen: Actually, this specific method makes sense to move to Test Suite Mgt. 

The others you mentioned probably do not belong in either, so I created two new codeunits for them and moved them there.

src/Tools/AI Test Toolkit/src/Logs/AITLogEntry.Codeunit.al
stkillen: Moved to AIT Test Data codeunit.

src/Tools/AI Test Toolkit/src/Logs/AITLogEntries.Page.al
stkillen: CU is widely used as codeunit around the product, but changed to codeunit.

src/Tools/AI Test Toolkit/src/Logs/AITLogEntries.Page.al
stkillen: Fixed

Groenbech96 (dismissed):
LGTM

darjoo (approved):

gggdttt (approved):

PR #1973, opened 2024-09-06
Re-enable Retention Policy tests

### LINKED ISSUE #1947: [Bug]: Retention Policy Filtering SetSelectedSystemId
### Describe the issue

In the Codeunit 3915 "Reten. Pol. Filtering Impl." in the procedure SetSelectedSystemId the else statement is on the wrong ir-statement. It should be for `if IsSelected then` and not for `if not SelectedSystemIds.ContainsKey(SelectedSystemId) then`.

https://github.com/microsoft/BCApps/blob/832c37a2fee659ce038c48d9016eebf610ebfcaa/src/System%20Application/App/Retention%20Policy/src/Apply%20Retention%20Policy/RetenPolFilteringImpl.Codeunit.al#L276

This influences the record counter and could lead to the wrong wrong outcome of the retention policy deletion.
For example many customers and partners have problem with the change log and maybe this could solve the issue. 

I think the list gets longer and even when removing the entries to delete, it will not be removed instead kept in the list to delete. Because the marks get removed but not the buffer, the counter gets wrong in the `ApplyRetentionPolicySubSetFilters` procedure of the same codeunit.

https://github.com/microsoft/BCApps/blob/832c37a2fee659ce038c48d9016eebf610ebfcaa/src/System%20Application/App/Retention%20Policy/src/Apply%20Retention%20Policy/RetenPolFilteringImpl.Codeunit.al#L120

### Expected behavior

The counter should be updated correct.

### Steps to reproduce

.

### Additional context

_No response_

grobyns:
#### Summary 
re-enabling retention policy tests

#### Work Item(s) 
Fixes [AB#488830](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/488830)
Fixes #1947


src/System Application/App/Retention Policy/app.json
```diff
@@ -74,8 +74,8 @@
   ],
   "internalsVisibleTo": [
     {
-      "id": "489a0bcb-0619-4bd1-b626-9f30dbe8af4d",
-      "name": "Retention Policy Test",
+      "id": "33003b8a-f6d8-4efe-af22-1a8cb8fbacbe",
+      "name": "Retencion Policy Test Library",
```
aholstrup1:
```suggestion
      "name": "Retention Policy Test Library",
```

Or Retención to give it more of a Spanish touch 😄

aholstrup1 (approved):

src/System Application/App/Retention Policy/app.json
grobyns: nice catch :P

aholstrup1 (approved):

mazhelez (approved):

haoranpb:
@grobyns @nikolakukrika  I know this is a bit old history, but can you help me understand why does this test codeunit needs `TestIsolation=Dsiabled`?

Tried it locally, the tests are indeed only passing when running it with this `codeunit 130451 "Test Runner - Isol. Disabled"`. But not sure I understand why after reading the code... (e.g. [TestLogError](https://github.com/microsoft/BCApps/blob/d238f297b45fc78a0a920e6d5fb7695ec46967be/src/System%20Application/Test/Retention%20Policy/src/RetentionPolicyLogTest.Codeunit.al#L76))

I assume this is because `asserterror` behaviors differently between Test Runners? I guess the part that confuses me: this has nothing to do with [TestIsolation](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/properties/devenv-testisolation-property) in the sense that tests can be isolated from each other

PR #1976, opened 2024-09-09
Use AppId function to delete orphaned extension data AB#534788

ehenriksen9:
Problem: Before apps not installed since 14.x might be missing a package ID. So we are unable to delete them by packageID

Fix: Delete by AppId

Fixes: [AB#534788](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/534788)

### DIFF
--- a/src/System Application/App/Extension Management/src/ExtensionInstallationImpl.Codeunit.al
+++ b/src/System Application/App/Extension Management/src/ExtensionInstallationImpl.Codeunit.al
@@ -53,7 +53,8 @@ codeunit 2500 "Extension Installation Impl"
         exit(not NAVAppInstalledApp.IsEmpty());
     end;
 
-    procedure IsInstalledByAppId(AppID: Guid): Boolean
+
+    procedure IsInstalledByAppId(AppId: Guid): Boolean
     var
         [SecurityFiltering(SecurityFilter::Ignored)]
         NAVAppInstalledApp: Record "NAV App Installed App";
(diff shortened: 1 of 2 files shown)


ehenriksen9:
> > ```
> > ```

kalberes (approved):

src/System Application/App/Extension Management/src/ExtensionInstallationImpl.Codeunit.al
```diff
@@ -241,6 +241,18 @@ codeunit 2500 "Extension Installation Impl"
...
+        AssertIsInitialized();
+        DotNetNavAppALInstaller.ALUninstallNavAppByAppId(AppId, ClearData, ClearSchema);
+
+        if IsInstalledByAppId(AppId) then
+            exit(false);
+
+        exit(true);
+    end;
+
     procedure UninstallExtensionWithConfirmDialog(PackageId: Guid; ClearData: Boolean; ClearSchema: Boolean): Boolean
```
qutreson, line 256: Do we need the App ID on these ones too? Or this is not a problem since installed apps are not from BC14 and older?

qutreson (approved):

src/System Application/App/Extension Management/src/ExtensionInstallationImpl.Codeunit.al
ehenriksen9: Yep I haven't seen any cases of installed apps missing package ID, i think it's not possible.

PR #1990, opened 2024-09-10
Delete TenantMediaSet in Batch and safeguard loading detached media

haoranpb:
#### Summary

The PR consists of two parts:
1. Safeguard loading detached media. Loading detached media to temp record could fail (observed in Telemetry) with duplicated ID, using `if ... then ...` so safeguard the insert.
2. Delete Tenant Media Set in batch and Commit between each batch. Just like what we did with Tenant Media, providing the ability to clean up even when there are large number of records.

The code change should be covered by test case: `EnsureDetachedMediaAndMediaSetAreCleanedUp`

#### Work Item(s)

Fixes AB#548323


darjoo (approved):

Groenbech96 (approved):
LGTM

PR #2025, opened 2024-09-12
Add ability to log warnings to Feature Telemetry

AleksanderGladkov:
Now it is only possible to log Usage, Error and Uptake.
Sometimes I need warnings in the features that I responsible for.

Fixes #548708

### DIFF
--- a/src/System Application/App/Telemetry/src/Feature Telemetry/FeatureTelemetryImpl.Codeunit.al
+++ b/src/System Application/App/Telemetry/src/Feature Telemetry/FeatureTelemetryImpl.Codeunit.al
@@ -42,6 +42,19 @@ codeunit 8704 "Feature Telemetry Impl."
         LogMessage(EventId, EventName, Verbosity::Normal, CallerCustomDimensions, UsageCustomDimensions, CallerModuleInfo, CallerCallStackModuleInfos);
     end;
 
+    procedure LogWarning(EventId: Text; FeatureName: Text; EventName: Text; WarningText: Text; CallerCustomDimensions: Dictionary of [Text, Text]; CallerModuleInfo: ModuleInfo)
+    var
+        WarningCustomDimensions: Dictionary of [Text, Text];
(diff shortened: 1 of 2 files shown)


src/System Application/App/Telemetry/src/Feature Telemetry/FeatureTelemetryImpl.Codeunit.al
```diff
@@ -42,6 +42,19 @@ codeunit 8704 "Feature Telemetry Impl."
         LogMessage(EventId, EventName, Verbosity::Normal, CallerCustomDimensions, UsageCustomDimensions, CallerModuleInfo, CallerCallStackModuleInfos);
     end;
 
+    procedure LogWarning(EventId: Text; FeatureName: Text; EventName: Text; WarningText: Text; CallerCustomDimensions: Dictionary of [Text, Text]; CallerModuleInfo: ModuleInfo)
```
grobyns, line 45: afair FeatureTelemetry is meant to be used for statistics... like, how much uptake do we have, how much usage do we have... I think the error bit was added to be able to see painpoints... 

For other things we have the Telemetry codeunit (same module here) and that is used for logging all sorts of telemetry, i.e. the normal stuff...

If we're adding warnings to feature telemetry, is there work to update the PowerBI dashboards etc? because if not, then I think you should use Telemetry and not FeatureTelemetry...

AleksanderGladkov: @grobyns Ok, thanks, I understand your point, but for me it is weird to write telemetry in my feature using two different codeunit. I see that it is ok to have errors in Feature Telemetry for statistics, but not ok to have warnings for statistics. I was not going to update PowerBI reports, I wanted to emit all telemetry from one place in my app, and as this module has "Feature" in its name, I thought it was supposed to be used in apps (extensions).

PR #2032, opened 2024-09-13
Edit In Excel: Fix metadata generation bug when caption name uses `en dash` instead of `em dash`

DenLilleMand:
#### Summary 
Apparently in the Norwegian translation and maybe other places captions use  `en dash` instead of the traditional `em dash`
, I'm not sure if this is on purpose but it does make our integrations a little more complicated to do nicely, e.g. now I will make sure we support it in Edit In Excel, but we could easily have other integrations that breaks on this or will have in the future.

Can see the following two screenshots where the first line is using `em dash` and the second line is using `en dash`, where the `en dash` is encoded and the `em dash` is removed.
Code:
image
Result:
image

#### Work Item(s) 
Fixes [AB#548471](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/548471/)


src/System Application/App/Edit in Excel/src/EditinExcelImpl.Codeunit.al
```diff
@@ -224,7 +224,9 @@ codeunit 1482 "Edit in Excel Impl."
         CurrentPosition := 1;
 
         while CurrentPosition <= StrLen(ConvertedName) do begin
-            if ConvertedName[CurrentPosition] in ['''', '+'] then begin
+            // Notice in the following line that – (en dash) is not a normal dash (em dash).
+            // We need to handle this here because at least the norwegian translation uses en dash.
+            if ConvertedName[CurrentPosition] in ['''', '+', '–'] then begin
```
encimita, line 231: Can we add a unit test hitting this path?
Nice that you have a comment but just want to make sure this doesn't slip somehow in the future.

encimita (approved):

src/System Application/App/Edit in Excel/src/EditinExcelImpl.Codeunit.al
DenLilleMand: @encimita added a test now

encimita (approved):

encimita (approved):

mazhelez (dismissed):

JesperSchulz (dismissed):

JesperSchulz (approved):

encimita (approved):

PR #2061, opened 2024-09-18
#1978 HasDatabaseUpgradeTag

### LINKED ISSUE #1978: [Bug]: UpgradeTag does not have a procedure for database upgrade tags
### Describe the issue

Codeunit 9999 "Upgrade Tag" does not have all procedures for database upgrade tags.the 

### Expected behavior

While we have 
- UpgradeTag.SetDatabaseUpgradeTag(TagName)

We don't have a corresponding procedure for HasDatabaseUpgradeTag. For this, developers must use HasUpgradeTag(TagName, '') to manually specify a blank company name.

While this is not challenging, it's hard to explain to new developers why we have a special Setter (as the SetUpgradeTag(TagName, '') would also work) but not for Getter.

### Steps to reproduce

-

### Additional context

_No response_

TKapitan:
#### Summary 
Adds HasDatabaseUpgradeTag() procedure to make the setters and getters the same.

#### Work Item(s) 
Fixes #1978

### DIFF
--- a/src/System Application/App/Upgrade Tags/src/UpgradeTag.Codeunit.al
+++ b/src/System Application/App/Upgrade Tags/src/UpgradeTag.Codeunit.al
@@ -44,6 +44,16 @@ codeunit 9999 "Upgrade Tag"
     end;
 
     /// <summary>
+    /// Verifies if the database upgrade tag exists.
+    /// </summary>
+    /// <param name="Tag">Tag code to check</param>
+    /// <returns>True if the Tag with given code exist.</returns>
+    procedure HasDatabaseUpgradeTag(Tag: Code[250]): Boolean
+    begin
+        exit(UpgradeTagImpl.HasDatabaseUpgradeTag(Tag));
+    end;
+
+    /// <summary>
     /// Sets the upgrade tag for database upgrades.
     /// </summary>
     /// <param name="NewTag">Tag code to save</param>
--- a/src/System Application/App/Upgrade Tags/src/UpgradeTagImpl.Codeunit.al
+++ b/src/System Application/App/Upgrade Tags/src/UpgradeTagImpl.Codeunit.al
@@ -25,6 +25,11 @@ codeunit 9996 "Upgrade Tag Impl."
         exit(HasUpgradeTag(Tag, CopyStr(CompanyName(), 1, MaxStrLen(ConstUpgradeTags.Company))));
     end;
 
+    procedure HasDatabaseUpgradeTag(Tag: Code[250]): Boolean
+    begin
+        exit(HasUpgradeTag(Tag, ''));
+    end;
+
     procedure HasUpgradeTag(Tag: Code[250]; TagCompanyName: Code[30]): Boolean
     var
         UpgradeTags: Record "Upgrade Tags";


src/System Application/App/Upgrade Tags/src/UpgradeTag.Codeunit.al
```diff
@@ -43,6 +43,16 @@ codeunit 9999 "Upgrade Tag"
         exit(UpgradeTagImpl.HasUpgradeTag(Tag, TagCompanyName));
     end;
 
+    /// <summary>
+    /// Verifies if the database upgrade tag exists.
+    /// </summary>
+    /// <param name="Tag">Tag code to check</param>
+    /// <returns>True if the Tag with given code exist.</returns>
```
NKarolak: I know that you only copied the existing pattern. However, I see no reason to copy grammatical errors as well:

```suggestion
    /// <returns>True if the Tag with the given code exists.</returns>
```

NKarolak (changes requested):

JesperSchulz (approved):

src/System Application/App/Upgrade Tags/src/UpgradeTag.Codeunit.al
```diff
@@ -43,6 +43,16 @@ codeunit 9999 "Upgrade Tag"
         exit(UpgradeTagImpl.HasUpgradeTag(Tag, TagCompanyName));
     end;
 
+    /// <summary>
+    /// Verifies if the database upgrade tag exists.
+    /// </summary>
+    /// <param name="Tag">Tag code to check</param>
+    /// <returns>True if the Tag with the given code exists.</returns>
+    procedure HasDatabaseUpgradeTag(Tag: Code[250]): Boolean
```
nikolakukrika, line 51: So the idea was to use SetUpgradeTag and HasUpgradeTag only, the code will find out from the context if the tag is per database or per company. Approving since we already have the SetDatabaseUpgradeTag, which is wrongly named IMO. It should be PerDatabaseUpgradeTag

src/System Application/App/Upgrade Tags/src/UpgradeTag.Codeunit.al
```diff
@@ -43,6 +43,16 @@ codeunit 9999 "Upgrade Tag"
         exit(UpgradeTagImpl.HasUpgradeTag(Tag, TagCompanyName));
     end;
 
+    /// <summary>
+    /// Verifies if the database upgrade tag exists.
```
nikolakukrika, line 47: per-databse

src/System Application/App/Upgrade Tags/src/UpgradeTag.Codeunit.al
```diff
@@ -43,6 +43,16 @@ codeunit 9999 "Upgrade Tag"
...
+    /// </summary>
+    /// <param name="Tag">Tag code to check</param>
```
nikolakukrika, line 49: Could you add a comment that the developer can use HasUpgradeTag if it is called from the upgrade context? If it is called from upgrade context e.g. OnUpgradePerDatabase we will automatically use perdatabase check and similar for percompany? 

Nothing wrong with using this one directly but it could be confusing if someone is reading the existing code.

nikolakukrika (approved):

StefanSosic (approved):

JesperSchulz:
@TKapitan, should we try to reel this one in? Not much pending, the way I see it. Nikola even approved.

src/System Application/App/Upgrade Tags/src/UpgradeTag.Codeunit.al
JesperSchulz: Seems to be a minor. Can be addressed in a chaser pack.

PR #2064, opened 2024-09-19
Added dotnet wrappers for Purview telemetry

attilatoury:
#### Summary

Added three DotNet types to enable the use of Purview telemetry (audit logging) in BaseApp (24.x).

Required for deliverables [AB#549389](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/549389) and [AB#548176](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/548176)

### DIFF
--- a/src/System Application/App/DotNet Aliases/src/dotnet.al
+++ b/src/System Application/App/DotNet Aliases/src/dotnet.al
@@ -729,6 +729,18 @@ dotnet
         type("Microsoft.Dynamics.Nav.Runtime.Designer.NavDesignerALCopyResponse"; "NavDesignerALCopyResponse")
         {
         }
+
+        type("Microsoft.Dynamics.Nav.Runtime.CustomerAudit.CustomerAuditLoggerALHelper"; "CustomerAuditLoggerALHelper")
+        {
+        }
+
+        type("Microsoft.Dynamics.Nav.Runtime.ALSecurityOperationResult"; "ALSecurityOperationResult")
+        {
+        }
+
+        type("Microsoft.Dynamics.Nav.Runtime.ALAuditCategory"; "ALAuditCategory")
+        {
+        }
     }
 
     assembly("Microsoft.Dynamics.Nav.OAuth")


mirkobiasini (approved):

AleksandricMarko (approved):

BardurKnudsen (approved):
ok

PredragMaricic (approved):

src/System Application/App/DotNet Aliases/src/dotnet.al
```diff
@@ -729,6 +729,18 @@ dotnet
         type("Microsoft.Dynamics.Nav.Runtime.Designer.NavDesignerALCopyResponse"; "NavDesignerALCopyResponse")
         {
         }
+
+        type("Microsoft.Dynamics.Nav.Runtime.CustomerAudit.CustomerAuditLoggerALHelper"; "CustomerAuditLoggerALHelper")
```
WaelAbuSeada, line 733: Where are these libraries consumed?

WaelAbuSeada (changes requested):

src/System Application/App/DotNet Aliases/src/dotnet.al
attilatoury: @WaelAbuSeada  They will be used by the deliverables linked in this PR's summary.

[Here](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_git/NAV/pullrequest/194895) and [here](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_git/NAV/pullrequest/193851).

WaelAbuSeada: We should not be using more dotnet in other apps, but rather add a wrapper to it in system module or create a platform function

attilatoury: @WaelAbuSeada The functions will be used both in BaseApp (see linked deliverables) and in SystemApp ([this PR](https://github.com/microsoft/BCApps/pull/2067)) so it was my understanding that they should be added in this file.

Adding a platform function was discussed but rejected.

@mirkobiasini Could you correct me or provide some details on why that is the case?

mirkobiasini: There is a platform function which is used from 25. However, for 24, we need to use dotnet interop not to introduce breaking changes. This is due the fact that the new platform function was introduced in 24.1. Thus, it requires runtime version >= 13.1. This can be a "breaking change" to customer code. That's why we had to use dotnet for 24.

WaelAbuSeada: If it is only for 24, then it is fine. This should not go into master or any other branches after 24

WaelAbuSeada (approved):
Approving on the condition that this is only for 24

PR #2065, opened 2024-09-19
Add CLEANSCHEMA symbols

grobyns:
#### Summary 
CLEANSCHEMA symbols 

#### Work Item(s) 
Fixes [AB#549296](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/549296)


PredragMaricic (approved):

.github/workflows/VerifyAppChanges.yaml
```diff
@@ -0,0 +1,21 @@
...
+  VerifyAppChanges:
+    runs-on: windows-latest
+    name: Verify App Changes
+    steps:
+      - name: checkout
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.sha }}
+
+      - uses: microsoft/BCApps/.github/actions/VerifyAppChanges@features/549296
```
mazhelez: As a last commit on this PR

```suggestion
      - uses: microsoft/BCApps/.github/actions/VerifyAppChanges@main
```

.github/workflows/powershell.yaml
```diff
@@ -16,8 +12,11 @@ permissions:
   contents: read
 
 jobs:
-  build:
-    name: PSScriptAnalyzer
+  PSScriptAnalyzer:
+    # https://github.com/microsoft/action-psscriptanalyzer
```
mazhelez:
```suggestion
    # https://github.com/microsoft/psscriptanalyzer-action
```

.github/actions/VerifyAppChanges/action.yaml
```diff
@@ -0,0 +1,13 @@
+name: Verify App Changes
```
aholstrup1: Sounds a little bit broad. This string will show up in the PR so it might be a good idea to be more specific about what exactly this check does. E.g. "Test Preprocessor Symbols" or "Verify Preprocessor Symbols"

.github/actions/VerifyAppChanges/action.yaml
```diff
@@ -0,0 +1,13 @@
...
+author: Microsoft Corporation
+description: Verifies the changes on AL files changed in a PR
```
aholstrup1: This could be more specific as well

.github/workflows/VerifyAppChanges.yaml
```diff
@@ -0,0 +1,21 @@
...
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  VerifyAppChanges:
+    runs-on: windows-latest
+    name: Verify App Changes
+    steps:
+      - name: checkout
+        uses: actions/checkout@v4
```
aholstrup1: Use a pinned depedency :) 
```suggestion
        uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
```

.github/workflows/powershell.yaml
```diff
@@ -1,8 +1,4 @@
-# https://github.com/microsoft/action-psscriptanalyzer
-# For more information on PSScriptAnalyzer in general, see
-# https://github.com/PowerShell/PSScriptAnalyzer
-
-name: PSScriptAnalyzer
+name: 'PS Goodies'
```
aholstrup1: How about "Powershell tests" or something 😄

.github/actions/VerifyAppChanges/TestPreprocessorSymbols.ps1
```diff
@@ -0,0 +1,47 @@
...
+# CLEANSCHEMA is on a 5y cycle starting from 26
+if ($CurrentVersion -le 26) {
+    $schemaLowerBound = 15
+} else {
+    $schemaLowerBound = ([math]::Floor(($CurrentVersion - 2) / 5) * 5) - 2
+}
+
+# Define the preprocessor symbols to check for
+$symbolConfigs = @(
+    @{stem = "CLEAN"; lowerBound = ($CurrentVersion - 4); upperBound = $mainVersion},
```
aholstrup1: Why not use this list of CLEAN tags? 
https://github.com/microsoft/BCApps/blob/1a67cfd41f42a3a58efce7a3fa1c8c92993e5d08/.github/AL-Go-Settings.json#L12C1-L23C5

If we calculate it like this based on the current version, we might be blocked from doing version updates

.github/actions/VerifyAppChanges/TestPreprocessorSymbols.ps1
```diff
@@ -0,0 +1,47 @@
...
+$mainVersion = Get-MaxAllowedObsoleteVersion
+
+# Get the major build version from the current branch
+$currentVersion = (Get-ConfigValue -Key "repoVersion" -ConfigType AL-Go) -split '\.' | Select-Object -First 1
+
+# CLEANSCHEMA is on a 5y cycle starting from 26
+if ($CurrentVersion -le 26) {
+    $schemaLowerBound = 15
+} else {
+    $schemaLowerBound = ([math]::Floor(($CurrentVersion - 2) / 5) * 5) - 2
```
Groenbech96: Maybe a comment to express this

Groenbech96 (commented):
Looks good, will leave to engineering systems to approve.

.github/actions/VerifyAppChanges/TestPreprocessorSymbols.ps1
grobyns: Yes, that is kind of the point. There should be no old tags left behind because no one will look at them, and this ensure that. Take the current state as example: We have main = 26 so that is the upperbound, and 22 as the lowerbound. We just cleaned up all v23 tags... When you move to 27, 23 will become the lowerbound... If there is any tag 22 or lower left then that would fail... that also means we had 2 releases to clean it up... At that point things should fail so they get cleaned up.

aholstrup1: Personally, I'm not a fan of blocking app version updates because we forgot to remove some old code. Queues are typically blocked when we update the app version in main so we want to minimize the amount of time it takes before queues can be opened again. 

Couldn't you achieve the same thing by: 
* Updating the allowed CLEAN versions when you clean up code. 
* Running this check against the list of allowed CLEAN tags

That way you make sure we clean up old code without blocking app version updates.

Groenbech96 (approved):
LGTM

.github/actions/VerifyAppChanges/TestPreprocessorSymbols.ps1
grobyns: You will run into the exact same issue because now you rely on a human to update the list... and the human will forget. It doesn't solve the problem.

aholstrup1 (dismissed):

aholstrup1 (approved):

PredragMaricic (approved):

PR #2104, opened 2024-09-25
Fix getting app ID for system objects

blrobl:
#### Summary 
GetAppIdForObject wouldn't work on System objects, since it's not on the "NAV App Installed App" record.

#### Work Item(s) 
Fixes [AB#548696](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/548696)

### DIFF
--- a/src/System Application/App/VS Code Integration/src/VsCodeIntegrationImpl.Codeunit.al
+++ b/src/System Application/App/VS Code Integration/src/VsCodeIntegrationImpl.Codeunit.al
@@ -11,7 +11,6 @@ codeunit 8333 "VS Code Integration Impl."
     InherentPermissions = X;
 
     var
-        AllObjWithCaption: Record AllObjWithCaption;
         UriBuilder: Codeunit "Uri Builder";
         VSCodeRequestHelper: DotNet VSCodeRequestHelper;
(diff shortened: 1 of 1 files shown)


src/System Application/App/VS Code Integration/src/VsCodeIntegrationImpl.Codeunit.al
```diff
@@ -120,7 +120,12 @@ codeunit 8333 "VS Code Integration Impl."
     local procedure GetAppIdForObject(ObjectType: Option; ObjectId: Integer): Text
     var
         NavAppInstalledApp: Record "NAV App Installed App";
+        EmptyGuid: Guid;
     begin
+        // Objects in the system app range
+        if ObjectId >= 2000000000 then
+            exit(EmptyGuid);
+
         if AllObjWithCaption.ReadPermission() then begin
```
grobyns, line 131: unrelated to this pr but AllObjWithCaption should really be a local... and with it you should skip the .Reset().
Right now there's no issue because the variable is only used here and higher up as a dummy... but that won't last... someone will abuse it...

grobyns (dismissed):

grobyns (approved):

PR #2108, opened 2024-09-26
[AI Test Tool] Add support for multiple iterations of suite

stkillen:
#### Summary 
Allow multiple iterations of the same suite in AI Test Tool. New action "Batch Run" allows specifying number of iterations.

#### Work Item(s) 
Fixes [AB#543079](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/543079)

### DIFF
--- a/src/Tools/AI Test Toolkit/src/TestSuite/AITBatchRunDialog.Page.al
+++ b/src/Tools/AI Test Toolkit/src/TestSuite/AITBatchRunDialog.Page.al
@@ -1,0 +1,45 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.TestTools.AITestToolkit;
+
+page 149043 "AIT Batch Run Dialog"
+{
+    Caption = 'Batch Run';
+    PageType = StandardDialog;
+    ApplicationArea = All;
+    UsageCategory = None;
+
+    layout
+    {
+        area(content)
+        {
+            group(Control1)
+            {
+                ShowCaption = false;
+
+                field(Iterations; Iterations)
+                {
+                    ApplicationArea = All;
+                    Caption = 'Number of iterations';
+                    ToolTip = 'Specifies the number of iterations to run.';
+                }
+            }
+        }
+    }
+
+    var
+        Iterations: Integer;
+
+    trigger OnInit()
+    begin
+        Iterations := 1;
+    end;
+
+    procedure GetNumberOfIterations(): Integer
+    begin
+        exit(Iterations);
+    end;
+}
--- a/src/Tools/AI Test Toolkit/src/TestSuite/AITTestSuite.Page.al
+++ b/src/Tools/AI Test Toolkit/src/TestSuite/AITTestSuite.Page.al
@@ -156,6 +156,30 @@ page 149031 "AIT Test Suite"
                     CurrPage.Update(false);
                 end;
             }
+            action(StartBatch)
+            {
+                Enabled = Rec.Status <> Rec.Status::Running;
+                Caption = 'Start Batch';
+                Image = ExecuteBatch;
+                ToolTip = 'Starts running the AI Test Suite, the specified number of times.';
+
+                trigger OnAction()
+                var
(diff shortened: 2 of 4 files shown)


src/Tools/AI Test Toolkit/src/TestSuite/AITBatchRunDialog.Page.al
```diff
@@ -0,0 +1,45 @@
...
+                    ApplicationArea = All;
+                    Caption = 'Number of iterations';
+                    ToolTip = 'Specifies the number of iterations to run.';
+                }
+            }
+        }
+    }
+
+    var
+        Iterations: Integer;
```
Groenbech96, line 34: How is this incremented?

src/Tools/AI Test Toolkit/src/TestSuite/AITTestSuite.Page.al
```diff
@@ -156,6 +156,30 @@ page 149031 "AIT Test Suite"
...
+
+                trigger OnAction()
+                var
+                    AITBatchRunDialog: Page "AIT Batch Run Dialog";
+                    Iterations: Integer;
+                begin
+                    CurrPage.Update(false);
+
+                    AITBatchRunDialog.LookupMode := true;
+                    if not (AITBatchRunDialog.RunModal() = ACTION::LookupOK) then
```
Groenbech96: <>

and remove not

src/Tools/AI Test Toolkit/src/TestSuite/AITTestSuite.Page.al
```diff
@@ -156,6 +156,30 @@ page 149031 "AIT Test Suite"
...
+                Caption = 'Start Batch';
+                Image = ExecuteBatch;
+                ToolTip = 'Starts running the AI Test Suite, the specified number of times.';
+
+                trigger OnAction()
+                var
+                    AITBatchRunDialog: Page "AIT Batch Run Dialog";
+                    Iterations: Integer;
+                begin
+                    CurrPage.Update(false);
```
Groenbech96, line 171: Why the need to update?

src/Tools/AI Test Toolkit/src/TestSuite/AITBatchRunDialog.Page.al
stkillen: It's set on the page. Default is 1

src/Tools/AI Test Toolkit/src/TestSuite/AITTestSuite.Page.al
stkillen: The normal start action updates the page before and after executing suite, this action should mirror that action.

The last one is to update the results after the run.

This one is probably not so necessary but I'm thinking this update is to update the results on the page before running, so it shows most recent results until this execution finishes (which can take a while). Execution is in a separate thread, so results could have been updated since accessing the page.

src/Tools/AI Test Toolkit/src/TestSuite/AITTestSuite.Page.al
stkillen:
```suggestion
                    if AITBatchRunDialog.RunModal() <> ACTION::LookupOK then
```

src/Tools/AI Test Toolkit/src/TestSuite/AITBatchRunDialog.Page.al
Groenbech96: So we cant have more than 1?

src/Tools/AI Test Toolkit/src/TestSuite/AITTestSuite.Page.al
Groenbech96: I guess i can understand, but could user not just hit f5 ? If they wanted? 
Once something completes, the update on 180 would ensure to update it, if i understand correctly.

Groenbech96 (approved):
Minor follow up questions. NIT

src/Tools/AI Test Toolkit/src/TestSuite/AITBatchRunDialog.Page.al
stkillen: Yes, iterations is a field on the page, you can modify it. See the field "Iterations" above. When you open the page, it's set to 1, then you can change it to the number of iterations that you want to run.

darjoo (approved):

PR #2112, opened 2024-09-27
Fixes #1970 Increasing Image File Size Limit to 50MB

### LINKED ISSUE #1970: [Bug]: Reasoning for Image Codeunit being limited to 5MB?
### Describe the issue

[Originally started here](https://www.yammer.com/dynamicsnavdev/#/Threads/show?threadId=2959960207736832).

[The following limit is implemented](https://github.com/microsoft/BCApps/blob/f591cf2931ed2df68a8b8066129ae9ffb1b059e4/src/System%20Application/App/Image/src/ImageImpl.Codeunit.al#L295).

This prevents processing larger images that may need to be cropped or altered in some fashion. This means we now have to make a request to an external service, process the return, etc., greatly increasing overhead and generally being slower, "locking" BC for more time.

### Expected behavior

Have the limit be configurable, removed, or higher.

### Steps to reproduce

pass in image that is larger than 5mb and get hit with error

### Additional context

typically we process labels of some sort from an external service, e.g. ucc labels, and some services will return all images in a single image we then need to split and process to store and print correctly. Because of this limit, we are unable to use the built in BC tools and instead have to rely on external services.

### Additional note

There may be a legitimate reason for the 5MB limit. This is more a hope that it can be changed in the future.

mjmatthiesen:
#### Summary 

Increase the file size limit from 5MB to 50MB so more image processing can be done within BC rather than externally

#### Work Item(s) 
Fixes #1970

Fixes [AB#550942](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/550942)

### DIFF
--- a/src/System Application/App/Image/src/ImageImpl.Codeunit.al
+++ b/src/System Application/App/Image/src/ImageImpl.Codeunit.al
@@ -292,7 +292,7 @@ codeunit 3970 "Image Impl."
         Size, Width, Height : Integer;
         MaxSize: Integer;
     begin
-        MaxSize := 5242880; // 5MB
+        MaxSize := 52428800; // 50MB
         if TempBlob.Length() > MaxSize then begin
             Session.LogMessage('0000FMA', StrSubstNo(ImageTooLargeTxt, TempBlob.Length()), Verbosity::Normal, DataClassification::SystemMetadata, TelemetryScope::ExtensionPublisher, 'Category', ImageCatTxt);
             System.Clear(TempBlob);


mjmatthiesen:
@JesperSchulz Hi Jesper, how do I get the AB work item number it's currently failing on? I didn't see that mentioned in the contributing.md doc. Thanks

JesperSchulz:
> @JesperSchulz Hi Jesper, how do I get the AB work item number it's currently failing on? I didn't see that mentioned in the contributing.md doc. Thanks

It's either automatically added when an idea already exists, or it's added when the "linked" label is added. I've done the latter here 😊 We'll review this PR asap.

JesperSchulz (approved):

StefanSosic (approved):

JesperSchulz:
We're still discussing what the possible side-effects of this change could be. This change will for sure increase the memory footprint of the service where this is extensively used.

Aleyenda (approved):

PR #2143, opened 2024-10-01
Fetching user settings should return the company specified in the user personalization if specified

AndreasMoth:
Currently we are specifying that all users are in the current company, however we should only fall back to this if the user personalization does not specify otherwise.
This fixes an issue on users list where all users are shown to be in the current company.

Fixes [AB#537109](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/537109)

### DIFF
--- a/src/System Application/App/User Settings/src/UserSettingsImpl.Codeunit.al
+++ b/src/System Application/App/User Settings/src/UserSettingsImpl.Codeunit.al
@@ -131,7 +131,7 @@ codeunit 9175 "User Settings Impl."
         UserSettingsRec."Locale ID" := UserPersonalization."Locale ID";
         UserSettingsRec."Time Zone" := UserPersonalization."Time Zone";
 
-        if CompanyName() <> '' then
+        if (UserPersonalization.Company = '') and (CompanyName() <> '') then
             UserSettingsRec.Company := CopyStr(CompanyName(), 1, 30)
         else
             UserSettingsRec.Company := UserPersonalization.Company;


src/System Application/App/User Settings/src/UserSettingsImpl.Codeunit.al
```diff
@@ -131,7 +131,7 @@ codeunit 9175 "User Settings Impl."
         UserSettingsRec."Locale ID" := UserPersonalization."Locale ID";
         UserSettingsRec."Time Zone" := UserPersonalization."Time Zone";
 
-        if CompanyName() <> '' then
+        if (UserPersonalization.Company = '') and (CompanyName() <> '') then
             UserSettingsRec.Company := CopyStr(CompanyName(), 1, 30)
         else
             UserSettingsRec.Company := UserPersonalization.Company;
```
Aleyenda, line 137: We will hit this code if UserPersonalization.Company is blank BUT CompanyName() is not. Wouldn't that make UserSettingsRec.Company blank?

AndreasMoth: If personalization company is blank and company name is not, then we copy from CompanyName which is not blank (since a company name is missing from user personalization), otherwise we use the company already specified in user personalization which of course could be blank (I.E. both company name and user personalization company is blank).

Aleyenda (approved):

grobyns (approved):

PR #2186, opened 2024-10-09
Added access to attachment ID in Codeunit "EMail Message"

### LINKED ISSUE #2038: [Bug]: Missing access to the Id field of the attachment in the “Email Message Attachment” table via the EMailMessage codeunit (8904)
### Describe the issue

There is currently no way to determine the ID of an attachment that is attached to a mail. The ID is currently stored in the internal table 8904 “Email Message Attachment”. As the table is internal, you can only access the attachments via the codeunit 8904 “Email Message” and iterate over the attachments with the functions Attachments_First() and/or Attachments_Next(). There are other methods in the codeunit that can be used to access the name of the attachment, for example. Unfortunately, several attachments can have the same name so that they cannot be clearly identified.

Identification would be possible via the Id field of the “EMail Message Attachment” table. However, there is no function in the implementation codeunit 8904 that returns the Id. Such an addition would provide considerable added value in identifying the correct attachment of the mail.
I would like to add to the code so that this is possible in the future.

### Expected behavior

I would expect to have a method available via codeunit 8904 “EMail Message” that returns the ID of the current attachment.

### Steps to reproduce

As you can see in the screenshot, you can access some methods in the codeunit, but the ID of the attachment remains internal.

image

### Additional context

_No response_

SPinkow:
#### Summary 
Access to the attachment ID has been added in the ???Email message??? codeunit. In addition to the name and other properties of the mail, the attachment ID is also available via the ???EMail Message??? codeunit 

#### Work Item(s) 
Fixes #2038 

Fixes [AB#555473](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/555473)


SPinkow:
@microsoft-github-policy-service agree company="GOB Software & Systeme"

NKarolak (approved):

pri-kise (approved):

JesperSchulz (approved):

darjoo (approved):

PR #2190, opened 2024-10-11
Fix unguarded record lookup which fails choosing email scenario with no email account

AndreasMoth:
If you open email scenarios and click assign scenarios, it fails if there is no email account due to an unguarded findfirst check.

Fixes [AB#551605](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/551605)


Groenbech96 (approved):
Dont you want to throw an actual error? This will make it silent.

gggdttt (approved):

JesperSchulz (approved):

PR #2217, opened 2024-10-17
Add build metadata to Extension Settings page

blrobl:
#### Summary 
Surface an extension's build metadata Extension Settings page. More specifically, the system that compiled it and the URL where the package is located. These properties are defined in the app.json and/or stamped by some CI/CD pipelines, like AL-Go for Github:

image

#### Work Item(s) 
Fixes [AB#547675](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/547675)

### DIFF
--- a/src/System Application/App/Extension Management/src/ExtensionSettings.Page.al
+++ b/src/System Application/App/Extension Management/src/ExtensionSettings.Page.al
@@ -117,6 +117,26 @@ page 2511 "Extension Settings"
                     ToolTip = 'Specifies the commit ID of the source code for the current version of the project.';
                 }
             }
+
+            group("Build app metadata")
+            {
+                Caption = 'Build app metadata';
+                Visible = IsBuildInformationAvailable;
+
+                field(BuildBy; BuildBy)
+                {
+                    Caption = 'Build by';
+                    Editable = false;
+                    ToolTip = 'Specifies the name of the system that assisted in the compilation of this extension.';
+                }
+                field(BuildUrl; BuildUrl)
+                {
+                    Caption = 'Build URL';
+                    Editable = false;
+                    ToolTip = 'Specifies the URL of where the package can be found.';
+                    ExtendedDatatype = URL;
+                }
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/System Application/App/Extension Management/src/ExtensionSettings.Page.al
```diff
@@ -117,6 +117,26 @@ page 2511 "Extension Settings"
...
             }
+
+            group("Build app metadata")
+            {
+                Caption = 'Build app metadata';
+                Visible = IsBuildInformationAvailable;
+
+                field(BuildBy; BuildBy)
+                {
+                    Caption = 'Build by';
```
EmilDamsbo, line 128: I know this matches the name of the table field, but it's "_built_ by" XYZ.

src/System Application/App/Extension Management/src/ExtensionSettings.Page.al
```diff
@@ -117,6 +117,26 @@ page 2511 "Extension Settings"
...
+                    Editable = false;
+                    ToolTip = 'Specifies the name of the system that assisted in the compilation of this extension.';
+                }
+                field(BuildUrl; BuildUrl)
+                {
+                    Caption = 'Build URL';
+                    Editable = false;
+                    ToolTip = 'Specifies the URL of where the package can be found.';
```
EmilDamsbo:
```suggestion
                    ToolTip = 'Specifies the URL where the package can be found.';
```

EmilDamsbo (changes requested):

src/System Application/App/Extension Management/src/ExtensionSettings.Page.al
blrobl: This is surfacing the build properties in the app.json, that's why it's "buld" and not "built":

image

blrobl: Added to PR description to avoid confusion.

src/System Application/App/Extension Management/src/ExtensionSettings.Page.al
blrobl: Also updated descriptions to match app.json ones.

EmilDamsbo (approved):

thpeder (approved):

PR #2218, opened 2024-10-17
[AI Test Tool] Only show dialog when test lines exist

stkillen:
#### Summary 
Only show dialog asking to update lines when:
- Input dataset is changed
- Test lines exist

#### Work Item(s) 
Fixes [AB#555181](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/555181)

### DIFF
--- a/src/Tools/AI Test Toolkit/src/TestSuite/AITTestSuite.Table.al
+++ b/src/Tools/AI Test Toolkit/src/TestSuite/AITTestSuite.Table.al
@@ -46,25 +46,6 @@ table 149030 "AIT Test Suite"
             TableRelation = "Test Input Group".Code;
             ValidateTableRelation = true;
             ToolTip = 'Specifies the dataset to be used by the test suite.';
-
-            trigger OnValidate()
-            var
-                AITTestMethodLine: Record "AIT Test Method Line";
-            begin
-                if Rec."Input Dataset" = xRec."Input Dataset" then
-                    exit;
-
-                AITTestMethodLine.SetRange("Test Suite Code", Rec.Code);
-
-                if AITTestMethodLine.IsEmpty() then
-                    exit;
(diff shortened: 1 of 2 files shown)


t-prda (approved):

nikolakukrika (commented):
xRec is previous record on the page. If there is no page, xRec is always the same as Rec. This happens if you trigger modify from AL code. For these reasons it is not good.to use it outside of the page as code becomes unpredictable. This change looks wrong. The code should be moved to the page

src/Tools/AI Test Toolkit/src/TestSuite/AITTestSuite.Table.al
```diff
@@ -51,11 +51,18 @@ table 149030 "AIT Test Suite"
             var
                 AITTestMethodLine: Record "AIT Test Method Line";
             begin
+                if Rec."Input Dataset" = xRec."Input Dataset" then
```
nikolakukrika: This is not good if modified from code it will always be the same value. Double check this because it is validate trigger but I believe if started from AL code then it will have the same value. 

We should not use xRec outside of the page context. To fix if they are the same do the xRec.FiND and then if they are the same exit. Othewise run the code. Alternative is to delete the check.

nikolakukrika (approved):

src/Tools/AI Test Toolkit/src/TestSuite/AITTestSuite.Table.al
stkillen: Moved the OnValidate trigger to the page. Probably also make more sense only to trigger when modified on the page.

t-prda (approved):

darjoo (approved):

PR #2247, opened 2024-10-22
Add procedures to Get/Remove for UriBuilder Query Paramaters/Flags

### LINKED ISSUE #2248: [Bug]: It's not possible to remove/get query parameters in UriBuilder
### Describe the issue

It was already reported in the other repo, where the suggestion was to create a PR:
https://github.com/microsoft/ALAppExtensions/issues/16460

So I did: https://github.com/microsoft/BCApps/pull/2247
But I think the validation needs a ticket within the same repository, so I'm creating this one. I hope this is the right approach, out of the issue templates, a Bug seemed to be the most fitting since I couldn't find a Microsoft idea and this had already been accepted in the other repo. 

### Expected behavior

UriBuilder should be able to Get/Add/Replace/Remove the queryparameter flags/parameters, but currently it can only Add/Replace them.

### Steps to reproduce

Add parameter to UriBuilder instance.
The only way to remove the parameter is to Initialize it again with the Uri from scratch and make sure you don't add the parameter again.

### Additional context

_No response_

bjarkihall:
#### Summary 
UriBuilder procedures can only Add/Replace Query Flags/Parameters, while the internal procedure can parse them and set them, so I want to be able to use that to only Get/Remove from the collection.

There are different ways to implement this and many ways to test this, I just wanted to propose this approach which reuses what's already there for AddQueryFlag/Parameter, figuring we might need something like this for the upcoming HttpClient mocking: https://www.yammer.com/dynamicsnavdev/#/threads/show?threadId=3029127257333760 (why expose a special QueryParameter dictionary if we already have Uri/UriBuilder codeunits that are supposed to expose these structures for us?)

The new procedures added are:

```al
procedure RemoveQueryFlag(Flag: Text; DuplicateAction: Enum "Uri Query Duplicate Behaviour")
procedure RemoveQueryFlag(Flag: Text)
procedure RemoveQueryParameter(ParameterKey: Text; ParameterValue: Text; DuplicateAction: Enum "Uri Query Duplicate Behaviour")
procedure RemoveQueryParameter(ParameterKey: Text; ParameterValue: Text)
procedure RemoveQueryParameters()
procedure GetQueryFlags(): List of [Text]
procedure GetQueryParameters(): Dictionary of [Text, List of [Text]]
procedure GetQueryParameter(ParameterKey: Text): List of [Text]
```

#### Work Item(s) 
Fixes #2248 
Fixes [AB#555623](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/555623)

### DIFF
--- a/src/System Application/App/URI/src/UriBuilder.Codeunit.al
+++ b/src/System Application/App/URI/src/UriBuilder.Codeunit.al
@@ -224,9 +224,9 @@ codeunit 3061 "Uri Builder"
     end;
 
     /// <summary>
-    /// Removes a flag to the query string of this UriBuilder. In case the same query flag exists already, the action in <paramref name="DuplicateAction"/> is taken.
+    /// Removes a flag from the query string of this UriBuilder. In case the same query flag exists already, the action in <paramref name="DuplicateAction"/> is taken.
     /// </summary>
-    /// <param name="Flag">A flag to Remove to the query string of this UriBuilder. This value will be encoded before being Removed to the URI query string. Cannot be empty.</param>
+    /// <param name="Flag">A flag to Remove from the query string of this UriBuilder. This value will be encoded before being Removed to the URI query string. Cannot be empty.</param>
     /// <param name="DuplicateAction">Specifies which action to take if the flag specified already exist.</param>
     /// <error>If the provided <paramref name="Flag"/> is empty.</error>
     /// <error>If the provided <paramref name="DuplicateAction"/> is <c>"Throw Error"</c> and the flag does not exist in the URI.</error>
@@ -238,9 +238,9 @@ codeunit 3061 "Uri Builder"
     end;
 
     /// <summary>
-    /// Removes a flag to the query string of this UriBuilder. In case the same query flag exists already, only one occurrence is kept.
+    /// Removes a flag from the query string of this UriBuilder. In case the same query flag exists already, only one occurrence is kept.
     /// </summary>
-    /// <param name="Flag">A flag to Remove to the query string of this UriBuilder. This value will be encoded before being Removed to the URI query string. Cannot be empty.</param>
+    /// <param name="Flag">A flag to Remove from the query string of this UriBuilder. This value will be encoded before being Removed to the URI query string. Cannot be empty.</param>
     /// <error>If the provided <paramref name="Flag"/> is empty.</error>
     /// <remarks>This function could alter the order of the existing query string parts. For example, if the previous URL was "https://microsoft.com?foo=bar&amp;john=doe" and the new flag is "contoso", the result could be "https://microsoft.com?john=doe&amp;foo=bar&amp;contoso".</remarks>
     procedure RemoveQueryFlag(Flag: Text)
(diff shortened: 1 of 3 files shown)


bjarkihall:
> Issue #2248 is not valid. Please make sure you link an **issue** that exists, is **open** and is **approved**.

I guess we have an edge case for the contribution workflow, since I had an issue/approval for the idea in another repo. I hope this is okay?

JesperSchulz:
Let's takt this for spin 🥳

bjarkihall:
Thanks @JesperSchulz, I was thinking about feedback on the signatures and if the `DuplicateAction` enum should be accepted and if `ShouldRemove` parameter should simply be added to `AddQueryParameterInternal` and `AddQueryFlag`, instead of roughly cloning these 2 methods into 2 new procedures which would only focus on the removal?
From there, the summary-comments and tests can also be made more detailed, I just wanted to start with a quick draft and tried to match the code which was already there.

JesperSchulz (approved):
Looks good to me!

JesperSchulz:
There seems to be a small issue in the tests during compilation. Once addressed, I think we're good to go 🥳

bjarkihall:
@JesperSchulz sorry about that, I was trying to use the opportunity to get AL help from Github Copilot to write the extra test scenario for me since it was a simple case, since I was told it was really smart about that kind of scenario, but turns out it wasn't.
I've fixed the compilation error. I'm also a bit surprised I didn't get feedback in vscode either, but I hope everything's okay now.

JesperSchulz:
Let's give this another run! 🚀

bjarkihall:
@JesperSchulz I noticed the tests were failing, since `Uri.GetAbsoluteUri` implicitly adds a forward-slash before the query-parameters, and this was taken care of in some of the existing tests before this PR:
image

But judging by how hard it is to catch this little hardcoded detail (for us humans and apparently AI as well), I changed them and the new tests so the "Init-URI" is closer to the final "Expected-URI", so they both have the forward-slash in all scenarios.
It added a little bit of diff from lines I didn't intend to change originally.

Hopefully it's finally ready for the pipelines now, I didn't want to spend too much time on the autotests initially, in case the PR (or parts of it) would be rejected on first review.

JesperSchulz:
> @JesperSchulz I noticed the tests were failing, since `Uri.GetAbsoluteUri` implicitly adds a forward-slash before the query-parameters, and this was taken care of in some of the existing tests before this PR: image
> 
> But judging by how hard it is to catch this little hardcoded detail (for us humans and apparently AI as well), I changed them and the new tests so the "Init-URI" is closer to the final "Expected-URI", so they both have the forward-slash in all scenarios. It added a little bit of diff from lines I didn't intend to change originally.
> 
> Hopefully it's finally ready for the pipelines now, I didn't want to spend too much time on the autotests initially, in case the PR (or parts of it) would be rejected on first review.

That's a fine strategy 😊 I don't mind at all taking a few turns to get things right. Let's try again and hope this time it passes 😎

JesperSchulz (approved):

nikolakukrika (approved):

PR #2275, opened 2024-10-29
[Email] Filtering for email retrievals

darjoo:
#### Summary 
- Adds a new codeunit "Email Retrieval Filters" for specifying the filters to be used when retrieving an email.
- Update APIs to work with both v2 and v3 interfaces
- Obsolete the v2 interface as it's the same just without the filtering.
- Update tests

#### Work Item(s) 
Fixes [AB#550315](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/550315)


src/System Application/App/Email/src/Connector/EmailRetrievalFilters.Codeunit.al
```diff
@@ -0,0 +1,109 @@
+namespace System.Email;
+
+/// <summary>
+/// This codeunit is used to store the filters for email connectors to use when retrieving emails.
+/// </summary>
+/// <remarks>
+/// Not all email connectors may use all the filters.
+/// </remarks>
+codeunit 4534 "Email Retrieval Filters"
```
nikolakukrika: It would be easier if this was a table subtype temporary. You could have the default values and could skip calling Initialized. It is hard to read. it would also be more extensible.

nikolakukrika (dismissed):

src/System Application/App/Email/src/Connector/EmailRetrievalFilters.Codeunit.al
darjoo: Initial design was a temporary table, but different from what you suggested from internal comms. 
Updated to use temp table which allows better extensibility now. 
Thanks!

src/System Application/App/Email/src/Email/EmailImpl.Codeunit.al
```diff
@@ -236,37 +260,76 @@ codeunit 8900 "Email Impl"
...
+#pragma warning restore AL0432
+#endif
 
-        exit(true);
+        Error(EmailconnectorDoesNotSupportReplyingErr);
     end;
-
+#if not CLEAN26
+#pragma warning disable AL0432
+    [Obsolete('Connector v2 is replaced by v3.', '26.0')]
```
NKarolak: Could you replace the obsolete message with an instruction what to use instead?

```suggestion
    [Obsolete('Replaced by CheckAndGetEmailConnectorv3.', '26.0')]
```

src/System Application/App/Email/src/Connector/EmailConnectorv2.Interface.al
```diff
@@ -10,6 +11,10 @@ namespace System.Email;
 /// </summary>
 interface "Email Connector v2" extends "Email Connector"
 {
+    ObsoleteReason = 'Replaced by EmailConnectorv3 which adds filtering for retrieval of emails';
```
NKarolak: I believe this is correct name:

```suggestion
    ObsoleteReason = 'Replaced by "Email Connector v3" which adds filtering for retrieval of emails';
```

src/System Application/App/Email/src/Email/Email.Codeunit.al
```diff
@@ -253,17 +253,30 @@ codeunit 8901 Email
...
     #region RetrieveEmails
-
+#if not CLEAN26
     /// <summary>
     /// Retrieves emails from the email account.
     /// </summary>
     /// <param name="EmailAccountId">The ID of the email account to use for sending the email.</param>
     /// <param name="EmailConnector">The email connector to use for sending the email.</param>
     /// <param name="EmailInbox">The return record of all new emails that were retrieved.</param>
+    [Obsolete('Replaced by RetrieveEmails with an additional Filter parameter.', '26.0')]
```
NKarolak:
```suggestion
    [Obsolete('Replaced by RetrieveEmails with an additional Filters parameter.', '26.0')]
```

src/System Application/App/Email/src/Connector/DefaultEmailConnectorv2.Codeunit.al
```diff
@@ -50,10 +57,18 @@ codeunit 8998 "Default Email Connector v2" implements "Email Connector v2"
 
     end;
 
+#if not CLEAN26
+    [Obsolete('Replaced by RetrieveEmails with an additional Filter parameter.', '26.0')]
```
NKarolak:
```suggestion
    [Obsolete('Replaced by RetrieveEmails with an additional Filters parameter.', '26.0')]
```

src/System Application/App/Email/src/Email/Inbox/EmailRetrievalFilters.Table.al
```diff
@@ -0,0 +1,63 @@
...
+table 8885 "Email Retrieval Filters"
+{
+    Access = Public;
+    TableType = Temporary;
+
+    fields
+    {
+        field(1; Id; Integer)
+        {
+            DataClassification = SystemMetadata;
```
NKarolak: What about setting DataClassification only once instead, in the table properties?

src/System Application/Test/Email/src/EmailTest.Codeunit.al
```diff
@@ -1412,12 +1413,14 @@ codeunit 134685 "Email Test"
...
         // [Then] An error is thrown that the connector does not support this operation
-        asserterror Email.RetrieveEmails(EmailAccount."Account Id", EmailAccount.Connector, EmailInbox);
+        asserterror Email.RetrieveEmails(EmailAccount."Account Id", EmailAccount.Connector, EmailInbox, EmailRetrievalFilters);
         Assert.ExpectedError('The selected email connector does not support retrieving emails');
     end;
 
+#if not CLEAN26
     [Test]
-    procedure RetrieveEmails()
+    [Obsolete('v2 connector is replaced by v3 connector.', '26.0')]
```
NKarolak, line 1421: Same for `RetrieveEmailsFailv2()`: Why would you obsolete test functions?

NKarolak (changes requested):

src/System Application/Test/Email/src/EmailTest.Codeunit.al
darjoo: I obsoleted test functions because I want the test to stay and continue working until the interface is removed. Otherwise I'd be removing the test function now and if v2 were to fail, for any reason, we wouldn't know.

NKarolak: I don't quite understand. Of course, the test functions should stay until the interface is removed. You already have put the compiler directives (#if not CLEAN26) around your v2 test functions, so they will be removed (only) together with the v2 interface. That's all correct, but also already sufficient. 
The obsolete attribute itself is required only to "inform" dependent extensions. I cannot imagine a scenario where anybody calls your test (!) functions in their code?

So what is it exactly you achieve with setting the Obsolete attribute here?

NKarolak: Ah, hold on - does it mean you have started to use the AppSourceCop for test apps as well? Then you could not delete the public test function before obsoleting it earlier.

NKarolak (approved):

src/System Application/Test/Email/src/EmailTest.Codeunit.al
darjoo: Ah, I see what you mean. Absolutely, you're correct where the obsoletion is unnecessary. 
I just prefer to do it when I'm coding as it allowed me to immediately see where I have usages of the function and update those as well. There's no process etc requirements for the tests like you mentioned.

src/System Application/App/Email/src/Message/EmailMessageImpl.Codeunit.al
```diff
@@ -53,11 +53,9 @@ codeunit 8905 "Email Message Impl."
     end;
 
     procedure Create(ToRecipients: Text; Subject: Text; Body: Text; HtmlFormatted: Boolean)
-    var
-        EmptyList: List of [Text];
     begin
 #pragma warning disable AA0205
-        Create(EmptyList, Subject, Body, HtmlFormatted, false);
+        Create(ToRecipients, Subject, Body, HtmlFormatted, false);
```
grobyns, line 57: AA0205:
Variables must be initialized before usage.

No more variable so why is the pragma still here?

src/System Application/App/Email/src/Account/EmailAccounts.Page.al
```diff
@@ -402,7 +402,7 @@ page 8887 "Email Accounts"
         CurrPage.Update(false);
     end;
 
-    local procedure FilterToConnectorv2Accounts(var EmailAccounts: Record "Email Account")
+    local procedure FilterToConnectorv2v3Accounts(var EmailAccounts: Record "Email Account")
     var
         IConnector: Interface "Email Connector";
```
grobyns, line 407: can we please not add i to interfaces... we don't add c to codeunits nor t to tables...

JeremyVyska: While I get where that is coming from, because it's true we don't prefix p for Params, nor CU to Codeunits, there might be some mangling of examples.

See:  https://youtu.be/PHmFqehrPG4?si=Kg1o5s-kiImOt1_k&t=638

Examples provided in the Launch events on the Interface include it.

But, it's also causing challenges with Prefix style affixes - before or after the I?  Both look weird.

Maybe there can be a nerdy as heck PGI "Interface prefix, to I or not to I?" or just a Yammer thread about it

ajkauffmann: > can we please not add i to interfaces... we don't add c to codeunits nor t to tables...

It's very commonly used in DotNet. I would really like us to follow these naming conventions.
https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-classes-structs-and-interfaces.

In short: add I as a prefix for Interfaces, but not C before classes.
image

darjoo: For the Object name, I agree that we shouldn't prefix it with `I` as we always know it is of type Interface when we have to define it. 

For the variable/parameter name, I have no particular preference for or against. But the preference I have is that we stay consistent throughout. Having it makes it clearer that it is an interface and that it is not of some other type, but what value + or - does it bring to just add it for interfaces? If we don't do it for tables, pages, codeunits.. Why do we start for interfaces besides similarity to how DotNet does?

Also, I don't think DotNet has a naming convention regarding the the variable/parameter name. It only has it for the object name.

```
interface ITest {...}
class Tester : ITest {...}

void SomeFunc()
{
    ITest test = new Tester();
}
```
As far as I can tell, from limited searching, the variable name wouldn't contain `I` nor any prefix to represent the variable as an Interface. 

I will leave this thread conversation open for now while I continue trying to fix my build failures :(

grobyns: My 2 cents: AL is a language onto its own. Yes we can peek and steal from other languages things that have proven 
 to work but we shouldn't do things simply because 'they' do it... 

The naming should convey meaning and add value. In my opinion, the i doesn't add either. When i call Foo.Bar() does it matter that Foo is a codeunit and not an interface or the other way around? Would i write code differently or have a better understanding when i see iFoo.Bar()? If the i is really needed, I'd say the thing is poorly named. 
Now, there are cases (and more so in modules) where we have a table, a page, codeunit and an interface, maybe even an enum all with the same name. When you need more than one of those in the same method naming becomes tricky. But even then I'd still not go for the i but prefer a more explicit naming of the use of each of these... 

This is a matter of taste and convention. Partners are always free to choose which conventions they follow. For MS code I strongly advise against the i prefix in both the object name and the variable name, and that is the prevailing pattern in our code now.

ajkauffmann: I agree with "don't do things because they do them", but they have a reason to do so.

Because interfaces are different from concrete objects and can represent many other objects, I find it useful to see that at a glance when reading code. Marking it with an "I" tells me immediately that I'm dealing with a blueprint and not the real implementation. I believe that's why other languages have this naming convention and why it also makes sense for AL.

So, let's agree to disagree here. I'm also strongly against using spaces and slashes or dots in object names. But that's another story. Let's not hijack this thread for a discussion about naming conventions.

nikolakukrika (approved):

WaelAbuSeada (approved):

PR #2289, opened 2024-10-30
New Version number 24.8

ankosiorek:
#### Work Item(s) 
Fixes #555843


mazhelez (approved):

aholstrup1:
Replacing this PR with #2291. AL-Go had a bug in the Increment Version Number logic which caused this PR Build to fail. That bug has been fixed now and the PR Build for #2291 works.

PR #2290, opened 2024-10-30
Registering Summarize capability

MonicaPAhuja:
#### Summary 
Registering Summarize capabilty

#### Work Item(s) 
Fixes #548693

### DIFF
--- a/src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
+++ b/src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
@@ -277,7 +277,7 @@ codeunit 7774 "Copilot Capability Impl"
     begin
         CopilotCapability := Enum::"Copilot Capability".FromInteger(Capability);
 
-        if CopilotCapability = Enum::"Copilot Capability"::Chat then
+        if CopilotCapability in [Enum::"Copilot Capability"::Chat, Enum::"Copilot Capability"::Summarize] then
             Silent := true;
 
         IsEnabled := AzureOpenAI.IsEnabled(CopilotCapability, Silent);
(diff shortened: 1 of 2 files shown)


src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
```diff
@@ -278,6 +278,10 @@ codeunit 7774 "Copilot Capability Impl"
         CopilotCapability: Enum "Copilot Capability";
     begin
         CopilotCapability := Enum::"Copilot Capability".FromInteger(Capability);
+
+        if CopilotCapability in [Enum::"Copilot Capability"::Chat, Enum::"Copilot Capability"::Summarize] then
+            Silent := true;
+
```
encimita: Why?
The `silent` parameter bubbles up all the way to the top of the stack, so we should just make sure the caller sets it to true instead of hardcoding true here.

darjoo: I assume Summarize is a platform capability, and as such, when you're invoking the event, you can set the caller to `true` like Enrico mentioned. This event should not set it for individual features, because there are circumstances that App's subscriber cannot account for.

MonicaPAhuja: Right, I actually checked the PR for platform and silent is true.
https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_git/Platform-C]ore/pullrequest/192723?path=/Platform/Client/Web/Script/TS/src/Adapters/CopilotSummary/CopilotSummaryAdapter.tsx&version=GBmaster&line=81&lineEnd=81&lineStartColumn=1&lineEndColumn=138&type=2&lineStyle=plain&_a=files&iteration=32&base=0

encimita (approved):

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
encimita: nice :)

darjoo (approved):

PR #2295, opened 2024-11-01
Fail if secret is empty as it could be missing

WaelAbuSeada:
Summary
Throwing an error if keyvault secret does not exist or is stored empty in the KV, as in both situation it is a broken secret. Doing it from app as Platform does not fail if the secret does not exist. 

Fixes #
[AB#544481](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/544481)

### DIFF
--- a/src/System Application/App/Azure Key Vault/src/AzureKeyVaultImpl.Codeunit.al
+++ b/src/System Application/App/Azure Key Vault/src/AzureKeyVaultImpl.Codeunit.al
@@ -30,18 +30,25 @@ codeunit 2202 "Azure Key Vault Impl."
         IsKeyVaultClientInitialized: Boolean;
         AzureKeyVaultTxt: Label 'Azure Key Vault', Locked = true;
         CertificateInfoTxt: Label 'Successfully constructed certificate from secret %1. Certificate thumbprint %2', Locked = true;
+        MissingSecretErr: Label '%1 is either missing or empty', Comment = '%1 = Secret Name.';
 
     [NonDebuggable]
     procedure GetAzureKeyVaultSecret(SecretName: Text; var Secret: Text)
     begin
         // Gets the secret as a Text from the key vault, given a SecretName.
         Secret := GetSecretFromClient(SecretName);
+
+        if Secret = '' then
+            Error(MissingSecretErr);
     end;
 
     [NonDebuggable]
     procedure GetAzureKeyVaultSecret(SecretName: Text; var Secret: SecretText)
     begin
         Secret := GetSecretFromClient(SecretName);
+
+        if Secret.IsEmpty() then
+            Error(MissingSecretErr);
     end;
 
     [NonDebuggable]


mazhelez (dismissed):

src/System Application/App/Azure Key Vault/src/AzureKeyVaultImpl.Codeunit.al
```diff
@@ -30,18 +30,25 @@ codeunit 2202 "Azure Key Vault Impl."
         IsKeyVaultClientInitialized: Boolean;
         AzureKeyVaultTxt: Label 'Azure Key Vault', Locked = true;
         CertificateInfoTxt: Label 'Successfully constructed certificate from secret %1. Certificate thumbprint %2', Locked = true;
+        MissingSecretErr: Label '%1 is either missing or empty', Comment = '%1 = Secret Name.';
```
AndreasMoth: "The secret %1 is either missing or empty."? To make it a bit more readable. Especially in cases where the secret name may be confused for something else being missing.

src/System Application/App/Azure Key Vault/src/AzureKeyVaultImpl.Codeunit.al
```diff
@@ -30,18 +30,25 @@ codeunit 2202 "Azure Key Vault Impl."
...
 
     [NonDebuggable]
     procedure GetAzureKeyVaultSecret(SecretName: Text; var Secret: Text)
     begin
         // Gets the secret as a Text from the key vault, given a SecretName.
         Secret := GetSecretFromClient(SecretName);
+
+        if Secret = '' then
+            Error(MissingSecretErr, SecretName);
```
AndreasMoth, line 42: This error will break features that rely on an empty string being returned in certain cases, such as OpenAI which assumes the secret will be empty if disabled and then it will log a nice error to telemetry and show a nice error to the user rather than the secret being missing:
image

src/System Application/App/Azure Key Vault/src/AzureKeyVaultImpl.Codeunit.al
```diff
@@ -30,18 +30,25 @@ codeunit 2202 "Azure Key Vault Impl."
...
         CertificateInfoTxt: Label 'Successfully constructed certificate from secret %1. Certificate thumbprint %2', Locked = true;
+        MissingSecretErr: Label '%1 is either missing or empty', Comment = '%1 = Secret Name.';
 
     [NonDebuggable]
     procedure GetAzureKeyVaultSecret(SecretName: Text; var Secret: Text)
     begin
         // Gets the secret as a Text from the key vault, given a SecretName.
         Secret := GetSecretFromClient(SecretName);
+
+        if Secret = '' then
```
AndreasMoth: perhaps: "if Secret.Trim() = '' then"?
I see we do that in many places, I assume there's a reason for this. Also if we do this change, we can fix all those calls to not make this check anymore :)

src/System Application/Test/Azure Key Vault/src/AzureKeyVaultTest.Codeunit.al
```diff
@@ -95,11 +96,12 @@ codeunit 135212 "Azure Key Vault Test"
 
         // [GIVEN] A configured Azure Key Vault
         MockAzureKeyvaultSecretProvider := MockAzureKeyvaultSecretProvider.MockAzureKeyVaultSecretProvider();
+        MockAzureKeyvaultSecretProvider.AddSecretMapping('some-secret', 'SecretFromKeyVault');
         AzureKeyVaultTestLibrary.SetAzureKeyVaultSecretProvider(MockAzureKeyvaultSecretProvider);
-        AzureKeyVaultTestLibrary.ClearSecrets();
 
         // [WHEN] The key vault is called with an unknown key
         asserterror AzureKeyVault.GetAzureKeyVaultSecret('somekeythatdoesnotexist', Secret);
+        Assert.ExpectedError(StrSubstNo(MissingSecretErr, 'somekeythatdoesnotexist'));
```
AndreasMoth, line 104: You should also test a secret that does exist but is blank. After all that's the functionality you are adding above. This always failed after all. I'm actually surprised you now get the new error since the failure should be on GetSecretFromClient (just like it failed before..)

src/System Application/App/Azure Key Vault/src/AzureKeyVaultImpl.Codeunit.al
WaelAbuSeada: Fair

src/System Application/Test/Azure Key Vault/src/AzureKeyVaultTest.Codeunit.al
WaelAbuSeada: Funny enough, this is literally a blank scenario, as I said above platform would return an empty string for a non existing secret.

src/System Application/App/Azure Key Vault/src/AzureKeyVaultImpl.Codeunit.al
WaelAbuSeada: This is actually added because the platform fetch secrets in a try/catch, so the old way will never fail unless the secret was not in the allowed list. Since we are getting rid of this pattern, the easist way to have a reasonable pattern is to throw an exception if secret is empty/not found. Either way, your feature should break if that secret does not exist or empty. 

Trimming is not a bad idea, let me add it.

gggdttt (approved):

src/System Application/App/Azure Key Vault/src/AzureKeyVaultImpl.Codeunit.al
```diff
@@ -30,18 +30,25 @@ codeunit 2202 "Azure Key Vault Impl."
...
+        if Secret.Trim() = '' then
+            Error(MissingSecretErr, SecretName);
     end;
 
     [NonDebuggable]
     procedure GetAzureKeyVaultSecret(SecretName: Text; var Secret: SecretText)
     begin
         Secret := GetSecretFromClient(SecretName);
+
+        if Secret.IsEmpty() then
```
AndreasMoth, line 50: Do you also want to trim this secret?

src/System Application/App/Azure Key Vault/src/AzureKeyVaultImpl.Codeunit.al
```diff
@@ -30,18 +30,25 @@ codeunit 2202 "Azure Key Vault Impl."
...
+        MissingSecretErr: Label 'The secret %1 is either missing or empty.', Comment = '%1 = Secret Name.';
 
     [NonDebuggable]
     procedure GetAzureKeyVaultSecret(SecretName: Text; var Secret: Text)
     begin
         // Gets the secret as a Text from the key vault, given a SecretName.
         Secret := GetSecretFromClient(SecretName);
+
+        if Secret.Trim() = '' then
+            Error(MissingSecretErr, SecretName);
```
AndreasMoth, line 42: Feel free to won't fix and close the PR, just wondering if you want to fix the references that also does this trim check. Now they are no longer needed and we can clean those up. Ex. in AI module:
image

AndreasMoth (approved):

src/System Application/App/Azure Key Vault/src/AzureKeyVaultImpl.Codeunit.al
gggdttt: 😮 Can secret contain space? （Just interested in this)

WaelAbuSeada: No

src/System Application/App/Azure Key Vault/src/AzureKeyVaultImpl.Codeunit.al
WaelAbuSeada: I think we can keep this for now as well

haoranpb (approved):

mazhelez (approved):

PR #2303, opened 2024-11-04
Fix GetTableUrl for DataPerCompany=false

### LINKED ISSUE #2262: [Bug]: GetTableUrl in TableInformation Cache Impl. is wrong for Table with DataPerCompany=false (e.g. Tenant Media)
### Describe the issue

The `Company Name` of cross company tables is populated with the string `(Cross-Company Data)`.
The problems is that the procedure `GetTableUrl` uses this placeholder string as  company name.

### Expected behavior

The url should be constructed  with the current company instead of the `(Cross-Company Data)` string.

### Steps to reproduce

Open the TableInformation Page and click on the a table that isn't company specific e.g. Tenant Media.

### Additional context

_No response_

pri-kise:
#### Summary 
Fix GetTableUrl for Tables with `DataPerCompany=false`

#### Work Item(s)
Fixes #2262

Fixes [AB#557526](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/557526)

### DIFF
--- a/src/System Application/App/Table Information/src/TableInformationCacheImpl.Codeunit.al
+++ b/src/System Application/App/Table Information/src/TableInformationCacheImpl.Codeunit.al
@@ -145,12 +145,9 @@ codeunit 8700 "Table Information Cache Impl."
     end;
 
     procedure GetTableUrl(Company: Text; TableNo: Integer): Text
-    var
-        TableMetadata: Record "Table Metadata";
     begin
-        TableMetadata.Get(TableNo);
-        if not TableMetadata.DataPerCompany then
-            Company := CompanyName(); // use the current company for the URL for the cases when table is not per company
+        if Company = CrossCompanyDataLbl then
+            Company := ''; // Clear the Company text if it is the CrossCompanyDataLbl to avoid errors with GetUrl
 
         exit(GetUrl(ClientType::Web, Company, ObjectType::Table, TableNo));
     end;
--- a/src/System Application/App/Table Information/src/TableInformationCacheImpl.Codeunit.al
+++ b/src/System Application/App/Table Information/src/TableInformationCacheImpl.Codeunit.al
@@ -6,7 +6,6 @@
 namespace System.DataAdministration;
 
 using System.Environment;
-using System.Reflection;
 
 codeunit 8700 "Table Information Cache Impl."
 {
--- a/src/System Application/App/Table Information/src/TableInformationCacheImpl.Codeunit.al
+++ b/src/System Application/App/Table Information/src/TableInformationCacheImpl.Codeunit.al
@@ -144,8 +144,11 @@ codeunit 8700 "Table Information Cache Impl."
     end;
 
     procedure GetTableUrl(Company: Text; TableNo: Integer): Text
+    var
+        TableMetadata: Record "Table Metadata";
     begin
-        if Company = '' then
+        TableMetadata.Get(TableNo);
+        if not TableMetadata.DataPerCompany then
             Company := CompanyName(); // use the current company for the URL for the cases when table is not per company
 
         exit(GetUrl(ClientType::Web, Company, ObjectType::Table, TableNo));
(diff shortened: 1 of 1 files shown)


JesperSchulz (dismissed):

src/System Application/App/Table Information/src/TableInformationCacheImpl.Codeunit.al
```diff
@@ -144,8 +145,11 @@ codeunit 8700 "Table Information Cache Impl."
     end;
 
     procedure GetTableUrl(Company: Text; TableNo: Integer): Text
+    var
+        TableMetadata: Record "Table Metadata";
     begin
-        if Company = '' then
+        TableMetadata.Get(TableNo);
+        if not TableMetadata.DataPerCompany then
```
nikolakukrika: What if Company is blank and the data is per company? Why cannot we always replace the blank company name with the current company?

This will also slow down the code since we will load the table metadata, which can be slow on the first call. I'm not sure how slow it is after the first call.

nikolakukrika (changes requested):

src/System Application/App/Table Information/src/TableInformationCacheImpl.Codeunit.al
pri-kise: This won't help.

The only thing that would help is to try a Company.Get(Company) to know if the `Company` text that was passed to the procedure is a valid company name.

I could then remove the check for the DataPerCompany. 

```al
if not CompanyRecord.Get(Company) then 
    Company := CompanyName(); 
```

nikolakukrika: @pri-kise If the company exist or not is a caller's responsibility to determine. This code should not fix it. If you ask for a non-existing company URL, you should get what you ask. 

If we have a problem with blank company name, why cannot we replace it with current company? We know that it exists.

Also keep in mind, we can open a bug for a platform to fix GetURL if we have issues with it. 
The system could easily be generating the links to the external environment - we generate the first part and then the part is replaced. We need to be careful with our assumptions.

pri-kise: The problem is not the Blank Company Name.

The problem is Company with the value of the label.
```al
CrossCompanyDataLbl: Label '(Cross-Company Data)'
```

I can change the code to check for the label `CrossCompanyDataLbl`.
....

nikolakukrika: I would say if the table is a cross-company table, we should replace this string with current company name. We know that the current company exists. String comparison should be rather fast.

grobyns: @pri-kise @nikolakukrika 

The procedure is used from 2 places. One is the 'real' Table Information (i.e. platform table) and the other is from the Table Information Cache (i.e. application table).
The code copies the 'real' thing into the cache and then has this:
```
                TableInformationCache.TransferFields(TableInformation);
                if TableInformationCache."Company Name" = '' then
                    TableInformationCache."Company Name" := CrossCompanyDataLbl;
                TableInformationCache.Insert();
```

In the 'real' table the Company Name would be blank. Does the url not work when given a blank? 
If I remember correctly, the procedure was created to avoid duplicate code when I added the cache table. The Cross Company lbl was added just to be more clear for the user (blank apparently was not obvious to users).

I would expect the url to work with blank so maybe instead of adding Company Name we should set it to blank?

nikolakukrika: @grobyns If the company name is blank, you will get the company defined in your user personalization. This could be OK if you want per-database data. So either omitting the company name or using the current company name should work IMO. Blank is probably better.

pri-kise: I updated my code.
Now I change the Company Text back to an empty string if it equals to the CrossCompanyLbl

nikolakukrika (approved):

JesperSchulz (approved):

PR #2317, opened 2024-11-05
Discussion: Introduce company initialize in System Application

AndreasMoth:
I have created this PR to start a discussion of whether we should have company initialize in System Application and see how it might look like. We currently have apps with tables that require setup data per company. So far we insert data on the fly, but should we always do this check or should it be done as part of initializing each company?
Either way, let's see.

System Application examples are: "AAD Application Setup", "Setup Azure AD Mgt. Provider", "Register Company Signal".

### DIFF
--- a/src/System Application/App/Company Initialize/CompanyInitialize.Codeunit.al
+++ b/src/System Application/App/Company Initialize/CompanyInitialize.Codeunit.al
@@ -1,0 +1,19 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Environment.Configuration;
+
+codeunit 900 "Company Initialize"
+{
+    [IntegrationEvent(false, false)]
+    internal procedure InitializeCompanySetup()
+    begin
+    end;
+
+    [IntegrationEvent(false, false)]
+    internal procedure InitializeCompany()
+    begin
+    end;
+}
--- a/src/System Application/App/Company Initialize/CompanyInitialize.Table.al
+++ b/src/System Application/App/Company Initialize/CompanyInitialize.Table.al
@@ -1,0 +1,33 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Environment.Configuration;
+
+table 900 "Company Initialize"
+{
+    DataClassification = SystemMetadata;
+
+    fields
+    {
+        field(1; "Primary Key"; Integer)
+        {
+        }
+        field(2; "Initialized Time"; DateTime)
+        {
+        }
+        field(3; "Initialized Version"; Code[20])
+        {
+        }
+    }
+
+    keys
+    {
+        key(Key1; "Primary Key")
+        {
+            Clustered = true;
+        }
+    }
+
+}
--- a/src/System Application/App/Company Initialize/CompanyInitializeImpl.Codeunit.al
+++ b/src/System Application/App/Company Initialize/CompanyInitializeImpl.Codeunit.al
@@ -1,0 +1,51 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Environment.Configuration;
+
+using System.Environment;
+
+codeunit 901 "Company Initialize Impl."
+{
+    Access = Internal;
+
+    procedure InitializeCompany()
+    var
+        CompanyInitialize: Record "Company Initialize";
+        CompanyInitializeCodeunit: Codeunit "Company Initialize";
+        ModuleInfo: ModuleInfo;
+    begin
+        CompanyInitialize."Initialized Time" := CurrentDateTime();
+
+        if NavApp.GetCurrentModuleInfo(ModuleInfo) then
+            CompanyInitialize."Initialized Version" := format(ModuleInfo.AppVersion());
+        CompanyInitialize.Insert();
+
+        CompanyInitializeCodeunit.InitializeCompanySetup();
+
+        CompanyInitializeCodeunit.InitializeCompany();
+    end;
+
+    [EventSubscriber(ObjectType::Codeunit, Codeunit::"System Initialization", 'OnAfterLogin', '', false, false)]
+    local procedure OnLoginInitializeCompany()
+    var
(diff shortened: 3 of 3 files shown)


src/System Application/App/Company Initialize/CompanyInitialize.Table.al
```diff
@@ -0,0 +1,33 @@
...
+    {
+        field(1; "Primary Key"; Integer)
+        {
+        }
+        field(2; "Initialized Time"; DateTime)
+        {
+        }
+        field(3; "Initialized Version"; Code[20])
+        {
+        }
```
haoranpb, line 22: One thing I'd add here is a field: `IsCompanyInitialized` (or something like that)

If you look at some code in BaseApp, we are leveraging the existences of GLSetup to determine this

src/System Application/App/Company Initialize/CompanyInitializeImpl.Codeunit.al
```diff
@@ -0,0 +1,51 @@
...
+    begin
+        CompanyInitialize."Initialized Time" := CurrentDateTime();
+
+        if NavApp.GetCurrentModuleInfo(ModuleInfo) then
+            CompanyInitialize."Initialized Version" := format(ModuleInfo.AppVersion());
+        CompanyInitialize.Insert();
+
+        CompanyInitializeCodeunit.InitializeCompanySetup();
+
+        CompanyInitializeCodeunit.InitializeCompany();
```
NKarolak, line 28: What's the gain of raising two identical events (except for their name) in a row?
And shouldn't their names start with `On*`?

src/System Application/App/Company Initialize/CompanyInitializeImpl.Codeunit.al
```diff
@@ -0,0 +1,51 @@
...
+    end;
+
+    [EventSubscriber(ObjectType::Codeunit, Codeunit::"System Initialization", 'OnAfterLogin', '', false, false)]
```
NKarolak, line 31:
```suggestion
    [EventSubscriber(ObjectType::Codeunit, Codeunit::"System Initialization", OnAfterLogin, '', false, false)]
```

src/System Application/App/Company Initialize/CompanyInitialize.Codeunit.al
```diff
@@ -0,0 +1,19 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Environment.Configuration;
+
+codeunit 900 "Company Initialize"
```
nikolakukrika, line 8: This should be moved to the platform system events. Whenever a new company is created they should trigger the initialize company event.

src/System Application/App/Company Initialize/CompanyInitialize.Table.al
```diff
@@ -0,0 +1,33 @@
...
+
+namespace System.Environment.Configuration;
+
+table 900 "Company Initialize"
+{
+    DataClassification = SystemMetadata;
+
+    fields
+    {
+        field(1; "Primary Key"; Integer)
```
nikolakukrika, line 14: This should be moved to the company table so we track for each company.
Keep in mind that it is by design that we can run it multiple times.

src/System Application/App/Company Initialize/CompanyInitializeImpl.Codeunit.al
```diff
@@ -0,0 +1,51 @@
...
+    var
+        CompanyInitialize: Record "Company Initialize";
+        CompanyInitializeCodeunit: Codeunit "Company Initialize";
+        ModuleInfo: ModuleInfo;
+    begin
+        CompanyInitialize."Initialized Time" := CurrentDateTime();
+
+        if NavApp.GetCurrentModuleInfo(ModuleInfo) then
+            CompanyInitialize."Initialized Version" := format(ModuleInfo.AppVersion());
+        CompanyInitialize.Insert();
```
nikolakukrika, line 24: We can initialize company many times, the codeunit 2 must be safe to run many times.   This logic is also wrong, it should be moved after Initialize company not before. We could partially commit and keep the record.

src/System Application/App/Company Initialize/CompanyInitialize.Codeunit.al
```diff
@@ -0,0 +1,19 @@
...
+{
+    [IntegrationEvent(false, false)]
+    internal procedure InitializeCompanySetup()
```
nikolakukrika, line 11: I prefer On for the events, it increases readability

src/System Application/App/Company Initialize/CompanyInitializeImpl.Codeunit.al
```diff
@@ -0,0 +1,51 @@
...
+    [EventSubscriber(ObjectType::Codeunit, Codeunit::"System Initialization", 'OnAfterLogin', '', false, false)]
+    local procedure OnLoginInitializeCompany()
+    var
+        CompanyInitialize: Record "Company Initialize";
+        ClientTypeManagement: Codeunit "Client Type Management";
+    begin
+        if not GuiAllowed() then
+            exit;
+
+        if ClientTypeManagement.GetCurrentClientType() = ClientType::Background then
```
nikolakukrika, line 40: The most important check that needs to be refactored is the check that we have on the GLSetup.Get(). We should introduce concepts of force initialization and a better check if it was implemented or not.

src/System Application/App/Company Initialize/CompanyInitialize.Codeunit.al
nikolakukrika: This is a nice opportunity to improve the flows.

src/System Application/App/Company Initialize/CompanyInitializeImpl.Codeunit.al
AndreasMoth: True, they should start with On. My thought on two events is that we have a lot of setup tables that are standalone and doesn't depend on anything else being setup and then we have other tables that does depend on the initial setup tables being filled in. With this you always subscribe to the first event if you're independent of any other setup and second if you are dependent. Of course this does not account for third or fourth level dependencies and there are also many other ways to deal with this kind of problem. It's simply trying to help a very complicated problem a little bit.

src/System Application/App/Company Initialize/CompanyInitialize.Table.al
AndreasMoth: Yes, I was using the fact that the table is empty as it not being initialized. Do you have a suggestion for when we want to insert a record but state that the company has not been initialized?

src/System Application/App/Company Initialize/CompanyInitialize.Table.al
AndreasMoth: Can you explain why we want to run this multiple times?

JesperSchulz:
Closing for now, as this won't get any traction at this moment.

PR #2320, opened 2024-11-05
Add SetBaseUrl procedure to AFSFileClient for customizable base URL

### LINKED ISSUE #2319: [BC Idea]: SetBaseUrl in codeunit 8950 "AFS File Client"
### BC Idea Link

https://experience.dynamics.com/ideas/idea/?ideaid=7ea58001-9f9b-ef11-95f5-7c1e5245ba58

### Description

Similar to codeunit 9053 "ABS Blob Client", codeunit 8950 "AFS File Client" should allow to set a different base url in order to allow app proxies or other resources.

This idea involves the creation of a new SetBaseUrl function in codeunit 8950 "AFS File Client".

Internal work item: [AB#556933](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/556933)

domingosbarreto:
#### Summary 
Add SetBaseUrl procedure to AFSFileClient for customizable base URL

#### Work Item(s)
Fixes #2319

### DIFF
--- a/src/System Application/App/Azure File Services API/src/AFSFileClient.Codeunit.al
+++ b/src/System Application/App/Azure File Services API/src/AFSFileClient.Codeunit.al
@@ -45,6 +45,17 @@ codeunit 8950 "AFS File Client"
     end;
 
     /// <summary>
+    /// The base URL to use when constructing the final URI.
+    /// If not set, the base URL is https://%1.file.core.windows.net where %1 is the storage account name.
+    /// </summary>
+    /// <remarks>Use %1 as a placeholder for the storage account name.</remarks>
+    /// <param name="BaseUrl">A valid URL string</param>
+    procedure SetBaseUrl(BaseUrl: Text)
+    begin
+        AFSFileClientImpl.SetBaseUrl(BaseUrl);
+    end;
+
+    /// <summary>
     /// Creates a file in the file share.
     /// This does not fill in the file content, it only initializes the file.
     /// </summary>


src/System Application/App/Azure File Services API/src/AFSFileClient.Codeunit.al
```diff
@@ -44,6 +44,17 @@ codeunit 8950 "AFS File Client"
         AFSFileClientImpl.Initialize(StorageAccount, FileShare, '', Authorization, APIVersion);
     end;
 
+    /// <summary>
+    /// The base URL to use when constructing the final URI.
+    /// If not set, the base URL is https://%1.file.core.windows.net where %1 is the storage account name.
+    /// </summary>
+    /// <remarks>Use %1 as a placeholder for the storage account name.</remarks>
+    /// <param name="BaseUrl">A valid URL string</param>
+    procedure SetBaseUrl(BaseUrl: Text)
```
mazhelez, line 53: Isn't it better to have an override of `Initialize` that accepts a base URL?

domingosbarreto: Designed it this way to maintain the existing logic in the ABS module. Let me know if you still think it's better to add an override, and if so, whether we should also add it to the `Initialize` method in the ABS module.

mazhelez: Let's keep it as it is to align with the ABS module.

mazhelez (approved):

PR #2342, opened 2024-11-12
[PowerBI] Decouple addin settings, default "Show zoom bar" to true

encimita:
#### Summary 
The Power BI Embedded control add-in currently supports a bunch of different settings that change the embed experience making it richer and more customizable.
These settings are currently controlled with the `SetSettings` add-in function, which takes a bunch of booleans. This is of course not very readable nor very extensible.

This PR cleans up the code for these settings by having a different function to set up each setting. It also flips the default for the "Show Status Bar" setting to true; this is for accessibility reasons, as full-page reports might not be accessible or "navigable" without the zoom controls.

#### Work Item(s) 
Fixes [AB#556723](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/556723)


src/System Application/App/ControlAddIns/Resources/PowerBIManagement/js/PowerBIManagement.js
```diff
@@ -411,13 +430,6 @@ function CompileSettings() {
         settingsObject.background = models.BackgroundType.Transparent;
     }
 
-    if (_forceFitToPage) {
```
mynjj: I get that `forceFitToPage` option is removed, but it wasn't removed from the obsoleted procedure. So it's technically breaking (semantically) anyone using it without the grace period.

I'm ok with breaking if you are too :smile:, I'm just pointing it out.

mynjj (approved):

dawidwoz (approved):

src/System Application/App/ControlAddIns/Resources/PowerBIManagement/js/PowerBIManagement.js
encimita: Let's respect the setting as long as it's still settable.

mynjj (approved):

AndersLarsenMicrosoft (commented):
all good. but remember i am an AL developer

PR #2348, opened 2024-11-13
[Open in VS Code] Fix dependencies not found

blrobl:
#### Summary 
Find() wouldn't iterate through the rows, FindSet does.

#### Work Item(s) 
Fixes [AB#548696](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/548696)

### DIFF
--- a/src/System Application/App/VS Code Integration/src/VsCodeIntegrationImpl.Codeunit.al
+++ b/src/System Application/App/VS Code Integration/src/VsCodeIntegrationImpl.Codeunit.al
@@ -74,7 +74,7 @@ codeunit 8333 "VS Code Integration Impl."
     var
         DependencyList: TextBuilder;
     begin
-        if NavAppInstalledApp.FindFirst() then
+        if NavAppInstalledApp.FindSet() then
             repeat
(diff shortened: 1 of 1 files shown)


src/System Application/App/VS Code Integration/src/VsCodeIntegrationImpl.Codeunit.al
```diff
@@ -74,7 +74,7 @@ codeunit 8333 "VS Code Integration Impl."
     var
         DependencyList: TextBuilder;
     begin
-        if NavAppInstalledApp.Find() then
+        if NavAppInstalledApp.FindFirst() then
```
grobyns:
```suggestion
        if NavAppInstalledApp.FindSet() then
```

grobyns: FYI, this cop rule should have triggered: 
https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/analyzers/codecop-aa0233

I don't know why it did not...

BazookaMusic (approved):

PR #2361, opened 2024-11-14
No. Series: Ability to extend filters on finding No. Series Lines when getting new numbers

### LINKED ISSUE #1362: No. Series: Ability to extend filters on finding No. Series Lines when getting new numbers
### Describe the issue

In the current implementation it is not possible to set **additional filters** when matching No. Series Lines are filtered on getting new numbers, see procedure [GetNoSeriesLine()](https://github.com/microsoft/BCApps/blob/main/src/Business%20Foundation/App/NoSeries/src/Single/NoSeriesImpl.Codeunit.al#L150), codeunit 304 "No. Series - Impl.".

The only way to extend the filtering right now is by subscribing to IntegrationEvent OnAfterSetNoSeriesCurrentLineFilters(), called in procedure [SetNoSeriesCurrentLineFilters()](https://github.com/microsoft/BCApps/blob/main/src/Business%20Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al#L91), codeunit 305 "No. Series - Setup Impl.".
Obviously the No. Series Lines are filtered there for UI purposes when setting up new No. Series and not for getting new numbers as in GetNoSeriesLine().

Before the movement of No. Series module to Business Foundation filters applied to No. Series Lines could be extended by subscribing to OnNoSeriesLineFilterOnBeforeFindLast() called in [SetNoSeriesLineFilter()](https://github.com/microsoft/BCApps/blob/main/src/Business%20Foundation/App/NoSeries/src/Legacy/NoSeriesManagement.Codeunit.al#L501) of codeunit 396 NoSeriesManagement, which is now obsoleted.

The filtering in SetNoSeriesLineFilter() served both for UI and getting new numbers purposes, whereas in the new solution the filtering is done in separate places (codeunits 304 and 305).
Therefore OnAfterSetNoSeriesCurrentLineFilters() does not completely replace former OnNoSeriesLineFilterOnBeforeFindLast() - there is no ability to extend filters on getting new numbers as it was in the old solution.

### Expected behavior

There should be the possibility of extending filters applied to No. Series Lines in [GetNoSeriesLine()](https://github.com/microsoft/BCApps/blob/main/src/Business%20Foundation/App/NoSeries/src/Single/NoSeriesImpl.Codeunit.al#L150), codeunit 304 "No. Series - Impl.", as a partial replacement of former OnNoSeriesLineFilterOnBeforeFindLast() in order to **unbreak** current subscribers.

The new IntegrationEvents could be raised in the same manner as the obsoleted OnNoSeriesLineFilterOnBeforeFindLast() is currently raised. Suggested signature: **OnGetNoSeriesLineOnBeforeFindLast(var NoSeriesLine: Record "No. Series Line")**

### Steps to reproduce

Use case: No. Series Lines are filtered by department of current user.

Example:

```
[EventSubscriber(ObjectType::Codeunit, Codeunit::NoSeriesManagement, OnNoSeriesLineFilterOnBeforeFindLast, '', false, false)]
local procedure NoSeriesManagement_OnNoSeriesLineFilterOnBeforeFindLast(var NoSeriesLine: Record "No. Series Line")
begin
	NoSeriesLine.SetFilter("Department", '%1|%2', '', UserSetup.GetDepartmentCode());
end;
```

### Additional context

_No response_

PeterDurrer:
#### Summary
Add Event for No. Series

#### Work Item(s) 
Fixes #1362

Fixes [AB#558102](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/558102)

### DIFF
--- a/src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
+++ b/src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
@@ -15,7 +15,7 @@ codeunit 305 "No. Series - Setup Impl."
         NumberFormatErr: Label 'The number format in %1 must be the same as the number format in %2.', Comment = '%1=No. Series Code,%2=No. Series Code';
         UnIncrementableStringErr: Label 'The value in the %1 field must have a number so that we can assign the next number in the series.', Comment = '%1 = New Field Name';
         NumberLengthErr: Label 'The number %1 cannot be extended to more than 20 characters.', Comment = '%1=No.';
-        CodeFieldChangedErr: Label 'The filter on %1 was altered by an event subscriber. This is a programming error. Please contact your partner to resolve the issue.\Original %1: %2\Modified Filter: %3';
+        CodeFieldChangedErr: Label 'The filter on %1 was altered by an event subscriber. This is a programming error. Please contact your partner to resolve the issue.\Original %1: %2\Modified Filter: %3', Comment = '%1=NoSeriesLine.FieldCaption("Series Code") %2=Original filter Value of NoSeriesLine."Series Code" %3=New filter Value of NoSeriesLine."Series Code"';
 
     procedure SetImplementation(var NoSeries: Record "No. Series"; Implementation: Enum "No. Series Implementation")
     var
@@ -138,8 +138,8 @@ codeunit 305 "No. Series - Setup Impl."
 
     procedure SetNoSeriesLineFilters(var NoSeriesLine: Record "No. Series Line"; NoSeriesCode: Code[20]; StartingDate: Date)
     var
-        NoSeries: Codeunit "No. Series";
         NoSeriesLine2: Record "No. Series Line";
+        NoSeries: Codeunit "No. Series";
     begin
         NoSeriesLine2.SetCurrentKey("Series Code", "Starting Date");
         NoSeriesLine2.SetRange("Starting Date", 0D, StartingDate);
--- a/src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
+++ b/src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
@@ -15,6 +15,7 @@ codeunit 305 "No. Series - Setup Impl."
         NumberFormatErr: Label 'The number format in %1 must be the same as the number format in %2.', Comment = '%1=No. Series Code,%2=No. Series Code';
         UnIncrementableStringErr: Label 'The value in the %1 field must have a number so that we can assign the next number in the series.', Comment = '%1 = New Field Name';
         NumberLengthErr: Label 'The number %1 cannot be extended to more than 20 characters.', Comment = '%1=No.';
+        CodeFieldChangedErr: Label 'The filter on Series Code was altered by an event subscriber. This is a programming error. Please contact your partner to resolve the issue.\Original Series Code: %1\Modified Filter: %2';
 
     procedure SetImplementation(var NoSeries: Record "No. Series"; Implementation: Enum "No. Series Implementation")
     var
@@ -98,7 +99,7 @@ codeunit 305 "No. Series - Setup Impl."
 #endif
     begin
         NoSeriesLine.Reset();
-        GetNoSeriesLineFilters(NoSeriesLine, NoSeriesRec.Code, WorkDate());
+        SetNoSeriesLineFilters(NoSeriesLine, NoSeriesRec.Code, WorkDate());
 #if not CLEAN24
 #pragma warning disable AL0432
         NoSeriesManagement.RaiseObsoleteOnNoSeriesLineFilterOnBeforeFindLast(NoSeriesLine);
@@ -135,18 +136,17 @@ codeunit 305 "No. Series - Setup Impl."
         exit(NoSeriesSingle.MayProduceGaps());
     end;
 
-    local procedure GetNoSeriesLineFilters(var NoSeriesLine: Record "No. Series Line"; NoSeriesCode: Code[20]; StartingDate: Date)
+    internal procedure SetNoSeriesLineFilters(var NoSeriesLine: Record "No. Series Line"; NoSeriesCode: Code[20]; StartingDate: Date)
     var
         NoSeries: Codeunit "No. Series";
         NoSeriesLine2: Record "No. Series Line";
-        CodeFieldChangedErr: Label 'Change of Series Code Field is on this Record not allowed.';
     begin
         NoSeriesLine2.SetCurrentKey("Series Code", "Starting Date");
         NoSeriesLine2.SetRange("Starting Date", 0D, StartingDate);
         NoSeriesLine2.SetRange("Series Code", NoSeriesCode);
-        RaiseSetAdditionalNoSeriesLineFilters(NoSeriesLine2);
-        If NoSeriesLine2."Series Code" <> NoSeriesCode then
-            Error(CodeFieldChangedErr); // Extensions should never change the code field range, this is a bug that developers should know immediately.
+        NoSeries.OnSetNoSeriesLineFilters(NoSeriesLine2);
+        If NoSeriesLine2.GetFilter("Series Code") <> NoSeriesCode then
+            Error(CodeFieldChangedErr, NoSeriesCode, NoSeriesLine2.GetFilter("Series Code")); // Extensions should never change the code field range, this is a bug that developers should know immediately.
 
         NoSeriesLine.SetCurrentKey("Series Code", "Starting Date");
         NoSeriesLine.CopyFilters(NoSeriesLine2);
@@ -403,10 +403,5 @@ codeunit 305 "No. Series - Setup Impl."
         if Rec."Sequence Name" <> '' then
             if NumberSequence.Exists(Rec."Sequence Name") then
                 NumberSequence.Delete(Rec."Sequence Name");
-    end;
-
-    [IntegrationEvent(false, false)]
-    internal procedure RaiseSetAdditionalNoSeriesLineFilters(var NoSeriesLine: Record "No. Series Line");
-    begin
     end;
 }
--- a/src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
+++ b/src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
@@ -89,7 +89,7 @@ codeunit 305 "No. Series - Setup Impl."
         NoSeries.MarkedOnly(true);
     end;
 
-    procedure SetNoSeriesCurrentLineFilters(var NoSeriesRec: Record "No. Series"; var NoSeriesLine: Record "No. Series Line"; ResetForDrillDown: Boolean)
+    local procedure SetNoSeriesCurrentLineFilters(var NoSeriesRec: Record "No. Series"; var NoSeriesLine: Record "No. Series Line"; ResetForDrillDown: Boolean)
     var
         NoSeries: Codeunit "No. Series";
 #if not CLEAN24
--- a/src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
+++ b/src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
@@ -101,6 +101,7 @@ codeunit 305 "No. Series - Setup Impl."
         NoSeriesLine.SetCurrentKey("Series Code", "Starting Date");
         NoSeriesLine.SetRange("Series Code", NoSeriesRec.Code);
         NoSeriesLine.SetRange("Starting Date", 0D, WorkDate());
+        NoSeries.OnGetNoSeriesLineOnBeforeFindLast(NoSeriesLine);
 #if not CLEAN24
 #pragma warning disable AL0432
         NoSeriesManagement.RaiseObsoleteOnNoSeriesLineFilterOnBeforeFindLast(NoSeriesLine);
--- a/src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
+++ b/src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
@@ -136,7 +136,7 @@ codeunit 305 "No. Series - Setup Impl."
         exit(NoSeriesSingle.MayProduceGaps());
     end;
 
-    internal procedure SetNoSeriesLineFilters(var NoSeriesLine: Record "No. Series Line"; NoSeriesCode: Code[20]; StartingDate: Date)
+    procedure SetNoSeriesLineFilters(var NoSeriesLine: Record "No. Series Line"; NoSeriesCode: Code[20]; StartingDate: Date)
     var
         NoSeries: Codeunit "No. Series";
         NoSeriesLine2: Record "No. Series Line";
--- a/src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
+++ b/src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
@@ -88,7 +88,7 @@ codeunit 305 "No. Series - Setup Impl."
         NoSeries.MarkedOnly(true);
     end;
 
-    local procedure SetNoSeriesCurrentLineFilters(var NoSeriesRec: Record "No. Series"; var NoSeriesLine: Record "No. Series Line"; ResetForDrillDown: Boolean)
+    procedure SetNoSeriesCurrentLineFilters(var NoSeriesRec: Record "No. Series"; var NoSeriesLine: Record "No. Series Line"; ResetForDrillDown: Boolean)
     var
         NoSeries: Codeunit "No. Series";
 #if not CLEAN24
--- a/src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
+++ b/src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
@@ -15,7 +15,7 @@ codeunit 305 "No. Series - Setup Impl."
         NumberFormatErr: Label 'The number format in %1 must be the same as the number format in %2.', Comment = '%1=No. Series Code,%2=No. Series Code';
         UnIncrementableStringErr: Label 'The value in the %1 field must have a number so that we can assign the next number in the series.', Comment = '%1 = New Field Name';
         NumberLengthErr: Label 'The number %1 cannot be extended to more than 20 characters.', Comment = '%1=No.';
-        CodeFieldChangedErr: Label 'The filter on Series Code was altered by an event subscriber. This is a programming error. Please contact your partner to resolve the issue.\Original Series Code: %1\Modified Filter: %2';
+        CodeFieldChangedErr: Label 'The filter on %1 was altered by an event subscriber. This is a programming error. Please contact your partner to resolve the issue.\Original %1: %2\Modified Filter: %3';
 
     procedure SetImplementation(var NoSeries: Record "No. Series"; Implementation: Enum "No. Series Implementation")
     var
@@ -145,8 +145,8 @@ codeunit 305 "No. Series - Setup Impl."
         NoSeriesLine2.SetRange("Starting Date", 0D, StartingDate);
         NoSeriesLine2.SetRange("Series Code", NoSeriesCode);
         NoSeries.OnSetNoSeriesLineFilters(NoSeriesLine2);
-        If NoSeriesLine2.GetFilter("Series Code") <> NoSeriesCode then
-            Error(CodeFieldChangedErr, NoSeriesCode, NoSeriesLine2.GetFilter("Series Code")); // Extensions should never change the code field range, this is a bug that developers should know immediately.
+        if NoSeriesLine2.GetFilter("Series Code") <> NoSeriesCode then
+            Error(CodeFieldChangedErr, NoSeriesLine2.FieldCaption("Series Code"), NoSeriesCode, NoSeriesLine2.GetFilter("Series Code"));
 
         NoSeriesLine.SetCurrentKey("Series Code", "Starting Date");
         NoSeriesLine.CopyFilters(NoSeriesLine2);
--- a/src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
+++ b/src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
@@ -98,10 +98,7 @@ codeunit 305 "No. Series - Setup Impl."
 #endif
     begin
         NoSeriesLine.Reset();
-        NoSeriesLine.SetCurrentKey("Series Code", "Starting Date");
-        NoSeriesLine.SetRange("Series Code", NoSeriesRec.Code);
-        NoSeriesLine.SetRange("Starting Date", 0D, WorkDate());
-        NoSeries.OnGetNoSeriesLineOnBeforeFindLast(NoSeriesLine);
+        GetNoSeriesLineFilters(NoSeriesLine, NoSeriesRec.Code, WorkDate());
 #if not CLEAN24
 #pragma warning disable AL0432
         NoSeriesManagement.RaiseObsoleteOnNoSeriesLineFilterOnBeforeFindLast(NoSeriesLine);
@@ -136,6 +133,23 @@ codeunit 305 "No. Series - Setup Impl."
     begin
         NoSeriesSingle := NoSeriesLine.Implementation;
         exit(NoSeriesSingle.MayProduceGaps());
+    end;
+
+    local procedure GetNoSeriesLineFilters(var NoSeriesLine: Record "No. Series Line"; NoSeriesCode: Code[20]; StartingDate: Date)
+    var
+        NoSeries: Codeunit "No. Series";
+        NoSeriesLine2: Record "No. Series Line";
+        CodeFieldChangedErr: Label 'Change of Series Code Field is on this Record not allowed.';
+    begin
+        NoSeriesLine2.SetCurrentKey("Series Code", "Starting Date");
+        NoSeriesLine2.SetRange("Starting Date", 0D, StartingDate);
+        NoSeriesLine2.SetRange("Series Code", NoSeriesCode);
+        RaiseSetAdditionalNoSeriesLineFilters(NoSeriesLine2);
+        If NoSeriesLine2."Series Code" <> NoSeriesCode then
+            Error(CodeFieldChangedErr); // Extensions should never change the code field range, this is a bug that developers should know immediately.
+
+        NoSeriesLine.SetCurrentKey("Series Code", "Starting Date");
+        NoSeriesLine.CopyFilters(NoSeriesLine2);
     end;
 
     procedure CalculateOpen(NoSeriesLine: Record "No. Series Line"): Boolean
@@ -389,5 +403,10 @@ codeunit 305 "No. Series - Setup Impl."
         if Rec."Sequence Name" <> '' then
             if NumberSequence.Exists(Rec."Sequence Name") then
                 NumberSequence.Delete(Rec."Sequence Name");
+    end;
+
+    [IntegrationEvent(false, false)]
+    internal procedure RaiseSetAdditionalNoSeriesLineFilters(var NoSeriesLine: Record "No. Series Line");
+    begin
     end;
 }
--- a/src/Business Foundation/App/NoSeries/src/Single/NoSeries.Codeunit.al
+++ b/src/Business Foundation/App/NoSeries/src/Single/NoSeries.Codeunit.al
@@ -362,4 +362,13 @@ codeunit 310 "No. Series"
     internal procedure OnAfterSetNoSeriesCurrentLineFilters(NoSeries: Record "No. Series"; var NoSeriesLine: Record "No. Series Line"; IsDrillDown: Boolean);
     begin
     end;
+
+    /// <summary>
+    /// Use this event to change the filters set on the No. Series Line record. These filters are used when searching the No. Series.
+    /// </summary>
+    /// <param name="NoSeriesLine">The No. Series Line to set filters on.</param>
+    [IntegrationEvent(false, false)]
+    internal procedure OnGetNoSeriesLineOnBeforeFindLast(var NoSeriesLine: Record "No. Series Line");
+    begin
+    end;
 }
--- a/src/Business Foundation/App/NoSeries/src/Single/NoSeries.Codeunit.al
+++ b/src/Business Foundation/App/NoSeries/src/Single/NoSeries.Codeunit.al
@@ -364,11 +364,11 @@ codeunit 310 "No. Series"
     end;
 
     /// <summary>
-    /// Use this event to change the filters set on the No. Series Line record. These filters are used when searching the No. Series.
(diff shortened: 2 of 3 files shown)


PeterDurrer:
> @PeterDurrer please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > ```
> 
> Contributor License Agreement

src/Business Foundation/App/NoSeries/src/Single/NoSeriesImpl.Codeunit.al
```diff
@@ -189,6 +190,7 @@ codeunit 304 "No. Series - Impl."
         if (NoSeriesLine."Line No." <> 0) and (NoSeriesLine."Series Code" = NoSeriesCode) and (NoSeriesLine."Starting Date" = NoSeriesLine2."Starting Date") then begin
             NoSeriesLine.CopyFilters(NoSeriesLine2);
             NoSeriesLine.SetRange("Line No.", NoSeriesLine."Line No.");
+            NoSeries.OnGetNoSeriesLineOnBeforeFindLast(NoSeriesLine);
```
AndreasMoth: This is not needed as we copy all filters from NoSeriesLine2, including the filters you would have set in the previous call.

src/Business Foundation/App/NoSeries/src/Single/NoSeriesImpl.Codeunit.al
```diff
@@ -200,14 +202,17 @@ codeunit 304 "No. Series - Impl."
...
+        if not LineFound then begin     
+            NoSeries.OnGetNoSeriesLineOnBeforeFindLast(NoSeriesLine);       
             NoSeriesManagement.RaiseObsoleteOnNoSeriesLineFilterOnBeforeFindLast(NoSeriesLine);
             LineFound := NoSeriesLine.FindLast();
         end;
 #pragma warning restore AL0432
 #else
-        if not LineFound then
+        if not LineFound then begin
+            NoSeries.OnGetNoSeriesLineOnBeforeFindLast(NoSeriesLine);
```
AndreasMoth: The NoSeriesLine has copied filters from NoSeriesLine2 above so this should not be needed.

src/Business Foundation/App/NoSeries/src/Single/NoSeriesImpl.Codeunit.al
PeterDurrer: You are right, only the first one is needed, I had included it because the RaiseObsoleteOnNoSeriesLineFilterOnBeforeFindLast was present there.

src/Business Foundation/App/NoSeries/src/Single/NoSeriesImpl.Codeunit.al
PeterDurrer: You are right, only the first one is needed, I had included it because the RaiseObsoleteOnNoSeriesLineFilterOnBeforeFindLast was present there.

src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
```diff
@@ -101,6 +101,7 @@ codeunit 305 "No. Series - Setup Impl."
         NoSeriesLine.SetCurrentKey("Series Code", "Starting Date");
         NoSeriesLine.SetRange("Series Code", NoSeriesRec.Code);
```
AndreasMoth: I would extract these three lines into a function in the setup impl. codeunit like below. This allows anyone to set filters without potentially messing up with the record. Pure setting of filters which is what we need and they have access to what code and date we are trying to filter to.
GetNoSeriesLineFilters(var NoSeriesLine: Record "No. Series Line", code, date)
var
  var NoSeriesLine2: Record "No. Series Line"
begin
	NoSeriesLine2.SetRange("Starting Date", 0D, date);
	NoSeriesLine2.SetRange("Series Code", code)
	RaiseSetAdditionalNoSeriesLineFilters(NoSeriesLine2)
	If NoSeriesLine2.<filter for code field> <> code then
                Error(CodeFieldChangedErr); // Extensions should never change the code field range, this is a bug that developers should know immediately.
	NoSeriesLine.CopyFilters(NoSeriesLine2);
end;

AndreasMoth: After extracting this, I think it looks pretty good :)

PeterDurrer: I hope I understood you correctly about outsourcing the code, otherwise just let me know what the function should look like.

src/Business Foundation/App/NoSeries/src/Single/NoSeriesImpl.Codeunit.al
```diff
@@ -177,6 +177,7 @@ codeunit 304 "No. Series - Impl."
         NoSeriesLine2.SetCurrentKey("Series Code", "Starting Date");
         NoSeriesLine2.SetRange("Series Code", NoSeriesCode);
         NoSeriesLine2.SetRange("Starting Date", 0D, UsageDate);
+        NoSeries.OnGetNoSeriesLineOnBeforeFindLast(NoSeriesLine2);
```
AndreasMoth: This should also call into GetNoSeriesLineFilters

AndreasMoth (approved):
Looks good, just need the other event to call the new GetNoSeriesLineFilters procedure as well.

src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
```diff
@@ -137,6 +135,23 @@ codeunit 305 "No. Series - Setup Impl."
...
+    var
+        NoSeries: Codeunit "No. Series";
+        NoSeriesLine2: Record "No. Series Line";
+        CodeFieldChangedErr: Label 'Change of Series Code Field is on this Record not allowed.';
+    begin
+        NoSeriesLine2.SetCurrentKey("Series Code", "Starting Date");
+        NoSeriesLine2.SetRange("Starting Date", 0D, StartingDate);
+        NoSeriesLine2.SetRange("Series Code", NoSeriesCode);
+        RaiseSetAdditionalNoSeriesLineFilters(NoSeriesLine2);
+        If NoSeriesLine2."Series Code" <> NoSeriesCode then
```
AndreasMoth: "if NoSeriesLine2.GetFilter("Series Code") <> NoSeriesCode then" - that's what I meant. Do verify this is what GetFilter actually returns, I didn't test this completely :)

src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
```diff
@@ -88,7 +88,7 @@ codeunit 305 "No. Series - Setup Impl."
         NoSeries.MarkedOnly(true);
     end;
 
-    local procedure SetNoSeriesCurrentLineFilters(var NoSeriesRec: Record "No. Series"; var NoSeriesLine: Record "No. Series Line"; ResetForDrillDown: Boolean)
+    procedure SetNoSeriesCurrentLineFilters(var NoSeriesRec: Record "No. Series"; var NoSeriesLine: Record "No. Series Line"; ResetForDrillDown: Boolean)
```
AndreasMoth: This no longer needs to be public

src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
```diff
@@ -137,6 +135,23 @@ codeunit 305 "No. Series - Setup Impl."
         exit(NoSeriesSingle.MayProduceGaps());
     end;
 
+    local procedure GetNoSeriesLineFilters(var NoSeriesLine: Record "No. Series Line"; NoSeriesCode: Code[20]; StartingDate: Date)
```
grobyns:
```suggestion
    procedure SetNoSeriesLineFilters(var NoSeriesLine: Record "No. Series Line"; NoSeriesCode: Code[20]; StartingDate: Date)
```

src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
```diff
@@ -137,6 +135,23 @@ codeunit 305 "No. Series - Setup Impl."
...
+    var
+        NoSeries: Codeunit "No. Series";
+        NoSeriesLine2: Record "No. Series Line";
+        CodeFieldChangedErr: Label 'Change of Series Code Field is on this Record not allowed.';
```
grobyns: We add labels as global variables.

```suggestion
        CodeFieldChangedErr: Label 'The filter on Series Code was altered by an event subscriber. This is a programming error. Please contact your partner to resolve the issue.\Original Series Code: %1\Modified Filter: %2';
```

src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
```diff
@@ -137,6 +135,23 @@ codeunit 305 "No. Series - Setup Impl."
...
+    begin
+        NoSeriesLine2.SetCurrentKey("Series Code", "Starting Date");
+        NoSeriesLine2.SetRange("Starting Date", 0D, StartingDate);
+        NoSeriesLine2.SetRange("Series Code", NoSeriesCode);
+        RaiseSetAdditionalNoSeriesLineFilters(NoSeriesLine2);
```
grobyns:
```suggestion
        OnSetNoSeriesLineFilters(NoSeriesLine2);
```

src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
```diff
@@ -137,6 +135,23 @@ codeunit 305 "No. Series - Setup Impl."
...
+        If NoSeriesLine2."Series Code" <> NoSeriesCode then
+            Error(CodeFieldChangedErr); // Extensions should never change the code field range, this is a bug that developers should know immediately.
```
grobyns:
```suggestion
            Error(CodeFieldChangedErr, NoSeriesCode, NoSeriesLine2.GetFilter("Series Code"); // Extensions should never change the code field range, this is a bug that developers should know immediately.
```

src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
```diff
@@ -389,4 +404,9 @@ codeunit 305 "No. Series - Setup Impl."
             if NumberSequence.Exists(Rec."Sequence Name") then
                 NumberSequence.Delete(Rec."Sequence Name");
     end;
+
+    [IntegrationEvent(false, false)]
+    internal procedure RaiseSetAdditionalNoSeriesLineFilters(var NoSeriesLine: Record "No. Series Line");
```
grobyns: The event publisher should be added in NoSeries.Codeunit.al, not here

src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
grobyns:
```suggestion
        NoSeries.OnSetNoSeriesLineFilters(NoSeriesLine2);
```

src/Business Foundation/App/NoSeries/src/Single/NoSeries.Codeunit.al
```diff
@@ -362,4 +362,13 @@ codeunit 310 "No. Series"
     internal procedure OnAfterSetNoSeriesCurrentLineFilters(NoSeries: Record "No. Series"; var NoSeriesLine: Record "No. Series Line"; IsDrillDown: Boolean);
     begin
     end;
+
+    /// <summary>
+    /// Use this event to change the filters set on the No. Series Line record. These filters are used when searching the No. Series.
+    /// </summary>
+    /// <param name="NoSeriesLine">The No. Series Line to set filters on.</param>
+    [IntegrationEvent(false, false)]
+    internal procedure OnGetNoSeriesLineOnBeforeFindLast(var NoSeriesLine: Record "No. Series Line");
```
grobyns:
```suggestion
    internal procedure OnSetNoSeriesLineFilters(var NoSeriesLine: Record "No. Series Line");
```

src/Business Foundation/App/NoSeries/src/Single/NoSeries.Codeunit.al
```diff
@@ -362,4 +362,13 @@ codeunit 310 "No. Series"
     internal procedure OnAfterSetNoSeriesCurrentLineFilters(NoSeries: Record "No. Series"; var NoSeriesLine: Record "No. Series Line"; IsDrillDown: Boolean);
     begin
     end;
+
+    /// <summary>
+    /// Use this event to change the filters set on the No. Series Line record. These filters are used when searching the No. Series.
+    /// </summary>
+    /// <param name="NoSeriesLine">The No. Series Line to set filters on.</param>
```
grobyns, line 370:
```suggestion
    /// <remarks>Changing the filter on the "Series Code" field is not allowed and will result in an error.</remarks>
    /// <param name="NoSeriesLine">The No. Series Line to set filters on.</param>
```

src/Business Foundation/App/NoSeries/src/Single/NoSeries.Codeunit.al
```diff
@@ -362,4 +362,13 @@ codeunit 310 "No. Series"
     internal procedure OnAfterSetNoSeriesCurrentLineFilters(NoSeries: Record "No. Series"; var NoSeriesLine: Record "No. Series Line"; IsDrillDown: Boolean);
     begin
     end;
+
+    /// <summary>
+    /// Use this event to change the filters set on the No. Series Line record. These filters are used when searching the No. Series.
```
grobyns:
```suggestion
    /// Use this event to set additional filters on the No. Series Line record. These filters are used when searching the No. Series.
```

src/Business Foundation/App/NoSeries/src/Single/NoSeriesImpl.Codeunit.al
grobyns:
```suggestion
        NoSeriesSetupImpl.SetNoSeriesLineFilters(NoSeriesLine2, NoSeriesCode, UsageDate);
```

grobyns (changes requested):

src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
```diff
@@ -137,6 +136,22 @@ codeunit 305 "No. Series - Setup Impl."
         exit(NoSeriesSingle.MayProduceGaps());
     end;
 
+    internal procedure SetNoSeriesLineFilters(var NoSeriesLine: Record "No. Series Line"; NoSeriesCode: Code[20]; StartingDate: Date)
```
grobyns:
```suggestion
    procedure SetNoSeriesLineFilters(var NoSeriesLine: Record "No. Series Line"; NoSeriesCode: Code[20]; StartingDate: Date)
```

grobyns: No need to add internal here, the whole codeunit is internal

src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
```diff
@@ -88,7 +89,7 @@ codeunit 305 "No. Series - Setup Impl."
         NoSeries.MarkedOnly(true);
     end;
 
-    local procedure SetNoSeriesCurrentLineFilters(var NoSeriesRec: Record "No. Series"; var NoSeriesLine: Record "No. Series Line"; ResetForDrillDown: Boolean)
+    procedure SetNoSeriesCurrentLineFilters(var NoSeriesRec: Record "No. Series"; var NoSeriesLine: Record "No. Series Line"; ResetForDrillDown: Boolean)
```
grobyns:
```suggestion
    local procedure SetNoSeriesCurrentLineFilters(var NoSeriesRec: Record "No. Series"; var NoSeriesLine: Record "No. Series Line"; ResetForDrillDown: Boolean)
```

grobyns:
> @grobyns Can you please resolve your comments that are know fine for you. Then i can see what i must do to finish the work. Thank you.

done.

We've been discussing the need to add a test or two to verify the error is thrown when the filter on series code changes to ensure future code modifications don't remove that logic.

src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
```diff
@@ -15,6 +15,7 @@ codeunit 305 "No. Series - Setup Impl."
         NumberFormatErr: Label 'The number format in %1 must be the same as the number format in %2.', Comment = '%1=No. Series Code,%2=No. Series Code';
         UnIncrementableStringErr: Label 'The value in the %1 field must have a number so that we can assign the next number in the series.', Comment = '%1 = New Field Name';
         NumberLengthErr: Label 'The number %1 cannot be extended to more than 20 characters.', Comment = '%1=No.';
+        CodeFieldChangedErr: Label 'The filter on Series Code was altered by an event subscriber. This is a programming error. Please contact your partner to resolve the issue.\Original Series Code: %1\Modified Filter: %2';
```
NKarolak: What about replacing the "Series Code" part with another placeholder, to be filled by line 149 with the current field caption?

src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
```diff
@@ -137,6 +136,22 @@ codeunit 305 "No. Series - Setup Impl."
...
+    procedure SetNoSeriesLineFilters(var NoSeriesLine: Record "No. Series Line"; NoSeriesCode: Code[20]; StartingDate: Date)
+    var
+        NoSeries: Codeunit "No. Series";
+        NoSeriesLine2: Record "No. Series Line";
+    begin
+        NoSeriesLine2.SetCurrentKey("Series Code", "Starting Date");
+        NoSeriesLine2.SetRange("Starting Date", 0D, StartingDate);
+        NoSeriesLine2.SetRange("Series Code", NoSeriesCode);
+        NoSeries.OnSetNoSeriesLineFilters(NoSeriesLine2);
+        If NoSeriesLine2.GetFilter("Series Code") <> NoSeriesCode then
```
NKarolak:
```suggestion
        if NoSeriesLine2.GetFilter("Series Code") <> NoSeriesCode then
```

src/Business Foundation/App/NoSeries/src/Single/NoSeries.Codeunit.al
```diff
@@ -362,4 +362,14 @@ codeunit 310 "No. Series"
     internal procedure OnAfterSetNoSeriesCurrentLineFilters(NoSeries: Record "No. Series"; var NoSeriesLine: Record "No. Series Line"; IsDrillDown: Boolean);
     begin
     end;
+
+    /// <summary>
+    /// Use this event to set additional filters set on the No. Series Line record. These filters are used when searching the No. Series.
```
NKarolak:
```suggestion
    /// Use this event to set additional filters on the No. Series Line record. These filters are used when searching the No. Series.
```

src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
```diff
@@ -137,6 +136,22 @@ codeunit 305 "No. Series - Setup Impl."
...
+            Error(CodeFieldChangedErr, NoSeriesCode, NoSeriesLine2.GetFilter("Series Code")); // Extensions should never change the code field range, this is a bug that developers should know immediately.
```
NKarolak: I believe the comment can be removed, as the content of `CodeFieldChangedErr` actually explains it quite well.
```suggestion
            Error(CodeFieldChangedErr, NoSeriesCode, NoSeriesLine2.GetFilter("Series Code"));
```

NKarolak (changes requested):

src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
PeterDurrer: resolved in newest commit.

src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
PeterDurrer: resolved in newest commit.

src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
PeterDurrer: resolved in newest commit.

NKarolak (approved):

src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
grobyns: We could do that. Placeholders make things more complicated for translations. Something the order of words changes depending on the language and if you use placeholders you can't easily fix that. I don't know if that would be an issue here...

grobyns: Just thinking about Spanish or French for example, the translation would put 'original' after series code.  With this change, that is no longer possible.
@NKarolak @PeterDurrer

PeterDurrer: @grobyns Sorry i forget the comments to explain the Placeholders, its know done. I Think the Idea of @NKarolak is good, its better to use the fieldcaption so if the Caption of field is changed you can find the field in the Message otherwise you Searching for the field but it captions is other than in the error.

NKarolak: @grobyns: Is it because all translations are automatic only, and there is no productive way to interact?
Usually, I'd explain all placeholders in the comment. The (human) translators then change the order if required: e.g., from `SomeWord %1` to `%1 SomeWord`, depending on the context. We do that for our apps without any problems.

PeterDurrer: @grobyns OK, i understand you the result of the Translator in Bing is this:
image

but the Problem you don't have the control if the FieldCaption changed, with the FieldCaption Placeholder you only one time manualy correct the Translation and the Message is all time correct.

grobyns: @PeterDurrer @NKarolak 
The problem is that a translator (machine or human) needs to know that when translating 'Original %1: %2' to Spanish they need to actually use '%1 original: %2'

in this case it seems pretty static per language but we have other cases where the order could change based on the value of the placeholders. 
In general we try to avoid:
- composite sentences. Typically you get this when you have the same message multiple times with only small variations. It's better to repeat the whole thing as that makes for much easier translations.
- multiple successive placeholders (like foo %1 %2 %3 bar). Again, it's hard to know whether the order should change.
- when a placeholder can really have only 1 value, then we shouldn't use a placeholder. Here, the %1 would always be 'Series Code'. A comment can be added stating that Series Code is the caption of the "Series Code" field. It's no guarantee but at least the info is there for translators.

grobyns: In any case, these are recommendations. If they were 'laws' we'd have cop rules wouldn't we 🤔 ? I'll leave it to you guys to decide what you think is best.

PeterDurrer: Thanks for the interesting conversation. Finally, I'll show you the link to [Microsoft Best Practices for User Messages](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-progress-windows-message-error-and-confirm-methods#best-practices-for-user-messages). It says that it's better to use the field caption if you can. So I'll leave this version. But I can understand both inputs and will think about it in my next code.

grobyns: I think that comment is about using Name v. Caption... and of course you should use Caption.

AndreasMoth (dismissed):

grobyns (approved):

PeterDurrer:
> > @PeterDurrer please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> > ```
> > @microsoft-github-policy-service agree [company="{your company}"]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > > Options:
> > > 
> > > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > > 
> > > ```
> > > @microsoft-github-policy-service agree
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > > 
> > > ```
> > > @microsoft-github-policy-service agree company="Microsoft"
> > > ```
> > 
> > 
> > Contributor License Agreement
> 
> @microsoft-github-policy-service agree company="Logico Solutions AG"

src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
NKarolak: @PeterDurrer I've got feedback from our internal translators: They agree placeholders should be used only if multiple values are possible. My information was outdated - today, automatic translation tools / AI are used first, and they still cannot process the commented placeholders correctly (yet).
So: from the translational point of view, it's indeed better to return to your original draft. Sorry for the previous confusion.

src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
```diff
@@ -15,6 +15,7 @@ codeunit 305 "No. Series - Setup Impl."
         NumberFormatErr: Label 'The number format in %1 must be the same as the number format in %2.', Comment = '%1=No. Series Code,%2=No. Series Code';
         UnIncrementableStringErr: Label 'The value in the %1 field must have a number so that we can assign the next number in the series.', Comment = '%1 = New Field Name';
         NumberLengthErr: Label 'The number %1 cannot be extended to more than 20 characters.', Comment = '%1=No.';
+        CodeFieldChangedErr: Label 'The filter on %1 was altered by an event subscriber. This is a programming error. Please contact your partner to resolve the issue.\Original %1: %2\Modified Filter: %3';
```
AndreasMoth: As the error below shows, whenever you add a label with placeholder, you must add a comment specifying what those placeholders are. These are necessary for translation teams to translate the string better. This is required for all strings with placeholders because some string can be very difficult for a translator to understand when they almost fully consist of these placeholders..

src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al
```diff
@@ -137,6 +136,22 @@ codeunit 305 "No. Series - Setup Impl."
         exit(NoSeriesSingle.MayProduceGaps());
     end;
 
+    procedure SetNoSeriesLineFilters(var NoSeriesLine: Record "No. Series Line"; NoSeriesCode: Code[20]; StartingDate: Date)
+    var
+        NoSeries: Codeunit "No. Series";
+        NoSeriesLine2: Record "No. Series Line";
```
AndreasMoth, line 141: See errors below, basically the sorting of variables must be fixed, Record comes before codeunit.

grobyns (approved):

JesperSchulz (approved):

PeterDurrer:
Dear @JesperSchulz , where can I find out from which BC version this will be available? I saw that 26.0 was set under milestones, but that will still be quite a while away. How do I find out whether this will be available in the next CU?

dominicstarkl:
@PeterDurrer @grobyns @AndreasMoth Thanks for your great implementation of this extensibility request - it fully meets our customers requirements 👍

Would it be possible to change this procedure from local to **public**?
- Codeunit 305 "No. Series - Setup Impl.", [SetNoSeriesCurrentLineFilters()](https://github.com/microsoft/BCApps/blob/8d2a7032ba4a06fd3173da76fb2d5462af5c4339/src/Business%20Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al#L92)

It would help us to invoke the **already public** procedure [DrillDown()](https://github.com/microsoft/BCApps/blob/8d2a7032ba4a06fd3173da76fb2d5462af5c4339/src/Business%20Foundation/App/NoSeries/src/Setup/NoSeriesSetupImpl.Codeunit.al#L29) without automatically running the no. series page.
We want to let the user choose from a no. series entry (LookupMode).
If this procedure would be public we could avoid code copying.

Thanks in advance for your appreciated reply!

Regards
Dominic

grobyns:
@dominicstarkl I don't quite understand what you're trying to do. Also, the drilldown method you point to is not public, it's internal.

dominicstarkl:
@grobyns
Hi Gert

As far as I can see there is no access modifier in procedure DrillDown. Could you show me your source code?

We want to select a No. Series Line from the possible No. Series Lines that show when you drill down from setup..
In DrillDown procedure we cannot specify the page behavior. We need to set LookUp mode to a page variable and run the page in our own.

Of course we could copy the code but it would be more future proof when this procedure was public and not local as it is right now.
As I can see this could be done without any interferences.

What do you think about it?

Regards
Dominic

dominicstarkl:
@grobyns Now I have just seen that this whole Codeunit is internal. So we should place the public access somewhere else.

grobyns:
@dominicstarkl I will take a look at the ask. I'm not convinced it will work as you expect. 

Regardless, you should open a new issue for this. This pull request is closed and we'll need a new issue to handle this follow up case.

dominicstarkl:
@grobyns Thanks Gert, I'll have a look at it

PR #2383, opened 2024-11-19
Add Privacy Notice IDs to telemetry statements

DenLilleMand:
#### Summary 

It is not possible in the current privacy notice telemetry to know which logs relate to what privacy notice IDs, which limits its usefulness

#### Work Item(s) 
Fixes [AB#549916](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/549916/)

### DIFF
--- a/src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
+++ b/src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
@@ -28,7 +28,7 @@ codeunit 1565 "Privacy Notice Impl."
         PrivacyNoticeAutoApprovedByAdminTelemetryTxt: Label 'The privacy notice was auto-approved by the admin', Locked = true;
         PrivacyNoticeAutoRejectedByAdminTelemetryTxt: Label 'The privacy notice was auto-rejected by the admin', Locked = true;
         PrivacyNoticeAutoApprovedByUserTelemetryTxt: Label 'The privacy notice was auto-approved by the user', Locked = true;
-        ShowingPrivacyNoticeTelemetryTxt: Label 'Showing privacy notice', Locked = true;
+        ShowingPrivacyNoticeTelemetryTxt: Label 'Showing privacy notice %1', Locked = true;
         PrivacyNoticeApprovalResultTelemetryTxt: Label 'Approval State after showing privacy notice: %1', Locked = true;
         CheckPrivacyNoticeApprovalStateTelemetryTxt: Label 'Checking privacy approval state for privacy notice %1', Locked = true;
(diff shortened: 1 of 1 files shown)


src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
```diff
@@ -23,20 +23,20 @@ codeunit 1565 "Privacy Notice Impl."
...
-        ConfirmPrivacyNoticeTelemetryTxt: Label 'Confirming privacy notice', Locked = true;
-        PrivacyNoticeAutoApprovedByAdminTelemetryTxt: Label 'The privacy notice was auto-approved by the admin', Locked = true;
-        PrivacyNoticeAutoRejectedByAdminTelemetryTxt: Label 'The privacy notice was auto-rejected by the admin', Locked = true;
-        PrivacyNoticeAutoApprovedByUserTelemetryTxt: Label 'The privacy notice was auto-approved by the user', Locked = true;
-        ShowingPrivacyNoticeTelemetryTxt: Label 'Showing privacy notice', Locked = true;
-        PrivacyNoticeApprovalResultTelemetryTxt: Label 'Approval State after showing privacy notice: %1', Locked = true;
-        CheckPrivacyNoticeApprovalStateTelemetryTxt: Label 'Checking privacy approval state', Locked = true;
-        AdminPrivacyApprovalStateTelemetryTxt: Label 'Admin privacy approval state: %1', Locked = true;
-        UserPrivacyApprovalStateTelemetryTxt: Label 'User privacy approval state: %1', Locked = true;
+        CreatePrivacyNoticeTelemetryTxt: Label 'Creating privacy notice %1', Locked = true;
```
encimita: The privacy notice table (`table 2000000237 "Privacy Notice"`) is a system table in our platform that partners can use as well.
Here we are effectively treating the ID of that table as System Metadata, but this might not be immediately clear to partners that might be using the table.

I suggest we also make a change to the system table `table 2000000237 "Privacy Notice"` (separate to this PR) to mark the ID field with the annotation `DataClassification = SystemMetadata;`, just to make the classification explicit.

encimita (dismissed):

src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
JesperSchulz: You need to add a comment for placeholders (see above).

JesperSchulz (dismissed):

encimita (approved):

JesperSchulz (approved):

PR #2391, opened 2024-11-23
Private/waabusea/move media to resources

WaelAbuSeada:
#### Summary 
Move media files to resources in system app

Fixes #460898


src/System Application/App/AI/app.json
```diff
@@ -71,6 +71,12 @@
       "publisher": "Microsoft",
       "version": "26.0.0.0"
     },
+    {
+      "id": "e31ad830-3d46-472e-afeb-1d3d35247943",
+      "name": "BLOB Storage",
```
darjoo, line 76: Why does the AI module now have a dependency to BLOB storage?

src/System Application/App/Guided Experience/app.json
```diff
@@ -88,6 +88,11 @@
       "id": "4af62c1c-2f6d-4574-ba34-bb5c149cdf93",
       "name": "Guided Experience Test",
       "publisher": "Microsoft"
+    },
+    {
+      "id": "d3433b68-4901-445f-9547-fdfeca57575a",
```
darjoo, line 93: I don't think we should encourage this, using internalsvisibleto can easily get out of hand.
Create an API in the GuidedExperience Facade and use ModuleInfo to block non-Microsoft from invoking it.

src/System Application/App/AI/src/Copilot/CopilotNotAvailable.Page.al
```diff
@@ -104,8 +105,14 @@ page 7771 "Copilot Not Available"
     end;
 
     local procedure LoadBanner()
+    var
+        GuidedExperience: Codeunit "Guided Experience";
     begin
-        if MediaResources.Get('COPILOTNOTAVAILABLE.PNG') then
-            BannerVisible := MediaResources."Media Reference".HasValue;
+        BannerVisible := GuidedExperience.LoadFileFromTenantMediaSet(TenantMediaSet, 'CopilotNotAvailable.png');
```
haoranpb, line 111: Not sure I understand, why are we loading Media using procedures from `GuidedExperience` here?

src/System Application/App/Email/src/Account/EmailAccountWizard.Page.al
```diff
@@ -539,18 +539,16 @@ page 8886 "Email Account Wizard"
     end;
 
     local procedure LoadTopBanners()
+    var
+        GuidedExperience: Codeunit "Guided Experience";
     begin
-        if MediaResourcesStandard.Get('ASSISTEDSETUP-NOTEXT-400PX.PNG') and
-            MediaResourcesDone.Get('ASSISTEDSETUPDONE-NOTEXT-400PX.PNG') and (CurrentClientType() = ClientType::Web)
-        then
-            TopBannerVisible := MediaResourcesDone."Media Reference".HasValue();
```
haoranpb, line 546: Same here, are we taking dependency from Email module to Guided Experience to load Media?

I see we have a System Module called [Image](https://github.com/microsoft/BCApps/blob/main/src/System%20Application/App/Image/src/Image.Codeunit.al), might be more suitable for this purpose

src/System Application/App/Guided Experience/src/Guided Experience/GuidedExperience.Codeunit.al
```diff
@@ -443,6 +444,37 @@ codeunit 1990 "Guided Experience"
...
+    /// <param name="TenantMediaSet">Tenant Media set record by var for standard image.</param>
+    /// <param name="FilePath">Path to file name.</param>
+    /// <param name="FileName">File name.</param>
+    /// <returns>True if the the image is loaded successfuly</returns>
+    internal procedure InsertSystemFileToTenantMediaSet(var TenantMediaSet: Record "Tenant Media Set"; FilePath: Text[100]; FileName: Text[250])
+    var
+    begin
+        GuidedExperienceImpl.InsertSystemFileToTenantMediaSet(TenantMediaSet, FilePath, FileName);
+    end;
+
```
haoranpb, line 477: Not sure how those procedures fit in Guided Experience, consider Image module per previous comment

haoranpb (commented):
I understand the part 
> "instead will be added as resources to the appropriate app."

But can you elaborate on this?
> Media files will not exists in "Media resources" table anymore

pri-kise:
Before the images are removed from the `MediaResources` please add good alternatives and wait for the cleanup the common period of time, that partners are able to refactor their wizards.

And please update the new documentation for wizard creation then, too.
https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-designing-navigate-pages
The same applies for the learning path.
Currenlty the MediaResources are used in the example for wizards.

As soon as the examples is updated and you provide a good mechanis to get those images in custom wizards then we're able to refactor our code and update common snippets (https://github.com/waldo1001/crs-al-language-extension/blob/master/snippets/page.json)

src/System Application/App/AI/src/Copilot/CopilotNotAvailable.Page.al
WaelAbuSeada: We are loading wizard images from tenant media set, there is a fine line of ownership I agree. but it is the best place I could find

src/System Application/App/Guided Experience/src/Guided Experience/GuidedExperience.Codeunit.al
WaelAbuSeada: First one, is strictly for banners in a wizard, and should definitely be there. The problem with the rest is that they are not really only for images, as resources can be anything in the future

src/System Application/App/Guided Experience/app.json
WaelAbuSeada: True, I was just considering if this is something we would like to make internal or not. Probably no need and I will switch it to public.

TKapitan:
As @pri-kise said, it needs to follow the proper obsoletion path. But I'm not sure how to notify partners in pipeline about this change. Without this, you can be sure there will be many complains once this is removed from tenant media as there is certainly hundreds of apps that are using these resources as it was (still is) part of the learning path and official examples how to build this kind of pages.

src/System Application/App/Guided Experience/src/Guided Experience/GuidedExperienceImpl.Codeunit.al
```diff
@@ -569,6 +571,44 @@ codeunit 1991 "Guided Experience Impl."
...
+    var
+        TenantMedia: Record "Tenant Media";
+    begin
+        TenantMedia.SetFilter(Description, FileName);
+        if TenantMedia.FindFirst() then
+            if TenantMediaSet.Get(TenantMedia.ID, TenantMedia.ID) then
+                exit(TenantMediaSet."Media ID".HasValue);
+    end;
+
+    procedure InsertSystemFileToTenantMediaSet(var TenantMediaSet: Record "Tenant Media Set"; FilePath: Text[100]; FileName: Text[250])
```
IceOnly, line 598: I think this procedure should be set to local. I would like a function that first looks in the tenant media, if there is nothing there, then it should be inserted from the resources. 
And then the FilePath is also useless, it is currently always "images/".

src/System Application/App/app.json
```diff
@@ -36,10 +36,13 @@
     "TranslationFile",
     "GenerateCaptions"
   ],
+  "resourceFolders": [
+    ".resources"
```
IceOnly, line 40: Are you using .ressource because ressources is not possible? At least that's what I've discovered. Dot folders are currently always for content that is not repository relevant. Like .alpackages, .vscode, .snapshots. I would prefer data here.

src/System Application/App/AI/app.json
IceOnly: Why is tihs resolved? It seems not to be needed.

src/System Application/App/Email/src/Account/EmailAccountWizard.Page.al
IceOnly: I also think that Guided Experince doesn't fit, but the existing image modules don't really either.

IceOnly (changes requested):

PR #2421, opened 2024-11-29
Bug 501006: [Bug] Page "Word Template Creation Wizard" (9995) - missing variable update leads to an error #23867

JesperSchulz:
#### Summary 
The issue was reported here: https://github.com/microsoft/ALAppExtensions/issues/23867
In essence, if you ever got the error message that more than 250 fields were selected and you tried to delete a row, the count wouldn't get updated and you couldn't get past it. Now you can.

#### Work Item(s) 
Fixes [AB#501006](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/501006)


Groenbech96 (approved):

JesperSchulz:
I actually don't fully understand why the current code isn't working, as the dictionary is supposed to be updated in the trigger. The new code works, but as should the old. I'll debug...

WaelAbuSeada (approved):

Groenbech96 (approved):

PR #2422, opened 2024-11-29
Unnecessary variable ShowLogo in Email Accounts Page

### LINKED ISSUE #2365: [Bug]: Unnecessary variabl ShowLogo in Email Accounts Page
### Describe the issue

As we have fixed it for:
https://github.com/microsoft/BCApps/pull/663

Variable Show Logo on Email Accounts page have no function:
Image

Image

It's always set to true.
We could do cleanup and remove it, behavior would stay same.

### Expected behavior

/

### Steps to reproduce

/

### Additional context

_No response_

StefanSosic:
#### Summary 
Variable Show Logo on Email Accounts page have no function:
image

image

It's always set to true.
We could do cleanup and remove it, behavior would stay same.

#### Work Item(s) 
Fixes #2365

Fixes [AB#559267](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/559267)

### DIFF
--- a/src/System Application/App/Email/src/Account/EmailAccounts.Page.al
+++ b/src/System Application/App/Email/src/Account/EmailAccounts.Page.al
@@ -34,7 +34,6 @@ page 8887 "Email Accounts"
         {
             repeater(Accounts)
             {
-                Visible = ShowLogo;
                 FreezeColumn = NameField;
 #pragma warning disable AW0009
                 field(LogoField; Rec.LogoBlob)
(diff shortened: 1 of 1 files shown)


JesperSchulz (approved):
Yes, indeed pointless.

NKarolak:
Is there any scenario where it would make sense to hide the logo by default?
If yes, then the variable should not be removed, but a setter function should be added instead (one to set it to false. If not used, the value will still be set to true by default).

In any case, removing `Visible = ShowLogo;` from the repeater control (as suggested) seems to be correct.

grobyns (approved):

PR #2439, opened 2024-12-03
Calculate preprocessor symbols

aholstrup1:
#### Summary 
Calculate CLEAN tags instead of using list 

#### Work Item(s) 
Fixes [AB#559429](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/559429)

### DIFF
--- a/.github/actions/TestPreprocessorSymbols/action.ps1
+++ b/.github/actions/TestPreprocessorSymbols/action.ps1
@@ -19,7 +19,7 @@ if ($CurrentVersion -le 26) {
 
 # Define the preprocessor symbols to check for
 $symbolConfigs = @(
-    @{stem = "CLEAN"; lowerBound = (Get-MinAllowedCleantagVersion); upperBound = $mainVersion},
+    @{stem = "CLEAN"; lowerBound = (Get-MinAllowedCleanTagVersion); upperBound = $mainVersion},
     @{stem = "CLEANSCHEMA"; lowerBound = $schemaLowerBound; upperBound = $mainVersion + 3} # next lowerbound, after cleanup should be 25, then
 )
 
--- a/.github/actions/TestPreprocessorSymbols/action.ps1
+++ b/.github/actions/TestPreprocessorSymbols/action.ps1
@@ -19,7 +19,7 @@ if ($CurrentVersion -le 26) {
 
 # Define the preprocessor symbols to check for
 $symbolConfigs = @(
-    @{stem = "CLEAN"; lowerBound = ($CurrentVersion - 4); upperBound = $mainVersion},
+    @{stem = "CLEAN"; lowerBound = (Get-MinAllowedCleantagVersion); upperBound = $mainVersion},
     @{stem = "CLEANSCHEMA"; lowerBound = $schemaLowerBound; upperBound = $mainVersion + 3} # next lowerbound, after cleanup should be 25, then
 )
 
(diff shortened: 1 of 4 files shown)


build/scripts/GuardingV2ExtensionsHelper.psm1
```diff
@@ -268,4 +268,18 @@ function Get-MaxAllowedObsoleteVersion() {
     return [System.Version]::Parse($alGoSettings.repoVersion).Major
 }
 
+function Get-MinAllowedCleantagVersion() {
```
mazhelez:
```suggestion
function Get-MinAllowedCleanTagVersion() {
```

.github/actions/TestPreprocessorSymbols/action.ps1
```diff
@@ -19,7 +19,7 @@ if ($CurrentVersion -le 26) {
 
 # Define the preprocessor symbols to check for
 $symbolConfigs = @(
-    @{stem = "CLEAN"; lowerBound = ($CurrentVersion - 4); upperBound = $mainVersion},
+    @{stem = "CLEAN"; lowerBound = (Get-MinAllowedCleantagVersion); upperBound = $mainVersion},
```
mazhelez:
```suggestion
    @{stem = "CLEAN"; lowerBound = (Get-MinAllowedCleanTagVersion); upperBound = $mainVersion},
```

build/scripts/PreCompileApp.ps1
```diff
@@ -53,6 +53,8 @@ if($appType -eq 'app')
                 else {
                     Compile-AppInBcContainer @tempParameters | Out-Null
                 }
+
+                $parameters.Value["preprocessorsymbols"] = Get-PreprocessorSymbols
```
mazhelez, line 57: If the clean tags are set here, what's the point of _cleanModePreprocessorSymbols_ setting?

mazhelez (commented):
Looks good. Left a comment that implies that _cleanModePreprocessorSymbols_ setting might not be needed anymore. 

I suppose it's required by AL-Go.
Can we instead set it so that _cleanModePreprocessorSymbols_ is updated every time there is a major version change? What do you think?

build/scripts/PreCompileApp.ps1
aholstrup1: None. They're not used. But AL-Go fails if we don't set anything. 

Maybe we should just write a warning [here](https://github.com/microsoft/AL-Go/blob/9dbee17f636ca9ba2649b45adc2b65f4fd66af84/Actions/RunPipeline/RunPipeline.ps1#L376) instead of throwing an error?

aholstrup1:
> Looks good. Left a comment that implies that cleanModePreprocessorSymbols setting might not be needed anymore.

> I suppose it's required by AL-Go.
> Can we instead set it so that cleanModePreprocessorSymbols is updated every time there is a major version change? What do you think?

I don't hate the idea, but it does imply more automation PRs and for that reason I'm hesitant.

build/scripts/PreCompileApp.ps1
mazhelez: This is a defensive check, so not sure about a warning. How would you fix such a warning?

But why automation PRs? 🤔 
My idea was just to add a check (in "Verify App Changes" maybe) that the `cleanModePreprocessorSymbols` are in order, based on the current version.
Then, to be on the safe side, `Get-PreprocessorSymbols` can dynamically add CLEAN preprocessor symbols for all versions between the current version and the version on _main_.

Generally speaking, CLEAN26 makes no sense in version 25. But so do all the clean builds we do all around, but _main_.

aholstrup1:
Closing for now

PR #2442, opened 2024-12-03
add Drilldown to facade

grobyns:
#### Summary 
Adds Drilldown to facade

#### Work Item(s) 
Fixes [AB#559484](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/559484)

### DIFF
--- a/src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetup.Codeunit.al
+++ b/src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetup.Codeunit.al
@@ -49,28 +49,4 @@ codeunit 299 "No. Series - Setup"
     begin
         NoSeriesSetupImpl.UpdateNoSeriesLine(NoSeriesLine, NewNo, NewFieldCaption);
     end;
-
-    /// <summary>
-    /// Opens the No. Series Lines page filtered on the Series Code focused on the applicable line based on the workdate.
-    /// </summary>
-    /// <param name="NoSeries">The No. Series record to drill down on.</param>
-    procedure DrillDown(var NoSeries: Record "No. Series")
-    var
-        NoSeriesSetupImpl: Codeunit "No. Series - Setup Impl.";
-    begin
-        NoSeriesSetupImpl.DrillDown(NoSeries);
-    end;
-
-    /// <summary>
-    /// Filters the No Series Lines and selects the currently applicable line based on the workdate.
-    /// </summary>
-    /// <param name="NoSeries">The No. Series for which to select a line.</param>
-    /// <param name="NoSeriesLine">The No Series Line is returned filtered and the currently applicable line is selected</param>
-    /// <param name="ResetFilters">If set to true, the filters on the Starting Date and Open fields are reset. The currently applicable line is still selected in the table.</param>
-    procedure SelectCurrentNoSeriesLine(var NoSeries: Record "No. Series"; var NoSeriesLine: Record "No. Series Line"; ResetFilters: Boolean)
-    var
-        NoSeriesSetupImpl: Codeunit "No. Series - Setup Impl.";
-    begin
-        NoSeriesSetupImpl.SelectCurrentNoSeriesLine(NoSeries, NoSeriesLine, ResetFilters);
-    end;
 }
--- a/src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetup.Codeunit.al
+++ b/src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetup.Codeunit.al
@@ -49,4 +49,28 @@ codeunit 299 "No. Series - Setup"
     begin
         NoSeriesSetupImpl.UpdateNoSeriesLine(NoSeriesLine, NewNo, NewFieldCaption);
     end;
+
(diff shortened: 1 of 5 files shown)


src/Business Foundation/App/NoSeries/src/Setup/NoSeriesSetup.Codeunit.al
```diff
@@ -49,4 +49,28 @@ codeunit 299 "No. Series - Setup"
...
+        NoSeriesSetupImpl.DrillDown(NoSeries);
+    end;
+
+    /// <summary>
+    /// Filters the No Series Lines and selects the currently applicable line based on the workdate.
+    /// </summary>
+    /// <param name="NoSeries">The No. Series for which to select a line.</param>
+    /// <param name="NoSeriesLine">The No Series Line is returned filtered and the currently applicable line is selected</param>
+    /// <param name="ResetFilters">If set to true, the filters on the Starting Date and Open fields are reset. The currently applicable line is still selected in the table.</param>
+    procedure SelectCurrentNoSeriesLine(var NoSeries: Record "No. Series"; var NoSeriesLine: Record "No. Series Line"; ResetFilters: Boolean)
```
AndreasMoth: I would expect this to return false if no line was found. Currently we just init a line, which seems a bit weird.. That also means this is not the currently applicable line, it's just a blank line..

AndreasMoth: I would skip the ResetFilters boolean. Also it doesn't reset any filters set by the partners through SetNoSeriesLineFilters or the other events. If we really want to support reset filters we should use another approach, but why even have this? Just hardcode it to false here.

AndreasMoth: There is no reason for NoSeries to be by var, in fact we should remove the var from the function.

grobyns: You need the reset for drilldown, lookup and whatnot… so we can do an overload, one that does and one that doesn’t, but we do need to reset… probably need true more than false

grobyns: I considered that… but since the underlying method has it I decided to keep it. I don’t remember why we added var originally …

grobyns: We can add a return value I think. We have to return a line always since we use the values on the page…

dominicstarkl:
@grobyns @AndreasMoth Wow, you are my heroes, I really didn't expect that great engagement 😊

src/Business Foundation/App/NoSeries/src/Single/NoSeries.Codeunit.al
```diff
@@ -326,6 +326,20 @@ codeunit 310 "No. Series"
...
     end;
 
+    /// <summary>
+    /// Filters the No Series Lines and selects the currently applicable line based on the workdate.
+    /// </summary>
+    /// <param name="NoSeries">The No. Series for which to select a line.</param>
+    /// <param name="NoSeriesLine">The No Series Line is returned filtered and the currently applicable line is selected</param>
+    /// <param name="ResetFilters">If set to true, the filters on the Starting Date and Open fields are reset. The currently applicable line is still selected in the table.</param>
+    /// <returns>Returns true if a line was selected, false if no line was selected.</returns>
+    procedure SelectCurrentNoSeriesLine(NoSeries: Record "No. Series"; var NoSeriesLine: Record "No. Series Line"; ResetFilters: Boolean): Boolean
```
AndreasMoth, line 336: Maybe call ResetFilters for ResetStartingDateAndOpenFilters :P. More seriously though, it's a bit confusing that it's simply called reset filters even though it only resets filters for two fields. Maybe just always put this to true and specify that is what this function does. Still seems odd to me. I get that we need these reset for later use, but this function shouldn't care about what it should be used for but what the name implies it should do, which isn't to reset filters just in case of other functions using it..
Maybe the other functions should reset these fields instead? That might make sense.. not sure, I haven't looked at that code.
Anyway this is not a big problem and feel free to won't fix this and check in.

AndreasMoth (approved):

dominicstarkl:
@grobyns @AndreasMoth What do you think will be the ETA?

src/Business Foundation/App/NoSeries/src/Single/NoSeries.Codeunit.al
grobyns: Yeah... unless someone proposes a nice name or solution, i'll leave it as is for now...

onbuyuka (approved):

grobyns:
> @grobyns @AndreasMoth What do you think will be the ETA?

@dominicstarkl The fix should be available in 25.3 when it ships.

dominicstarkl:
@grobyns Thank you very much Gert 😊

dominicstarkl:
@grobyns Thank you very much Gert 😊

PR #2443, opened 2024-12-03
[Copilot] Copilot capabilities can depend on privacy notices

msft-sam:
#### Summary
In addition to using AOAI, Copilot capabilities may need to integrate with other services that depend on a privacy notice.
This PR adds a new event capabilities can subscribe to which provide a list of required privacy notices that must be accepted before the capability can be enabled.

#### Work Item(s)
Fixes [AB#502565](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/502565)

### DIFF
--- a/src/System Application/App/AI/src/Copilot/CopilotCapEarlyPreview.Page.al
+++ b/src/System Application/App/AI/src/Copilot/CopilotCapEarlyPreview.Page.al
@@ -90,6 +90,8 @@ page 7770 "Copilot Cap. Early Preview"
                 trigger OnAction()
                 begin
                     if Dialog.Confirm(ActivateEarlyPreviewTxt, false) then begin
+                        if not Rec.EnsurePrivacyNoticesApproved() then
+                            exit;
                         Rec.Status := Rec.Status::Active;
                         Rec.Modify(true);
 
@@ -176,10 +178,12 @@ page 7770 "Copilot Cap. Early Preview"
     end;
 
     local procedure SetActionsEnabled()
+    var
+        CopilotCapability: Codeunit "Copilot Capability";
     begin
         if CopilotCapabilityImpl.IsAdmin() then begin
             ActionsEnabled := (Rec.Capability.AsInteger() <> 0) and DataMovementEnabled;
-            CapabilityEnabled := Rec.Status = Rec.Status::Active;
+            CapabilityEnabled := CopilotCapability.IsCapabilityActive(Rec.Capability, Rec."App Id");
         end
         else begin
             ActionsEnabled := false;
--- a/src/System Application/App/Privacy Notice/src/SystemPrivacyNoticeReg.Codeunit.al
+++ b/src/System Application/App/Privacy Notice/src/SystemPrivacyNoticeReg.Codeunit.al
@@ -6,7 +6,7 @@
 namespace System.Privacy;
 
 /// <summary>
-/// This codeunit registers platform level privacy notices and provides procedures to consistently reference the privcay notices.
+/// This codeunit registers platform level privacy notices and provides procedures to consistently reference the privacy notices.
 /// </summary>
 codeunit 1566 "System Privacy Notice Reg."
 {
--- a/src/System Application/App/AI/src/Copilot/CopilotCapabilitiesGA.Page.al
+++ b/src/System Application/App/AI/src/Copilot/CopilotCapabilitiesGA.Page.al
@@ -89,6 +89,9 @@ page 7774 "Copilot Capabilities GA"
 
                 trigger OnAction()
                 begin
+                    if not Rec.EnsurePrivacyNoticesApproved() then
+                        exit;
+
                     Rec.Status := Rec.Status::Active;
                     Rec.Modify(true);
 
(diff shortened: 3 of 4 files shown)


src/System Application/App/Privacy Notice/src/PrivacyNotices.Page.al
```diff
@@ -159,6 +159,7 @@ page 1565 "Privacy Notices"
 
     trigger OnAfterGetRecord()
     begin
+        Rec.CalcFields(Rec.Enabled, Rec.Disabled);
```
encimita, line 162: Minor: since we are doing it for every record, do we gain some perf if we do `Rec.SetAutoCalcFields` in `OnOpenPage` instead?

src/System Application/App/Privacy Notice/src/SystemPrivacyNoticeReg.Codeunit.al
```diff
@@ -7,7 +7,6 @@ namespace System.Privacy;
 
 codeunit 1566 "System Privacy Notice Reg."
 {
```
encimita, line 12: We need docs comment for the codeunit since now it's externally visible (up to you whether we need doc comments for procedures as well, since they are quite self explanatory)

src/System Application/App/AI/src/Copilot/CopilotCapEarlyPreview.Page.al
```diff
@@ -35,7 +35,7 @@ page 7770 "Copilot Cap. Early Preview"
                     Editable = false;
                     Width = 30;
                 }
-                field(Status; Rec.Status)
+                field(Status; Rec.EvaluatedStatus())
```
encimita: nit: why not `EvaluateStatus`?

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
```diff
@@ -133,13 +133,27 @@ codeunit 7774 "Copilot Capability Impl"
...
             exit(false);
 
-        exit(CopilotSettings.Status = Enum::"Copilot Status"::Active);
+        CopilotCapabilityCU.OnGetRequiredPrivacyNotices(CopilotCapability, AppId, RequiredPrivacyNotices);
+
+        if (CopilotSettings.Status <> Enum::"Copilot Status"::Active) or (RequiredPrivacyNotices.Count() <= 0) then
+            exit(CopilotSettings.Status = Enum::"Copilot Status"::Active);
+
+        // check privacy notices
+        foreach RequiredPrivacyNotice in RequiredPrivacyNotices do
```
encimita, line 152: Do we need some telemetry here (verbose or info) to help us debug in the future?

src/System Application/App/AI/src/Copilot/CopilotSettings.Table.al
```diff
@@ -51,4 +53,36 @@ table 7775 "Copilot Settings"
...
+        if Rec.Status <> Rec.Status::Active then
+            exit(Rec.Status);
+
+        if CopilotCapability.IsCapabilityActive(Rec.Capability, Rec."App Id") then
+            exit(Rec.Status::Active)
+        else
+            exit(Rec.Status::Inactive);
+    end;
+
+    procedure EnsurePrivacyNoticesApproved(): Boolean
```
encimita, line 70: `CheckPrivacyNoticesApproved` or `ArePrivacyNoticesApproved`

Ensure to me means you are creating something if it's not there..

encimita: Oh wait is that what we actually want?

encimita: Where do we plan to use this?

encimita (approved):

darjoo (approved):
Enrico has the comments I'd add. So look at those :D

src/System Application/App/AI/src/Copilot/CopilotSettings.Table.al
msft-sam: When you enable the capability, we ensure the privacy notice is approved (in the page parts), if it is not, you will see the disclaimer and be required to approve, since admins can only enable the capability, the approval will be for the org.

(I put the code here to avoid repeating it over all the pages)

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
msft-sam: I had considered it, but [`PrivacyNotice`](https://github.com/microsoft/BCApps/blob/main/src/System%20Application/App/Privacy%20Notice/src/PrivacyNoticeImpl.Codeunit.al#L121) already emits a load of telemetry here so I think we would just be duplicating it. Let me know if you think it would still be valuable to add though!

src/System Application/App/Privacy Notice/src/PrivacyNotices.Page.al
msft-sam: This doesn't seem to work for me, I was surprised I needed to add this though, so perhaps there has been some plat regression. I'll log a bug to look more into it.

encimita: Mmmm maybe some quirk with it being a system table. Worth investigating, not blocking since we don't have thousands of privacy notices anyway :)

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
encimita: Fair!

src/System Application/App/AI/src/Copilot/CopilotSettings.Table.al
encimita: Yeah that's fine but I cannot find where we are calling `EnsurePrivacyNoticesApproved` at all (but might be github search that is acting weirdly)

msft-sam: No you are right I had a bad merge 😅

encimita (approved):

darjoo (approved):

Issue #989, opened 2024-04-25
Add support of cookies to the Rest Client module


ajkauffmann:
### Describe the issue

Version v24 adds support for cookies to the HTTP calls. The Rest Client module currently doesn't support this.

### Expected behavior

The Rest Client module should also support adding and reading cookies.

### Steps to reproduce

The Rest Client module has no functions to add or read cookies.

### Additional context

The Rest Client module supports working directly with the underlying HttpClient, and the HttpRequestMessage and HttpResponseMessage. In that way, cookies are supported. But it would be more developer-friendly if the module itself exposes functions to work with cookies.

### I will provide a fix for a bug

- [X] I will provide a fix for a bug

JesperSchulz:
Of course approved!

pri-kise:
I think that this issue can be closed as it's solved.
@JesperSchulz

JesperSchulz:
> I think that this issue can be closed as it's solved. [@JesperSchulz](https://github.com/JesperSchulz)

Indeed. Link was missing. Link added.

PR #2474, opened 2024-12-09
[Rest Client] Support for Cookies + Fix secret headers + Collectible errors

### LINKED ISSUE #1479: [Bug]: HttpContent.SetHeader failing when secret header already exists
### Describe the issue

The "Http Content" codeunit throws an error message when content is defined from SecretText.  
```
Error message: 
Cannot add value because the header 'Content-Type' does not support multiple values.

AL call stack: 
"Http Content Impl."(CodeUnit 2355).SetHeader line 12 - System Application by Microsoft
"Http Content Impl."(CodeUnit 2355).SetContentTypeHeader line 4 - System Application by Microsoft
"Http Content Impl."(CodeUnit 2355).SetContent line 6 - System Application by Microsoft
"Http Content"(CodeUnit 2354).SetContent line 2 - System Application by Microsoft
"Http Content"(CodeUnit 2354).Create line 2 - System Application by Microsoft\"Http Content"(CodeUnit 2354).Create line 2 - System Application by Microsoft
```
This happens because inside "Http Content Impl.".SetHeader procedure `Headers.Contains(Name)` condition will return false for a secret header, which will be automatically created when we set a content from the SecretText value. 

### Expected behavior

Content will be defined with a new header without error.

### Steps to reproduce

```
codeunit 50000 SetHeader
{
    trigger OnRun()
    var
        HttpContent: Codeunit "Http Content";
    begin
        HttpContent.Create(SecretStrSubstNo('anything'));
    end;
}
```

### Additional context

_No response_

ajkauffmann:
#### Summary 
Adding support for cookies to the Rest Client
Fixing an issue with secret headers
Adding collectible error behavior
Redesigning internal code for better maintainability and uptaking new AL features like the keyword `this`

#### Work Item(s) 
Fixes #1479 #989 

Fixes [AB#560010](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/560010)


src/System Application/App/Rest Client/src/ExceptionHandling/RestClientExceptionBuilder.Codeunit.al
```diff
@@ -0,0 +1,28 @@
+namespace System.RestClient;
+
+codeunit 2362 "Rest Client Exception Builder"
+{
```
pri-kise, line 8: Public Codeunits should be explict with the Access property.
And public Codeunits require a documentation header.

pri-kise (changes requested):

src/System Application/App/Rest Client/src/ExceptionHandling/RestClientExceptionBuilder.Codeunit.al
ajkauffmann: Didn't know about the requirement of Access = Public on all public codeunits. It's a default property anyway.
Other objects didn't have this as well. So I've now added it to all objects for the sake of unity and clarity.

pri-kise: At least this pattern was used when the emphasized the Facade Pattern in the System application. 

Will you add an XML Summary to the enum values and the procedures in the new Exception Codeunit, too?
(I think the term "documentation header" was a little bit unclear)

src/System Application/Test/Rest Client/src/RequestMessageTests.Codeunit.al
```diff
@@ -145,12 +145,86 @@ codeunit 134972 "Request Message Tests"
...
         ContentHeaders.GetValues('X-Custom-Header', ContentHeaderValues);
-        Assert.AreEqual(ContentHeaderValues.Get(1), 'My Request Header', 'The custom header value is not correct.');
+        Assert.AreEqual('My Request Header', ContentHeaderValues.Get(1), 'The custom header value is not correct.');
     end;
 
+    [Test]
+    [NonDebuggable]
+    procedure TestAddSecretRequestHeader()
+    var
+        ALHttpRequestMessage: Codeunit "Http Request Message";
```
pri-kise, line 158: Did you intent to use the prefix `AL` for the AL HttpRequestMessage Type?
Then it's currently the other way.

It's the same on other test functions

ajkauffmann: The intent is to name the built-in HttpRequestMessage type just HttpRequestMessage (as I was always used to) and prefix AL for the codeunit wrappers (they are actually written in AL). But you can also do this they other way around of course, I know. At least it is consistent in all the testing code I believe.

src/System Application/App/Rest Client/src/ExceptionHandling/RestClientExceptionBuilder.Codeunit.al
ajkauffmann: Added documentation to the functions and enum.

pri-kise:
@ajkauffmann do I understand the changes with `this` correctly that we're now able to use the instance of the `HttpContent` when we use the method `Create`.
e.g. the following code

```al
HttpContent := HttpContent.Create('Dummy');
```
can now be simplified to 

```al
HttpContent.Create('Dummy');
```

If this is correct then maybe

src/System Application/App/Rest Client/src/ExceptionHandling/RestClientExceptionBuilder.Codeunit.al
pri-kise: I don't want to annoy you, since I really appreciate your contribution, but I think that the XML Summary should contain the lines for parameters.

ajkauffmann:
> @ajkauffmann do I understand the changes with `this` correctly that we're now able to use the instance of the `HttpContent` when we use the method `Create`. e.g. the following code
> 
> ```al
> HttpContent := HttpContent.Create('Dummy');
> ```
> 
> can now be simplified to
> 
> ```al
> HttpContent.Create('Dummy');
> ```

It's actually the other way around. Since Create() returns the instance by means of `this`, it can be assigned to a variable or parameter. If assigned to the same variable, like `HttpContent := HttpContent.Create('Dummy');` then there is technically no difference with just `HttpContent.Create('Dummy');`.

src/System Application/App/Rest Client/src/ExceptionHandling/RestClientExceptionBuilder.Codeunit.al
ajkauffmann: Eh... yes of course... Done!

JesperSchulz:
Looks like this one already is ready for a review by the product group! Let me find a Rest Client expert, as I'm definitely not one of those 😝

src/System Application/App/Rest Client/src/HttpContentImpl.Codeunit.al
```diff
@@ -113,31 +198,52 @@ codeunit 2355 "Http Content Impl."
...
+
+    procedure AsJsonObject() ReturnValue: JsonObject
+    var
+        Json: Text;
+    begin
+        Json := AsText();
+        if Json <> '' then
+            if not ReturnValue.ReadFrom(AsText()) then
+                ThrowInvalidJsonException();
+    end;
```
JesperSchulz: NIT: personal preference is to always reduce nesting where possible:

```suggestion
    procedure AsJsonObject() ReturnValue: JsonObject
    var
        Json: Text;
    begin
        Json := AsText();
        if Json == '' then
            exit;
        if not ReturnValue.ReadFrom(AsText()) then
            ThrowInvalidJsonException();
    end;
```

Doesn't really matter in such small functions, but it's good practice to make code more easily readable. But feel free to disregard here.

src/System Application/Test/Rest Client/app.json
```diff
@@ -35,7 +35,7 @@
   "idRanges": [
     {
       "from": 134970,
-      "to": 134974
+      "to": 134975
```
JesperSchulz: This ID is already used by codeunit 134975 "ERM Dimension Report". 134977 seems to be available.
```suggestion
      "to": 134977
```

JesperSchulz (approved):
Nice one. Looks fine to me!

src/System Application/Test/Rest Client/src/HttpExceptionTests.Codeunit.al
```diff
@@ -0,0 +1,290 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Test.RestClient;
+
+using System.RestClient;
+using System.TestLibraries.Utilities;
+
+codeunit 134975 "Http Exception Tests"
```
JesperSchulz:
```suggestion
codeunit 134977 "Http Exception Tests"
```

JesperSchulz: Let me just accept those changes and get a new build running.

JesperSchulz (dismissed):

src/System Application/App/Rest Client/src/HttpContentImpl.Codeunit.al
```diff
@@ -32,9 +116,9 @@ codeunit 2355 "Http Content Impl."
     var
         Headers: HttpHeaders;
     begin
-        if not HttpContent.GetHeaders(Headers) then begin
-            HttpContent.Clear();
-            HttpContent.GetHeaders(Headers);
+        if not this.HttpContent.GetHeaders(Headers) then begin
```
grobyns: echoing a comment made on Viva Engage: 
[Peter D. Jørgensen](https://www.yammer.com/dynamicsnavdev/#/users/1869009403904) – [36 minutes ago](https://www.yammer.com/dynamicsnavdev/#/Threads/show?threadId=3085674977689600)

Why are the global variables addressed as "this.<global var>" ?

Should we also start doing that, when calling functions in the current object?

src/System Application/App/Rest Client/src/HttpContentImpl.Codeunit.al
ajkauffmann: Done. I also removed duplicate usage of AsText() while at it.

JesperSchulz (approved):
Love it.

src/System Application/App/Rest Client/src/HttpContentImpl.Codeunit.al
grobyns: Just as a note, the current MS internal guideline is to only use 'this' when required and to not use it when it is not required. 

This came about after a few pr's where 'this' was used everywhere, consistently, and the feedback was it made the code less readable...

src/System Application/Test/Rest Client/src/RestClientTests.Codeunit.al
```diff
@@ -417,17 +482,160 @@ codeunit 134971 "Rest Client Tests"
...
+        HttpResponseMessage: Codeunit "Http Response Message";
+    begin
+        // [SCENARIO] Test GET request with cookies
+
+        // [GIVEN] An initialized Rest Client
+        HttpClientHandler.Initialize();
+        RestClient.Initialize(HttpClientHandler);
+
+        // [WHEN] The Get method is called
+        HttpResponseMessage := RestClient.Get('https://postman-echo.com/get');
```
onbuyuka, line 503: We can't have this. MS blocked Postman due to security concerns, we cannot have tests calling Postman. Actually we cannot have tests calling an external service at all, I'll comment on that later

onbuyuka: Calling this API most likely would not be a risk but we shouldn't take a change

src/System Application/App/Rest Client/src/HttpContentImpl.Codeunit.al
ajkauffmann: > echoing a comment made on Viva Engage: [Peter D. Jørgensen](https://www.yammer.com/dynamicsnavdev/#/users/1869009403904) – [36 minutes ago](https://www.yammer.com/dynamicsnavdev/#/Threads/show?threadId=3085674977689600)
> 
> Why are the global variables addressed as "this." ?
> 
> Should we also start doing that, when calling functions in the current object?

It's a good point, and I hoped to force a discussion by using it. Seems to work. 😉

The first reason to use the keyword `this` for referencing globals in the same object is that it's mentioned in the documentation: https://learn.microsoft.com/en-us/dynamics365/release-plan/2024wave2/smb/dynamics365-business-central/use-this-keyword-codeunit-self-reference#feature-details. There is even a Codecop rule AA0248 with an AL code action. The documentation also states that the System Application has been modified to use `this`.

The second reason is that I started by using the same name for parameters and global variables and then distinguished by using `this` to refer to the global variable. This is an example where `this` really would make sense. But that failed on Codecop rule AA0244, and I had no choice but to rename the parameters. But I kept `this` for referencing global variables.

I'm still not sure if this is really what we want or what problem it actually solves. It would help when parameter names are the same as global variables, but the Codecop rule already prevented that.

onbuyuka (changes requested):
Tests calling external service (httpbin) have been causing us issues, when the service is down for any reason, our tests fail and block all the gates. We are working on HTTP mocking in the compiler scenarios, maybe we should wait until that's in place and refactor these tests?

"Request changes" is for Postman calls

src/System Application/Test/Rest Client/src/RestClientTests.Codeunit.al
ajkauffmann: The other tests also call an external service: httpbin.org which has been accepted before.
Testing this module is close to impossible without actually calling an external service. Rest assured I did the tests. How should we work around this?

JesperSchulz: The way I see it, there are two options:
1) Change postman call to something else and disable the test to be run during test automation for now. The test can be run manually when code is churned.
2) Wait for the mocking capabilities to come along (should be soon) and refactor tests to uptake those.

I'd vote for 1, with planned work to do 2 when the time comes. I stand by my "approved" verdict.

src/System Application/App/Rest Client/src/HttpContentImpl.Codeunit.al
JesperSchulz: Yeah, I'm not getting in the middle of this discussion 🤣 My take has always been: if it improves code readability, use it. Use your common sense. If your code get unnecessarily cluttered, then don't use it. Unfortunately that doesn't always work with _rules_, which so many are so fond of. I hence tend to get into trouble with my view on how to use such things 😝 So I'll leave it to others to decide on the rules, and then I'll blindly follow them!

src/System Application/App/Rest Client/app.json
```diff
@@ -34,11 +34,11 @@
   "idRanges": [
     {
       "from": 2350,
-      "to": 2361
+      "to": 2362
     }
   ],
   "target": "OnPrem",
-  "runtime": "12.0",
+  "runtime": "14.0",
```
nikolakukrika: Please remove the runtime version. We do not define it for Microsoft owned code.

src/System Application/Test/Rest Client/src/RestClientTests.Codeunit.al
ajkauffmann: The challenge is to find another service I can use for testing cookies. Tried httpbin.org first, but it didn't work because of an automatic redirect loosing cookies. It was planning to test this specific scenario and maybe report it as a bug.
To avoid further delays, I searched for another service and found postman-echo.com. I'm open to alternatives. Or should we build our own testing service, e.g., an Azure function, for testing purposes?

Do I need to do anything in the code to disable the test during test automation?

src/System Application/App/Rest Client/src/HttpContentImpl.Codeunit.al
ajkauffmann: Now that the internal guidelines appear to have changed, I'll remove `this` everywhere it is used to reference global vars. May I suggest making this internal guideline public, please? The documentation currently says something else.

nikolakukrika: If you are getting hit by AA0244, you must rename the variable. The code is not readable, with multiple variables named the same.
We do not use this in the code right now. Usage of this is an exception. We should either apply this codeword to all codebase or use it very sparsely.

In this case, please rename the global variable so it is not named the same as the built in keyword. Local parameters should have a different name too. It helps with the readability and understanding the code.

src/System Application/Test/Rest Client/src/RestClientTests.Codeunit.al
JesperSchulz: Please disable any test which communicates outside of the test runner here: https://github.com/microsoft/BCApps/tree/main/src/System%20Application/Test/DisabledTests

We've blocked the use of the Postman cloud solution / app, which this isn't directly related to. I'd prefer not mention postman anywhere, but I'm pretty sure that what you've used isn't part of the security initiative.

Let's disable the tests with REST calls and fix them when we have the means to do so.

src/System Application/App/Rest Client/src/HttpContentImpl.Codeunit.al
ajkauffmann: Removed all usage of `this` for referencing global variables.
Also renamed global variables in case they had the same name as the built-in keyword.

src/System Application/App/Rest Client/app.json
ajkauffmann: Runtime has been removed from app.json

src/System Application/Test/Rest Client/src/RestClientTests.Codeunit.al
ajkauffmann: I've looked into the DisabledTests folder and found a recent PR that removed the Rest Client tests from the disabled list.
https://github.com/microsoft/BCApps/pull/2142

@JesperSchulz  Do you want this to be added again? Or only for the new tests?

JesperSchulz: Let's only disable the tests we must - not the entire codeunit.

JesperSchulz:
And this is why we don't run tests with external dependencies in our CI:

Error: Function TestWithUseResponseCookies Assert.AreEqual failed. Expected:<200> (Integer). Actual:<503> (Integer). The response status code should be 200.

Let's disable those, and push forward 🥳

src/System Application/Test/Rest Client/src/RestClientTests.Codeunit.al
ajkauffmann: I've added the tests for cookies to the disabled tests and kept the already existing list.

JesperSchulz (approved):

WaelAbuSeada (approved):
Looks fine.

PR #2480, opened 2024-12-09
Bug 557680: [Instability][NAV.master] Tests are unstable in task RunALTests_CZ_Bucket1. (AUTODETECTED) - T001_PostingDateIsInNotAllowedPeriodInGLSetup

BardurKnudsen:
#### Summary 
Often tests are run in parallel, and these library functions cause race conditions (=unstable tests) since they don't consider other processes. The change is just to *try* to insert the new no. series record.

#### Work Item(s) 
Fixes #AB557680

### DIFF
--- a/src/Business Foundation/Test Library/NoSeries/src/LibraryNoSeries.Codeunit.al
+++ b/src/Business Foundation/Test Library/NoSeries/src/LibraryNoSeries.Codeunit.al
@@ -11,7 +11,7 @@ codeunit 134510 "Library - No. Series"
         NoSeries.Code := NoSeriesCode;
         NoSeries.Description := NoSeriesCode;
         NoSeries."Default Nos." := true;
-        NoSeries.Insert();
+        if NoSeries.Insert() then;
     end;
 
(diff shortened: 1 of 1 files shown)


src/Business Foundation/Test Library/NoSeries/src/LibraryNoSeries.Codeunit.al
```diff
@@ -11,7 +11,7 @@ codeunit 134510 "Library - No. Series"
         NoSeries.Code := NoSeriesCode;
         NoSeries.Description := NoSeriesCode;
         NoSeries."Default Nos." := true;
-        NoSeries.Insert();
+        if NoSeries.Insert() then;
```
AndreasMoth, line 14: What about exiting this instead? if you expect a new no. series it's good to get an error, if you don't care whether it already exists then you can do if-then around this call.

AndreasMoth: Same below

BardurKnudsen:
Skipping this change

PR #2524, opened 2024-12-12
Application Insights warning on llm model deprecation

christian-andersen-msft:
#### Summary 

Use of deprecated llm models in extensions will be logged in 3rd party Application Insights with a warning

image

#### Work Item(s) 
Fixes [AB#548931](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/548931)


haoranpb (approved):

src/System Application/App/AI/src/Azure OpenAI/AOAIDeploymentsImpl.Codeunit.al
```diff
@@ -15,11 +16,15 @@ codeunit 7769 "AOAI Deployments Impl"
...
     var
+        Telemetry: Codeunit Telemetry;
         UnableToGetDeploymentNameErr: Label 'Unable to get deployment name, if this is a third party capability you must specify your own deployment name. You may need to contact your partner.';
         GPT4oLatestLbl: Label 'gpt-4o-latest', Locked = true;
         GPT4oPreviewLbl: Label 'gpt-4o-preview', Locked = true;
         GPT4oMiniLatestLbl: Label 'gpt-4o-mini-latest', Locked = true;
         GPT4oMiniPreviewLbl: Label 'gpt-4o-mini-preview', Locked = true;
+        DeprecatedDeployments: Dictionary of [Text, Date];
+        DeprecationDatesInitialized: Boolean;
+        DeprecationMessageLbl: Label 'Deployment %1 deprecated from %2. Check out codeunit 7768 AOAI Deployments', Comment = 'Telemetry message where %1 is the name of the deployment and %2 is the date of deprecation';
```
encimita: Do we translate these strings or do we emit partner telemetry always in english?
If we translate them, you are good, but if we emit in english, you need to lock this string for translations, by adding `, Locked = true` after you close the quote of the comment

src/System Application/App/AI/src/Azure OpenAI/AOAIDeploymentsImpl.Codeunit.al
```diff
@@ -103,12 +108,55 @@ codeunit 7769 "AOAI Deployments Impl"
...
     end;
 
+    // Initializes dictionary of deprecated models
+    local procedure InitializeDeploymentDeprecationDates()
+    begin
+        if DeprecationDatesInitialized then
+            exit;
+
+        // Add deprecated deployments with their deprecation dates here
+        DeprecatedDeployments.Add(Turbo0301SaasLbl, DMY2Date(1, 11, 2024));
```
encimita, line 123: This label is defined in the file, but it's between the `#if not CLEAN25` and the `#endif` preprocessor directive. That means it will be automatically removed in a couple of major updates. At that point, your code will no longer compile, so you need to add the lines for this model and the other obsoleted models within an `#if not CLEAN25` and an `#endif` as well.

We can talk about this in person if you like :)

src/System Application/App/AI/src/Azure OpenAI/AOAIDeploymentsImpl.Codeunit.al
encimita: Let's make the text a bit more dev friendly.
`We detected usage of the Azure OpenAI deployment "%1". This model is obsoleted starting %2 and the quality of your results might vary after that date. Check out codeunit 7768 AOAI Deployments to find the supported deployments.`

encimita (approved):

src/System Application/App/AI/src/Azure OpenAI/AOAIDeploymentsImpl.Codeunit.al
```diff
@@ -103,12 +109,61 @@ codeunit 7769 "AOAI Deployments Impl"
...
+        DeprecatedDeployments.Add(GPT4oLatestLbl, DMY2Date(13, 5, 2025));
+        DeprecatedDeployments.Add(GPT4oPreviewLbl, DMY2Date(13, 5, 2025));
+        DeprecatedDeployments.Add(GPT4oMiniLatestLbl, DMY2Date(18, 7, 2025));
+        DeprecatedDeployments.Add(GPT4oMiniPreviewLbl, DMY2Date(18, 7, 2025));
+
+        DeprecationDatesInitialized := true;
+    end;
+
+    // Application Insights telemetry on deprecated models
+    local procedure LogDeprecationTelemetry(DeploymentName: Text)
```
msft-sam, line 131: Did we consider routing this through the `CopilotTelemetry` codeunit? It's a single instance codeunit so we would only initialize the dict once, and we could also then emit information about the current Copilot capability that is used (so you will be able to see from telemetry which capability is using the deprecated model)

msft-sam: An other way could be to add a procedure to the codeunit to get the current capability and include it in the telemetry message

src/System Application/App/AI/src/Azure OpenAI/AOAIDeploymentsImpl.Codeunit.al
```diff
@@ -103,12 +109,61 @@ codeunit 7769 "AOAI Deployments Impl"
...
+#if not CLEAN25
+        DeprecatedDeployments.Add(GPT4LatestLbl, DMY2Date(1, 11, 2024));
+        DeprecatedDeployments.Add(GPT4PreviewLbl, DMY2Date(1, 11, 2024));
+        DeprecatedDeployments.Add(GPT35TurboLatestLbl, DMY2Date(1, 11, 2024));
+        DeprecatedDeployments.Add(GPT35TurboPreviewLbl, DMY2Date(1, 11, 2024));
+        DeprecatedDeployments.Add(Turbo0301SaasLbl, DMY2Date(1, 11, 2024));
+        DeprecatedDeployments.Add(GPT40613SaasLbl, DMY2Date(1, 11, 2024));
+        DeprecatedDeployments.Add(Turbo0613SaasLbl, DMY2Date(1, 11, 2024));
+#endif
+        DeprecatedDeployments.Add(GPT4oLatestLbl, DMY2Date(13, 5, 2025));
```
darjoo: Do we have actual obsoletion dates for these? Why have they been added?

src/System Application/App/AI/src/Azure OpenAI/AOAIDeploymentsImpl.Codeunit.al
```diff
@@ -28,6 +33,7 @@ codeunit 7769 "AOAI Deployments Impl"
         Turbo0301SaasLbl: Label 'turbo-0301', Locked = true;
         GPT40613SaasLbl: Label 'gpt4-0613', Locked = true;
         Turbo0613SaasLbl: Label 'turbo-0613', Locked = true;
+
```
darjoo: NIT: Remove empty line

src/System Application/App/AI/src/Azure OpenAI/AOAIDeploymentsImpl.Codeunit.al
christian-andersen-msft: The lines with the 1st of November were added on that date and were already deprecated. The future dates were announced by OpenAI and added now to ensure that they'll be ready on the deprecation date. They can be added or changed later if the date changes, but its a manual task.

darjoo: I understand for the obsoleted ones (GPT40613, Turbo0631...) but for GPT4oLatest, Preview etc, we don't have any obsoletion dates.

christian-andersen-msft: I believe the current dates are based on this list: https://learn.microsoft.com/en-us/azure/ai-services/openai/concepts/model-retirements although it seems like they dont match exactly at the moment.

haoranpb (dismissed):

src/System Application/App/AI/src/Azure OpenAI/AOAIDeploymentsImpl.Codeunit.al
darjoo: Since it's the "earliest" they might, don't put them in at all. Especially because, what are they suppose to do if they receive "Deprecation date 13 May 2025", what is the mitigation step? There is none.

christian-andersen-msft: Adding a deprecation date to the list will not result in an error but act as a notification about a model in use could be deprecated on a certain date even though there is no replacement at the current point in time. The future deprecation dates can be removed from this build, but my fear is that we won't add them in time when the final deprecation dates are announced. So I would argue that its better to make the isv aware of the upcoming task to switch the models as soon as we have an estimated date, but if you think we can be on top of adding these models in time we can remove all future dates for now.

encimita: Our current AI infrastructure relies on some "special" Azure OpenAI subscriptions that are managed in a slightly different way than the normal Azure OpenAI subscriptions available to the public.

This results in sometimes different dates and timelines for model deprecation.

I checked our internal wiki page and it seems like we are using GPT 4o 2024-05-13, and our internal subscriptions don't have an obsoletion date for that yet. Same for GPT 4o mini (version 2024-07-18).

Therefore, I agree with Darrick and suggest we leave out deprecation notices for `GPT4oLatestLbl`, `GPT4oPreviewLbl`, `GPT4oMiniLatestLbl`, `GPT4oMiniPreviewLbl` for now!

encimita (commented):
Looks good!

Left a followup to Darrick's comment, I think we should fix that and then we are ready for merging the PR!

darjoo (approved):

PR #2599, opened 2024-12-23
[Draft] AL-Go setup for Extensions

aholstrup1:
#### Summary 
Add AL-Go configuration for putting extensions on BCApps. In addition to this there's still some internal wiring and an AL-Go change needed before source code can be put on BCApps. 

#### Work Item(s) 
Related to AB#555754

### DIFF
--- a/build/projects/Extensions (W1)/.AL-Go/settings.json
+++ b/build/projects/Extensions (W1)/.AL-Go/settings.json
@@ -1,0 +1,27 @@
+{
+    "projectName": "Extensions (W1)",
+    "appFolders":  [
+        "../../../src/Extensions/W1/*/app"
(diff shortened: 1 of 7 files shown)


build/projects/Extensions (W1)/.AL-Go/settings.json
```diff
@@ -0,0 +1,27 @@
+{
+    "projectName": "Extensions (W1)",
+    "appFolders":  [
+        "../../../src/Extensions/W1/*/app"
```
mazhelez: "Add-ons" instead of "Extensions"?

I remember there was a discussion about the naming before.

ankosiorek (approved):

PR #2678, opened 2025-01-07
Adding permission set

MonicaPAhuja:
#### Summary 
Adding Copilot summary Permission set

#### Work Item(s) 
Fixes [AB#561802](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/561802/)

### DIFF
--- a/src/System Application/App/System Permissions/permissions/CopilotSummary.PermissionSet.al
+++ b/src/System Application/App/System Permissions/permissions/CopilotSummary.PermissionSet.al
@@ -1,0 +1,15 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Security.AccessControl;
+
+permissionset 9680 "Copilot Summary"
+{
+    Access = Public;
+    Assignable = true;
+    Caption = 'Copilot Summary';
+
+    Permissions = system "Allow Copilot Summary" = X;
+}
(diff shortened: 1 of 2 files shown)


src/System Application/App/System Permissions/permissions/CopilotSummary.PermissionSet.al
```diff
@@ -0,0 +1,15 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Security.AccessControl;
+
+permissionset 9680 "Copilot Summary"
```
encimita: I think for newly created permission sets we are trying to stick to a naming convention that wants them to end with particles that specify the type of permission this grants.
[Internal reference here.](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_wiki/wikis/Wiki/8037/Permissions) See for example other permission sets in the same folder.

Older permission sets don't stick to this mostly for legacy and backwards compatibility reasons.

In this case, I'd suggest to go for `Copilot Summary - Exec`.

encimita (approved):
Double check the permission set name :)

src/System Application/App/System Permissions/permissions/CopilotSummary.PermissionSet.al
encimita: (seems like we are stricter on permission set name length these days, so ignore this if it doesn't fit!)

MonicaPAhuja: Yes, the length is the reason

encimita (approved):

stkillen (approved):

PR #2746, opened 2025-01-13
[AI Test Tool] Trim ToText output

stkillen:
#### Summary 
Trim ToText output in Test Input Json and Test Output Json.

#### Work Item(s) 
Fixes [AB#555687](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/555687)

### DIFF
--- a/src/Tools/Test Framework/Test Runner/src/DataDrivenTest/DataOutputs/TestOutputJson.Codeunit.al
+++ b/src/Tools/Test Framework/Test Runner/src/DataDrivenTest/DataOutputs/TestOutputJson.Codeunit.al
@@ -199,10 +199,11 @@ codeunit 130462 "Test Output Json"
         TextOutput: Text;
     begin
         TestJson.WriteTo(TextOutput);
-        if TextOutput = 'null' then
-            exit('');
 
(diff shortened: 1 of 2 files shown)


darjoo (dismissed):

src/Tools/Test Framework/Test Runner/src/DataDrivenTest/DataOutputs/TestOutputJson.Codeunit.al
```diff
@@ -202,6 +202,8 @@ codeunit 130462 "Test Output Json"
         if TextOutput = 'null' then
             exit('');
 
+        TextOutput := TextOutput.TrimStart('"').TrimEnd('"');
```
AndreasMoth, line 203: Should we do this before the test above for null? I.E. if it's '"null"' then that's still null.

AndreasMoth: Also, that check for null should also be .ToLower

AndreasMoth (approved):

darjoo (approved):

gggdttt (approved):

PR #2764, opened 2025-01-14
 Automatic access verification for AOAI services to develop and run on CAPI/managed AI resources

christian-andersen-msft:
Fixes [AB#535826](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/535826)


src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -25,6 +26,10 @@ codeunit 7767 "AOAI Authorization"
         [NonDebuggable]
         ManagedResourceDeployment: Text;
         ResourceUtilization: Enum "AOAI Resource Utilization";
+        [NonDebuggable]
```
encimita, line 31: `[NonDebuggable]` is an attribute you can add before a method or variable to indicate that the AL debugger should not show values or break execution for that variable or method.

In this case, you are tagging FirstPartyAuthorization as nondebuggable, but since it doesn't represent any private or sensitive data, I think we can remove this attribute here

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -5,7 +5,7 @@
 namespace System.AI;
 
 using System;
-
+using System.Telemetry;
```
encimita, line 8: Minor, for readability leave an empty line after the `using`s

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAI.Codeunit.al
```diff
@@ -105,6 +105,20 @@ codeunit 7771 "Azure OpenAI"
         AzureOpenAIImpl.SetManagedResourceAuthorization(ModelType, Endpoint, Deployment, ApiKey, ManagedResourceDeployment);
```
encimita, line 107: You also need to obsolete this function, by adding the Obsolete annotation before the `procedure` definition.
For example in this case something like:
```
[Obsolete('Using Managed AI resources now requires different input parameters. Use the other overload for SetManagedResourceAuthorization instead.', '26.0')]
```

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAI.Codeunit.al
```diff
@@ -105,6 +105,20 @@ codeunit 7771 "Azure OpenAI"
...
     end;
 
+    /// <summary>
+    /// Sets the managed Azure OpenAI API authorization to use for a specific model type.
+    /// This will send the Azure OpenAI call to the deployment specified in <paramref name="ManagedResourceDeployment"/>, and will use the other parameters to verify that you have access to Azure OpenAI.
+    /// </summary>
+    /// <param name="ModelType">The model type to set authorization for.</param>
+    /// <param name="AOAIAccountName">Azure OpenAI account name)</param>
```
encimita: Minor: In case of parameters that are not super obvious, I like to add a sample of how they look like in the documentation comment. For example here you could do: `An Azure OpenAI account name. For example, if your Azure OpenAI endpoint is "contoso.openai.azure.com", use "contoso" for this parameter.`

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -57,6 +62,22 @@ codeunit 7767 "AOAI Authorization"
         Deployment := NewDeployment;
         ApiKey := NewApiKey;
         ManagedResourceDeployment := NewManagedResourceDeployment;
+        MicrosoftManagedAuthorization := true;
+    end;
+
+    [NonDebuggable]
+    procedure SetMicrosoftManagedAuthorization(AOAIAccountName: Text; NewApiKey: SecretText; NewManagedResourceDeployment: Text)
```
encimita: Mmmm I suggest we keep the pattern that this file was using before (=> the `Set...` functions are just setting values, and then we check that the values are valid inside `IsConfigured`).

This saves us from possible bugs if some other part of the code is changing the authorization variables (and also delays checking the access and the validity of the values until they are actually needed).

Which means, here you just need to set the values for RecourceUtilization, ManagedResourceDeployment, and AOAIAccountName to some global variable. And then you can move the call to `VerifyAOAIAccount` to inside `IsConfigured` instead, and you don't need the three boolean variables.

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -115,5 +138,131 @@ codeunit 7767 "AOAI Authorization"
...
+
+    local procedure VerifyAOAIAccount(AOAIAccountName: Text; NewApiKey: Text): Boolean
+    var
+        Notif: Notification;
+        IsVerified: Boolean;
+        GracePeriod: Duration;
+        CachePeriod: Duration;
+        TruncatedAccountName: Text[100];
+    begin
+        GracePeriod := 15 * 60 * 1000;//14 * 24 * 60 * 60 * 1000; // 2 weeks in milliseconds
```
encimita: I guess these values are for testing?

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -115,5 +138,131 @@ codeunit 7767 "AOAI Authorization"
...
+        CachePeriod := 1 * 60 * 1000;//24 * 60 * 60 * 1000; // 1 day in milliseconds
+
+        TruncatedAccountName := CopyStr(AOAIAccountName, 1, 100);
+
+        if IsAccountVerifiedWithinPeriod(TruncatedAccountName, CachePeriod) then begin
+            Message('Verification skipped (within cache period).');
```
encimita: I guess messages here are also for testing purposes. I think it's still a good idea to log this to telemetry though, so later you can add some `Session.LogMessage` instead of the `Message`s.

src/System Application/App/AI/src/Azure OpenAI/AOAIAccountVerificationLog.Table.al
```diff
@@ -0,0 +1,35 @@
+namespace System.AI;
+
+table 7767 "AOAIAccountVerificationLog"
```
darjoo:
```suggestion
table 7767 "AOAI Account Verification Log"
```

Add spaces to the name. It is still within the object name limitation

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
darjoo: Agreed, `[NonDebuggable]` doesn't provide any value here as `FirstPartyAuthorization` doesn't contain anything private or sensitive.

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -57,6 +62,22 @@ codeunit 7767 "AOAI Authorization"
...
+        MicrosoftManagedAuthorization := true;
+    end;
+
+    [NonDebuggable]
+    procedure SetMicrosoftManagedAuthorization(AOAIAccountName: Text; NewApiKey: SecretText; NewManagedResourceDeployment: Text)
+    var
+        IsVerified: Boolean;
+    begin
+        ClearVariables();
+        IsVerified := VerifyAOAIAccount(AOAIAccountName, NewApiKey.Unwrap());
```
darjoo: Do not unwrap `NewApiKey` here.

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -115,5 +138,131 @@ codeunit 7767 "AOAI Authorization"
...
+        IsSuccessful: Boolean;
+        UrlFormatTxt: Label 'https://%1.openai.azure.com/openai/models?api-version=2024-06-01', Locked = true;
+    begin
+        Url := StrSubstNo(UrlFormatTxt, AOAIAccountName);
+
+        HttpContent.GetHeaders(ContentHeaders);
+        if ContentHeaders.Contains('Content-Type') then
+            ContentHeaders.Remove('Content-Type');
+        ContentHeaders.Add('Content-Type', 'application/json');
+        ContentHeaders.Add('api-key', NewApiKey);
```
darjoo, line 183: `ContentHeaders` should be able to accept SecretText directly. `NewApiKey` should be SecretText all the way down here.

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -115,5 +138,131 @@ codeunit 7767 "AOAI Authorization"
...
+    local procedure PerformAOAIAccountVerification(AOAIAccountName: Text; NewApiKey: Text): Boolean
+    var
+        HttpClient: HttpClient;
+        HttpRequestMessage: HttpRequestMessage;
+        HttpResponseMessage: HttpResponseMessage;
+        HttpContent: HttpContent;
+        ContentHeaders: HttpHeaders;
+        Url: Text;
+        IsSuccessful: Boolean;
+        UrlFormatTxt: Label 'https://%1.openai.azure.com/openai/models?api-version=2024-06-01', Locked = true;
```
msft-sam, line 164: Would it make sense to handle this part of the request in platform? We handle other things there like ensuring the calls are aligned on api-version (I know we spoke about this previously so just wondering if there was a better reason for doing this in sys app)

Now we will have to remember to maintain both app and platform when moving to a newer api-version (and in app we would have to HF all minor releases).

Additionally, the verification log could be a system table so all the tracking and gating could be handled there

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
christian-andersen-msft: Yes, would like to avoid prolonging the test period to two weeks

src/System Application/App/AI/src/Azure OpenAI/AOAIAccountVerificationLog.Table.al
```diff
@@ -0,0 +1,35 @@
...
+
+table 7767 "AOAI Account Verification Log"
+{
+    Caption = 'AOAI Account Verification Log';
+    Access = Internal;
+    Extensible = false;
+    InherentEntitlements = RIMDX;
+    InherentPermissions = X;
+    DataPerCompany = false;
+    ReplicateData = false;
```
NKarolak: For the human eye, could we sort the properties alphabetically, please?
Compare with  https://github.com/microsoft/BCApps/blob/main/src/System%20Application/App/AI/src/Copilot/CopilotSettings.Table.al

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -115,5 +138,178 @@ codeunit 7767 "AOAI Authorization"
         Clear(Deployment);
         Clear(ManagedResourceDeployment);
         Clear(ResourceUtilization);
+        Clear(FirstPartyAuthorization);
+        clear(SelfManagedAuthorization);
```
NKarolak:
```suggestion
        Clear(SelfManagedAuthorization);
```

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -115,5 +138,178 @@ codeunit 7767 "AOAI Authorization"
...
+        Message('Function SaveVerificationTime called. Verification successful. Record saved.');
+        exit(true);
+    end;
+
+    local procedure IsAccountVerifiedWithinPeriod(AccountName: Text[100]; Period: Duration): Boolean
+    var
+        Rec: Record "AOAI Account Verification Log";
+        IsVerified: Boolean;
+    begin
+        ;
```
NKarolak:
```suggestion
```

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -115,5 +138,178 @@ codeunit 7767 "AOAI Authorization"
...
+    begin
+        Notif.Message := MessageLbl;
+        Notif.Scope := NotificationScope::LocalScope;
+        Notif.Send();
+    end;
+
+    local procedure LogTelemetry(AccountName: Text; VerificationDate: Date)
+    var
+        Telemetry: Codeunit Telemetry;
+        MessageLbl: Label 'Azure Open AI authorization failed for account %1 on %2 because it is not authorized to access AI services. The connection will be terminated within 2 weeks if not rectified', Comment = 'Telemetry message where %1 is the name of the Azure Open AI account name and %2 is the date where verification has taken place';
```
NKarolak:
```suggestion
        MessageLbl: Label 'Azure Open AI authorization failed for account %1 on %2 because it is not authorized to access AI services. The connection will be terminated within 2 weeks if not rectified.', Comment = 'Telemetry message where %1 is the name of the Azure Open AI account name and %2 is the date where verification has taken place';
```

src/System Application/App/AI/src/Azure OpenAI/AOAIAccountVerificationLog.Table.al
```diff
@@ -0,0 +1,35 @@
...
+    {
+        field(1; AccountName; Text[100])
+        {
+            Caption = 'Account Name';
+            DataClassification = CustomerContent;
+        }
+
+        field(2; LastSuccessfulVerification; DateTime)
+        {
+            Caption = 'Access Verified';
```
NKarolak: The Caption naming indicates the field was a Boolean. What about "Last Successful Verification" instead?

NKarolak (changes requested):

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
christian-andersen-msft: The reason for keeping everything in BC Apps is that the PM chose to implement the simplest option first and then evaluate later if it was worth splitting up for maintainability.

NKarolak (approved):

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAI.Codeunit.al
```diff
@@ -100,11 +100,26 @@ codeunit 7771 "Azure OpenAI"
     /// Deployment would look like: gpt-35-turbo-16k
     /// </remarks>
     [NonDebuggable]
+    [Obsolete('Using Managed AI resources now requires different input parameters. Use the other overload for SetManagedResourceAuthorization instead.', '26.0')]
```
encimita, line 104: You also need to wrap this procedure (including docs comments) within preprocessor symbols:
```
#if not CLEAN26
    <obsoleted procedure code>
#endif
```

encimita: This will make sure our automations remove the code after the obsoletion period has passed. It's also the root cause of the failure you see in the automated tests for this PR.

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
```diff
@@ -207,6 +207,21 @@ codeunit 7772 "Azure OpenAI Impl"
         end;
```
encimita, line 131: Since you are obsoleting a function in the facade codeunit, you are leaving some code in this codeunit unused after the obsoletion period has passed.
In particular the old overload of `SetManagedResourceAuthorization` will not be called by anyone anymore and it should hence be removed after the obsoletion period has passed.

This codeunit is marked with `access=internal`, which means you don't need to explicitly obsolete the function with an `[Obsolete()]` tag because noone can reference this code outside of its own AL extension.
But you still need to make sure we don't leave unused code in the repos after the automations clean up the obsoleted code.

The way to do it here is to wrap the old overload of `SetManagedResourceAuthorization` in the tags:
```
#if not CLEAN26
    <code for the old overload of SetManagedResourceAuthorization>
#endif
```
This way, when we remove the code from the other codeunit because it's wrapped in CLEAN tags (see my other comment), this will also be removed (our automations will just remove whatever is inside the CLEAN tags, there is no smartness there, so it's up to the developer to decide what to put inside these tags).

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -57,6 +77,19 @@ codeunit 7767 "AOAI Authorization"
         Deployment := NewDeployment;
         ApiKey := NewApiKey;
         ManagedResourceDeployment := NewManagedResourceDeployment;
+        MicrosoftManagedAuthorizationWithDeployment := true;
```
encimita: The other overload of SetMicrosoftManagedAuthorization is now unused after the function in the other codeunit is removed.

So you need to wrap the old overload of SetMicrosoftManagedAuthorization  into
```
#if not CLEAN26
    <old overload>
#endif
```

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -24,26 +27,43 @@ codeunit 7767 "AOAI Authorization"
...
+                exit(SelfManagedAuthorization);
             Enum::"AOAI Resource Utilization"::"Microsoft Managed":
-                exit((Deployment <> '') and (Endpoint <> '') and (not ApiKey.IsEmpty()) and (ManagedResourceDeployment <> '') and AzureOpenAiImpl.IsTenantAllowlistedForFirstPartyCopilotCalls());
+                if MicrosoftManagedAuthorizationWithAOAIAccount then begin
+                    AOAIAccountIsVerified := VerifyAOAIAccount(AOAIAccountName, ApiKey);
+                    exit(AOAIAccountIsVerified and AzureOpenAiImpl.IsTenantAllowlistedForFirstPartyCopilotCalls());
+                end
+                else
+                    if MicrosoftManagedAuthorizationWithDeployment then
+                        exit(AzureOpenAiImpl.IsTenantAllowlistedForFirstPartyCopilotCalls());
```
encimita: With this change, we are no longer checking anywhere that the variables are not empty.

I suggest we don't add the 4 new booleans at all, and instead we rely on the existence of account name or not (for example).

Example pseudo-code:

```
Enum::"AOAI Resource Utilization"::"Microsoft Managed":
    if (AOAIAccountName <> '') and (ManagedResourceDeployment <> '')  and (not ApiKey.IsEmpty()) then
        exit(VerifyAOAIAccount(AOAIAccountName, ApiKey) and AzureOpenAiImpl.IsTenantAllowlistedForFirstPartyCopilotCalls())
    else
        exit((Deployment <> '') and (Endpoint <> '') and (not ApiKey.IsEmpty()) and (ManagedResourceDeployment <> '') and AzureOpenAiImpl.IsTenantAllowlistedForFirstPartyCopilotCalls());
```

You could even go one step further and make sure the old verification code is cleaned up automatically after the obsoletion period has passed

Example pseudo-code:


```
#if CLEAN26
    Enum::"AOAI Resource Utilization"::"Microsoft Managed":
        exit((AOAIAccountName <> '') and (ManagedResourceDeployment <> '')  and (not ApiKey.IsEmpty()) and VerifyAOAIAccount(AOAIAccountName, ApiKey) and AzureOpenAiImpl.IsTenantAllowlistedForFirstPartyCopilotCalls());
#else
    Enum::"AOAI Resource Utilization"::"Microsoft Managed":
        if (AOAIAccountName <> '') and (ManagedResourceDeployment <> '')  and (not ApiKey.IsEmpty()) then
            exit(VerifyAOAIAccount(AOAIAccountName, ApiKey) and AzureOpenAiImpl.IsTenantAllowlistedForFirstPartyCopilotCalls())
        else
            exit((Deployment <> '') and (Endpoint <> '') and (not ApiKey.IsEmpty()) and (ManagedResourceDeployment <> '') and AzureOpenAiImpl.IsTenantAllowlistedForFirstPartyCopilotCalls());
#endif
```

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -113,7 +148,173 @@ codeunit 7767 "AOAI Authorization"
...
+        end;
+
+        // Outside CACHE period
+        AccountVerified := PerformAOAIAccountVerification(AccountName, NewApiKey);
+
+        if not AccountVerified then begin
+            if VerificationLog.Get(TruncatedAccountName) then
+                RemainingGracePeriod := GracePeriod - (CurrentDateTime - VerificationLog.LastSuccessfulVerification)
+            else
+                RemainingGracePeriod := GracePeriod;
```
encimita: If there is no entry in verification log, then the remaining grace period should be 0.

It means that azure account was never verified and hence they are not entitled to grace period.

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -113,7 +148,173 @@ codeunit 7767 "AOAI Authorization"
...
+        CustomDimensions: Dictionary of [Text, Text];
+    begin
+        CustomDimensions.Add('AccountName', AccountName);
+        CustomDimensions.Add('VerificationDate', Format(VerificationDate));
+
+        Telemetry.LogMessage(
+            '0000AA1', // Event ID
+            StrSubstNo(LogMessage, AccountName, VerificationDate),
+            Verbosity::Warning,
+            DataClassification::SystemMetadata,
```
encimita: DataClassification::CustomerContent

The account name is customer content

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -113,7 +148,173 @@ codeunit 7767 "AOAI Authorization"
...
+        if not AccountVerified then begin
+            if VerificationLog.Get(TruncatedAccountName) then
+                RemainingGracePeriod := GracePeriod - (CurrentDateTime - VerificationLog.LastSuccessfulVerification)
+            else
+                RemainingGracePeriod := GracePeriod;
+
+            // Within GRACE period
+            if IsAccountVerifiedWithinPeriod(TruncatedAccountName, GracePeriod) then begin
+                ShowUserNotification(StrSubstNo(AuthFailedWithinGracePeriodUserNotificationLbl, FormatDurationAsDays(RemainingGracePeriod)));
+                LogTelemetry(AccountName, Today, StrSubstNo(AuthFailedWithinGracePeriodLogMessageLbl, AccountName, Today, FormatDurationAsDays(RemainingGracePeriod)));
```
encimita, line 265: you are doing StrSubstNo inside LogTelemetry already. It seems like we don't need it here.

encimita (changes requested):

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
christian-andersen-msft: In the privacy review it has been determined to be OII. I've updated the data classification to reflect that

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
christian-andersen-msft: The reason is was done like this is because the LogTelemetry procedure was being reused for two different log messages, one with 2 and one with 3 variables. This is not very clear in the current code and I've updated it to make it easier to read and more generic by moving the StrSubstNo out of the LogTelemetry prodcure and requrie the message to be preformatted when calling the procedure

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
christian-andersen-msft: Good catch. Validation now fails if verification failed and no current record exist

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -113,7 +144,169 @@ codeunit 7767 "AOAI Authorization"
...
+        HttpClient: HttpClient;
+        HttpRequestMessage: HttpRequestMessage;
+        HttpResponseMessage: HttpResponseMessage;
+        HttpContent: HttpContent;
+        ContentHeaders: HttpHeaders;
+        Url: Text;
+        IsSuccessful: Boolean;
+        UrlFormatTxt: Label 'https://%1.openai.azure.com/openai/models?api-version=2024-06-01', Locked = true;
+    begin
+        Url := StrSubstNo(UrlFormatTxt, AOAIAccountNameToVerify);
```
encimita, line 172: I didn't notice this before, but we want to do some validation here, such as that the name only contains allowed characters (I assume the account name can only be alphanumeric?).

What I am thinking is that if the caller provides `mypage.mydomain.com/fakeVerificationEndpoint?querystring=` then this line of code returns `https://mypage.mydomain.com/fakeVerificationEndpoint?querystring=.openai.azure.com/openai/models?api-version=2024-06-01` , which of course we don't want.

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
encimita: @christian-andersen-msft seems like you wrapped the wrong function in #CLEAN26 tags inside AOAIAuthorization.Codeunit.al (you wrapped SetSelfManagedAuthorization instead of the old overload of SetManagedResourceAuthorization)

christian-andersen-msft: Clean tag moved to the correct procedure

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
christian-andersen-msft: Thats a good point. I've now added validation of aoai account name and url before validation of account

encimita (approved):

encimita (approved):

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -113,7 +147,207 @@ codeunit 7767 "AOAI Authorization"
...
+    local procedure VerifyAOAIAccount(AccountName: Text; NewApiKey: SecretText): Boolean
+    var
+        VerificationLog: Record "AOAI Account Verification Log";
+        AccountVerified: Boolean;
+        GracePeriod: Duration;
+        CachePeriod: Duration;
+        TruncatedAccountName: Text[100];
+        IsWithinCachePeriod: Boolean;
+        RemainingGracePeriod: Duration;
+        AuthFailedWithinGracePeriodLogMessageLbl: Label 'Azure Open AI authorization failed for account %1 on %2 because it is not authorized to access AI services. The connection will be terminated within %3 if not rectified', Comment = 'Telemetry message where %1 is the name of the Azure Open AI account name, %2 is the date where verification has taken place, and %3 is the remaining time until the grace period expires';
```
NKarolak, line 236: I've seen other telemetry message labels being locked. Was this forgotten here?

christian-andersen-msft: Those four messages are for logging with the ISV and for user messages and were indented to be translated. The Internal telemetry messages which are locked are prefixed with 'Telemetry' are located at code unit level.

NKarolak (approved):

jonasblunck (approved):

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAI.Codeunit.al
```diff
@@ -91,18 +91,13 @@ codeunit 7771 "Azure OpenAI"
     /// This will send the Azure OpenAI call to the deployment specified in <paramref name="ManagedResourceDeployment"/>, and will use the other parameters to verify that you have access to Azure OpenAI.
     /// </summary>
     /// <param name="ModelType">The model type to set authorization for.</param>
-    /// <param name="Endpoint">The endpoint to use to verify access to Azure OpenAI. This is used only for verification, not for actual Azure OpenAI calls.</param>
-    /// <param name="Deployment">The deployment to use to verify access to Azure OpenAI. This is used only for verification, not for actual Azure OpenAI calls.</param>
+    /// <param name="AOAIAccountName">Azure OpenAI account name)</param>
```
dawidwoz: Minor - can be ignored - unnecessary `)`

dawidwoz (approved):

encimita (approved):

PR #2834, opened 2025-01-23
Azure Document Intelligence Module

Groenbech96:
Azure Document Intelligence support
- OnPrem for now. Uptake available for 1p apps.
- Move capabilities to Capabilities Impl Codeunit. 
- Introduce Azure AI Service Enum.
- Hide Azure Doc Int. capabilities on the copilot pages based on PM feedback.

[AB#563692](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/563692)


src/System Application/App/AI/src/DocumentIntelligence/ADIModelType.Enum.al
```diff
@@ -0,0 +1,29 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+namespace System.AI;
```
encimita: intentional?

src/System Application/App/AI/src/DocumentIntelligence/AzureDIImpl.Codeunit.al
```diff
@@ -0,0 +1,115 @@
...
+        end;
+
+        FeatureTelemetry.LogUsage('0000OLN', AzureDocumentIntelligenceCapabilityTok, TelemetryAnalyzeInvoiceCompletedLbl, Enum::"AL Telemetry Scope"::All, CustomDimensions);
+    end;
+
+    [TryFunction]
+    [NonDebuggable]
+    local procedure SendRequest(Base64Data: Text; ModelType: Enum "ADI Model Type"; CallerModuleInfo: ModuleInfo; var Result: Text)
+    var
+        ALCopilotFunctions: DotNet ALCopilotFunctions;
```
Jose-agg: @Groenbech96 you are missing the 
using System;

src/System Application/App/AI/src/DocumentIntelligence/ADIModelType.Enum.al
```diff
@@ -0,0 +1,29 @@
...
+    /// Invoice model type.
+    /// </summary>
+    value(0; Invoice)
+    {
+    }
+
+    /// <summary>
+    /// Recepit model type.
+    /// </summary>
+    value(1; Recepit)
```
NKarolak:
```suggestion
    /// Receipt model type.
    /// </summary>
    value(1; Receipt)
```

src/System Application/App/AI/src/DocumentIntelligence/AzureDIImpl.Codeunit.al
```diff
@@ -0,0 +1,135 @@
...
+    /// <param name="CallerModuleInfo">The module info of the caller.</param>
+    /// <returns>The analyzed result.</returns>
+    [NonDebuggable]
+    procedure AnalyzeReceipt(Base64Data: Text; CallerModuleInfo: ModuleInfo) Result: Text
+    var
+        CustomDimensions: Dictionary of [Text, Text];
+    begin
+        AddTelemetryCustomDimensions(CustomDimensions, CallerModuleInfo);
+
+        if not SendRequest(Base64Data, Enum::"ADI Model Type"::Recepit, CallerModuleInfo, Result) then begin
```
NKarolak:
```suggestion
        if not SendRequest(Base64Data, Enum::"ADI Model Type"::Receipt, CallerModuleInfo, Result) then begin
```

src/System Application/App/AI/src/DocumentIntelligence/AzureDIImpl.Codeunit.al
```diff
@@ -0,0 +1,135 @@
...
+        ALCopilotCapability: DotNet ALCopilotCapability;
+        ALCopilotResponse: DotNet ALCopilotOperationResponse;
+        ErrorMsg: Text;
+    begin
+        ClearLastError();
+        ALCopilotCapability := ALCopilotCapability.ALCopilotCapability(CallerModuleInfo.Publisher(), CallerModuleInfo.Id(), Format(CallerModuleInfo.AppVersion()), AzureDocumentIntelligenceCapabilityTok);
+        case ModelType of
+            Enum::"ADI Model Type"::Invoice:
+                ALCopilotResponse := ALCopilotFunctions.GenerateInvoiceIntelligence(GenerateJson(Base64Data), ALCopilotCapability);
+            Enum::"ADI Model Type"::Recepit:
```
NKarolak:
```suggestion
            Enum::"ADI Model Type"::Receipt:
```

NKarolak (changes requested):
Latin is nice, but maybe not popular enough ;-)

src/System Application/App/AI/src/Azure AI Document Intelligence/AzureDI.Codeunit.al
```diff
@@ -0,0 +1,65 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+namespace System.Azure.DI;
+
+using System.AI;
+
+/// <summary>
+/// Azure Document Intelligence implementation.
+/// </summary>
+codeunit 7780 "Azure DI"
```
darjoo:
```suggestion
codeunit 7780 "Azure Document Intelligence"
```

Lets use the full name. It fits within the 30char limit.

src/System Application/App/AI/src/Azure AI Document Intelligence/AzureDIImpl.Codeunit.al
```diff
@@ -0,0 +1,156 @@
...
+    /// <returns>The analyzed result.</returns>
+    [NonDebuggable]
+    procedure AnalyzeInvoice(Base64Data: Text; CallerModuleInfo: ModuleInfo) Result: Text
+    var
+        CustomDimensions: Dictionary of [Text, Text];
+    begin
+        AddTelemetryCustomDimensions(CustomDimensions, CallerModuleInfo);
+
+        if not SendRequest(Base64Data, Enum::"ADI Model Type"::Invoice, CallerModuleInfo, Result) then begin
+            FeatureTelemetry.LogError('0000OLK', AzureDocumentIntelligenceCapabilityTok, TelemetryAnalyzeInvoiceFailureLbl, GetLastErrorText(), '', Enum::"AL Telemetry Scope"::All, CustomDimensions);
```
darjoo, line 60: Is it possible for GetLastErrorText, which would typically be the error message from ADI, to contain any customer content information?
I don't think so, and it should be fine, but just double check.

src/System Application/App/AI/src/Azure AI Document Intelligence/AzureDIImpl.Codeunit.al
```diff
@@ -0,0 +1,156 @@
...
+
+        CustomDimensions.Add('Capability', AzureDocumentIntelligenceCapabilityTok);
+        CustomDimensions.Add('AppId', CallerModuleInfo.Id);
+        CustomDimensions.Add('Publisher', CallerModuleInfo.Publisher);
+        CustomDimensions.Add('UserLanguage', Format(GlobalLanguage()));
+
+        GlobalLanguage(SavedGlobalLanguageId);
+    end;
+
+    local procedure GenerateJsonForSingleInput(Base64: Text): Text
```
darjoo, line 118: Missing `[NonDebuggable]` and on facade, if we consider the Base64 something that should be hidden.
Otherwise, the `[NonDebuggable]` on `AnalyzeInvoice` in `AzureDIImpl` is not needed either. There are no secrets & unwraps there.

src/System Application/App/AI/src/Azure AI Document Intelligence/AzureDIImpl.Codeunit.al
darjoo: Might be nice to have the status code of the failure as well.

src/System Application/App/AI/src/Azure AI Document Intelligence/ADIModelType.Enum.al
```diff
@@ -0,0 +1,29 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+namespace System.AI;
```
darjoo: I'd switch this to DI namespace well.
```suggestion
namespace System.AI.DI;
```

darjoo: Perhaps make it System.AI.DocumentIntelligence.
More explicit naming.

src/System Application/App/AI/src/Azure AI Document Intelligence/AzureDI.Codeunit.al
```diff
@@ -0,0 +1,65 @@
...
+        AzureDIImpl: Codeunit "Azure DI Impl.";
+
+
+    /// <summary>
+    /// Analyze the invoice.
+    /// </summary>
+    /// <param name="Base64Data">Data to analyze.</param>
+    /// <returns>The analyzed result.</returns>
+    [Scope('OnPrem')]
+    procedure AnalyzeInvoice(Base64Data: Text): Text
```
darjoo: Will customers be able to use their own subscription in the future when we open this up? Or are we making them go through Managed Resources?

darjoo (commented):
Good changes for introducing ADI, minor comments.

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
```diff
@@ -156,6 +194,148 @@ codeunit 7774 "Copilot Capability Impl"
...
+        BlockList, TelemtryTok : Text;
+    begin
+        if not EnvironmentInformation.IsSaaSInfrastructure() then
+            exit(true);
+
+        NavApp.GetCurrentModuleInfo(ModuleInfo);
+        if ModuleInfo.Publisher <> 'Microsoft' then
+            exit(true);
+
+        case ServiceType of
```
darjoo: Do we need this? Difference in telemetry category.

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
```diff
@@ -156,6 +194,148 @@ codeunit 7774 "Copilot Capability Impl"
...
+        PrivacyNotice: Codeunit "Privacy Notice";
+        AzureOpenAIImpl: Codeunit "Azure OpenAI Impl";
+        AzureDIImpl: Codeunit "Azure DI Impl.";
+        CopilotNotAvailable: Page "Copilot Not Available";
+        PrivacyNoticeApprovalState: Enum "Privacy Notice Approval State";
+    begin
+        case ServiceType of
+            Enum::"Azure AI Service Type"::"Azure OpenAI":
+                PrivacyNoticeApprovalState := PrivacyNotice.GetPrivacyNoticeApprovalState(AzureOpenAIImpl.GetAzureOpenAICategory(), false);
+            Enum::"Azure AI Service Type"::"Azure Document Intelligence":
```
darjoo: Where and how is this privacy notice being controlled from? 

It doesn't look like it's used, and will always be the default("Not Set") unless they specifically go into the privacy notice page to turn off. 

If users do turn the privacy notice off, the check on line 312 and 313 will not be correct. As those open a page linking back to the Copilot & agents capabilities page about activation of the capability which would be misleading and incorrect.

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
Groenbech96: Not necessarily. I dont know if we have reports that are dependent on the tok for openai. 

I considered renaming to Azure AI Services. Maybe for now this is ok, to ensure we dont break any downstream.

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
Groenbech96: When you call GenerateText or GenerateinvoiceAnalysis, code calls IsCapabilityEnabled, which in turn calls this check privacy notice. To ensure we split between the services. So from AzureOpenAI Impl we call this with the Azure OpenAI service enum type, and vice verca from Azure Document intellgience.

msft-sam: Note that you can define required privacy notices for your capability - so the capability is only enabled once the privacy notices have been accepted, maybe you can depend on that instead of this?
Here is how it is done for [chat](https://github.com/microsoft/BCApps/blob/main/src/System%20Application/App/AI/src/Copilot/CopilotCapabilityInstall.Codeunit.al#L58C5-L74C9)

Groenbech96: There would be a problem in when a 3party defines a capability that relies on these.

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
```diff
@@ -747,15 +596,29 @@ codeunit 7772 "Azure OpenAI Impl"
...
+    begin
+        exit(AzureOpenAiTxt);
+    end;
+
+    [EventSubscriber(ObjectType::Codeunit, Codeunit::"Privacy Notice", 'OnRegisterPrivacyNotices', '', false, false)]
+    local procedure CreatePrivacyNoticeRegistrations(var TempPrivacyNotice: Record "Privacy Notice" temporary)
+    begin
+        TempPrivacyNotice.Init();
+        TempPrivacyNotice.ID := GetAzureOpenAICategory();
+        TempPrivacyNotice."Integration Service Name" := GetAzureOpenAICategory();
```
Groenbech96:
```suggestion
        TempPrivacyNotice."Integration Service Name" := AzureOpenAiTxt;
```

src/System Application/App/AI/src/Azure AI Document Intelligence/AzureDIImpl.Codeunit.al
```diff
@@ -0,0 +1,145 @@
...
+    begin
+        exit(AzureAiDocumentIntelligenceTxt);
+    end;
+
+    [EventSubscriber(ObjectType::Codeunit, Codeunit::"Privacy Notice", OnRegisterPrivacyNotices, '', false, false)]
+    local procedure CreatePrivacyNoticeRegistrations(var TempPrivacyNotice: Record "Privacy Notice" temporary)
+    begin
+        TempPrivacyNotice.Init();
+        TempPrivacyNotice.ID := GetAzureAIDocumentIntelligenceCategory();
+        TempPrivacyNotice."Integration Service Name" := GetAzureAIDocumentIntelligenceCategory();
```
Groenbech96:
```suggestion
        TempPrivacyNotice."Integration Service Name" := AzureAiDocumentIntelligenceTxt;
```

Groenbech96 (commented):
.

darjoo (dismissed):

src/System Application/App/AI/src/Azure AI Document Intelligence/AzureDocumentIntelligence.Codeunit.al
```diff
@@ -0,0 +1,76 @@
...
+    begin
+        NavApp.GetCallerModuleInfo(CallerModuleInfo);
+        AzureDIImpl.RegisterCopilotCapability(CopilotCapability, CopilotAvailability, LearnMoreUrl, CallerModuleInfo);
+    end;
+
+    /// <summary>
+    /// Sets the copilot capability that the API is running for.
+    /// </summary>
+    /// <param name="CopilotCapability">The copilot capability to set.</param>
+    [NonDebuggable]
```
mynjj: is [NonDebuggable] needed?

mynjj (approved):

src/System Application/App/AI/src/Azure AI Document Intelligence/ADIModelType.Enum.al
```diff
@@ -0,0 +1,29 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+namespace System.AI.DocumentIntelligence;
+
+/// <summary>
+/// The supported model types for Azure OpenAI.
```
encimita: Minor, I guess:
```suggestion
/// The supported model types for Azure Document Intelligence.
```

src/System Application/App/AI/src/Copilot/CopilotCapEarlyPreview.Page.al
```diff
@@ -16,7 +16,7 @@ page 7770 "Copilot Cap. Early Preview"
     Editable = false;
     Extensible = false;
     SourceTable = "Copilot Settings";
-    SourceTableView = where(Availability = const("Early Preview"));
+    SourceTableView = where(Availability = const("Early Preview"), "Service Type" = const("Azure AI Service Type"::"Azure OpenAI"));
```
encimita, line 19: Will the DI capabilities be exposed somewhere else to allow admin control over them?

src/System Application/App/AI/src/Azure AI Document Intelligence/AzureDocumentIntelligence.Codeunit.al
```diff
@@ -0,0 +1,75 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+namespace System.AI.DocumentIntelligence;
+
+using System.AI;
+
+/// <summary>
+/// Azure Document Intelligence implementation.
```
encimita:
```suggestion
/// Provides functionality to invoke Azure Document Intelligence services.
```

src/System Application/App/AI/src/Azure AI Document Intelligence/AzureDocumentIntelligence.Codeunit.al
```diff
@@ -0,0 +1,75 @@
...
+    /// <summary>
+    /// Sets the copilot capability that the API is running for.
+    /// </summary>
+    /// <param name="CopilotCapability">The copilot capability to set.</param>
+    procedure SetCopilotCapability(CopilotCapability: Enum "Copilot Capability")
+    var
+        CallerModuleInfo: ModuleInfo;
+    begin
+        NavApp.GetCallerModuleInfo(CallerModuleInfo);
+        AzureDIImpl.SetCopilotCapability(CopilotCapability, CallerModuleInfo, AzureDIImpl.GetAzureAIDocumentIntelligenceCategory());
```
encimita: Do we foresee that we will have a wide range of options for the third parameter of SetCopilotCapability?

Right now, it seems like the two values we are using are the same as in the enum `AzureAIServiceType` (formatted differently).
So, consider if you want instead to make this parameter typed with that enum instead.

encimita (approved):

src/System Application/App/AI/src/Azure AI Document Intelligence/AzureDocumentIntelligence.Codeunit.al
Groenbech96: I did not want to change existing telemetry emission. So kept it as the text instead of enum name.

src/System Application/App/AI/src/Copilot/CopilotCapEarlyPreview.Page.al
Groenbech96: Not in this release. Since functions to analyze is onprem, we just keep it pretty simple for now. 

Idea is to have AI capabilities page, and migrate the ML AI features under there, since we are not allowed to use Gen AI (copilot) page (PM decision).

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
```diff
@@ -38,19 +51,25 @@ codeunit 7774 "Copilot Capability Impl"
...
             Error(AlreadyRegisteredErr);
 
+        Clear(CopilotSettings);
         CopilotSettings.Init();
         CopilotSettings.Capability := CopilotCapability;
         CopilotSettings."App Id" := CallerModuleInfo.Id();
         CopilotSettings.Publisher := CopyStr(CallerModuleInfo.Publisher, 1, MaxStrLen(CopilotSettings.Publisher));
         CopilotSettings.Availability := CopilotAvailability;
         CopilotSettings."Learn More Url" := LearnMoreUrl;
+        CopilotSettings."Service Type" := AzureAIServiceType;
```
msft-sam, line 72: If I have a capability that makes use of both AOAI and ADI, would I need to register two capabilities?

Do we need to have a related table to store the 1..n mapping of capability to supported service types?
That way I can register the one capability and indicate which AI services it consumes

msft-sam: I guess a related question is, if I disable a capability that supports multiple services, would it disable the entire capability, or just for that particular service.

I think for now we could just disable the entire capability (keeps the logic for the capability simple)

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
```diff
@@ -751,15 +600,29 @@ codeunit 7772 "Azure OpenAI Impl"
...
+    procedure GetAzureOpenAICategory(): Code[50]
+    begin
+        exit(AzureOpenAiTxt);
+    end;
+
+    [EventSubscriber(ObjectType::Codeunit, Codeunit::"Privacy Notice", 'OnRegisterPrivacyNotices', '', false, false)]
+    local procedure CreatePrivacyNoticeRegistrations(var TempPrivacyNotice: Record "Privacy Notice" temporary)
+    begin
+        TempPrivacyNotice.Init();
+        TempPrivacyNotice.ID := GetAzureOpenAICategory();
```
msft-sam, line 633: While they are the same value, it might make sense to differentiate them, i.e. `GetAzureOpenAIPrivacyNoticeId` (for the privacy notice id_ and `GetAzureOpenAICategory` (for telemetry).

For things which differentiate based on the AI service type, it might be nice to also have something that returns the enum

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
```diff
@@ -62,9 +81,38 @@ codeunit 7774 "Copilot Capability Impl"
         Telemetry.LogMessage('0000LDV', TelemetryRegisteredNewCopilotCapabilityLbl, Verbosity::Normal, DataClassification::OrganizationIdentifiableInformation, Enum::"AL Telemetry Scope"::All, CustomDimensions);
     end;
 
+    procedure SetCopilotCapability(Capability: Enum "Copilot Capability"; CallerModuleInfo: ModuleInfo; AzureAIServiceName: Text)
```
msft-sam: Similar to the previous comment, would it make more sense if `AzureAIServiceName` was the `AzureAIServiceType` enum

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
```diff
@@ -62,9 +81,38 @@ codeunit 7774 "Copilot Capability Impl"
...
+    var
+        CopilotTelemetry: Codeunit "Copilot Telemetry";
+        Language: Codeunit Language;
+        SavedGlobalLanguageId: Integer;
+        CustomDimensions: Dictionary of [Text, Text];
+        ErrorMessage: Text;
+    begin
+        if not IsCapabilityRegistered(Capability, CallerModuleInfo.Id()) then begin
```
msft-sam, line 93: Should the check here also assert the AI service type?

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
```diff
@@ -150,6 +199,140 @@ codeunit 7774 "Copilot Capability Impl"
...
+            end;
+
+            exit(false);
+        end;
+
+        exit(CheckPrivacyNoticeState(Silent, Capability));
+    end;
+
+    [NonDebuggable]
+    local procedure IsTenantAllowedToUseAOAI(): Boolean
```
msft-sam, line 267: Should this restriction also apply to ADI?

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
Groenbech96: Lets look at that problem when it comes.

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
Groenbech96: Added GetAzureOpenAIServiceName for the label for the service name. Not going to look at the AI service type before i have a use case.

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
Groenbech96: Good point. Fixed

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
Groenbech96: Lets keep it simple for now and not do that.

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
Groenbech96: No, we are not going to use this mechanism, as it is onprem and only for 1p uptake.

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
```diff
@@ -62,9 +81,40 @@ codeunit 7774 "Copilot Capability Impl"
...
+        CopilotSettings.SetLoadFields(Status);
+        CopilotSettings.Get(Capability, CallerModuleInfo.Id());
+        if CopilotSettings.Status = Enum::"Copilot Status"::Inactive then begin
+            ErrorMessage := StrSubstNo(CapabilityNotEnabledErr, Capability);
+            Error(ErrorMessage);
+        end;
+        CopilotTelemetry.SetCopilotCapability(Capability, CallerModuleInfo.Id());
+    end;
+
     procedure ModifyCapability(CopilotCapability: Enum "Copilot Capability"; CopilotAvailability: Enum "Copilot Availability"; LearnMoreUrl: Text[2048]; CallerModuleInfo: ModuleInfo)
```
darjoo, line 116: IsCapabilityRegister and others do not use ServiceType, is that purely for not showing in the Copilot & AI capabilities page? 

Will DI capabilities not be modifiable? If the same module registers a capability on both ADI & AOAI, how is that suppose to work out? Same for unregister.

src/System Application/App/AI/src/Copilot/Interfaces/IAIServiceName.Interface.al
```diff
@@ -0,0 +1,25 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+namespace System.AI;
+
+/// <summary>
+/// Interface for providing naming information for a given AI service.
+/// </summary>
+interface IAIServiceName
```
grobyns: i'll keep saying it... in .al we don't prefix object names with their type... no C for codeunit, no I for interface...

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
Groenbech96: Since PK is app id and enum name we dont rely on the AIServiceType for that.

darjoo: That limits a capability to only one ServiceType. If that's fine, then all good. We can't change the PK now anyway :D

Groenbech96: Yes. Discussed with Sam in other thread. We would need a larger refactor, and introduce a new concept if we want multiple capabilities in one "feature". So something like "AI Feature" that has multiple capabilities under it. But thats for later.

src/System Application/App/AI/src/Copilot/Interfaces/IAIServiceName.Interface.al
Groenbech96: nit wf.

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
msft-sam: > That limits a capability to only one ServiceType. If that's fine, then all good. We can't change the PK now anyway :D

since we don't gate on the registered service type through each AI service, the scenarios won't be blocked - it just means the capability will only be managed/visible from one location, think that's fine for now. If we need to support a 1..n mapping of capability to service types then yeah we will need another table (also fine), but perhaps it is not needed (will see how requirements change 😄 )

src/System Application/App/AI/src/Copilot/AzureAIServiceType.Enum.al
```diff
@@ -0,0 +1,32 @@
...
+        Implementation = IAIServiceName = "Azure OpenAI Impl";
+    }
+
+    /// <summary>
+    /// Azure Document Intelligence service type.
+    /// </summary>
+    value(1; "Azure Document Intelligence")
+    {
+        Caption = 'Azure Document Intelligence';
+        Implementation = IAIServiceName = "Azure OpenAI Impl";
```
msft-sam: ADI?

grobyns:
You are free to ignore the comment but what you consider nitpicking, I do not... If I did, I would have said so or not commented at all.

src/System Application/App/AI/src/Copilot/Interfaces/IAIServiceName.Interface.al
darjoo: Not a nit, please fix.
We don't prefix `I` anywhere in the product unless it's platform interfaces that they added as that's a dotnet practice. This is keeping consistency within AL. 

Yes, I did find places that do it and there are only 2
- Edocs
- Email - SMTP (Which is a mistake back then)

This goes against how it has been in BC, and we shouldn't add more just because it feels like a nit.

Groenbech96: I bow before you 💯 
Fixed.

darjoo (approved):

PR #2921, opened 2025-02-11
Exposing application environment settings.

mynjj:
Exposing the EnvironmentSettings for Microsoft published callers.

[AB#565525](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/565525)

### DIFF
--- a/src/System Application/App/Environment Information/src/EnvironmentInformationImpl.Codeunit.al
+++ b/src/System Application/App/Environment Information/src/EnvironmentInformationImpl.Codeunit.al
@@ -148,6 +148,16 @@ codeunit 3702 "Environment Information Impl."
         exit(NavTenantSettingsHelper.GetLinkedPowerPlatformEnvironmentId());
     end;
 
+    procedure GetEnvironmentSetting(SettingName: Text): Text
+    var
+        ModuleInfo: ModuleInfo;
+    begin
+        NavApp.GetCallerModuleInfo(ModuleInfo);
+        if ModuleInfo.Publisher <> 'Microsoft' then
+            exit('');
+        exit(NavTenantSettingsHelper.GetEnvironmentSetting(SettingName));
+    end;
+
     [InternalEvent(false)]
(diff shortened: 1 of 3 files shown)


Groenbech96:
What does GetEnvironmentSetting actually return?

src/System Application/App/Environment Information/src/EnvironmentInformationImpl.Codeunit.al
```diff
@@ -148,6 +148,16 @@ codeunit 3702 "Environment Information Impl."
         exit(NavTenantSettingsHelper.GetLinkedPowerPlatformEnvironmentId());
     end;
 
+    procedure GetEnvironmentSetting(SettingName: Text): Text
+    var
+        ModuleInfo: ModuleInfo;
+    begin
+        NavApp.GetCallerModuleInfo(ModuleInfo);
```
darjoo: This will always be Microsoft publisher.
This needs to be on the facade for it to work properly.

darjoo (changes requested):
Why do you even need this?
Do we know the exact values that EnvironmentSettings can contain? 
From a security pov, I don't like this.

mynjj:
@darjoo 
> Why do you need this?

To control the visibility of the e-document flows that will not be visible in v26.0 when released, but that we still want feedback from private previews organized by the PM. The approach that was agreed on was to change a setting from Cloud Manager.

I added OnPrem, and the caller module filtering calls from "Microsoft" published apps.

Alternatively, we need an exception to call dotnet directly from the edocs app - that could be better than adding this, should I do that and abandon this PR?

src/System Application/App/Environment Information/src/EnvironmentInformationImpl.Codeunit.al
mynjj: thanks, moved.

mynjj:
paused until it goes through security review

darjoo (dismissed):

PR #2949, opened 2025-02-12
[Privacy] Microsoft Learn privacy notice

msft-sam:
#### Summary
- Updates the privacy notice dialog for Microsoft Learn to make it clear it is not a third party service.
- No longer provides consent by default for all cases, platform will determine if the environment can connect by default
- Environments being upgraded from 25.x will be enabled by default, this ensures there is no breaking change for existing environments.

screenshot displaying the updated privacy notice

#### Work Item(s) 
Fixes [AB#502565](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/502565)

### DIFF
--- a/src/System Application/App/DotNet Aliases/src/dotnet.al
+++ b/src/System Application/App/DotNet Aliases/src/dotnet.al
@@ -2147,6 +2147,10 @@ dotnet
         type("Microsoft.Dynamics.Nav.Service.CopilotApi.AL.ALCopilotQuotaDetails"; ALCopilotQuotaDetails)
         {
         }
+
+        type("Microsoft.Dynamics.Nav.Service.CopilotApi.AL.ALMicrosoftLearnFunctions"; ALMicrosoftLearnFunctions)
+        {
+        }
     }
     assembly("Microsoft.Dynamics.Nav.DataSearch")
     {
--- a/src/System Application/App/Privacy Notice/app.json
+++ b/src/System Application/App/Privacy Notice/app.json
@@ -16,6 +16,12 @@
       "name": "Upgrade Tags",
       "publisher": "Microsoft",
       "version": "26.0.0.0"
+    },
+    {
+      "id": "7e3b999e-1182-45d2-8b82-d5127ddba9b2",
+      "publisher": "Microsoft",
+      "name": "DotNet Aliases",
+      "version": "26.0.0.0"
     }
   ],
   "screenshots": [],
--- a/src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
+++ b/src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
@@ -345,9 +345,11 @@ codeunit 1565 "Privacy Notice Impl."
     local procedure ShouldApproveByDefault(IntegrationID: Text): Boolean
     var
         SystemPrivacyNoticeReg: Codeunit "System Privacy Notice Reg.";
+        MicrosoftLearnServiceInGeo: Boolean;
     begin
         if CheckIntegrationIDEquality(SystemPrivacyNoticeReg.GetMicrosoftLearnID(), IntegrationID) then
-            exit(false);
+            if (SystemPrivacyNoticeReg.TryGetMicrosoftLearnInGeoSupport(MicrosoftLearnServiceInGeo)) then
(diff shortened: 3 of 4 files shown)


src/System Application/App/Privacy Notice/src/MicrosoftLearn/MicrosoftLearnPrivacyNotice.Page.al
```diff
@@ -0,0 +1,116 @@
...
+        }
+    }
+    actions
+    {
+        area(Processing)
+        {
+            action(Reject)
+            {
+                ApplicationArea = All;
+                Caption = 'Back';
```
darjoo, line 71: Is this the correct term? Both actions end up closing the page. So it's not really back nor continue.

msft-sam: When the page is closed the event subscription is handled and then the privacy notice state is checked. As back only closes the page the privacy notice is still rejected and the flow will stop, continue is only enabled with the "Allow data movement" field is set, clicking continue then approves the privacy notice and then closes the page, thus the privacy notice impl. will allow the flow to continue.

darjoo (approved):

src/System Application/App/Privacy Notice/src/SystemUpgradePrivacyNotices.Codeunit.al
```diff
@@ -34,14 +35,36 @@ codeunit 1567 "System Upgrade Privacy Notices"
...
+
+        UpgradeTag.SetUpgradeTag(GetMicrosoftLearnPrivacyNoticeUpgradeTag());
+    end;
+
     [EventSubscriber(ObjectType::Codeunit, Codeunit::"Upgrade Tag", OnGetPerCompanyUpgradeTags, '', false, false)]
     local procedure RegisterPerCompanyTags(var PerCompanyUpgradeTags: List of [Code[250]])
     begin
         PerCompanyUpgradeTags.Add(GetTeamsPrivacyNoticeUpgradeTag());
+        PerCompanyUpgradeTags.Add(GetMicrosoftLearnPrivacyNoticeUpgradeTag());
     end;
```
encimita, line 59: Privacy notice and privacy notice approval tables are set as `DataPerCompany = false`, so technically this could go into the per-database tags, but I don't think it matters that much?

src/System Application/App/Privacy Notice/src/MicrosoftLearn/MicrosoftLearnPrivacyNotice.Codeunit.al
```diff
@@ -0,0 +1,35 @@
...
+
+codeunit 1569 "Microsoft Learn Privacy Notice"
+{
+    Access = Internal;
+    InherentEntitlements = X;
+    InherentPermissions = X;
+
+    var
+        ShowingPrivacyNoticeTelemetryTxt: Label 'Showing privacy notice for Microsoft Learn', Locked = true;
+        TelemetryCategoryTxt: Label 'Privacy Notice Microsoft Learn', Locked = true;
```
encimita, line 16: Minor, if the telemetry category is defined in AL I usually prefer to prefix it with AL, e.g. `AL Privacy Notice Microsoft Learn`.
I think around 60% of the AL code is doing this (which is a shame imho) but I still point it out whenever I spot it in PRs. But you are welcome to not fix it, I don't have a strong opinion :)

encimita (approved):

src/System Application/App/Privacy Notice/src/MicrosoftLearn/MicrosoftLearnPrivacyNotice.Codeunit.al
msft-sam: I've kept it consistent with the PA Priv notice - no idea if we have queries that depend on this but since we are inconsistent through the product I guess it is better to be consistent with the other privacy notices at least..

src/System Application/App/Privacy Notice/src/SystemUpgradePrivacyNotices.Codeunit.al
msft-sam: I noticed this as well, will avoid churning before 26 but will check post branching if we can safely do this

PR #2956, opened 2025-02-13
[Agents] Show notification in the Agents list page when no agents available in environment

darjoo:
#### Summary 
Show a notification in the agents page when there are no agents available to be configured.

#### Work Item(s) 
Fixes [AB#564870](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/564870)

### DIFF
--- a/src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
+++ b/src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
@@ -416,6 +416,7 @@ codeunit 4301 "Agent Impl."
         NoAgentsNotification.Scope(NotificationScope::LocalScope);
         NoAgentsNotification.AddAction(NoAgentsAvailableNotificationLearnMoreLbl, Codeunit::"Agent Impl.", 'OpenNoAgentsLearnMore');
 
+        if NoAgentsNotification.Recall() then;
         NoAgentsNotification.Send();
     end;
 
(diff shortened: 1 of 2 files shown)


src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
```diff
@@ -407,8 +407,30 @@ codeunit 4301 "Agent Impl."
...
+    procedure ShowNoAgentsAvailableNotification()
+    var
+        NoAgentsNotification: Notification;
+    begin
+        NoAgentsNotification.Id(NoAgentsAvailableNotificationGuidLbl);
+        NoAgentsNotification.Message(NoAgentsAvailableNotificationLbl);
+        NoAgentsNotification.Scope(NotificationScope::LocalScope);
+        NoAgentsNotification.AddAction(NoAgentsAvailableNotificationLearnMoreLbl, Codeunit::"Agent Impl.", 'OpenNoAgentsLearnMore');
+
+        NoAgentsNotification.Send();
```
nikolakukrika, line 420: It is good to do a blank recall before sending to avoid stacking notifications.

src/System Application/App/Agent/Setup/AgentList.Page.al
```diff
@@ -90,4 +90,14 @@ page 4316 "Agent List"
             }
         }
     }
+
+    trigger OnOpenPage()
```
nikolakukrika, line 94: I think we also need to check if the capability is turned on. If there are agents but the capability is off, then we could raise a different notification.

nikolakukrika (dismissed):

src/System Application/App/Agent/Setup/AgentList.Page.al
darjoo: When they try to set up the agent, they'd get the dialog to enable the capability. I don't think we should change/move the user's flow away from that or add complexity.

nikolakukrika (approved):

stkillen (approved):

PR #2962, opened 2025-02-14
Retention Policy: Convert total number of deleted records to BigInteger

AndreasMoth:
In some very rare cases, we delete more entries than can be stored in an Integer, hence I'm upgrading all the deletion fields to BigInteger to make us future proof.

Fixes [AB#562945](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/562945)

### DIFF
--- a/src/System Application/App/Retention Policy/src/Apply Retention Policy/RetenPolDeletingParam.Table.al
+++ b/src/System Application/App/Retention Policy/src/Apply Retention Policy/RetenPolDeletingParam.Table.al
@@ -56,11 +56,13 @@ table 3907 "Reten. Pol. Deleting Param"
         /// <summary>
         /// Indicates the maximum number of records to be deleted.
         /// </summary>
+#pragma warning disable AS0004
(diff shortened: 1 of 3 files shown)


src/System Application/App/Retention Policy/src/Apply Retention Policy/RetenPolDeletingParam.Table.al
```diff
@@ -56,7 +56,7 @@ table 3907 "Reten. Pol. Deleting Param"
         /// <summary>
         /// Indicates the maximum number of records to be deleted.
         /// </summary>
-        field(4; "Max. Number of Rec. to Delete"; Integer)
+        field(4; "Max. Number of Rec. to Delete"; BigInteger)
```
AndreasMoth, line 60: @grobyns to confirm this is not breaking partners or at least very unlikely to do so and fine to check in.

PR #2991, opened 2025-02-18
Automatic access verification for AOAI services

christian-andersen-msft:
#### Summary 
Fixes table permissions 
#### Work Item(s) 
Fixes [AB#565982](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/565982)


src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -14,8 +16,10 @@ codeunit 7767 "AOAI Authorization"
     Access = Internal;
     InherentEntitlements = X;
     InherentPermissions = X;
+    Permissions = tabledata "AOAI Account Verification Log" = RIMD;
```
darjoo: We're not giving more than indirect, please be consistent here as well.
`rimd`

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -113,7 +146,207 @@ codeunit 7767 "AOAI Authorization"
...
+    local procedure VerifyAOAIAccount(AccountName: Text; NewApiKey: SecretText): Boolean
+    var
+        VerificationLog: Record "AOAI Account Verification Log";
+        AccountVerified: Boolean;
+        GracePeriod: Duration;
+        CachePeriod: Duration;
+        TruncatedAccountName: Text[100];
+        IsWithinCachePeriod: Boolean;
+        RemainingGracePeriod: Duration;
+        AuthFailedWithinGracePeriodLogMessageLbl: Label 'Azure Open AI authorization failed for account %1 on %2 because it is not authorized to access AI services. The connection will be terminated within %3 if not rectified', Comment = 'Telemetry message where %1 is the name of the Azure Open AI account name, %2 is the date where verification has taken place, and %3 is the remaining time until the grace period expires';
```
darjoo: Place all these labels to global scope.

src/System Application/App/AI/src/Azure OpenAI/AOAIAccountVerificationLog.Table.al
```diff
@@ -0,0 +1,35 @@
+namespace System.AI;
+
+table 7767 "AOAI Account Verification Log"
```
darjoo: I don't see any deletion of this log. It's used as a cache for a certain period but nothing to removed.

I know this table won't grow exorbitantly but small things like this add up in the customer's databases

darjoo: Secondly, why do we need a table? This can easily be stored in IsolatedStorage to be loaded again. Please change the design.

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -113,7 +146,207 @@ codeunit 7767 "AOAI Authorization"
...
+    begin
+        if VerificationLog.Get(AccountName) then begin
+            IsVerified := CurrentDateTime - VerificationLog.LastSuccessfulVerification <= Period;
+            exit(IsVerified);
+        end;
+
+        exit(false);
+    end;
+
+    local procedure SaveVerificationTime(AccountName: Text[100])
```
darjoo, line 278: Related to removal of the table. Use isolatedstorage.

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -113,7 +146,207 @@ codeunit 7767 "AOAI Authorization"
...
+            VerificationLog.LastSuccessfulVerification := CurrentDateTime;
+            VerificationLog.Insert()
+        end;
+    end;
+
+    local procedure ShowUserNotification(Message: Text)
+    var
+        Notif: Notification;
+    begin
+        Notif.Message := Message;
```
darjoo, line 354: Please give each unique notification a unique guid id.

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -113,7 +146,207 @@ codeunit 7767 "AOAI Authorization"
...
+        if IsWithinCachePeriod then begin
+            Session.LogMessage('0000OLS', TelemetryAccessWithinCachePeriodTxt, Verbosity::Normal, DataClassification::SystemMetadata, TelemetryScope::ExtensionPublisher, 'Category', AzureOpenAIImpl.GetAzureOpenAICategory());
+            exit(true);
+        end;
+
+        // Outside CACHE period
+        AccountVerified := PerformAOAIAccountVerification(AccountName, NewApiKey);
+
+        if not AccountVerified then begin
+            if VerificationLog.Get(TruncatedAccountName) then
```
darjoo: What exactly is the point of the grace period if it's basically a cache period still? You can still show all 3 messages using 1 date.

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -113,7 +146,207 @@ codeunit 7767 "AOAI Authorization"
...
+        end;
+
+        Url := StrSubstNo(UrlFormatTxt, AOAIAccountNameToVerify);
+
+        if not IsValidUrl(Url, TrustedDomainTxt) then begin
+            Session.LogMessage('0000OQM', TelemetryInvalidAOAIUrlTxt, Verbosity::Warning, DataClassification::SystemMetadata, TelemetryScope::ExtensionPublisher, 'Category', AzureOpenAIImpl.GetAzureOpenAICategory());
+            exit(false);
+        end;
+
+        HttpContent.GetHeaders(ContentHeaders);
```
darjoo, line 178: Extract the Http stuff into a local function please, to keep the function clean interms of what it does.

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -113,7 +146,207 @@ codeunit 7767 "AOAI Authorization"
...
+    local procedure PerformAOAIAccountVerification(AOAIAccountNameToVerify: Text; NewApiKey: SecretText): Boolean
+    var
+        HttpClient: HttpClient;
+        HttpRequestMessage: HttpRequestMessage;
+        HttpResponseMessage: HttpResponseMessage;
+        HttpContent: HttpContent;
+        ContentHeaders: HttpHeaders;
+        Url: Text;
+        IsSuccessful: Boolean;
+        UrlFormatTxt: Label 'https://%1.openai.azure.com/openai/models?api-version=2024-06-01', Locked = true;
```
darjoo, line 168: Are we going to remember to update the api-version? Must we have it there, what happens if the api-version isn't added?

darjoo (changes requested):

src/System Application/App/AI/src/Azure OpenAI/AOAIAccountVerificationLog.Table.al
christian-andersen-msft: It would only be a single line per account which will most likely never change so it's very unlikely that the table size will go beyond a single line

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
christian-andersen-msft: The cache period is the time since last successful verification and the grace period is the time since last unsuccessful verification, when the cache duration is reached a new web request is required and when the grace period is reached verification fails. So its completely different functionality extracted from the same date.

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
christian-andersen-msft: The api-version is required to reach the endpoint. It doesn't change often and would be available even after newer versions are out. If this api version becomes deprecated and it needs upgrading it might require code changes anyways.

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
christian-andersen-msft: I would argue that all lines in this procedure are related to the http call. All logic not related to construction the http call is already outside this function.

src/System Application/App/AI/src/Azure OpenAI/AOAIAccountVerificationLog.Table.al
christian-andersen-msft: good idea, I've changed it to IsolatedStorage in the latest commit

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -185,134 +187,135 @@ codeunit 7767 "AOAI Authorization"
...
+            // Never verified - no GRACE period
+            if not IsVerificationTimeSet(AccountName) then begin
+                ShowUserNotification('Account has never been verified: ' + TruncatedAccountName);
                 exit(false);
+            end;
 
             // Within GRACE period
             if IsAccountVerifiedWithinPeriod(TruncatedAccountName, GracePeriod) then begin
+                RemainingGracePeriod := GetRemainingGracePeriod(TruncatedAccountName, GracePeriod);
+                ShowUserNotification(StrSubstNo('Account is within grace period. Remaining: %1', FormatDurationAsDays(RemainingGracePeriod)));
```
darjoo: This notification is unnecessary. The next one should be sufficient.

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -185,134 +187,135 @@ codeunit 7767 "AOAI Authorization"
...
                 ShowUserNotification(StrSubstNo(AuthFailedWithinGracePeriodUserNotificationLbl, FormatDurationAsDays(RemainingGracePeriod)));
                 LogTelemetry(AccountName, Today, StrSubstNo(AuthFailedWithinGracePeriodLogMessageLbl, AccountName, Today, FormatDurationAsDays(RemainingGracePeriod)));
                 Session.LogMessage('0000OLT', StrSubstNo(TelemetryAccessTokenWithinGracePeriodTxt, FormatDurationAsDays(RemainingGracePeriod)), Verbosity::Warning, DataClassification::SystemMetadata, TelemetryScope::ExtensionPublisher, 'Category', AzureOpenAIImpl.GetAzureOpenAICategory());
                 exit(true);
             end
             // Outside GRACE period
             else begin
+                ShowUserNotification('Account is outside grace period: ' + TruncatedAccountName);
```
darjoo: Same with this

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -185,134 +187,135 @@ codeunit 7767 "AOAI Authorization"
...
                 ShowUserNotification(AuthFailedOutsideGracePeriodUserNotificationLbl);
                 LogTelemetry(AccountName, Today, StrSubstNo(AuthFailedOutsideGracePeriodLogMessageLbl, AccountName, Today));
                 Session.LogMessage('0000OLU', TelemetryAccessTokenOutsideCachePeriodTxt, Verbosity::Warning, DataClassification::SystemMetadata, TelemetryScope::ExtensionPublisher, 'Category', AzureOpenAIImpl.GetAzureOpenAICategory());
                 exit(false);
             end;
         end;
 
+        ShowUserNotification('Account verification successful: ' + TruncatedAccountName);
```
darjoo: Please test the entire thing again. I do not think you mean to use this function in the manner you do.

These are notifications being shown to the user. and for example. This one and the one in the next line in `SaveVerificationTime`, why is `Verification time saved for account:` being shown to the user as a notification?

darjoo: A number of them seem like telemetry related rather than something the user needs to know.
I'd think all this happens in the background except for when it fails then the user needs to solve the issue. Which to me, should be errors rather than just notifications.

darjoo (changes requested):

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
christian-andersen-msft: This is just for debugging purposes. Deploying a test for the changed storage feature.

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
christian-andersen-msft: debugging

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
christian-andersen-msft: The debugging ShowUserNotificaitons will be removed when my current test are done

encimita (approved):

MonicaPAhuja (approved):

PR #3063, opened 2025-02-25
[Agents] Allow discarding input messages

esardaya:
#### Summary 

Allow discarding input messages in the Agent Task Timeline.

#### Work Item(s) 
Fixes [AB#566842](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/566842)

### DIFF
--- a/src/System Application/App/Agent/TaskPane/AgentTaskTimeline.Page.al
+++ b/src/System Application/App/Agent/TaskPane/AgentTaskTimeline.Page.al
@@ -162,6 +162,7 @@ page 4307 "Agent Task Timeline"
 
     local procedure SetTaskTimelineDetails()
     var
+        AgentTaskMessage: Record "Agent Task Message";
         InStream: InStream;
         ConfirmationLogEntryType: Enum "Agent Task Log Entry Type";
         LogEntryId: Integer;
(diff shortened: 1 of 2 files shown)


qutreson (approved):

src/System Application/App/Agent/TaskPane/AgentTaskTimeline.Page.al
```diff
@@ -240,7 +241,16 @@ page 4307 "Agent Task Timeline"
             "Agent Task Log Entry Type"::"User Intervention Request":
                 ConfirmationStatusOption := ConfirmationStatusOption::ReviewConfirmationRequired;
             "Agent Task Log Entry Type"::"User Intervention":
-                ConfirmationStatusOption := ConfirmationStatusOption::ReviewConfirmed;
+                if (Rec.Type = Rec.Type::InputMessage) or (Rec.Type = Rec.Type::OutputMessage) then begin
+                    ConfirmationStatusOption := ConfirmationStatusOption::ReviewConfirmed;
+                    if AgentTaskMessage.Get(Rec."Primary Page Record ID") then
+                        if AgentTaskMessage.Status = AgentTaskMessage.Status::Discarded then begin
```
nikolakukrika, line 247: What if we cannot get the record or if the status is different? Is it OK to exit? Should we log telemetry?

nikolakukrika (approved):

src/System Application/App/Agent/TaskPane/AgentTaskTimeline.Page.al
esardaya: We default to ReviewConfirmed, which is an acceptable status in this case.
There should be no failure to get the record since the platform also does the same before returning the virtual record.
However, it's a good idea to start logging possible issues into telemetry in general for the timeline and other parts of the agent app.
Let's create a work item and address it in general.

PR #3106, opened 2025-02-26
[Word Template] Fix unexpected behavior of AddUnrelatedTable

### LINKED ISSUE #3105: [Bug]: "Word Template".AddUnrelatedTable method not working as expected
### Describe the issue

The method `AddUnrelatedTable` from the `Word Template` codeunit has the following 2 issues, which I don't believe are by design:
1) The method does not return the result from the implementation codeunit. This results in callers always thinking that the method failed to add the record in the table.
2) In the implementation, the specified WordTemplateCode does not actually get set on the new "Word Templates Related Table" record. This results in the related table being unusable in the Word Template.

### Expected behavior

- The method should return true if the record was added; otherwise, false. (which would match the method documentation)
- The created "Word Template Related Table" record should contain the Word Template Code that was given to the method.

### Steps to reproduce

The following code demonstrates the issue and is reproducible. This will only work in the context of a Microsoft App since the problematic table is internal. One can use recordrefs to observe the same result.

```
    local procedure IssueRepro()
    var
        WordTemplateRec: Record "Word Template";
        WordTemplatesRelatedTable: Record "Word Templates Related Table"; // (Access = Internal)
        WordTemplate: Codeunit "Word Template";
        Result: Boolean;
        MyTemplateCodeLbl: Label 'MY TEMPLATE', Locked = true;
        MyTemplateNameLbl: Label 'My Template', Locked = true;
    begin
        WordTemplateRec.Init();
        WordTemplateRec.Code := MyTemplateCodeLbl;
        WordTemplateRec.Name := MyTemplateNameLbl;
        WordTemplateRec."Table ID" := Database::Microsoft.Sales.Customer.Customer;
        WordTemplateRec.Insert(true);

        CompanyInformation.Get();
        Result := WordTemplate.AddUnrelatedTable(WordTemplateRec.Code, 'COMPA', Database::"Company Information", CompanyInformation.SystemId);
        // Result should be true, but the method always returns false

        WordTemplatesRelatedTable.Get(WordTemplateRec.Code, Database::"Company Information");
        // This call will currently always fail because the Code is empty on the created record.
        // This *could* be by design, but I have a hard time believing that. As is, the created record
        // is never usable by the word template (nor does it appear in the UI since Code is empty).
    end;
```

### Additional context

Further evidence that this behavior is not part of the original design is that in the initial check for `IsRelatedTableCodeAndIdUnique`, the `WordTemplateCode` is used as a filter on the `Code` field.

Image

Image

zeande:
#### Summary
This change addresses two issues that were present in the AddUnrelatedTable method.
  1. The result of the implementation method call is returned so that callers know if the record was successfully added.
  2. The "Word Templates Related Table" record that is created has the word template code assigned to it that was provided by the caller.

#### Work Item(s) 
Fixes #3105

### DIFF
--- a/src/System Application/App/Word Templates/src/WordTemplate.Codeunit.al
+++ b/src/System Application/App/Word Templates/src/WordTemplate.Codeunit.al
@@ -297,16 +297,16 @@ codeunit 9987 "Word Template"
     end;
 
     /// <summary>
-    /// Add unrelated table.
+    /// Adds an unrelated table entry for the specified <paramref name="WordTemplateCode"/>.
     /// </summary>
     /// <param name="WordTemplateCode">The code of an existing parent Word template.</param>
     /// <param name="PrefixCode">The code of the unrelated table to add.</param>
     /// <param name="UnrelatedTableId">The ID of the unrelated table to add.</param>
     /// <param name="RecordSystemId">The system id of the record to add.</param>
-    /// <returns>True if the unrelated table was added, false otherwise.</returns>
+    /// <returns>True if the unrelated table was added; otherwise, false.</returns>
     procedure AddUnrelatedTable(WordTemplateCode: Code[30]; PrefixCode: Code[5]; UnrelatedTableId: Integer; RecordSystemId: Guid): Boolean
     begin
-        WordTemplateImpl.AddUnrelatedTable(WordTemplateCode, PrefixCode, UnrelatedTableId, RecordSystemId);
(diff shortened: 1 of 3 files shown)


zeande:
> @zeande please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.

@microsoft-github-policy-service agree [company="SwissSalary Ltd."]

zeande:
Published as draft until linked issue is approved.

zeande:
Validation seems to be broken since the linked item was not approved at the time I opened the draft. Closing this PR and creating a new one. See #3135

PR #3135, opened 2025-03-03
[Word Template] Fix unexpected behavior of AddUnrelatedTable

### LINKED ISSUE #3105: [Bug]: "Word Template".AddUnrelatedTable method not working as expected

zeande:
#### Summary
This change addresses two issues that were present in the AddUnrelatedTable method.
  1. The result of the implementation method call is returned so that callers know if the record was successfully added.
  2. The "Word Templates Related Table" record that is created has the word template code assigned to it that was provided by the caller.

#### Work Item(s) 
Fixes #3105 

Fixes [AB#567861](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/567861)

### DIFF
--- a/src/System Application/Test/Word Templates/src/WordTemplatesTest.Codeunit.al
+++ b/src/System Application/Test/Word Templates/src/WordTemplatesTest.Codeunit.al
@@ -25,7 +25,7 @@ codeunit 130443 "Word Templates Test"
 
     [Test]
     [TransactionModel(TransactionModel::AutoRollback)]
-    procedure AddUnrelatedTable_CreatesRecord()
+    procedure AddUnrelatedTableCreatesRecord()
     var
         WordTemplatesRelatedTable: Record "Word Templates Related Table";
         Any: Codeunit Any;
@@ -58,7 +58,7 @@ codeunit 130443 "Word Templates Test"
 
     [Test]
     [TransactionModel(TransactionModel::AutoRollback)]
-    procedure AddUnrelatedTable_ReturnsTrue_IfCreated()
+    procedure AddUnrelatedTableReturnsTrueIfCreated()
     var
         WordTemplatesRelatedTable: Record "Word Templates Related Table";
         Any: Codeunit Any;
@@ -81,7 +81,7 @@ codeunit 130443 "Word Templates Test"
     [Test]
     [HandlerFunctions('MessageHandler')]
     [TransactionModel(TransactionModel::AutoRollback)]
-    procedure AddUnrelatedTable_ReturnsFalse_IfNotCreated()
+    procedure AddUnrelatedTableReturnsFalseIfNotCreated()
     var
         WordTemplatesRelatedTable: Record "Word Templates Related Table";
         Any: Codeunit Any;
--- a/src/System Application/Test/Word Templates/src/WordTemplatesTest.Codeunit.al
+++ b/src/System Application/Test/Word Templates/src/WordTemplatesTest.Codeunit.al
@@ -91,8 +91,8 @@ codeunit 130443 "Word Templates Test"
         RelatedTableSystemId: Guid;
         Result: Boolean;
     begin
-        // [SCENARIO 3105] Calling AddUnrelatedTable returns false if related table already e
-        i
+        // [SCENARIO 3105] Calling AddUnrelatedTable returns false if related table already exists
+
         // [GIVEN] A Word Template Code
         TemplateCode := Any.AlphabeticText(10);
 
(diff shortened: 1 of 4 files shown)


src/System Application/Test/Word Templates/src/WordTemplatesTest.Codeunit.al
```diff
@@ -23,6 +23,98 @@ codeunit 130443 "Word Templates Test"
...
+        Result := WordTemplate.AddUnrelatedTable(Any.AlphabeticText(10), Any.AlphabeticText(5), Any.IntegerInRange(1, 100), CreateGuid());
+
+        // [THEN] The method result is true
+        Assert.IsTrue(Result, 'Unexpected method return value.');
+    end;
+
+    [Test]
+    [HandlerFunctions('MessageHandler')]
+    [TransactionModel(TransactionModel::AutoRollback)]
+    procedure AddUnrelatedTable_ReturnsFalse_IfNotCreated()
```
stkillen: Please keep naming convention for procedures

```suggestion
    procedure AddUnrelatedTableReturnsFalseIfNotCreated()
```

src/System Application/Test/Word Templates/src/WordTemplatesTest.Codeunit.al
```diff
@@ -23,6 +23,98 @@ codeunit 130443 "Word Templates Test"
...
+        WordTemplatesRelatedTable.Get(TemplateCode, Database::"Word Templates Table");
+
+        // [THEN] The values match the provided parameters
+        Assert.AreEqual(PrefixCode, WordTemplatesRelatedTable."Related Table Code", WordTemplatesRelatedTable.FieldName("Related Table Code"));
+        Assert.AreEqual(RelatedTableSystemId, WordTemplatesRelatedTable."Source Record ID", WordTemplatesRelatedTable.FieldName("Source Record ID"));
+    end;
+
+    [Test]
+    [TransactionModel(TransactionModel::AutoRollback)]
+    procedure AddUnrelatedTable_ReturnsTrue_IfCreated()
```
stkillen: Please keep naming convention for procedures

```suggestion
    procedure AddUnrelatedTableReturnsTrueIfCreated()
```

src/System Application/Test/Word Templates/src/WordTemplatesTest.Codeunit.al
```diff
@@ -23,6 +23,98 @@ codeunit 130443 "Word Templates Test"
         PermissionsMock: Codeunit "Permissions Mock";
         Assert: Codeunit "Library Assert";
 
+    [Test]
+    [TransactionModel(TransactionModel::AutoRollback)]
+    procedure AddUnrelatedTable_CreatesRecord()
```
stkillen: Please keep naming convention for procedures

```suggestion
    procedure AddUnrelatedTableCreatesRecord()
```

stkillen (commented):
Looks great - can we just fix the naming of the test methods?

JesperSchulz (approved):

PR #3195, opened 2025-03-07
Adjust CLEAN tags

mazhelez:
Adjust CLEAN tags

[AB#568604](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/568604)

### DIFF
--- a/.github/AL-Go-Settings.json
+++ b/.github/AL-Go-Settings.json
@@ -16,6 +16,7 @@
       ],
       "settings": {
         "preprocessorSymbols": [
+          "CLEAN23",
           "CLEAN24",
           "CLEAN25",
           "CLEAN26",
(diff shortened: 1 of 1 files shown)


.github/AL-Go-Settings.json
```diff
@@ -16,16 +16,10 @@
...
       "settings": {
         "preprocessorSymbols": [
-          "CLEAN17",
-          "CLEAN18",
-          "CLEAN19",
-          "CLEAN20",
-          "CLEAN21",
-          "CLEAN22",
-          "CLEAN23",
           "CLEAN24",
```
aholstrup1, line 20: I think CLEAN23 is technically still allowed (even though people shouldn't be using it of course)

aholstrup1 (dismissed):

.github/AL-Go-Settings.json
mazhelez: I added it back. Only because we will be moving other apps as well.

aholstrup1 (approved):

stkillen (approved):

PR #3225, opened 2025-03-11
[Copilot] Allow capabilities to run in background

encimita:
#### Summary 
When the AL developer toolkit was introduced, our AI systems were still in an early stage, and we did not have an infrastructure capable of supporting high volumes of calls, or throttling capabilities, or use cases that required running AI capabilities in the background.
Now we have a strong Copilot backend that can throttle high usage, measure consumption, as well as Agents that run in the background and can use AI to provide huge business values for customers and users of the system.
Removing the gating that doesn't allow using capabilities in background sessions.

#### Work Item(s) 
Fixes [AB#568894](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/568894)


EvgenijKorovin (approved):

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
```diff
@@ -501,14 +501,11 @@ codeunit 7772 "Azure OpenAI Impl" implements "AI Service Name"
...
-    local procedure GuiCheck(AOAIAuthorization: Codeunit "AOAI Authorization")
+    local procedure GuiCheck()
     begin
         if GuiAllowed() then
             exit;
 
-        if AOAIAuthorization.GetResourceUtilization() = Enum::"AOAI Resource Utilization"::"Self-Managed" then
-            exit;
-
         Error(CapabilityBackgroundErr);
```
darjoo: This doesn't remove the gating mechanism though. It enforces it for Self-Managed.
We should remove this function check entirely to allow the call to run in the background.

mynjj:
:heart:

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
```diff
@@ -502,14 +503,17 @@ codeunit 7772 "Azure OpenAI Impl" implements "AI Service Name"
...
+        ClientTypeManagement: Codeunit "Client Type Management";
     begin
-        if GuiAllowed() then
-            exit;
-
         if AOAIAuthorization.GetResourceUtilization() = Enum::"AOAI Resource Utilization"::"Self-Managed" then
             exit;
 
-        Error(CapabilityBackgroundErr);
+        if ClientTypeManagement.GetCurrentClientType() in [ClientType::Api, ClientType::OData, ClientType::ODataV4, ClientType::SOAP, ClientType::Management] then
```
darjoo, line 512: Is there a reason to block API/OData? Why is that different from a scheduled task session?

encimita: The main reason is that we don't yet have a use case for that, so keeping it locked reduces the surface for bugs that might cause repeated calls to the LLM.

PR #3277, opened 2025-03-14
Fix media cleanup runtime error

### LINKED ISSUE #3271: [Bug]: Can't cleanup media
### Describe the issue

If I use the codeunit 1929 "Media Cleanup Runner" to cleanup my media data the "Job Queue" runs on an error.
The problem is in codeunit 1928 "Media Cleanup Impl." at function "SplitListIntoSubLists" at the "round" command. This command will round "rounds up or down to the nearest value" if the parameter "Direction" is not set. 
Example:
List 322 entry, split by 100 will result in [[1, 100], [101, 200], [201, 300], [301, 322]]
The function "SplitListIntoSubLists" will split a list in sublists by (at example) 100 entry.
So the function round by 322 counted entry down to 3 and runs 4 times in the loop.
If we have 351 entry, it runds up to 4 an runs 5 times in the loop. That is one time to mutch.

### Expected behavior

I suggest to use the parameter "Direction" from the round function with value "<" to round down every time.
```
   local procedure SplitListIntoSubLists(var InputList: List of [Guid]; SubListCount: Integer; var SplitList: List of [List of [Guid]])
    var
        Math: Codeunit Math;
        ListNumber: Integer;
        SubList: List of [Guid];
        From: Integer;
        ToInt: Integer;
    begin
        // old code
        // for ListNumber := 0 to Round(InputList.Count() / SubListCount, 1) do begin
        // new code
       for ListNumber := 0 to Round(InputList.Count() / SubListCount, 1, '<') do begin
            Clear(SubList);
            From := ListNumber * SubListCount + 1;
            ToInt := Math.Min(SubListCount, InputList.Count() - ListNumber * SubListCount);
            SubList := InputList.GetRange(From, ToInt);
            SplitList.Add(SubList);
        end;
    end;
```

### Steps to reproduce

Use the data administration in BC to find to clean tennant media.
See that the number of records is divid by 100 has an remainder of 50 or more.
Clean the tennant media by using Codeunit 1929 "Media Cleanup Runner" in "Job Queue".

### Additional context

The importent thing is the round command that give back diffrent result if there 149 or 150 entrys:
**149 Entrys**
Round(149/100, 1) --> 149/100 = 1,49 --> 1,49 round(1) = 1
**150 Entrys**
Round(150/100, 1) --> 150/100 = 1,5 --> 1,5 round(1) = **2**

If you use the parameter "direction" with "<", the command will give the same result in every case:
**149 Entrys**
Round(149/100, 1, '<') --> 149/100 = 1,49 --> 1,49 round(1, '<') = 1
**150 Entrys**
Round(150/100, 1, '<') --> 150/100 = 1,5 --> 1,5 round(1, '<') = **1**

Drakonian:
#### Summary 

Fix media cleanup runtime error
Update SplitListIntoSubLists procedure to get consistent result
Tests for edge cases for SplitListIntoSubLists and Media Cleanup

#### Work Item(s) 
Fixes #3271

Fixes [AB#571724](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/571724)

### DIFF
--- a/src/System Application/App/Data Administration/src/MediaCleanupImpl.Codeunit.al
+++ b/src/System Application/App/Data Administration/src/MediaCleanupImpl.Codeunit.al
@@ -5,7 +5,6 @@
 
 namespace System.DataAdministration;
 
-using System.Utilities;
 using System.Environment;
 
 codeunit 1928 "Media Cleanup Impl."
@@ -245,20 +244,24 @@ codeunit 1928 "Media Cleanup Impl."
     end;
 
     // 322, 100 will result in [[1, 100], [101, 200], [201, 300], [301, 322]]
-    local procedure SplitListIntoSubLists(var InputList: List of [Guid]; SubListCount: Integer; var SplitList: List of [List of [Guid]])
+    procedure SplitListIntoSubLists(var InputList: List of [Guid]; SubListCount: Integer; var SplitList: List of [List of [Guid]])
     var
-        Math: Codeunit Math;
         ListNumber: Integer;
-        SubList: List of [Guid];
-        From: Integer;
-        ToInt: Integer;
+        FromIndex: Integer;
+        SubSize: Integer;
+        NumberOfBatches: Integer;
+        TempSubList: List of [Guid];
     begin
-        for ListNumber := 0 to Round(InputList.Count() / SubListCount, 1) do begin
-            Clear(SubList);
-            From := ListNumber * SubListCount + 1;
-            ToInt := Math.Min(SubListCount, InputList.Count() - ListNumber * SubListCount);
-            SubList := InputList.GetRange(From, ToInt);
-            SplitList.Add(SubList);
+        NumberOfBatches := (InputList.Count() + SubListCount - 1) div SubListCount;
+
+        for ListNumber := 1 to NumberOfBatches do begin
+            Clear(TempSubList);
+            FromIndex := (ListNumber - 1) * SubListCount + 1;
+            SubSize := InputList.Count() - (FromIndex - 1);
+            if SubSize > SubListCount then
+                SubSize := SubListCount;
+            TempSubList := InputList.GetRange(FromIndex, SubSize);
+            SplitList.Add(TempSubList);
         end;
     end;
 }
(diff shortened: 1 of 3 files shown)


src/System Application/App/Data Administration/src/MediaCleanupImpl.Codeunit.al
```diff
@@ -245,20 +244,24 @@ codeunit 1928 "Media Cleanup Impl."
     end;
 
     // 322, 100 will result in [[1, 100], [101, 200], [201, 300], [301, 322]]
-    local procedure SplitListIntoSubLists(var InputList: List of [Guid]; SubListCount: Integer; var SplitList: List of [List of [Guid]])
+    procedure SplitListIntoSubLists(var InputList: List of [Guid]; SubListCount: Integer; var SplitList: List of [List of [Guid]])
     var
-        Math: Codeunit Math;
```
pri-kise, line 250: Since the Math Codeunit is now removed in this procedure.
Could you please check if the dependency can be removed?

src/System Application/Test/Data Administration/src/MediaCleanupTest.Codeunit.al
```diff
@@ -199,6 +288,146 @@ codeunit 135018 "Media Cleanup Test"
         LibraryAssert.IsFalse(TenantMediaSetup.IsEmpty(), 'Normal tenant media set is also affected.');
     end;
```
pri-kise, line 290: Those tests looks good to me, but does this test app have access to the internals of the data administration app?

src/System Application/App/Data Administration/src/MediaCleanupImpl.Codeunit.al
Drakonian: What dependency?
I've already removed `using System.Utilities;`

pri-kise: The dependency in the app.json of the Data Administration app

src/System Application/Test/Data Administration/src/MediaCleanupTest.Codeunit.al
Drakonian: Interesting, it didn't seem to be, but it compiled successfully for me. Using F5 or NewDevEnv with -RebuildApps.

Anyway, I added internalVisibleTo.

pri-kise (approved):

JesperSchulz (approved):

darjoo (approved):

PR #3287, opened 2025-03-17
No try functions in Page Summary

mynjj:
#### Summary 
In the Payables Agent, the agent can finish in the Purchase Invoice page, this triggers the summarization in this page within a try function.

Try functions don't allow write transactions to be started, so if the onopen logic requires write transactions, it crashes, in this particular case inserting a notification is crashing it.

#### Work Item(s) 
Fixes [AB#575510](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/575510)


src/System Application/App/Page Summary Provider/src/PageSummaryProviderImpl.Codeunit.al
```diff
@@ -192,20 +192,19 @@ codeunit 2717 "Page Summary Provider Impl."
         end;
 
         // Get summary fields
-        if not TryGetPageSummaryFields(PageId, RecId, Bookmark, ResultJsonObject, IncludeBinaryData) then
+        if not GetPageSummaryFields(PageId, RecId, Bookmark, ResultJsonObject, IncludeBinaryData) then
```
nikolakukrika, line 195: Maybe we should get the ability to run this code from an isolated event or via codeunit so it does not throw an error, so we have a way of safe running the code.

nikolakukrika (approved):

src/System Application/App/Page Summary Provider/src/PageSummaryProviderImpl.Codeunit.al
mynjj: the code in the platform that this dotnet calls to seems to already be catching plenty of the exceptions, so I think we were just being overly careful - but let's see what client experience thinks about it 😄 - I can do that too

mynjj: resolving, the exception handling was moved to platform

mynjj:
pending platform uptake

MonicaPAhuja (approved):

PR #3481, opened 2025-04-08
Fix installed extension version mismatch

blrobl:
#### Summary 
There is an extension version mismatch between the page Installed Extensions and Extension setting.

The problem arises because the "Extension Settings" page uses the App Id when navigated to from the "Installed Extensions" page, without distinguishing amongst the different runtime packages. 

#### Work Item(s) 
Fixes AB#567978

### DIFF
--- a/src/System Application/App/Extension Management/src/ExtensionSettings.Page.al
+++ b/src/System Application/App/Extension Management/src/ExtensionSettings.Page.al
@@ -145,6 +145,7 @@ page 2511 "Extension Settings"
         PublishedApplication: Record "Published Application";
     begin
         PublishedApplication.SetRange(ID, Rec."App ID");
+        PublishedApplication.SetRange(Installed, true);
         PublishedApplication.SetRange("Tenant Visible", true);
 
         if PublishedApplication.FindFirst() then begin


src/System Application/App/Extension Management/src/ExtensionSettings.Page.al
```diff
@@ -181,6 +183,14 @@ page 2511 "Extension Settings"
         CanManageExtensions := ExtensionInstallationImpl.CanManageExtensions();
     end;
 
+    local procedure SetInstalledFilter(var PublishedApplication: Record "Published Application")
+    begin
+        PublishedApplication.SetRange(Installed, true);
```
thloke: Is this correct? The definition of `Installed` in `Published Application` is that it exists in the `Installed Application` table.

https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_git/Platform-Core?path=/Platform/Symbols/System/System%20Tables/Table%20-%20Published%20Application.al&version=GBmaster&_a=contents&line=111&lineStyle=plain&lineEnd=119&lineStartColumn=1&lineEndColumn=10

This means that if **any** tenant has this app installed, it will be marked as installed here. Not necessarily the tenant you're looking at. For PTEs this is fine, but for Global apps this will be wrong.

thloke: The bigger problem here is that "Published Application" is an app database table, not a tenant one.

thloke (commented):
Please test in a multitenant setting first.

src/System Application/App/Extension Management/src/ExtensionSettings.Page.al
```diff
@@ -146,6 +146,9 @@ page 2511 "Extension Settings"
     begin
         PublishedApplication.SetRange(ID, Rec."App ID");
         PublishedApplication.SetRange("Tenant Visible", true);
+        PublishedApplication.SetRange("Tenant ID", TenantID());
```
thloke: This is only populated for PTEs, since it doesn't make sense for Global apps.

I'm pretty sure you can't get the information you need only from the Published Application table, but you need to cross-refer to the Installed Application table at least.

SBalslev (approved):

PR #3520, opened 2025-04-15
Support extensions in BCApps

aholstrup1:
This PR adds the ability to build and test extensions/add-ons in the BCApps repository. 

In default/translated mode the build steps are: 
1. Uninstall all apps (but keep them published in the environment)
2. Compile the app/test apps using the System App/Business Foundation from BCApps and the Base App from the BC Artifact. 
3. Publish the apps into the container
4. Generate demo data
5. Re-install all the apps that are currently published but not installed. 
6. Run tests

In CLEAN mode the build steps are: 
1. Create a compiler folder 
2. Grab the source code from the BC Artifact and recompile all the external dependencies in CLEAN mode. 
3. Compile the add-ons in CLEAN mode

**UseProjectDependencies**
By default the extensions are built with useProjectDependencies set to true but in some cases this may need to be set to false. This PR supports that configuration as well. In that case the System Application / Business Foundation is taken from the BC Artifact when we build the add-ons instead of using the ones built in the BCApps repo. 

Additionally, the "Update AL-Go Projects" workflow would help ensure that useProjectDependencies is reverted to true again. 

Example workflow run (Ignore the test failures for now): https://github.com/aholstrup1/BCApps/actions/runs/14708621591
Example workflow run with useProjectDependencies = false (Ignore the test failures for now): https://github.com/aholstrup1/BCApps/actions/runs/14710628005

**Workflow Items**
[AB#555754](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/555754)
[AB#555741](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/555741)

### DIFF
--- a/build/scripts/AppExtensionsHelper.psm1
+++ b/build/scripts/AppExtensionsHelper.psm1
@@ -1,10 +1,10 @@
 function GetSourceCodeFromArtifact() {
     param(
-        [string] $App,
+        [string] $AppName,
         [string] $TempFolder
     )
-    $sourceArchive = Get-ChildItem -Path $TempFolder -Recurse -Filter "$App.Source.zip" -ErrorAction SilentlyContinue
-    $sourceCodeFolder = "$TempFolder/$($App -replace " ", "_")Source"
+    $sourceArchive = Get-ChildItem -Path $TempFolder -Recurse -Filter "$AppName.Source.zip" -ErrorAction SilentlyContinue
+    $sourceCodeFolder = "$TempFolder/$($AppName -replace " ", "_")Source"
     # Return source code folder if it exists
     if (Test-Path $sourceCodeFolder) {
         Write-Host "Source code folder already exists: $sourceCodeFolder"
@@ -33,13 +33,13 @@ function GetSourceCodeFromArtifact() {
         Download-Artifacts -artifactUrl $artifactVersion -basePath $TempFolder -includePlatform | Out-Null
 
         # Unzip it
-        $sourceArchive = Get-ChildItem -Path $TempFolder -Recurse -Filter "$App.Source.zip"
+        $sourceArchive = Get-ChildItem -Path $TempFolder -Recurse -Filter "$AppName.Source.zip"
     }
 
     $sourceArchive | Expand-Archive -Destination $sourceCodeFolder
 
     if (-not (Test-Path $sourceCodeFolder)) {
-        Write-Error "Could not find the source code for $App"
+        Write-Error "Could not find the source code for $AppName"
         throw
     }
 
@@ -59,7 +59,7 @@ function GetSourceCodeFromArtifact() {
 #>
 function Build-App() {
     param(
-        [string] $App,
+        [string] $AppName,
         [hashtable] $CompilationParameters
     )
     # Set up temp folder if not already set
@@ -76,9 +76,9 @@ function Build-App() {
 
     # If app is already there then skip it
     if (Test-Path $CompilationParameters["appOutputFolder"]) {
-        $appSymbolsExist = Get-ChildItem -Path $CompilationParameters["appOutputFolder"] | Where-Object { $_.Name -like "Microsoft_$($App)*.app" }
+        $appSymbolsExist = Get-ChildItem -Path $CompilationParameters["appOutputFolder"] | Where-Object { $_.Name -like "Microsoft_$($AppName)*.app" }
         if ($appSymbolsExist) {
-            Write-Host "$App is already in the symbols folder. Skipping recompilation"
+            Write-Host "$AppName is already in the symbols folder. Skipping recompilation"
             return
         }
     }
@@ -88,13 +88,13 @@ function Build-App() {
     $sourceCodeFolder = $null
     # If we are using project dependencies we will try to find the source code within the repository
     if ($alGoSettings.useProjectDependencies) {
-        Write-Host "Get source code for $App"
+        Write-Host "Get source code for $AppName"
         Import-Module $PSScriptRoot\EnlistmentHelperFunctions.psm1
         $sourceCodeFolder = Get-ChildItem -Path (Get-BaseFolder) -Filter "app.json" -Recurse | ForEach-Object {
-            $appName = (Get-Content -Path $_.FullName | ConvertFrom-Json).Name
-            if ($appName -eq $App) {
+            $appJsonName = (Get-Content -Path $_.FullName | ConvertFrom-Json).Name
+            if ($appJsonName -eq $AppName) {
                 $sourceCodeFolder = Split-Path $_.FullName -Parent
-                Write-Host "Found code for $App in $sourceCodeFolder"
+                Write-Host "Found code for $AppName in $sourceCodeFolder"
                 return $sourceCodeFolder
             }
         }
@@ -102,7 +102,7 @@ function Build-App() {
 
     # If we didn't find the source code in the repository, we will try to get it from the artifact
     if (-not $sourceCodeFolder) {
-        $sourceCodeFolder = GetSourceCodeFromArtifact -App $App -TempFolder $script:tempFolder
+        $sourceCodeFolder = GetSourceCodeFromArtifact -AppName $AppName -TempFolder $script:tempFolder
     }
 
     # Update the CompilationParameters
@@ -115,7 +115,7 @@ function Build-App() {
     $CompilationParameters["EnablePerTenantExtensionCop"] = $false
     $CompilationParameters.Remove("ruleset")
 
-    Write-Host "Recompile $App with parameters"
+    Write-Host "Recompile $AppName with parameters"
     foreach ($key in $CompilationParameters.Keys) {
         Write-Host "$key : $($CompilationParameters[$key])"
     }
--- a/build/scripts/InstallMissingDependencies.ps1
+++ b/build/scripts/InstallMissingDependencies.ps1
@@ -1,4 +1,4 @@
-[Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSReviewUnusedParameter', 'parameters', Justification = 'The parameter is not used, but it''s script needs to match this format')]
+[Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSReviewUnusedParameter', 'parameters', Justification = 'The parameter is not used, but the script needs to match this format')]
 Param(
     [hashtable] $parameters,
     [string[]] $dependenciesToInstall = @()


.github/workflows/UpdateALGoProjects.yaml
```diff
@@ -0,0 +1,65 @@
...
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@4d991eb9b905ef189e4c376166672c3f2f230481 # v2.11.0
+        with:
+          egress-policy: audit
+
+      - name: Get Official Branches
+        id: getOfficialBranches
+        uses: microsoft/BCApps/.github/actions/GetGitBranches@main
+
```
mazhelez:
```suggestion
```

.github/workflows/UpdateALGoProjects.yaml
```diff
@@ -0,0 +1,65 @@
...
+        with:
+          include: "['main', 'releases/*']"
+
+  UpdateBCArtifactVersion:
```
mazhelez:
```suggestion
  UpdateALGoProjects:
```

.github/workflows/UpdateALGoProjects.yaml
```diff
@@ -0,0 +1,65 @@
...
+    name: "[${{ matrix.branch }}] Update BC Artifact Version"
```
mazhelez:
```suggestion
    name: "[${{ matrix.branch }}] Update AL-Go projects"
```

Or something like it.

.github/workflows/UpdateALGoProjects.yaml
```diff
@@ -0,0 +1,65 @@
...
+        with:
+          ref: ${{ matrix.branch }}
+
+      - uses: actions/create-github-app-token@v1
+        id: app-token
+        with:
+          app-id: ${{ vars.APP_ID }}
+          private-key: ${{ secrets.PRIVATE_KEY }}
+
+      - name: Update BC Artifact Version
```
mazhelez:
```suggestion
      - name: Update AL-Go Project Dependencies
```

build/scripts/AppExtensionsHelper.psm1
```diff
@@ -0,0 +1,264 @@
...
+        # Test that artifact is a url
+        if ($artifact -notmatch "^https?://") {
+            Write-Error "Artifact is not a valid URL: $artifact"
+            throw
+        }
+
+        $artifactVersion = $artifact -replace "/[^/]+$", "/w1"
+
+        # Download the artifact that contains the source code for those apps
+        Download-Artifacts -artifactUrl $artifactVersion -basePath $TempFolder -includePlatform | Out-Null
```
mazhelez, line 33: Is there a way to cache this? Downloading it all the time seems unnecessary.

Alternatively, what about just switching to /w1 artifact altogether?

aholstrup1: > Is there a way to cache this? Downloading it all the time seems unnecessary.


You mean in the .githubcache or? We only download it once in the CLEAN mode build for add-ons. We don't download it at all in the other build modes 

> Alternatively, what about just switching to /w1 artifact altogether?

Ah yeah... You mean in the repo settings? Rather than using the base artifact?

mazhelez: >You mean in the .githubcache or? We only download it once in the CLEAN mode build for add-ons. We don't download it at all in the other build modes

The cache that can be shared between dependencies: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows

> Ah yeah... You mean in the repo settings? Rather than using the base artifact?

Indeed.

aholstrup1: > The cache that can be shared between dependencies: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows

I can't call the cache action since the .yaml files are controlled by AL-Go 🤔 Our cache is also already full so I'm not sure we'll end up getting a lot out of caching base artifacts AND w1 artifacts 
https://github.com/microsoft/BCApps/actions/caches

We could switch to using the W1 artifact all over but that would also mean all the other projects would download the W1 artifact. Since the W1 artifact is 2,5 times bigger than the base artifact, it might end up hurting us more than it benefits us 🤷‍♂️

build/scripts/ImportTestDataInBcContainer.ps1
```diff
@@ -0,0 +1,36 @@
...
+        [string]$DemodataApp = "Contoso Coffee Demo Dataset"
+    )
+    Install-AppFromContainer -ContainerName $ContainerName -DependenciesToInstall @($DemodataApp)
+
+    Write-Host "Initializing company in container $ContainerName"
+    Invoke-NavContainerCodeunit -Codeunitid 2 -containerName $ContainerName -CompanyName $CompanyName
+
+    if ($SetupData) {
+        Write-Host "Setting up demo data in container $ContainerName"
+        Invoke-NavContainerCodeunit -Codeunitid 2 -containerName $ContainerName -CompanyName $CompanyName -MethodName "CreateSetupDemoData"
```
haoranpb, line 21:
```suggestion
        Invoke-NavContainerCodeunit -Codeunitid 5193 -containerName $ContainerName -CompanyName $CompanyName -MethodName "CreateSetupDemoData"
```

build/scripts/ImportTestDataInBcContainer.ps1
```diff
@@ -0,0 +1,36 @@
...
+    } else {
+        Write-Host "Importing configuration package"
+        Invoke-NavContainerCodeunit -CodeunitId 5193 -containerName $containerName -CompanyName $CompanyName -MethodName "CreateAllDemoData"
+        Invoke-NavContainerCodeunit -CodeunitId 5140 -containerName $containerName -CompanyName $CompanyName -MethodName "DeleteWarehouseEmployee"
+    }
+
+    Invoke-NavContainerCodeunit -CodeunitId 5691 -containerName $ContainerName -CompanyName $CompanyName
```
haoranpb:
```suggestion
    if ($SetupData) {
        Write-Host "Generating Setup Demo Data in container $ContainerName"
        Invoke-NavContainerCodeunit -Codeunitid 5193 -containerName $ContainerName -CompanyName $CompanyName -MethodName "CreateSetupDemoData"
    } else {
        Write-Host "Generating All Demo Data in container $ContainerName"
        Invoke-NavContainerCodeunit -CodeunitId 5193 -containerName $containerName -CompanyName $CompanyName -MethodName "CreateAllDemoData"
        Invoke-NavContainerCodeunit -CodeunitId 5140 -containerName $containerName -CompanyName $CompanyName -MethodName "DeleteWarehouseEmployee"
    }

    Invoke-NavContainerCodeunit -CodeunitId 5691 -containerName $ContainerName -CompanyName $CompanyName
```

build/scripts/ImportTestDataInBcContainer.ps1
```diff
@@ -0,0 +1,36 @@
...
+)
+
+Import-Module $PSScriptRoot\AppExtensionsHelper.psm1
+
+function Invoke-ContosoDemoTool() {
+    param(
+        [string]$ContainerName,
+        [string]$CompanyName = (Get-NavDefaultCompanyName),
+        [switch]$SetupData = $false,
+        [string]$DemodataApp = "Contoso Coffee Demo Dataset"
```
haoranpb: You will probably need a few downstream apps as well (search for `ContosoDemoData` in `groups.json`).

They are not explicitly called but the code is run by subscribing some events.

---

We have an upcoming discussion about the implicit dependency between tests and demo data.

aholstrup1: Moved some logic around now such that all the demo data apps are installed when we generate demo data

build/scripts/AppExtensionsHelper.psm1
```diff
@@ -0,0 +1,269 @@
...
+        Build a dependency app from source code and place it in the symbols folder for the app.
+    .Description
+        This function will build a dependency app from source code and place it in the symbols folder for the app.
+        The source code is downloaded from the artifact and the app is built with the same parameters as the main app.
+    .Parameter App
+        The name of the app to build.
+    .Parameter CompilationParameters
+        The parameters to use for the compilation of the app. This should be the same as the parameters used for the main app.
+#>
+function Build-Dependency() {
```
mazhelez: This sounds more like `Build-App` function. I understand where the name comes from, but nothing is the code suggests a dependency.

build/scripts/AppExtensionsHelper.psm1
```diff
@@ -0,0 +1,269 @@
...
+        [Parameter(Mandatory = $true)]
+        [string] $ContainerName,
+        [Parameter(Mandatory = $true, ParameterSetName = "ByAppFilePath")]
+        [string] $AppFilePath,
+        [Parameter(Mandatory = $true, ParameterSetName = "ByAppName")]
+        [string] $AppName
+    )
+    if ($PSCmdlet.ParameterSetName -eq "ByAppName") {
+        Write-Host "[Install App from file] - Searching for app file with name: $AppName"
+        $allApps = (Invoke-ScriptInBCContainer -containerName $ContainerName -scriptblock { Get-ChildItem -Path "C:\Applications\" -Filter "*.app" -Recurse })
```
mazhelez, line 150: `C:\Applications\` seems quite specific, where does it come from? Is it some BCContainerHelper-specific location?

build/scripts/AppExtensionsHelper.psm1
```diff
@@ -0,0 +1,269 @@
...
+        $AppFilePath = $allApps | Where-Object { $($_.BaseName) -like "*$($AppName)" } | ForEach-Object { $_.FullName }
```
mazhelez: Isn't the app files format `<publisher>_<app name>_<app version>`? What is the assumed format here?

build/scripts/AppExtensionsHelper.psm1
```diff
@@ -0,0 +1,269 @@
...
+        The name of the container to install the dependencies in.
+    .Parameter DependenciesToInstall
+        The list of dependencies to install.
+#>
+function Install-AppFromContainer() {
+    param(
+        [Parameter(Mandatory = $true)]
+        [string] $ContainerName,
+        [Parameter(Mandatory = $true)]
+        [string[]] $DependenciesToInstall
```
mazhelez: Just `$AppsToInstall`?

build/projects/Add-Ons (W1)/.AL-Go/settings.json
```diff
@@ -0,0 +1,36 @@
...
+    ],
+    "installOnlyReferencedApps": false,
+    "ConditionalSettings": [
+        {
+            "buildModes": [
+                "Clean"
+            ],
+            "settings": {
+                "doNotRunTests": true,
+                "useCompilerFolder": true,
```
mazhelez, line 22: There are scripts overrides to deal with running tests, etc., but these settings suggest tests are only built and with a compiler folder, not a container.

build/projects/Add-Ons (W1)/.AL-Go/PreCompileApp.ps1
```diff
@@ -0,0 +1,20 @@
+Param(
+    [string] $appType,
+    [ref] $compilationParams
+)
+
+if ($ENV:BuildMode -eq 'Clean') {
+
+    # If useProjectDependencies is set to false, we need to remove the app symbols folder so we recompile everything.
+    $settings = Get-Content (Join-Path $PSScriptRoot "settings.json" -Resolve) | ConvertFrom-Json
```
mazhelez: Aren't AL-Go settings (`$env:settings`) available here?

build/projects/Add-Ons (W1)/.AL-Go/PreCompileApp.ps1
```diff
@@ -0,0 +1,20 @@
...
+    if ($settings.useProjectDependencies -eq $false) {
+        $symbolsFolder = $compilationParams.Value["appSymbolsFolder"]
+        if (Test-Path $symbolsFolder) {
+            Remove-Item -Path $symbolsFolder\* -Recurse -Force -Verbose
```
mazhelez: Wouldn't that mean that the external dependencies will be recompiled for every add-on app? There should be a smarter way to recompile them once.

build/scripts/PreCompileApp.ps1
```diff
@@ -69,3 +77,8 @@ if($appType -eq 'app')
         }
     }
 }
+
+if($useCompilerFolder) {
```
mazhelez: Not sure why this is needed. The assembly probing paths are set when using a compiler folder.

build/scripts/AppExtensionsHelper.psm1
mazhelez: Ok, then let's see how it goes as it is, and optimize if needed.

build/scripts/AppExtensionsHelper.psm1
aholstrup1: It's a folder on the container. It's also where containerhelper finds test toolkit apps when you run Import-TestToolkitToBcContainer: 
https://github.com/microsoft/navcontainerhelper/blob/be02a51c131e67c37a38db855c3ac9e59e5d69db/HelperFunctions.ps1#L450

Added a small comment on this

build/projects/Add-Ons (W1)/.AL-Go/PreCompileApp.ps1
aholstrup1: Good point. It should be. I'll use that instead

build/scripts/PreCompileApp.ps1
aholstrup1: It was needed because containerhelper automatically would pick .NET 9 as an assemblyproblingpath. I'm guessing that was fixed with this change: https://github.com/microsoft/navcontainerhelper/commit/c846102da530b72ebc0e30de026abfc938363b02

So we might not need this anymore

build/projects/Add-Ons (W1)/.AL-Go/settings.json
aholstrup1: That is indeed the case for CLEAN mode. In order to run tests in CLEAN mode we'd need to run on a self-hosted runner. Otherwise, we can't publish the CLEAN base application. 

We need the script overrides for default/translated mode since we do run tests in those buildmodes.

build/scripts/AppExtensionsHelper.psm1
aholstrup1: Added a comment on the assumed format

build/projects/Add-Ons (W1)/.AL-Go/PreCompileApp.ps1
aholstrup1: True 😄 That's not ideal. I'll see if I can address that

build/scripts/AppExtensionsHelper.psm1
```diff
@@ -0,0 +1,249 @@
...
+        This function will build an app from source code and place it in the symbols folder for the app.
+        The source code is downloaded from the artifact and the app is built with the same parameters as the main app.
+    .Parameter App
+        The name of the app to build.
+    .Parameter CompilationParameters
+        The parameters to use for the compilation of the app. This should be the same as the parameters used for the main app.
+#>
+function Build-App() {
+    param(
+        [string] $App,
```
mazhelez: Minor: naming. How about `$Name` or `$AppName`?

build/scripts/InstallMissingDependencies.ps1
```diff
@@ -0,0 +1,15 @@
+[Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSReviewUnusedParameter', 'parameters', Justification = 'The parameter is not used, but it''s script needs to match this format')]
```
mazhelez:
```suggestion
[Diagnostics.CodeAnalysis.SuppressMessageAttribute('PSReviewUnusedParameter', 'parameters', Justification = 'The parameter is not used, but the script needs to match this format')]
```

build/scripts/ImportTestDataInBcContainer.ps1
```diff
@@ -0,0 +1,43 @@
...
+    Write-Host "Initializing company in container $ContainerName"
+    Invoke-NavContainerCodeunit -Codeunitid 2 -containerName $ContainerName -CompanyName $CompanyName
+
+    if ($SetupData) {
+        Write-Host "Generating Setup Demo Data in container $ContainerName"
+        Invoke-NavContainerCodeunit -Codeunitid 5193 -containerName $ContainerName -CompanyName $CompanyName -MethodName "CreateSetupDemoData"
+    } else {
+        Write-Host "Generating All Demo Data in container $ContainerName"
+        Invoke-NavContainerCodeunit -CodeunitId 5193 -containerName $containerName -CompanyName $CompanyName -MethodName "CreateAllDemoData"
+        Invoke-NavContainerCodeunit -CodeunitId 5140 -containerName $containerName -CompanyName $CompanyName -MethodName "DeleteWarehouseEmployee"
```
mazhelez, line 22: Where do all these come from? How did you know what codeunits and methods to run here?

mazhelez (approved):

build/projects/Add-Ons (W1)/.AL-Go/PreCompileApp.ps1
```diff
@@ -0,0 +1,21 @@
...
+
+if ($ENV:BuildMode -eq 'Clean') {
+    Import-Module (Join-Path $PSScriptRoot "../../../scripts/AppExtensionsHelper.psm1" -Resolve)
+    $scriptPath = Join-Path $PSScriptRoot "../../../scripts/PreCompileApp.ps1" -Resolve
+
+    $symbolsPath = Join-Path $compilationParams.Value["compilerFolder"] 'symbols'
+    if (Test-Path $symbolsPath) {
+        # Remove all .app files except System.app in the symbols folder from the compiler folder
+        # This is to ensure that we don't end up using any of those apps as dependencies in the clean build
+        $appFiles = Get-ChildItem -Path $symbolsPath -Filter *.app -Recurse | Where-Object { $_.Name -ne 'System.app' }
```
haoranpb, line 14: Not sure I get this. Why not remove System App as well?

build/projects/Add-Ons (W1)/.AL-Go/customSettings.json
```diff
@@ -0,0 +1,20 @@
+{
+    "ExternalAppDependencies" : [
+        "System Application",
+        "Business Foundation", 
+        "Base Application", 
+        "Application",
+        "Statistical Accounts",
+        "Sustainability"
```
haoranpb, line 8: I assume this two apps are here because of they are needed for Demo Data?

build/projects/Add-Ons (W1)/.AL-Go/settings.json
```diff
@@ -0,0 +1,38 @@
...
+    "appFolders": [
+        "../../../src/Add-Ons/W1/*/App"
+    ],
+    "testFolders": [
+        "../../../src/Add-Ons/W1/*/Test",
+        "../../../src/Add-Ons/W1/*/Test Library",
+        "../../../src/Add-Ons/W1/*/Demo Data"
+    ],
+    "buildModes": [
+        "Translated"
```
haoranpb, line 12: Curious: what language do we use in build?

haoranpb (commented):
Note that not all Application Add-Ons have dependency on Application/BaseApp.

e.g. `"Email - Outlook REST API"` only need `System App`. Not sure if that changes thing.

---

Another thing is Demo Data: can we add an option to skip the generation of Demo Data?

As discussed, we plan to introduce the capabilities to run Unit Tests without Demo Data. We will handle the uptake when we get there, but it'd probably be easier to have the ability to skip Demo Data step

build/projects/Add-Ons (W1)/.AL-Go/customSettings.json
aholstrup1: No, it's because the first two apps we're going to open-source take dependency on them :)

build/projects/Add-Ons (W1)/.AL-Go/settings.json
aholstrup1: It's W1 so... English I assume 😄

build/scripts/ImportTestDataInBcContainer.ps1
aholstrup1: It's really just a copy from here: 
https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_git/NAV?path=/Eng/Core/Lib/NavAppDev.psm1&version=GBmaster&line=349&lineEnd=350&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents

Maybe it's something we could consider making part of AL-Go eventually but for now I opted for just keeping it as a script override in BCApps

build/projects/Add-Ons (W1)/.AL-Go/PreCompileApp.ps1
aholstrup1: With all the other dependencies, we recompile them from source code. We can't do that with the platform symbols though so that's the one dependency we'd like to keep

haoranpb (approved):

mazhelez (approved):

PR #3539, opened 2025-04-16
[Master] Include security filter when import permissionset from xml

gggdttt:
#### Work Item(s) 
Fixes [[AB#573207](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/573207)]

### DIFF
--- a/src/System Application/Test/Permission Sets/src/PermissionImportExportTests.Codeunit.al
+++ b/src/System Application/Test/Permission Sets/src/PermissionImportExportTests.Codeunit.al
@@ -422,6 +422,66 @@ codeunit 132437 "Permission Import Export Tests"
         LibraryAssert.AreEqual(TenantPermission."Delete Permission", TenantPermission."Delete Permission"::" ", 'Delete permission is not set correctly.');
     end;
 
+
+    [Test]
+    [Scope('OnPrem')]
+    procedure ExportImportTenantPermissionSetUsingOverwriteOptionWithSecurityFilter()
+    var
+        TenantPermission: Record "Tenant Permission";
+        TenantPermissionSet: Record "Tenant Permission Set";
+        PermissionSetRelation: Codeunit "Permission Set Relation";
+        TempBlobOriginal: Codeunit "Temp Blob";
+        ImportPermissionSets: XmlPort "Import Permission Sets";
+        OutStr: OutStream;
+        InStr: InStream;
+        ZeroGuid: Guid;
+        AppId: Guid;
+        NewRoleId: Code[20];
+        NewName: Text;
+    begin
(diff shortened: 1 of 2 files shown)


src/System Application/Test/Permission Sets/src/PermissionImportExportTests.Codeunit.al
```diff
@@ -422,6 +422,66 @@ codeunit 132437 "Permission Import Export Tests"
         LibraryAssert.AreEqual(TenantPermission."Delete Permission", TenantPermission."Delete Permission"::" ", 'Delete permission is not set correctly.');
```
gggdttt, line 422: Why do we use ` ImportPermissionSets.SetUpdatePermissions(false);`  instead of `ImportPermissionSets.SetUpdatePermissions(true);` in the test case `ExportImportTenantPermissionSetUsingOverwriteOption`?

f we want to UsingOverwriteOption, we should set it to 'true'.

stkillen: The naming of the function is not great. `SetUpdatePermissions(true)` means that we should update the existing permission set (e.g. not overwrite it) and `SetUpdatePermissions(false)` means overwrite it entirely, by deleting the existing permissions and importing the permission set again.

src/System Application/Test/Permission Sets/src/PermissionImportExportTests.Codeunit.al
```diff
@@ -422,6 +422,66 @@ codeunit 132437 "Permission Import Export Tests"
...
+        LibraryAssert.IsTrue(TenantPermission.Get(ZeroGuid, TenantPermissionSet."Role ID", TenantPermission."Object Type"::"Table Data", Database::"Tenant Permission"), 'Included permission to Test permission set is missing');
+
+        // [WHEN] Change the security filter again
+        Evaluate(TenantPermission."Security Filter", '');
+        TenantPermission.Modify();
+
+        // [WHEN] Import the original tenant permission set that was exported, and user choose to overwrite the permission set
+        TempBlobOriginal.CreateInStream(InStr);
+        ImportPermissionSets.SetSource(InStr);
+        ImportPermissionSets.SetUpdatePermissions(true);
```
stkillen: This should work with both true and false

gggdttt: @stkillen  That is strange because if I set it to false the security filter is not imported.

gggdttt: I tried it again, you are right, it work with both true and false.

darjoo (approved):

haoranpb (approved):

PR #3563, opened 2025-04-24
Add IsUserDelegated to AzureAdGraphUser

stkillen:
#### Summary 
Add `IsUserDelegated` to AzureAdGraphUser.

Pending platform uptake that supports `UserAccountHelper.IsUserDelegated()`

#### Work Item(s) 
Fixes [AB#542745](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/542745/?view=edit)


JesperSchulz (dismissed):

src/System Application/App/Azure AD User/src/AzureADGraphUserImpl.Codeunit.al
```diff
@@ -359,6 +372,11 @@ codeunit 9011 "Azure AD Graph User Impl."
         exit(LanguageId);
     end;
 
+    [InternalEvent(false)]
+    local procedure OnIsUserDelegated(var IsDelegatedUser: Boolean; var Handled: Boolean)
```
nikolakukrika, line 376: Why do we need an event?
I would skip this event, there should not be need to test this from AL.

nikolakukrika (dismissed):

src/System Application/App/Azure AD User/src/AzureADGraphUserImpl.Codeunit.al
stkillen: Yes there is a need. E.g. in Job Queue tests in BaseApp. Possibly other scenarios as well to verify that it works as intended when the user is delegated. I added it to the test library now which should make it clearer.

This is an internal event.

nikolakukrika (approved):

JesperSchulz (approved):

PR #3572, opened 2025-04-25
[26.0]Include security filter when import permissionset from xml

gggdttt:
#### Summary 
Include security filter when import permissionset from xml
#### Work Item(s) 
Fixes [AB#575215](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/575215)


gggdttt:
Decide not to backport to 26.0 because the dependencies are only backported to 26.1

PR #3613, opened 2025-04-30
26.x: No try functions in Page Summary

mynjj:
#### Summary 
In the Payables Agent, the agent can finish in the Purchase Invoice page, this triggers the summarization in this page within a try function.

Try functions don't allow write transactions to be started, so if the onopen logic requires write transactions, it crashes, in this particular case inserting a notification is crashing it.

#### Work Item(s) 
Fixes [AB#575512](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/575512)


Jose-agg (approved):

AndersLarsenMicrosoft (approved):
looks al good

PR #3626, opened 2025-05-01
Refactoring Rest Client Tests to use HttpClient Mocking instead of an actual service

nikolakukrika:
#### Summary 
We must refactor the Rest Client tests not to use actual services. The tests are fragile and we must not call actual endpoints from our automation gates. It is a horrible practice that must be avoided at all costs. We are introducing the cost on the endpoint. No automated test should target actual live services. This can only be done with semiautomated or manual testing. 

#### Work Item(s) 
Fixes [AB#564620](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/564620)

### DIFF
--- a/src/System Application/Test/Rest Client/src/MockRestClientService.Codeunit.al
+++ b/src/System Application/Test/Rest Client/src/MockRestClientService.Codeunit.al
@@ -1,0 +1,111 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Test.RestClient;
+using System.TestLibraries.Utilities;
+
+codeunit 134965 "Mock Rest Client Service"
+{
+    var
+        Assert: Codeunit "Library Assert";
+        BaseURLTxt: Label 'https://localhost';
+        GetUrlTxt: Label 'https://localhost/get';
+        PostUrlTxt: Label 'https://localhost/post';
+        PatchUrlTxt: Label 'https://localhost/patch';
+        PutUrlTxt: Label 'https://localhost/put';
+        DeleteUrlTxt: Label 'https://localhost/delete';
+        ResponseTxt: Text;
+        ExpectedRequestQueryParameters: Dictionary of [Text, Text];
+
+    procedure GetBaseURL(): Text
+    begin
+        exit(BaseURLTxt);
+    end;
+
+    procedure GetGetUrl(): Text
+    begin
+        exit(GetUrlTxt);
+    end;
+
+    procedure GetPostUrl(): Text
+    begin
+        exit(PostUrlTxt);
+    end;
+
+    procedure GetPatchUrl(): Text
+    begin
+        exit(PatchUrlTxt);
+    end;
+
+    procedure GetPutUrl(): Text
+    begin
+        exit(PutUrlTxt);
+    end;
+
+    procedure GetDeleteUrl(): Text
+    begin
+        exit(DeleteUrlTxt);
+    end;
+
+    procedure SetResponse(Response: Text)
+    begin
+        ResponseTxt := Response;
+    end;
+
+    procedure SetQueryParameters(QueryParameters: Dictionary of [Text, Text])
+    begin
+        ExpectedRequestQueryParameters := QueryParameters;
+    end;
+
+    procedure HandleRequest(Request: TestHttpRequestMessage; var Response: TestHttpResponseMessage): Boolean
+    begin
+        if not Request.Path.StartsWith(BaseURLTxt) then
+            exit;
+
+        if Request.Path.StartsWith(GetUrlTxt) then begin
+            if Request.QueryParameters.Count() > 0 then
+                VerifyRequestQueryParameters(Request.QueryParameters);
+
+            Response.Content.WriteFrom(ResponseTxt);
+            Response.HttpStatusCode := 200;
+            exit;
+        end;
+
+        if Request.Path = PostUrlTxt then begin
+            Response.Content.WriteFrom(ResponseTxt);
+            Response.HttpStatusCode := 200;
+            exit;
+        end;
+
+        if Request.Path = PatchUrlTxt then begin
+            Response.Content.WriteFrom(ResponseTxt);
+            Response.HttpStatusCode := 200;
+            exit;
+        end;
+
+        if Request.Path = PutUrlTxt then begin
+            Response.Content.WriteFrom(ResponseTxt);
+            Response.HttpStatusCode := 200;
+            exit;
+        end;
+
+        if Request.Path = DeleteUrlTxt then begin
+            Response.Content.WriteFrom(ResponseTxt);
+            Response.HttpStatusCode := 200;
+            exit;
+        end;
+    end;
+
+    local procedure VerifyRequestQueryParameters(RequestQueryParameters: Dictionary of [Text, Text])
+    var
+        QueryParameter: Text;
+    begin
+        Assert.AreEqual(ExpectedRequestQueryParameters.Count(), RequestQueryParameters.Count(), 'Wrong number of query parameters in request.');
+        foreach QueryParameter in ExpectedRequestQueryParameters.Keys() do begin
+            Assert.IsTrue(RequestQueryParameters.ContainsKey(QueryParameter), 'Missing query parameter: ' + QueryParameter);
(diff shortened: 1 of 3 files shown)


src/System Application/Test/Rest Client/src/MockRestClientService.Codeunit.al
```diff
@@ -0,0 +1,111 @@
...
+        if Request.Path.StartsWith(GetUrlTxt) then begin
+            if Request.QueryParameters.Count() > 0 then
+                VerifyRequestQueryParameters(Request.QueryParameters);
+
+            Response.Content.WriteFrom(ResponseTxt);
+            Response.HttpStatusCode := 200;
+            exit;
+        end;
+
+        if Request.Path = PostUrlTxt then begin
```
WaelAbuSeada, line 75: Should you check content value and type? Headers?

src/System Application/Test/Rest Client/src/MockRestClientService.Codeunit.al
```diff
@@ -0,0 +1,111 @@
...
+            Response.HttpStatusCode := 200;
+            exit;
+        end;
+
+        if Request.Path = DeleteUrlTxt then begin
+            Response.Content.WriteFrom(ResponseTxt);
+            Response.HttpStatusCode := 200;
+            exit;
+        end;
+    end;
```
WaelAbuSeada, line 105: Should there be a default value if the url doesnt exist? 404?

WaelAbuSeada (dismissed):

src/System Application/Test/Rest Client/src/MockRestClientService.Codeunit.al
```diff
@@ -0,0 +1,111 @@
...
+{
+    var
+        Assert: Codeunit "Library Assert";
+        BaseURLTxt: Label 'https://localhost';
+        GetUrlTxt: Label 'https://localhost/get';
+        PostUrlTxt: Label 'https://localhost/post';
+        PatchUrlTxt: Label 'https://localhost/patch';
+        PutUrlTxt: Label 'https://localhost/put';
+        DeleteUrlTxt: Label 'https://localhost/delete';
+        ResponseTxt: Text;
```
AndreasMoth: I prefer only calling labels Txt. When I see ResponseTxt I expect it's a label not a text variable. I understand it's a text but it's confusing and could lead to wrong assumptions about whether it's hardcoded or changable, which in worst case could lead to security issues.

src/System Application/Test/Rest Client/src/MockRestClientService.Codeunit.al
```diff
@@ -0,0 +1,111 @@
...
+    begin
+        ExpectedRequestQueryParameters := QueryParameters;
+    end;
+
+    procedure HandleRequest(Request: TestHttpRequestMessage; var Response: TestHttpResponseMessage): Boolean
+    begin
+        if not Request.Path.StartsWith(BaseURLTxt) then
+            exit;
+
+        if Request.Path.StartsWith(GetUrlTxt) then begin
```
AndreasMoth: Instead of all these if - then - exit statements, I would simply have a case statement and then no need for exit;

src/System Application/Test/Rest Client/src/MockRestClientService.Codeunit.al
nikolakukrika: We cannot, we do not have the access to headers. It is because of Security : )

src/System Application/Test/Rest Client/src/MockRestClientService.Codeunit.al
```diff
@@ -0,0 +1,111 @@
...
+            if Request.QueryParameters.Count() > 0 then
+                VerifyRequestQueryParameters(Request.QueryParameters);
```
AndreasMoth: We should call VerifyRequestQueryParameters regardless of whether there are any query parameters or not in the request. after all we may expect some and then the difference in count should error out.

src/System Application/Test/Rest Client/src/MockRestClientService.Codeunit.al
nikolakukrika: I was thinking to let it fall through if it cannot handle the URL. I will throw an error because the unexpected URL should fail the test. We do not test 404.

src/System Application/Test/Rest Client/src/MockRestClientService.Codeunit.al
```diff
@@ -0,0 +1,111 @@
...
+    begin
+        ResponseTxt := Response;
+    end;
+
+    procedure SetQueryParameters(QueryParameters: Dictionary of [Text, Text])
+    begin
+        ExpectedRequestQueryParameters := QueryParameters;
+    end;
+
+    procedure HandleRequest(Request: TestHttpRequestMessage; var Response: TestHttpResponseMessage): Boolean
```
AndreasMoth, line 62: This functionality only handles a single request. Sometimes we have functionality which will send multiple requests. That is not possible to handle with how this currently works.
Some pairing of "return this response for this url" and "expect these parameters for this url". Keep the simple SetResponse though for the response to anything. But add SetResponse which also takes base url.

src/System Application/Test/Rest Client/src/RestClientTests.Codeunit.al
```diff
@@ -11,13 +11,16 @@ using System.TestLibraries.Utilities;
 codeunit 134971 "Rest Client Tests"
 {
     Subtype = Test;
+    TestHttpRequestPolicy = BlockOutboundRequests;
```
AndreasMoth, line 14: Why is this not default for all tests? At least it should be configurable on the app.json file that you don't want any of your tests to be able to call outside BC.

nikolakukrika: It will become a default in 1.5 years. We need to update the scripts to block the HttpRequests, it is a server setting...

src/System Application/Test/Rest Client/src/MockRestClientService.Codeunit.al
nikolakukrika: Fixed by introducing an error.

src/System Application/Test/Rest Client/src/MockRestClientService.Codeunit.al
nikolakukrika: @AndreasMoth I hate case(true) it must not be used. Multiple ifs are better.

src/System Application/Test/Rest Client/src/MockRestClientService.Codeunit.al
nikolakukrika: Fixed, thanks

src/System Application/Test/Rest Client/src/MockRestClientService.Codeunit.al
nikolakukrika: Fixed by adding library variable storage. 
This specific Mock is designed to be used only by this test, it is not designed as a generic library. 

It is hard to queue the Query parameters for example. I would avoid creating complexity for this specific test.

src/System Application/Test/Rest Client/src/MockRestClientService.Codeunit.al
nikolakukrika: Fixed, thanks

darjoo (approved):

WaelAbuSeada (approved):

ajkauffmann:
> @ajkauffmann, how do you like _this_? 😊

It's definitely the way to avoid real outgoing calls. This was not possible before, so I think it's a good improvement.
More tests need to be refactored though. But it's a good start!

Do you want me to cover the other tests?

JesperSchulz:
> > @ajkauffmann, how do you like _this_? 😊
> 
> It's definitely the way to avoid real outgoing calls. This was not possible before, so I think it's a good improvement. More tests need to be refactored though. But it's a good start!
> 
> Do you want me to cover the other tests?

If you've got an interest in fixing the other tests, I'd say that would be awesome! @nikolakukrika, do you have active plans to continue this work? In that case, we should coordinate to not do double work...

nikolakukrika:
> > @ajkauffmann, how do you like _this_? 😊
> 
> It's definitely the way to avoid real outgoing calls. This was not possible before, so I think it's a good improvement. More tests need to be refactored though. But it's a good start!
> 
> Do you want me to cover the other tests?

@ajkauffmann I'm not aware of other tests, but if you know of some, please make a PR in a similar way. 
We use the codeunit as a mock for web service. It should use Library Variable storage to ensure that the web requests happened and that we got the expected number of web requests - not more or less.

PR #3655, opened 2025-05-07
SharePoint Graph API support

### LINKED ISSUE #3198: [BC Idea]: SharePoint - add graph API support
### BC Idea Link

https://experience.dynamics.com/ideas/idea/?ideaid=94a9b2ef-66fb-ef11-a4dd-6045bdb0bb28

### Description

I noticed that the current SharePoint module only supports REST SharePoint API.
As far as I understand, this is an outdated method of working with SharePoint .

I suggest adding graph API for SharePoint module.

SharePoint REST API: https://learn.microsoft.com/en-us/sharepoint/dev/sp-add-ins/get-to-know-the-sharepoint-rest-service
SharePoint Graph API: https://learn.microsoft.com/en-us/graph/api/resources/sharepoint

Perhaps there is some reason why a legacy REST API was integrated instead of the Graph API?

But, I am ready to implement Graph API support.

My main motivation is that the Graph API is constantly evolving and contains features that are not found in the REST SharePoint API. 

Drakonian:
#### Summary 

This is the first version of the SharePoint Graph API module that leverages the Microsoft Graph module.

After analyzing the current REST implementation, I concluded that it cannot be changed without breaking backward compatibility. In addition, the REST and Graph APIs have completely different data models and sets of fields, which makes it impossible to reuse the same tables for entities that REST uses.

Therefore, adding a separate SharePoint Graph Client seemed the simplest and most natural solution. I also considered factory/builder patterns and dependency injection, but for the reasons mentioned above these approaches did not seem optimal, they don’t fit well and would greatly complicate the overall use of the SharePoint module.

I have already tested this code locally in various scenarios, and it behaves well. My plans are:

* Detail review of all the code with improvements where necessary
* Gathering feedback from the community and Microsoft
* Test pagination
* Test large file uploads
* Restructuring the folders so that REST and Graph reside in separate, peer-level folders
* Writing tests

Examples of how to use:
```AL
local procedure GetDrivesFromSite()
var
    TempGraphDrive: Record "SharePoint Graph Drive" temporary;
    SharePointGraphClient: Codeunit "SharePoint Graph Client";
    GraphAuthorization: Codeunit "Graph Authorization";
    GraphAuthInterface: Interface "Graph Authorization";
begin
    GraphAuthInterface := GraphAuthorization.CreateAuthorizationWithClientCredentials(
        TenantIdValue,
        ClientIdValue,
        ClientSecretValue,
        GraphScopeLbl);

    SharePointGraphClient.Initialize(SharePointUrlValue, GraphAuthInterface);

    if not SharePointGraphClient.GetDrives(TempGraphDrive) then
        Message('No drives found.');

    if TempGraphDrive.FindSet() then
        repeat
            Message('Drive Name: %1\Type: %2\Owner: %3\Quota Used: %4 of %5\Created: %6',
                TempGraphDrive.Name,
                TempGraphDrive.DriveType,
                TempGraphDrive.OwnerName,
                TempGraphDrive.QuotaUsed,
                TempGraphDrive.QuotaTotal,
                TempGraphDrive.CreatedDateTime);
        until TempGraphDrive.Next() = 0;
end;
```
```AL
local procedure GetDrivesFromSite()
var
    TempGraphDriveItem: Record "SharePoint Graph Drive Item" temporary;
    SharePointGraphClient: Codeunit "SharePoint Graph Client";
    GraphAuthorization: Codeunit "Graph Authorization";
    GraphAuthInterface: Interface "Graph Authorization";
    FileInStream: InStream;
    FileName: Text;
    FolderPath: Text;
begin
    GraphAuthInterface := GraphAuthorization.CreateAuthorizationWithClientCredentials(
        TenantIdValue,
        ClientIdValue,
        ClientSecretValue,
        GraphScopeLbl);

    SharePointGraphClient.Initialize(SharePointUrlValue, GraphAuthInterface);

    FolderPath := 'TestFolder'; //To root drive folder as we not specified DriveId
    if not UploadIntoStream('Select a file to upload', '', '', FileName, FileInStream) then
        exit;

    if SharePointGraphClient.UploadFile(FolderPath, FileName, FileInStream, TempGraphDriveItem) then
        Message('File uploaded successfully: %1', TempGraphDriveItem.Name)
    else
        ShowError();
end;
```

#### Work Item(s) 
Fixes #3198

### DIFF
--- a/src/System Application/App/SharePoint/permissions/SharePointApiObjects.PermissionSet.al
+++ b/src/System Application/App/SharePoint/permissions/SharePointApiObjects.PermissionSet.al
@@ -10,5 +10,5 @@ permissionset 9100 "SharePoint API - Objects"
     Access = Internal;
     Assignable = false;
 
-    Permissions = codeunit "SharePoint Client" = X;
+    Permissions = codeunit "SharePoint Client" = X, codeunit "SharePoint Graph Client" = X;
 }
--- a/src/System Application/App/SharePoint/src/graph/SharePointGraphClient.Codeunit.al
+++ b/src/System Application/App/SharePoint/src/graph/SharePointGraphClient.Codeunit.al
@@ -1,0 +1,544 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Integration.Sharepoint;
+
+using System.Integration.Graph;
+using System.Integration.Graph.Authorization;
+
+/// <summary>
+/// Provides functionality for interacting with SharePoint through Microsoft Graph API.
+/// </summary>
+codeunit 9119 "SharePoint Graph Client"
+{
+    Access = Public;
+
+    var
+        SharePointGraphClientImpl: Codeunit "SharePoint Graph Client Impl.";
+
+    #region Initialization
+
+    /// <summary>
+    /// Initializes SharePoint Graph client.
+    /// </summary>
+    /// <param name="NewSharePointUrl">SharePoint site URL.</param>
+    /// <param name="GraphAuthorization">The Graph API authorization to use.</param>
+    procedure Initialize(NewSharePointUrl: Text; GraphAuthorization: Interface "Graph Authorization")
+    begin
+        SharePointGraphClientImpl.Initialize(NewSharePointUrl, GraphAuthorization);
+    end;
+
+    /// <summary>
+    /// Initializes SharePoint Graph client with a specific API version.
+    /// </summary>
+    /// <param name="NewSharePointUrl">SharePoint site URL.</param>
+    /// <param name="ApiVersion">The Graph API version to use.</param>
+    /// <param name="GraphAuthorization">The Graph API authorization to use.</param>
+    procedure Initialize(NewSharePointUrl: Text; ApiVersion: Enum "Graph API Version"; GraphAuthorization: Interface "Graph Authorization")
+    begin
+        SharePointGraphClientImpl.Initialize(NewSharePointUrl, ApiVersion, GraphAuthorization);
+    end;
+
+    /// <summary>
+    /// Initializes SharePoint Graph client with a custom base URL.
+    /// </summary>
+    /// <param name="NewSharePointUrl">SharePoint site URL.</param>
+    /// <param name="BaseUrl">The custom base URL for Graph API.</param>
+    /// <param name="GraphAuthorization">The Graph API authorization to use.</param>
+    procedure Initialize(NewSharePointUrl: Text; BaseUrl: Text; GraphAuthorization: Interface "Graph Authorization")
+    begin
+        SharePointGraphClientImpl.Initialize(NewSharePointUrl, BaseUrl, GraphAuthorization);
+    end;
+
+    #endregion
+
+    #region Lists
+
+    /// <summary>
+    /// Gets all lists from the SharePoint site.
+    /// </summary>
+    /// <param name="GraphLists">Collection of the result (temporary record).</param>
+    /// <returns>True if the operation was successful; otherwise - false.</returns>
+    procedure GetLists(var GraphLists: Record "SharePoint Graph List" temporary): Boolean
+    begin
+        exit(SharePointGraphClientImpl.GetLists(GraphLists));
+    end;
+
+    /// <summary>
+    /// Gets all lists from the SharePoint site.
+    /// </summary>
+    /// <param name="GraphLists">Collection of the result (temporary record).</param>
+    /// <param name="GraphOptionalParameters">A wrapper for optional header and query parameters</param>
+    /// <returns>True if the operation was successful; otherwise - false.</returns>
+    procedure GetLists(var GraphLists: Record "SharePoint Graph List" temporary; GraphOptionalParameters: Codeunit "Graph Optional Parameters"): Boolean
+    begin
+        exit(SharePointGraphClientImpl.GetLists(GraphLists, GraphOptionalParameters));
+    end;
+
+    /// <summary>
+    /// Gets a SharePoint list by ID.
+    /// </summary>
+    /// <param name="ListId">ID of the list to retrieve.</param>
+    /// <param name="GraphList">Record to store the result.</param>
+    /// <returns>True if the operation was successful; otherwise - false.</returns>
+    procedure GetList(ListId: Text; var GraphList: Record "SharePoint Graph List" temporary): Boolean
+    begin
+        exit(SharePointGraphClientImpl.GetList(ListId, GraphList));
+    end;
+
+    /// <summary>
+    /// Gets a SharePoint list by ID.
+    /// </summary>
+    /// <param name="ListId">ID of the list to retrieve.</param>
+    /// <param name="GraphList">Record to store the result.</param>
+    /// <param name="GraphOptionalParameters">A wrapper for optional header and query parameters.</param>
+    /// <returns>True if the operation was successful; otherwise - false.</returns>
+    procedure GetList(ListId: Text; var GraphList: Record "SharePoint Graph List" temporary; GraphOptionalParameters: Codeunit "Graph Optional Parameters"): Boolean
+    begin
+        exit(SharePointGraphClientImpl.GetList(ListId, GraphList, GraphOptionalParameters));
+    end;
+
+    /// <summary>
+    /// Creates a new SharePoint list.
+    /// </summary>
+    /// <param name="DisplayName">Display name for the list.</param>
+    /// <param name="Description">Description for the list.</param>
+    /// <param name="GraphList">Record to store the result.</param>
+    /// <returns>True if the operation was successful; otherwise - false.</returns>
+    procedure CreateList(DisplayName: Text; Description: Text; var GraphList: Record "SharePoint Graph List" temporary): Boolean
+    begin
+        exit(SharePointGraphClientImpl.CreateList(DisplayName, Description, GraphList));
+    end;
+
+    /// <summary>
+    /// Creates a new SharePoint list.
+    /// </summary>
+    /// <param name="DisplayName">Display name for the list.</param>
+    /// <param name="ListTemplate">Template for the list (genericList, documentLibrary, etc.)</param>
+    /// <param name="Description">Description for the list.</param>
+    /// <param name="GraphList">Record to store the result.</param>
+    /// <returns>True if the operation was successful; otherwise - false.</returns>
+    procedure CreateList(DisplayName: Text; ListTemplate: Text; Description: Text; var GraphList: Record "SharePoint Graph List" temporary): Boolean
+    begin
+        exit(SharePointGraphClientImpl.CreateList(DisplayName, ListTemplate, Description, GraphList));
+    end;
+
+    #endregion
+
+    #region List Items
+
+    /// <summary>
+    /// Gets items from a SharePoint list.
+    /// </summary>
+    /// <param name="ListId">ID of the list.</param>
+    /// <param name="GraphListItems">Collection of the result (temporary record).</param>
+    /// <returns>True if the operation was successful; otherwise - false.</returns>
+    procedure GetListItems(ListId: Text; var GraphListItems: Record "SharePoint Graph List Item" temporary): Boolean
+    begin
+        exit(SharePointGraphClientImpl.GetListItems(ListId, GraphListItems));
+    end;
+
+    /// <summary>
+    /// Gets items from a SharePoint list.
+    /// </summary>
+    /// <param name="ListId">ID of the list.</param>
+    /// <param name="GraphListItems">Collection of the result (temporary record).</param>
+    /// <param name="GraphOptionalParameters">A wrapper for optional header and query parameters.</param>
+    /// <returns>True if the operation was successful; otherwise - false.</returns>
+    procedure GetListItems(ListId: Text; var GraphListItems: Record "SharePoint Graph List Item" temporary; GraphOptionalParameters: Codeunit "Graph Optional Parameters"): Boolean
+    begin
+        exit(SharePointGraphClientImpl.GetListItems(ListId, GraphListItems, GraphOptionalParameters));
+    end;
+
+    /// <summary>
+    /// Creates a new item in a SharePoint list.
+    /// </summary>
+    /// <param name="ListId">ID of the list.</param>
+    /// <param name="FieldsJsonObject">JSON object containing the fields for the new item.</param>
+    /// <param name="GraphListItem">Record to store the result.</param>
+    /// <returns>True if the operation was successful; otherwise - false.</returns>
+    procedure CreateListItem(ListId: Text; FieldsJsonObject: JsonObject; var GraphListItem: Record "SharePoint Graph List Item" temporary): Boolean
+    begin
+        exit(SharePointGraphClientImpl.CreateListItem(ListId, FieldsJsonObject, GraphListItem));
+    end;
+
+    /// <summary>
+    /// Creates a new item in a SharePoint list with a simple title.
+    /// </summary>
+    /// <param name="ListId">ID of the list.</param>
+    /// <param name="Title">Title for the new item.</param>
+    /// <param name="GraphListItem">Record to store the result.</param>
+    /// <returns>True if the operation was successful; otherwise - false.</returns>
+    procedure CreateListItem(ListId: Text; Title: Text; var GraphListItem: Record "SharePoint Graph List Item" temporary): Boolean
+    begin
+        exit(SharePointGraphClientImpl.CreateListItem(ListId, Title, GraphListItem));
+    end;
+
+    #endregion
+
+    #region Drive and Items
+
+    /// <summary>
+    /// Gets the default document library (drive) for the site.
+    /// </summary>
+    /// <param name="DriveId">ID of the default drive.</param>
+    /// <returns>True if the operation was successful; otherwise - false.</returns>
+    procedure GetDefaultDrive(var DriveId: Text): Boolean
+    begin
+        exit(SharePointGraphClientImpl.GetDefaultDrive(DriveId));
+    end;
+
+    /// <summary>
+    /// Gets all drives (document libraries) available on the site with detailed information.
+    /// </summary>
+    /// <param name="GraphDrives">Collection of the result (temporary record).</param>
+    /// <returns>True if the operation was successful; otherwise - false.</returns>
+    procedure GetDrives(var GraphDrives: Record "SharePoint Graph Drive" temporary): Boolean
+    begin
+        exit(SharePointGraphClientImpl.GetDrives(GraphDrives));
+    end;
+
+    /// <summary>
+    /// Gets all drives (document libraries) available on the site with detailed information.
+    /// </summary>
+    /// <param name="GraphDrives">Collection of the result (temporary record).</param>
+    /// <param name="GraphOptionalParameters">A wrapper for optional header and query parameters.</param>
+    /// <returns>True if the operation was successful; otherwise - false.</returns>
+    procedure GetDrives(var GraphDrives: Record "SharePoint Graph Drive" temporary; GraphOptionalParameters: Codeunit "Graph Optional Parameters"): Boolean
+    begin
+        exit(SharePointGraphClientImpl.GetDrives(GraphDrives, GraphOptionalParameters));
+    end;
+
+    /// <summary>
+    /// Gets a drive (document library) by ID with detailed information.
+    /// </summary>
+    /// <param name="DriveId">ID of the drive to retrieve.</param>
+    /// <param name="GraphDrive">Record to store the result.</param>
+    /// <returns>True if the operation was successful; otherwise - false.</returns>
+    procedure GetDrive(DriveId: Text; var GraphDrive: Record "SharePoint Graph Drive" temporary): Boolean
+    begin
+        exit(SharePointGraphClientImpl.GetDrive(DriveId, GraphDrive));
+    end;
+
+    /// <summary>
+    /// Gets a drive (document library) by ID with detailed information.
+    /// </summary>
+    /// <param name="DriveId">ID of the drive to retrieve.</param>
+    /// <param name="GraphDrive">Record to store the result.</param>
+    /// <param name="GraphOptionalParameters">A wrapper for optional header and query parameters.</param>
+    /// <returns>True if the operation was successful; otherwise - false.</returns>
+    procedure GetDrive(DriveId: Text; var GraphDrive: Record "SharePoint Graph Drive" temporary; GraphOptionalParameters: Codeunit "Graph Optional Parameters"): Boolean
+    begin
+        exit(SharePointGraphClientImpl.GetDrive(DriveId, GraphDrive, GraphOptionalParameters));
+    end;
+
+    /// <summary>
+    /// Gets the default document library (drive) for the site with detailed information.
+    /// </summary>
+    /// <param name="GraphDrive">Record to store the result.</param>
+    /// <returns>True if the operation was successful; otherwise - false.</returns>
+    procedure GetDefaultDrive(var GraphDrive: Record "SharePoint Graph Drive" temporary): Boolean
+    begin
+        exit(SharePointGraphClientImpl.GetDefaultDrive(GraphDrive));
+    end;
+
+    /// <summary>
+    /// Gets items in the root folder of the default drive.
+    /// </summary>
+    /// <param name="GraphDriveItems">Collection of the result (temporary record).</param>
+    /// <returns>True if the operation was successful; otherwise - false.</returns>
+    procedure GetRootItems(var GraphDriveItems: Record "SharePoint Graph Drive Item" temporary): Boolean
+    begin
+        exit(SharePointGraphClientImpl.GetRootItems(GraphDriveItems));
+    end;
+
+    /// <summary>
+    /// Gets items in the root folder of the default drive.
+    /// </summary>
+    /// <param name="GraphDriveItems">Collection of the result (temporary record).</param>
+    /// <param name="GraphOptionalParameters">A wrapper for optional header and query parameters.</param>
+    /// <returns>True if the operation was successful; otherwise - false.</returns>
+    procedure GetRootItems(var GraphDriveItems: Record "SharePoint Graph Drive Item" temporary; GraphOptionalParameters: Codeunit "Graph Optional Parameters"): Boolean
+    begin
+        exit(SharePointGraphClientImpl.GetRootItems(GraphDriveItems, GraphOptionalParameters));
+    end;
+
+    /// <summary>
+    /// Gets children of a folder by the folder's ID.
+    /// </summary>
+    /// <param name="FolderId">ID of the folder.</param>
+    /// <param name="GraphDriveItems">Collection of the result (temporary record).</param>
+    /// <returns>True if the operation was successful; otherwise - false.</returns>
+    procedure GetFolderItems(FolderId: Text; var GraphDriveItems: Record "SharePoint Graph Drive Item" temporary): Boolean
+    begin
+        exit(SharePointGraphClientImpl.GetFolderItems(FolderId, GraphDriveItems));
+    end;
+
+    /// <summary>
+    /// Gets children of a folder by the folder's ID.
+    /// </summary>
+    /// <param name="FolderId">ID of the folder.</param>
+    /// <param name="GraphDriveItems">Collection of the result (temporary record).</param>
+    /// <param name="GraphOptionalParameters">A wrapper for optional header and query parameters.</param>
+    /// <returns>True if the operation was successful; otherwise - false.</returns>
+    procedure GetFolderItems(FolderId: Text; var GraphDriveItems: Record "SharePoint Graph Drive Item" temporary; GraphOptionalParameters: Codeunit "Graph Optional Parameters"): Boolean
+    begin
+        exit(SharePointGraphClientImpl.GetFolderItems(FolderId, GraphDriveItems, GraphOptionalParameters));
+    end;
+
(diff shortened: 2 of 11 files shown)


Drakonian:
@pri-kise

Drakonian:
> Very interesting contribution! Please allow me to take a much closer look at your code in the coming days. Unfortunately I will be on the road next week, so this might take me a short while to get to - but I'll be on it soon!

Thank you, I’ll look forward to your feedback -)

src/System Application/App/SharePoint/src/graph/SharePointGraphClient.Codeunit.al
```diff
@@ -0,0 +1,600 @@
...
+        SharePointGraphClientImpl.Initialize(NewSharePointUrl, ApiVersion, GraphAuthorization);
+    end;
+
+    /// <summary>
+    /// Initializes SharePoint Graph client with a custom base URL.
+    /// </summary>
+    /// <param name="NewSharePointUrl">SharePoint site URL.</param>
+    /// <param name="BaseUrl">The custom base URL for Graph API.</param>
+    /// <param name="GraphAuthorization">The Graph API authorization to use.</param>
+    procedure Initialize(NewSharePointUrl: Text; BaseUrl: Text; GraphAuthorization: Interface "Graph Authorization")
```
pri-kise, line 60: i think that we should be able to provide a httpclienthandler via the Initialize procedures here.
first of all that can help us with testing and more important a parnter app can provide their own http client handler that you don't have to allow web communication for the system app, but it's sufficient to allow http communication for the partners custom app.

pri-kise (changes requested):
First of all: I really appreciate the work you have done for this PR.

I'm unsure if maybe some of the functionalities in the graph helper codeunits should maybe better be placed into the graph client module. 
I only took a short look into the new code so far. I will need some more time to look into the changes in more detail.

Update: 
I'm not sure if it's really helpful to provide a boolean value as return value on all those functions.
How should anyone know why the got `false`.

I think that it would be more helpful to return errors when something didn't went well.

Furthermore on your planned improvements for pagination:
I'm not sure if this is possible with the GraphOptionalParameters, but I think that it would be helpful if we could specify a range from outside, too. 
What I mean is that sometimes we might want to control the pagination steps from outside.

src/System Application/App/SharePoint/src/graph/SharePointGraphClient.Codeunit.al
Drakonian: Yes, I am just now working on the tests and in the process I noticed that Http Client Handler is missing.

Will commit it together with tests soon.

Drakonian:
@pri-kise 

Thank you for such a helpful review of the PR. Your remarks are very useful.

1. I added an HTTP client handler, of course, it is absolutely essential here.
2. I completely agree with your point about the boolean, it's a relic from an older code version where I was experimenting with building a SharePoint Graph module (trying to merge with the SharePoint REST module). I prefer the approach used in the Azure Blob Storage API, where a method returns an object containing detailed information about the operation, so I’ve applied a similar pattern here.

Regarding managing pagination externally, that’s an excellent observation, but I’m still considering the best way to implement it. I’ll think through an approach that is convenient.

I believe pagination would be best handled at the Microsoft Graph module level, this needs some further thought.

My plans: at the moment I intend to keep improving the code and finish writing the tests.

JesperSchulz:
@Drakonian, wow, that's impressive work! I must admit that I got "code review fatigue" about 1/3 into the review process - that's A LOT of code 🤣 I will need to see the code in action, rather than just read my way through it.
1) I will try to compile and use your module updates
2) I will wait for your tests to fully understand how the code is to be used in praxis

Generally though, I have no objections of accepting your changes. They seem to follow all of our best practices for system application modules!

tinestaric:
@Drakonian, to echo @JesperSchulz, really, really nice work! And it's precisely what I'll need soon! Thank you! :D

I'll take this module for a spin; however, I see a couple of gaps in the code.

1. UploadLargeFile sends the file by chunks, but there's no DownloadLargeFile equivalent. I recently discovered that BC has a limit on the size of the response content HttpClient can receive, set to 150 MB by default (controlled by the NavHttpClientMaxResponseContentSize server setting). That means that you can't download files larger than that, so we'd need a chunking option for download as well. 

2. I'm missing any Delete operations for either Files or Folders

3. No existence check operations for files or folders

And there are two nice-to-haves, which I can work around with a combination of existing functions, but it would be cleaner to use the Graph API directly
4. Copy operation (but I can use Download+Upload)
5. Move operation - via PATCH parentReference (but Download+Upload+Delete gets me there as well)

src/System Application/App/SharePoint/src/graph/SharePointGraphClient.Codeunit.al
```diff
@@ -0,0 +1,631 @@
...
+
+    /// <summary>
+    /// Uploads a file to a folder on the default drive.
+    /// </summary>
+    /// <param name="FolderPath">Path to the folder (e.g., 'Documents').</param>
+    /// <param name="FileName">Name of the file to upload.</param>
+    /// <param name="FileInStream">Content of the file.</param>
+    /// <param name="GraphDriveItem">Record to store the result.</param>
+    /// <returns>An operation response object containing the result of the operation.</returns>
+    procedure UploadFile(FolderPath: Text; FileName: Text; var FileInStream: InStream; var GraphDriveItem: Record "SharePoint Graph Drive Item" temporary): Codeunit "SharePoint Graph Response"
```
tinestaric: I'd argue against passing Streams as var. It's a by-reference type and you never actually change the reference itself. 
It's not a big deal on uploads, but on Downloads I now can't do 
DownloadFile(FileId, TempBlob.CreateOutStream()), as it complains that the parameter should be an assignable stream variable.

src/System Application/App/SharePoint/src/graph/SharePointGraphClientImpl.Codeunit.al
```diff
@@ -0,0 +1,1517 @@
...
+
+        // Use default drive ID if none specified
+        if DriveId = '' then begin
+            EnsureDefaultDriveId();
+            EffectiveDriveId := DefaultDriveId;
+        end else
+            EffectiveDriveId := DriveId;
+
+        // Remove leading slash if present
+        if FolderPath.StartsWith('/') then
```
tinestaric: nit, this could be a .TrimStart('/')

Drakonian:
@tinestaric

I tried to reuse your code as much as possible so that your time would not be wasted.

I also added automated tests, but only for part of the functionality. I still need to test everything thoroughly and complete automated tests. (I will move library/handlers codeunits for tests in separate application later)

You could also help me with testing if you want to speed up the process.

tinestaric:
@Drakonian 
Thank you, and yes. Let me know when it's ready for testing, and I'll go ahead.

src/System Application/Test Library/SharePoint/src/graph/SharePointGraphTestLibrary.Codeunit.al
```diff
@@ -0,0 +1,41 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Test.Integration.Sharepoint;
+
+using System.RestClient;
+
+codeunit 132975 "SharePoint Graph Test Library"
+{
+    EventSubscriberInstance = Manual;
```
duichwer: Has there been some redesign?
I don't see any event subscribers here.
When this property is removed then the BindSubscription needs to removed, too.

src/System Application/App/SharePoint/src/graph/models/SharePointGraphDriveItem.Table.al
```diff
@@ -0,0 +1,103 @@
...
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Integration.Sharepoint;
+
+/// <summary>
+/// Represents a SharePoint drive item (file or folder) as returned by Microsoft Graph API.
+/// </summary>
+table 9132 "SharePoint Graph Drive Item"
+{
+    Access = Public;
```
duichwer, line 13: Do we those new tables to extensible (at the first release)?

What I mean is, if it weren't better to make the new tables extensible false at first

duichwer (changes requested):

src/System Application/Test Library/SharePoint/src/graph/SharePointGraphTestLibrary.Codeunit.al
Drakonian: Yes, that's what's left after one of my redesigns. Thank you for noticing, it's certainly not necessary.

src/System Application/App/SharePoint/src/graph/models/SharePointGraphDriveItem.Table.al
Drakonian: To be honest, I am not even sure. Are there any best practices? Why prevent others from extending the temporary tables?

tinestaric:
@Drakonian 
I tested the following scenarios using your framework, and they all work without a problem.
List Files/Folders
Download File
Create File
Copy File
Move File
File Exists
Delete File
Create Folder
Folder Exists
Delete Folder

The one notable thing I did spot (but it is not directly related to your changes) is that using the Graph API is considerably slower than the REST API when listing files/folders.

The issue is that the REST API retrieves all at once, while the Graph API paginates requests. Pagination is fine, but the default page size set on GetAllPages in Graph Client Impl. is 100, and there's no way to change it because GraphPaginationData isn't passed as a parameter. 

In my tests, for a folder of 850 files, I'm noticing about 3-4 seconds slower performance.

It shouldn't be worth changing in the scope of this PR, but for anyone coming across this PR, it might be worth noting.

duichwer:
> @Drakonian I tested the following scenarios using your framework, and they all work without a problem. List Files/Folders Download File Create File Copy File Move File File Exists Delete File Create Folder Folder Exists Delete Folder
> 
> The one notable thing I did spot (but it is not directly related to your changes) is that using the Graph API is considerably slower than the REST API when listing files/folders.
> 
> The issue is that the REST API retrieves all at once, while the Graph API paginates requests. Pagination is fine, but the default page size set on GetAllPages in Graph Client Impl. is 100, and there's no way to change it because GraphPaginationData isn't passed as a parameter.
> 
> In my tests, for a folder of 850 files, I'm noticing about 3-4 seconds slower performance.
> 
> It shouldn't be worth changing in the scope of this PR, but for anyone coming across this PR, it might be worth noting.

@tinestaric aren't you able to set the page size via `GraphOptionalParameters: Codeunit "Graph Optional Parameters"` ?

Then the page size of the optional parameters is used here:
https://github.com/microsoft/BCApps/blob/5c5e2189cca4c553228e60be28c5e27bf1bd8021/src/System%20Application/App/MicrosoftGraph/src/GraphClientImpl.Codeunit.al#L107

Drakonian:
@tinestaric

Thank you for your discovery and testing. Overall, I agree that improving pagination most likely relates to the context of the Graph API module.

Therefore, I suggest moving forward with this module.
I will create separate issue/PR regarding your discovery in the Graph API module a little later.

I think we are close to completing this PR.
What do you think about it? 
@JesperSchulz

src/System Application/App/SharePoint/permissions/SharePointApiObjects.PermissionSet.al
```diff
@@ -10,5 +10,5 @@ permissionset 9100 "SharePoint API - Objects"
     Access = Internal;
     Assignable = false;
 
-    Permissions = codeunit "SharePoint Client" = X;
+    Permissions = codeunit "SharePoint Client" = X, codeunit "SharePoint Graph Client" = X;
```
JesperSchulz:
```suggestion
    Permissions = codeunit "SharePoint Client" = X,
                           codeunit "SharePoint Graph Client" = X;
```

JesperSchulz:
@Drakonian, I think the new compilation issues make a lot more sense ;-) Let's get on those!

src/System Application/App/SharePoint/src/graph/models/SharePointGraphDriveItem.Table.al
JesperSchulz: This is a hot topic. Our "best practice" is to keep new objects non-extensible, to allow us to make changes more easily in the beginning, should that be necessary. Public and extensible objects very quickly get taken dependency on, requiring deprecation etc. if breaking changes are needed.
Of course, the best solution would be to design for proper extensibility from the get-go. So if you're certain you're objects should be extensible in a certain way, by all means, implement that. If you're not sure, wait until someone requests it - also to then understand the extensibility requirement!

Drakonian:
@JesperSchulz

Great, could you trigger pipeline again?)

Drakonian:
@JesperSchulz

Progress is being made 😄 
I added the missing dependencies to the SharePoint Test application.

Drakonian:
@JesperSchulz

The next step, I would gladly check locally, but as it is ⌛

src/System Application/Test/DisabledTests/SharePointGraphAdvancedTest.json
```diff
@@ -0,0 +1,7 @@
+[
+    {
+        "codeunitId":  132985,
```
JesperSchulz, line 3: Just double-checking: we're going to check this in with this disabled test?

Drakonian: @JesperSchulz

I can play with indexes and lists of data as @pri-kise suggested

Drakonian: @JesperSchulz

It will just require bunch of changes for this PR

JesperSchulz: I'm fine with having that test disabled, if the quality of the released code is fine in spite of this. I've we're checking in with a known product defect, the story is different ;-)

Drakonian: This is more likely a shortcoming of the SharePoint Graph Test Library that I created, because I did not consider sequential API calls. The function itself works as intended.

In that case, I can create a separate PR to remove this DisabledTest, with the necessary changes, so we do not slow down this PR.

What do you think?

@JesperSchulz @pri-kise

JesperSchulz: Yes, that'd be preferable. Let's get this already rather large PR merged.

src/System Application/App/SharePoint/src/graph/models/SharePointGraphDriveItem.Table.al
Drakonian: Okay, I agree that it is safer to set `extensible = false` for new tables, thanks for context )

Drakonian:
@JesperSchulz @pri-kise

I set `extensible = false` for new temporary tables, as it's more safe approach
I also removed field level DataClassification from fields, I think SystemMetadata is correct classification as these table are temporary.

JesperSchulz:
Re-triggering CI. Closing/re-opening.

Drakonian:
@JesperSchulz

I'm not sure why the CA build has an error. Is there anything I need to do?

JesperSchulz:
> @JesperSchulz
> 
> This is a great opportunity, and I would be happy to participate!
> 
> Any time on Tuesday or Wednesday works for me.
> 
> Would you like to see a short demo at the meeting? Or do you have any other expectations for the meeting?

Love it! Let's book you in for Tuesday! Could you PM me your email address, then I'll send you an invite (LinkedIn, Viva, etc.).
The expectation is a short demo and quick walkthrough of the module's architecture. Then we can have a short discussion and afterwards we hope for sign-offs from the engineers 😊

src/System Application/App/SharePoint/src/graph/SharePointGraphClientImpl.Codeunit.al
```diff
@@ -0,0 +1,1941 @@
...
+
+    /// <summary>
+    /// Common initialization logic shared by all Initialize overloads.
+    /// </summary>
+    /// <param name="NewSharePointUrl">SharePoint site URL.</param>
+    local procedure InitializeCommon(NewSharePointUrl: Text)
+    begin
+        // If we have a new URL, clear the cached IDs so they'll be re-acquired
+        if SharePointUrl <> NewSharePointUrl then begin
+            SharePointUrl := NewSharePointUrl;
```
darjoo, line 158: We need to validate this Url to ensure it points to a sharepoint url. This ensures we're not going to accidently send details (data or credentials) to any external url that is not sharepoint.

See the following commit for an example, in the shopify file.
https://github.com/microsoft/BCApps/commit/a0c202c54ef582f10c0772a5b4875604d3282e62#diff-db42a5093fe95f67d41540be171811f9ceb98a2ecb044dca0ed185f041eaa9a9

src/System Application/App/SharePoint/src/graph/helpers/SharePointGraphParser.Codeunit.al
```diff
@@ -0,0 +1,422 @@
...
+        JsonArray := JsonToken.AsArray();
+        exit(ParseListCollection(JsonArray, GraphLists));
+    end;
+
+    /// <summary>
+    /// Parses a JSON array into a collection of SharePoint Graph List records.
+    /// </summary>
+    /// <param name="JsonArray">The JSON array to parse.</param>
+    /// <param name="GraphLists">The temporary record to populate.</param>
+    /// <returns>True if successfully parsed; otherwise false.</returns>
```
darjoo: Slightly misleading.
You will never get a false, it is either `True` or exception.

src/System Application/App/SharePoint/src/graph/helpers/SharePointGraphParser.Codeunit.al
```diff
@@ -0,0 +1,422 @@
...
+    /// <summary>
+    /// Parses a JSON object into a SharePoint Graph List record.
+    /// </summary>
+    /// <param name="JsonListObject">The JSON object to parse.</param>
+    /// <param name="GraphList">The record to populate.</param>
+    /// <returns>True if successfully parsed; otherwise false.</returns>
+    procedure ParseListItem(JsonListObject: JsonObject; var GraphList: Record "SharePoint Graph List" temporary): Boolean
+    var
+        JsonToken: JsonToken;
+        DtToken: JsonToken;
```
darjoo: What does Dt here stand for? I initially thought it is DateTime but it's used for template below.

src/System Application/App/SharePoint/src/graph/helpers/SharePointGraphParser.Codeunit.al
```diff
@@ -0,0 +1,422 @@
...
+        exit(ParseDriveItemCollection(JsonArray, DriveId, GraphDriveItems));
+    end;
+
+    /// <summary>
+    /// Parses a JSON array into a collection of SharePoint Graph Drive Item records.
+    /// </summary>
+    /// <param name="JsonArray">The JSON array to parse.</param>
+    /// <param name="DriveId">The ID of the drive the items belong to.</param>
+    /// <param name="GraphDriveItems">The temporary record to populate.</param>
+    /// <returns>True if successfully parsed; otherwise false.</returns>
```
darjoo: Similar here, True or exception only.

src/System Application/App/SharePoint/src/graph/helpers/SharePointGraphParser.Codeunit.al
```diff
@@ -0,0 +1,422 @@
...
+        JsonArray := JsonToken.AsArray();
+        exit(ParseDriveCollection(JsonArray, GraphDrives));
+    end;
+
+    /// <summary>
+    /// Parses a JSON array into a collection of SharePoint Graph Drive records.
+    /// </summary>
+    /// <param name="JsonArray">The JSON array to parse.</param>
+    /// <param name="GraphDrives">The temporary record to populate.</param>
+    /// <returns>True if successfully parsed; otherwise false.</returns>
```
darjoo: Another, True or exception

src/System Application/App/SharePoint/src/graph/helpers/SharePointGraphReqHelper.Codeunit.al
```diff
@@ -0,0 +1,650 @@
...
+    /// <param name="ContentRange">The content range header value (e.g., "bytes 0-1023/5000").</param>
+    /// <param name="ResponseJson">The JSON response.</param>
+    /// <returns>True if the chunk was uploaded successfully; otherwise false.</returns>
+    procedure UploadChunk(UploadUrl: Text; var ChunkContent: InStream; ContentRange: Text; var ResponseJson: JsonObject): Boolean
+    var
+        RestClient: Codeunit "Rest Client";
+        HttpContent: Codeunit "Http Content";
+        HttpResponseMessage: Codeunit "Http Response Message";
+    begin
+        // Important: For upload sessions, we don't use GraphClient
```
darjoo: Should this just be in the remarks documentation above rather than here?

src/System Application/App/SharePoint/src/graph/models/SharePointGraphDrive.Table.al
```diff
@@ -0,0 +1,96 @@
...
+    {
+        field(1; Id; Text[250])
+        {
+            Caption = 'Id';
+            Description = 'Unique identifier of the drive';
+        }
+        field(2; Name; Text[250])
+        {
+            Caption = 'Name';
+            Description = 'Name of the drive (document library)';
```
darjoo, line 30: This is not SystemMetaData. It is CustomerContent.

src/System Application/App/SharePoint/src/graph/models/SharePointGraphDrive.Table.al
```diff
@@ -0,0 +1,96 @@
...
+            Description = 'Type of drive (personal, business, documentLibrary)';
+        }
+        field(4; WebUrl; Text[2048])
+        {
+            Caption = 'Web URL';
+            Description = 'URL to access the drive in a web browser';
+        }
+        field(5; OwnerName; Text[250])
+        {
+            Caption = 'Owner Name';
```
darjoo, line 45: This is not SystemMetaData. It is CustomerContent if not EUPII.

src/System Application/App/SharePoint/src/graph/models/SharePointGraphDrive.Table.al
```diff
@@ -0,0 +1,96 @@
...
+            Description = 'Display name of the drive owner';
+        }
+        field(6; OwnerEmail; Text[250])
+        {
+            Caption = 'Owner Email';
```
darjoo, line 51: This is not SystemMetaData. It is CustomerContent if not EUPII.

src/System Application/App/SharePoint/src/graph/models/SharePointGraphDrive.Table.al
```diff
@@ -0,0 +1,96 @@
...
+            Description = 'Date and time when the drive was created';
+        }
+        field(8; LastModifiedDateTime; DateTime)
+        {
+            Caption = 'Last Modified Date Time';
+            Description = 'Date and time when the drive was last modified';
+        }
+        field(9; Description; Text[2048])
+        {
+            Caption = 'Description';
```
darjoo, line 67: This is not SystemMetaData. It is CustomerContent.

src/System Application/App/SharePoint/src/graph/models/SharePointGraphDriveItem.Table.al
```diff
@@ -0,0 +1,89 @@
...
+            Description = 'Unique identifier of the drive item';
+        }
+        field(2; DriveId; Text[250])
+        {
+            Caption = 'Drive Id';
+            Description = 'ID of the parent drive';
+        }
+        field(3; Name; Text[250])
+        {
+            Caption = 'Name';
```
darjoo, line 34: This is not SystemMetaData. It is CustomerContent.

src/System Application/App/SharePoint/src/graph/models/SharePointGraphDriveItem.Table.al
```diff
@@ -0,0 +1,89 @@
...
+            Description = 'Name of the item (file or folder name)';
+        }
+        field(4; ParentId; Text[250])
+        {
+            Caption = 'Parent Id';
+            Description = 'ID of the parent folder';
+        }
+        field(5; Path; Text[2048])
+        {
+            Caption = 'Path';
```
darjoo, line 45: This is not SystemMetaData. It is CustomerContent.

src/System Application/App/SharePoint/src/graph/models/SharePointGraphList.Table.al
```diff
@@ -0,0 +1,81 @@
...
+    fields
+    {
+        field(1; Id; Text[250])
+        {
+            Caption = 'Id';
+            Description = 'Unique identifier of the list';
+        }
+        field(2; DisplayName; Text[250])
+        {
+            Caption = 'Display Name';
```
darjoo, line 29: This is not SystemMetaData. It is CustomerContent.

src/System Application/App/SharePoint/src/graph/models/SharePointGraphList.Table.al
```diff
@@ -0,0 +1,81 @@
...
+            Description = 'Name of the list for display purposes';
+        }
+        field(3; Name; Text[250])
+        {
+            Caption = 'Name';
```
darjoo, line 35: This is not SystemMetaData. It is CustomerContent.

src/System Application/App/SharePoint/src/graph/models/SharePointGraphList.Table.al
```diff
@@ -0,0 +1,81 @@
...
+            Description = 'Name of the list';
+        }
+        field(4; Description; Text[2048])
+        {
+            Caption = 'Description';
```
darjoo, line 41: This is not SystemMetaData. It is CustomerContent.

src/System Application/App/SharePoint/src/graph/models/SharePointGraphListItem.Table.al
```diff
@@ -0,0 +1,128 @@
...
+            Description = 'Unique identifier of the list item';
+        }
+        field(2; ListId; Text[250])
+        {
+            Caption = 'List Id';
+            Description = 'ID of the parent list';
+        }
+        field(3; Title; Text[250])
+        {
+            Caption = 'Title';
```
darjoo, line 34: This is not SystemMetaData. It is CustomerContent.

src/System Application/App/SharePoint/src/graph/SharePointGraphClientImpl.Codeunit.al
```diff
@@ -0,0 +1,2095 @@
...
+        if JsonResponse.Get('id', JsonToken) then
+            DefaultDriveId := JsonToken.AsValue().AsText();
+    end;
+
+    local procedure ExtractHostName(Url: Text): Text
+    var
+        UriBuilder: Codeunit "Uri Builder";
+    begin
+        UriBuilder.Init(Url);
+        exit(UriBuilder.GetHost()); // Returns contoso.sharepoint.com
```
darjoo: I assume the comment is suppose to be an example?
Right now it sounds like it will return that specific host.

src/System Application/Test Library/SharePoint/src/graph/SharePointGraphAuthSpy.Codeunit.al
```diff
@@ -0,0 +1,30 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+namespace System.Test.Integration.Sharepoint;
+
+using System.Integration.Graph.Authorization;
+using System.RestClient;
+
+codeunit 132974 "SharePoint Graph Auth Spy" implements "Graph Authorization"
```
darjoo: What is Spy?

Drakonian: As far I know in testing terminology, a spy is a test double that records whether it was called, maybe that's not great word choice.

I can rename to "SharePoint Graph Auth Mock" if you think it will be better fit

src/System Application/App/SharePoint/src/graph/SharePointGraphClientImpl.Codeunit.al
Drakonian: Correct, fixed for mor clear understanding
I also can remove it you think it's good idea

src/System Application/Test Library/SharePoint/src/graph/SharePointGraphAuthSpy.Codeunit.al
Drakonian: Name updated to "SharePoint Graph Auth Mock"

src/System Application/App/SharePoint/src/graph/SharePointGraphClientImpl.Codeunit.al
```diff
@@ -0,0 +1,2095 @@
...
+    /// Common initialization logic shared by all Initialize overloads.
+    /// </summary>
+    /// <param name="NewSharePointUrl">SharePoint site URL.</param>
+    local procedure InitializeCommon(NewSharePointUrl: Text)
+    var
+        Uri: Codeunit Uri;
+    begin
+        // Validate that the URL points to a SharePoint domain to prevent sending credentials to external URLs
+        if not Uri.IsValidURIPattern(NewSharePointUrl, SharePointUrlPatternLbl) then begin
+            Session.LogMessage('', StrSubstNo(InvalidSharePointUrlErr, NewSharePointUrl), Verbosity::Error, DataClassification::CustomerContent, TelemetryScope::ExtensionPublisher, 'Category', GraphSharePointCategoryLbl);
```
Drakonian: It seems we missed Event Id here

PR #3688, opened 2025-05-14
[Draft] [IPD] [Email] Http Mock Test for Email

gggdttt:
#### Work Item(s) 
Fixes [AB#419737](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/419737)

### DIFF
--- a/src/System Application/Test/Email/app.json
+++ b/src/System Application/Test/Email/app.json
@@ -94,6 +94,12 @@
       "name": "Email - Outlook REST API",
       "publisher": "Microsoft",
       "version": "27.0.0.0"
+    },
+    {  
+      "id": "b5034210-b258-4983-8858-f5cbfd54cb35",
+      "name": "Library Outlook REST API",
+      "publisher": "Microsoft",
+      "version": "27.0.0.0"
     }
   ],
   "screenshots": [],
--- a/src/System Application/Test/Email/app.json
+++ b/src/System Application/Test/Email/app.json
@@ -82,11 +82,27 @@
       "name": "Retention Policy",
       "publisher": "Microsoft",
       "version": "27.0.0.0"
+    },
+    {
+      "id": "aceb66c8-472e-437c-81d3-27e6c07d0f14",
+      "name": "Email - Microsoft 365 Connector",
+      "publisher": "Microsoft",
+      "version": "27.0.0.0"
+  },
+    {
+      "id": "e6328152-bb29-4664-9dae-3bc7eaae1fd8",
+      "name": "Email - Outlook REST API",
+      "publisher": "Microsoft",
+      "version": "27.0.0.0"
     }
   ],
   "screenshots": [],
   "platform": "27.0.0.0",
   "idRanges": [
+    {
+      "from": 134055,
+      "to": 134055
+    },
     {
       "from": 134487,
       "to": 134487
--- a/src/System Application/Test/Email/src/EmailDefaultAttachmentTests.Codeunit.al
+++ b/src/System Application/Test/Email/src/EmailDefaultAttachmentTests.Codeunit.al
@@ -50,7 +50,7 @@ codeunit 134853 "Email Default Attachment Tests"
         EmailScenarioMock.AddMapping(Enum::"Email Scenario"::"Test Email Scenario", AccountId, Enum::"Email Connector"::"Test Email Connector");
 
         // [When] calling GetEmailAccount
-        // [Then] true is retuned and the email account is as expected
+        // [Then] true is returned and the email account is as expected
         Assert.IsTrue(EmailScenario.GetEmailAccount(Enum::"Email Scenario"::"Test Email Scenario", EmailAccount), 'There should be an email account');
         Assert.AreEqual(AccountId, EmailAccount."Account Id", 'Wrong account ID');
         Assert.AreEqual(Enum::"Email Connector"::"Test Email Connector", EmailAccount.Connector, 'Wrong connector');
@@ -92,7 +92,7 @@ codeunit 134853 "Email Default Attachment Tests"
         AddAttachment('Attachment2', InStream, Enum::"Email Scenario"::"Test Email Scenario", false);
 
         // [When] calling GetEmailAccount
-        // [Then] true is retuned and the email account is as expected
+        // [Then] true is returned and the email account is as expected
         Assert.IsTrue(EmailScenario.GetEmailAccount(Enum::"Email Scenario"::"Test Email Scenario", EmailAccount), 'There should be an email account');
         Assert.AreEqual(AccountId, EmailAccount."Account Id", 'Wrong account ID');
         Assert.AreEqual(Enum::"Email Connector"::"Test Email Connector", EmailAccount.Connector, 'Wrong connector');
--- a/src/System Application/Test/Email/src/EmailScenarioTest.Codeunit.al
+++ b/src/System Application/Test/Email/src/EmailScenarioTest.Codeunit.al
@@ -36,7 +36,7 @@ codeunit 134693 "Email Scenario Test"
         Initialize();
 
         // [When] calling GetEmailAccount
-        // [Then] false is retuned
+        // [Then] false is returned
         Assert.IsFalse(EmailScenario.GetEmailAccount(Enum::"Email Scenario"::"Test Email Scenario", EmailAccount), 'There should not be any account');
     end;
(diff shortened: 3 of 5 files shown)


src/System Application/Test Library/Email/src/HttpMockEmailMgnt.Codeunit.al
```diff
@@ -0,0 +1,116 @@
...
+    var
+        ConnectorMock: Codeunit "Connector Mock";
+    begin
+        ConnectorMock.AddAccount(EmailAccount, Connector);
+    end;
+
+    procedure HasAtLeastOneEmailInSentEmail(EmailMessageId: Guid): Boolean
+    begin
+        SentEmail.SetRange("Message Id", EmailMessageId);
+        exit(SentEmail.FindFirst());
```
haoranpb:
```suggestion
        exit(not SentEmail.IsEmpty());
```

src/System Application/Test Library/Email/src/HttpMockEmailMgnt.Codeunit.al
```diff
@@ -0,0 +1,116 @@
...
+        exit(ExpectedConnector = SentEmail.Connector);
+    end;
+
+    procedure SetupEmailOutbox(EmailMessageId: Guid; Connector: Enum "Email Connector"; EmailAccountId: Guid; EmailDescription: Text; EmailAddress: Text[250]; UserSecurityId: Code[50])
+    var
+        EmailOutbox: Record "Email Outbox";
+    begin
+        EmailOutbox."Message Id" := EmailMessageId;
+        EmailOutbox.Insert();
+        EmailOutbox.Connector := Connector;
```
haoranpb: Consider using Validate instead of Value assignment

src/System Application/Test Library/Email/src/HttpMockEmailMgnt.Codeunit.al
```diff
@@ -0,0 +1,116 @@
...
+        exit(EmailError.Count());
+    end;
+
+    procedure GetEmailOutBoxStatusWithMessageId(MessageId: Guid): Enum "Email Status"
+    var
+        EmailOutbox: Record "Email Outbox";
+    begin
+        EmailOutbox.SetRange("Message Id", MessageId);
+        if not EmailOutbox.FindFirst() then
+            Error('No Email Outbox record found for the given Message Id.');
```
haoranpb: use label?

haoranpb (approved):

ankosiorek (dismissed):

ChethanT (dismissed):

src/System Application/Test Library/Email/src/HttpMockEmailMgnt.Codeunit.al
```diff
@@ -0,0 +1,116 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.TestLibraries.Email;
+
+using System.Email;
+
+codeunit 134056 "Http Mock Email Mgnt."
```
nikolakukrika: Mgt is a nasty word, so is helper. It means the codeunit does not have a clear purpose but it is a list of hacks/helpers.
This codeunit looks like Library - Email.  I do not like mocking individual objects in the libraries. If you want to mock the actual object, then the name is Http Email Mgt Mock. I do not like mocking the actual objects like this in general as it is better to get the proper concepts implemented.

src/System Application/Test Library/Email/src/HttpMockEmailMgnt.Codeunit.al
```diff
@@ -0,0 +1,116 @@
...
+
+
+    procedure AddAccount(var EmailAccount: Record "Email Account"; Connector: Enum "Email Connector")
+    var
+        ConnectorMock: Codeunit "Connector Mock";
+    begin
+        ConnectorMock.AddAccount(EmailAccount, Connector);
+    end;
+
+    procedure HasAtLeastOneEmailInSentEmail(EmailMessageId: Guid): Boolean
```
nikolakukrika: SentEmailExists?

src/System Application/Test Library/Email/src/HttpMockEmailMgnt.Codeunit.al
```diff
@@ -0,0 +1,116 @@
...
+    begin
+        SentEmail.SetRange("Message Id", EmailMessageId);
+        exit(not SentEmail.IsEmpty());
+    end;
+
+    procedure CheckSentEmailDescription(EmailMessageId: Guid; ExpectedDescription: Text): Boolean
```
nikolakukrika: Why do we not have the GetSentEmailDescription? It is easier to reuse.
The check does not help much, we could in the test do Assert.AreEqual or different ways of checking

src/System Application/Test Library/Email/src/HttpMockEmailMgnt.Codeunit.al
```diff
@@ -0,0 +1,116 @@
...
+    begin
+        SentEmail.SetRange("Message Id", EmailMessageId);
+        if not SentEmail.FindFirst() then
+            Error(NoSentEmailRecordFoundErr);
+        exit(ExpectedDescription = SentEmail.Description);
+    end;
+
+
+    procedure CheckSentEmailFrom(EmailMessageId: Guid; ExpectedFrom: Text): Boolean
```
nikolakukrika: Same for other Check methods, they should be changed to Get, it is more flexible.

src/System Application/Test Library/Email/src/HttpMockEmailMgnt.Codeunit.al
```diff
@@ -0,0 +1,116 @@
...
+    end;
+
+    procedure RunEmailDispatcher(EmailMessageId: Guid)
+    var
+        EmailOutbox: Record "Email Outbox";
+    begin
+        EmailOutbox.SetRange("Message Id", EmailMessageId);
+        if not EmailOutbox.FindFirst() then
+            Error(NoSentEmailRecordFoundErr);
+        Codeunit.Run(Codeunit::"Email Dispatcher", EmailOutbox);
```
nikolakukrika: Should this one be configurable? We could mock this codeunit.

src/System Application/Test Library/Email/src/HttpMockEmailMgnt.Codeunit.al
```diff
@@ -0,0 +1,116 @@
...
+        exit(EmailError.Count());
+    end;
+
+    procedure GetEmailOutBoxStatusWithMessageId(MessageId: Guid): Enum "Email Status"
+    var
+        EmailOutbox: Record "Email Outbox";
+    begin
+        EmailOutbox.SetRange("Message Id", MessageId);
+        if not EmailOutbox.FindFirst() then
+            Error('No Email Outbox record found for the given Message Id.');
```
nikolakukrika: Tests are not using Errors in the code. We use Assert.Error so we know from stack trace that it is not a product that is throwing an error but a test. Please update other error messages too.

src/System Application/Test Library/Email/src/HttpMockEmailMgnt.Codeunit.al
gggdttt: @nikolakukrika  My origin idea is we should not return the internal table's value directly.  But  maybe it is safe for Email table. Let me change it..

nikolakukrika: @gggdttt This is a test context, it is a library. We are not exposing anything new, the page is listing emails, there should be helper methods that expose the values. 
I do not find this as a security issue as we are in a test context, however we need to be careful in general when writing the test libraries.

src/System Application/Test Library/Email/src/HttpMockEmailMgnt.Codeunit.al
gggdttt: We can configure it from our test by change the value for the corresponding EmailMessage.
The reason why I add this is "Email Dispatcher" is internal. We need one way to call it in our test.

darjoo (approved):

PR #3760, opened 2025-05-27
Update codecop ruleset

mohammedtabish0:
#### Summary
Update ruleset for Autoformat codecop rules 471, 472, 473 and 474.

### DIFF
--- a/src/rulesets/CodeCop.ruleset.json
+++ b/src/rulesets/CodeCop.ruleset.json
@@ -1886,19 +1886,19 @@
                   },
                   {
                       "id":  "AA0471",
-                      "action":  "Error"
+                      "action":  "Info"
                   },
                   {
                       "id":  "AA0472",
-                      "action":  "Error"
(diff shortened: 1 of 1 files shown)


src/rulesets/CodeCop.ruleset.json
```diff
@@ -1886,19 +1886,19 @@
                   },
                   {
                       "id":  "AA0471",
-                      "action":  "Error"
+                      "action":  "Info"
```
grobyns: We don't use Info or Warning. You can have Error or None. If you make it 'None' you should add a justification as to why this rule will never apply. 

If this is a temporary change and not a permanent one, you should make the change in 'ruleset.json' instead.

mohammedtabish0: Thanks, updated. Have added some justification for `why` these rules are let me know if justification should be for making rule `None` but these lines kind of tell that as well.

PR #3784, opened 2025-06-01
Add support for agent suggestions and free text instructions in the agent task pane

eridanurce:
Fixes
[AB#579777](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/579777)
[AB#579778](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/579778)

### DIFF
--- a/src/System Application/App/Agent/Interaction/AgentTaskImpl.Codeunit.al
+++ b/src/System Application/App/Agent/Interaction/AgentTaskImpl.Codeunit.al
@@ -113,6 +113,21 @@ codeunit 4300 "Agent Task Impl."
         CreateUserIntervention(UserInterventionRequestEntry, '', SelectedSuggestionId);
     end;
 
+    procedure CreateUserIntervention(UserInterventionRequestEntry: Record "Agent Task Log Entry"; UserInput: Text; SelectedSuggestionId: Text)
+    var
+        SelectedSuggestionIdInt: Integer;
+    begin
+        if SelectedSuggestionId <> '' then
+            if Evaluate(SelectedSuggestionIdInt, SelectedSuggestionId) then begin
+                CreateUserIntervention(UserInterventionRequestEntry, UserInput, SelectedSuggestionIdInt);
+                exit;
+            end
+            else
+                Session.LogMessage('0000PKA', StrSubstNo(InvalidSelectedSuggestionIdErr, SelectedSuggestionId), Verbosity::Error, DataClassification::SystemMetadata, TelemetryScope::ExtensionPublisher, 'Category', CategoryTok);
+
+        CreateUserIntervention(UserInterventionRequestEntry, UserInput);
+    end;
+
     procedure CreateUserIntervention(UserInterventionRequestEntry: Record "Agent Task Log Entry"; UserInput: Text; SelectedSuggestionId: Integer)
     var
         AgentTask: Record "Agent Task";
@@ -217,4 +232,6 @@ codeunit 4300 "Agent Task Impl."
         MessageTextMustBeProvidedErr: Label 'You must provide a message text.';
         AreYouSureThatYouWantToRestartTheTaskQst: Label 'Are you sure that you want to restart the task?';
         AreYouSureThatYouWantToStopTheTaskQst: Label 'Are you sure that you want to stop the task?';
+        InvalidSelectedSuggestionIdErr: Label 'Invalid SelectedSuggestionId: %1', Comment = '%1 - SelectedSuggestionId', Locked = true;
+        CategoryTok: Label 'Agent App', Locked = true;
 }
--- a/src/System Application/App/Agent/Interaction/AgentTaskImpl.Codeunit.al
+++ b/src/System Application/App/Agent/Interaction/AgentTaskImpl.Codeunit.al
@@ -233,5 +233,5 @@ codeunit 4300 "Agent Task Impl."
         AreYouSureThatYouWantToRestartTheTaskQst: Label 'Are you sure that you want to restart the task?';
         AreYouSureThatYouWantToStopTheTaskQst: Label 'Are you sure that you want to stop the task?';
         InvalidSelectedSuggestionIdErr: Label 'Invalid SelectedSuggestionId: %1', Comment = '%1 - SelectedSuggestionId', Locked = true;
-        CategoryTok: Label 'Agent App', Locked = true;
+        CategoryTok: Label 'Agents', Locked = true;
 }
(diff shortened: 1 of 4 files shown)


esardaya (dismissed):

qasimikram (dismissed):

qasimikram (approved):

src/System Application/App/Agent/TaskPane/AgentTaskDetails.Page.al
```diff
@@ -65,6 +75,12 @@ page 4313 "Agent Task Details"
         }
     }
 
+    trigger OnOpenPage()
+    begin
+        SelectedSuggestionId := '';
```
nikolakukrika: Minor - CLEAR(SelectedSuggestionId), same for other

src/System Application/App/Agent/TaskPane/AgentTaskDetails.Page.al
```diff
@@ -23,6 +21,18 @@ page 4313 "Agent Task Details"
     {
         area(content)
         {
+            field(SelectedSuggestionId; SelectedSuggestionId)
+            {
+                Editable = true;
+                Caption = 'Selected Suggestion ID';
+                ToolTip = 'Specifies the selected suggestion ID for the user intervention request.';
+            }
```
nikolakukrika, line 29: If this field is being set by the user, we should do the evaluate on the validate trigger of this field.
Otherwise the logic should be moved to the codeunit, we should not have it on 2 pages.

src/System Application/App/Agent/TaskPane/AgentTaskDetails.Page.al
```diff
@@ -23,6 +21,18 @@ page 4313 "Agent Task Details"
...
+            field(UserMessage; UserMessage)
+            {
+                Editable = true;
+                Caption = 'Additional Instructions';
```
nikolakukrika, line 33: Should this one be multiline?

src/System Application/App/Agent/TaskPane/AgentTaskTimeline.Page.al
```diff
@@ -197,6 +212,11 @@ page 4307 "Agent Task Timeline"
                 GlobalAnnotations.Read(InStream);
                 Clear(InStream);
             end;
+            if Rec."Last User Intervention Details".HasValue then begin
```
nikolakukrika: Minor - HasValue**()**.

nikolakukrika (approved):

src/System Application/App/Agent/TaskPane/AgentTaskDetails.Page.al
eridanurce: There's custom logic on the client for these task pane fields, so the multiline here would not matter.

src/System Application/App/Agent/TaskPane/AgentTaskDetails.Page.al
```diff
@@ -11,9 +11,7 @@ page 4313 "Agent Task Details"
     ApplicationArea = All;
     SourceTable = "Agent Task Timeline Step Det.";
     Caption = 'Agent Task Timeline Step Details';
-    Editable = false;
```
darjoo, line 14: Are `SelectedSuggestionId` and `UserMessage` the only two that should be editable?
The repeater should then contain the `editable=false;` property.

src/System Application/App/Agent/TaskPane/AgentTaskDetails.Page.al
```diff
@@ -126,7 +138,8 @@ page 4313 "Agent Task Details"
     var
         AgentTaskImpl: Codeunit "Agent Task Impl.";
         ClientContext: BigText;
-        UserMessage: Text;
+        UserMessage: Text[250];
+        SelectedSuggestionId: Text[3];
```
darjoo, line 139: What is this used for? Do the end user know it should be 3 chars and what they should be?

eridanurce: This is used for suggestions; the field is not visible to users, we render clickable suggestions instead. When a user has selected one of the suggestions and clicks confirm, the client code populates this field with the ID of that suggestion and then invokes the "Confirm" action. Users don't type into this field

src/System Application/App/Agent/TaskPane/AgentTaskDetails.Page.al
eridanurce: thanks, added it for the repeater on both AgentTaskDetails and AgentTaskTimeline pages

src/System Application/App/Agent/TaskPane/AgentTaskDetails.Page.al
eridanurce: Discussed on teams, the client code writes the ID of the selected suggestion into this field right before the Confirm action is invoked, so the validation can stay there.
I've refactored to move the suggestion id validation logic to the codeunit.

src/System Application/App/Agent/Interaction/AgentTaskImpl.Codeunit.al
```diff
@@ -217,4 +232,6 @@ codeunit 4300 "Agent Task Impl."
         MessageTextMustBeProvidedErr: Label 'You must provide a message text.';
         AreYouSureThatYouWantToRestartTheTaskQst: Label 'Are you sure that you want to restart the task?';
         AreYouSureThatYouWantToStopTheTaskQst: Label 'Are you sure that you want to stop the task?';
+        InvalidSelectedSuggestionIdErr: Label 'Invalid SelectedSuggestionId: %1', Comment = '%1 - SelectedSuggestionId', Locked = true;
+        CategoryTok: Label 'Agent App', Locked = true;
```
nikolakukrika: Do we have the category already? It would be good to reuse. 
We can either reuse the same as the platform, or name it the same as platform with AL prefix, to indicate it is coming from AL. For me it is better to align unless we have a need to distinguish AL from platform.

nikolakukrika (dismissed):

src/System Application/App/Agent/Interaction/AgentTaskImpl.Codeunit.al
eridanurce: Have aligned with platform category: "Agents"

esardaya (approved):

nikolakukrika (approved):

PR #3785, opened 2025-06-02
Pdf to Img

Groenbech96:
#### Summary 

Adds a function to convert pdf to image.
Module will be build out later for e-invocing scenarios.

#### Work Item(s) 
Fixes [AB#578388](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/578388)


src/System Application/App/Document Processor/README.md
```diff
@@ -0,0 +1,85 @@
...
+procedure Example(PdfStream: InStream)
+var
+    TempBlob: Codeunit "Temp Blob";
+    Success: Boolean;
+    PdfHelper: Codeunit "PDF Helper Impl";
+begin
+    Success := PdfHelper.GetInvoiceAttachmentStream(PdfStream, TempBlob);
+    if Success then
+        Message('Invoice extracted successfully');
+end;
```
NKarolak: All code samples should be enclosed with ` ``` ` to be formatted; see for example https://github.com/microsoft/BCApps/blob/main/src/System%20Application/App/Image/README.md?plain=1

src/System Application/App/Document Processor/src/PDFHelperImpl.Codeunit.al
```diff
@@ -0,0 +1,65 @@
...
+        // Use a shared stream and reset the read pointer to beginning of stream.
+        SharedDocumentStream := DocumentStream;
+        if SharedDocumentStream.Position > 1 then
+            SharedDocumentStream.ResetPosition();
+
+        MemoryStream := MemoryStream.MemoryStream();
+        MemoryStream := SharedDocumentStream;
+
+        ConvertImageFormatToPdfTargetDevice(ImageFormat, PdfTargetDevice);
+        ImageMemoryStream := PdfConverterInstance.ConvertPage(PdfTargetDevice, MemoryStream, PageNumber, 0, 0, 0); // apply default heighth, width and resolution
```
NKarolak:
```suggestion
        ImageMemoryStream := PdfConverterInstance.ConvertPage(PdfTargetDevice, MemoryStream, PageNumber, 0, 0, 0); // apply default height, width and resolution
```

src/System Application/App/Document Processor/src/PDFHelperImpl.Codeunit.al
```diff
@@ -0,0 +1,65 @@
...
+        case ImageFormat of
+            ImageFormat::PNG:
+                PdfTargetDevice := PdfTargetDevice.PngDevice;
+            ImageFormat::JPEG:
+                PdfTargetDevice := PdfTargetDevice.JpegDevice;
+            ImageFormat::TIFF:
+                PdfTargetDevice := PdfTargetDevice.TiffDevice;
+            ImageFormat::BMP:
+                PdfTargetDevice := PdfTargetDevice.BmpDevice;
+            ImageFormat::GIF:
```
NKarolak:
```suggestion
            ImageFormat::Png:
                PdfTargetDevice := PdfTargetDevice.PngDevice;
            ImageFormat::Jpeg:
                PdfTargetDevice := PdfTargetDevice.JpegDevice;
            ImageFormat::Tiff:
                PdfTargetDevice := PdfTargetDevice.TiffDevice;
            ImageFormat::Bmp:
                PdfTargetDevice := PdfTargetDevice.BmpDevice;
            ImageFormat::Gif:
```

src/System Application/App/Document Processor/src/PDFHelperImpl.Codeunit.al
```diff
@@ -0,0 +1,65 @@
...
+                PdfTargetDevice := PdfTargetDevice.GifDevice;
+            else
+                Error('Unsupported image format: %1', ImageFormat);
```
NKarolak: Shouldn't this be a translatable label?

src/System Application/Test/Document Processor/src/DocumentProcessorTest.Codeunit.al
```diff
@@ -0,0 +1,38 @@
...
+using System.Utilities;
+using System.TestLibraries.Utilities;
+
+
+codeunit 132601 "Document Processor Test"
+{
+    Access = Internal;
+    Subtype = Test;
+    TestPermissions = Disabled;
+
```
NKarolak:
```suggestion
```

src/System Application/Test/Document Processor/src/DocumentProcessorTest.Codeunit.al
```diff
@@ -0,0 +1,38 @@
...
+        PdfInstream, ImageStream : InStream;
+    begin
+        // Setup
+        NavApp.GetResource('test.pdf', PdfInstream, TextEncoding::UTF8);
+        TempBlob.CreateInStream(ImageStream);
+        PdfHelper.ConvertPdfToImage(PdfInstream, ImageStream, ImageFormat::Png, 1);
+        Assert.AreNotEqual(0, TempBlob.Length(), LengthErr);
+    end;
+
+
```
NKarolak:
```suggestion
```

src/System Application/App/Document Processor/app.json
```diff
@@ -0,0 +1,37 @@
+{
+  "id": "2c7afcfb-9625-4ace-b972-5889d045ce8f",
+  "name": "Document Processor",
```
nikolakukrika: Consider naming PDF Document if we only want to limit this module to PDF documents. I think it is valid, we could introduce a module for each of the specific documents we process, like Word, PDF...

Maybe we do not need to limit the module to extracting the data only, but to PDF documents.

If we want the module that extracts the data from all different documents it should be renamed to Document Data Extractor or something like this, maybe Document to Image?

PDF Document is better for me as it would be easier to use.

NKarolak (approved):

src/System Application/App/Document Processor/src/PDFHelper.Codeunit.al
```diff
@@ -0,0 +1,33 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+namespace System.IO;
+
+using System.Utilities;
+
+/// <summary>
+/// Codeunit that provides helper functions for PDF processing.
+/// </summary>
+codeunit 3110 "PDF Helper"
```
nikolakukrika: No helpers, helpers means this is a hack. We introduced debt here.
Management/Mgt. means less hacks, we do not have a clear purpose. 

I would name the file PDF Document

src/System Application/App/Document Processor/src/PDFHelper.Codeunit.al
```diff
@@ -0,0 +1,33 @@
...
+        PDFHelperImpl: Codeunit "PDF Helper Impl";
+
+    /// <summary>
+    /// This procedure is used to convert a PDF file to an image.
+    /// </summary>
+    /// <param name="PdfInStream">Stream of the PDF file.</param>
+    /// <param name="ImageStream">Stream of the image file.</param>
+    /// <param name="ImageFormat">Image format to convert the PDF to.</param>
+    /// <param name="PageNumber">Page number to convert.</param>
+    procedure ConvertPdfToImage(PdfInStream: InStream; var ImageStream: InStream; ImageFormat: Enum "Image Format"; PageNumber: Integer)
```
nikolakukrika: I believe this one should be more class like.
I would add LoadPDFDocument and remove one parameter from PDFInStream

nikolakukrika: ConvertPageToImage, no need to repeat PDF and we are missing the Page to indicate it is one page.
Page number should go first.

We should get an overload that returns the first page I believe if that is the use case.
Do we need a method to return how many pages are in the document? Otherwise how could we loop and get all images?

src/System Application/App/Document Processor/src/PDFHelperImpl.Codeunit.al
```diff
@@ -0,0 +1,68 @@
...
+    procedure ConvertPdfToImage(DocumentStream: InStream; var ImageStream: InStream; ImageFormat: Enum "Image Format"; PageNumber: Integer): Boolean
+    var
+        PdfConverterInstance: DotNet PdfConverter;
+        PdfTargetDevice: DotNet PdfTargetDevice;
+        MemoryStream: DotNet MemoryStream;
+        ImageMemoryStream: DotNet MemoryStream;
+        SharedDocumentStream: InStream;
+    begin
+        // Empty stream, no actions possible on the stream so return immediately
+        if DocumentStream.Length < 1 then
```
nikolakukrika: I do not think this is a proper way of checking if the stream has the data in it.

src/System Application/App/Document Processor/src/PDFHelperImpl.Codeunit.al
```diff
@@ -0,0 +1,68 @@
...
+            exit(false);
+
+        // Use a shared stream and reset the read pointer to beginning of stream.
+        SharedDocumentStream := DocumentStream;
+        if SharedDocumentStream.Position > 1 then
+            SharedDocumentStream.ResetPosition();
+
+        MemoryStream := MemoryStream.MemoryStream();
+        MemoryStream := SharedDocumentStream;
```
nikolakukrika: There were big issues with direct assigments of memory streams, we have removed them from the product. I cannot remember why exactly.

I believe CopyStream is better. Example in codeunit 3009 DotNet_MemoryStream, we also use it in the other places.

src/System Application/App/Document Processor/src/PDFHelperImpl.Codeunit.al
```diff
@@ -0,0 +1,68 @@
...
+
+        ConvertImageFormatToPdfTargetDevice(ImageFormat, PdfTargetDevice);
+        ImageMemoryStream := PdfConverterInstance.ConvertPage(PdfTargetDevice, MemoryStream, PageNumber, 0, 0, 0); // apply default height, width and resolution
```
nikolakukrika: Minor - Should we introduce variables, default height, width and resolution? We could remove the comment, but it is 3 extra variables, so decide yourself.

src/System Application/Test/Document Processor/src/DocumentProcessorTest.Codeunit.al
```diff
@@ -0,0 +1,36 @@
...
+        PdfHelper: Codeunit "PDF Helper";
+        TempBlob: Codeunit "Temp Blob";
+        ImageFormat: Enum "Image Format";
+        PdfInstream, ImageStream : InStream;
+    begin
+        // Setup
+        NavApp.GetResource('test.pdf', PdfInstream, TextEncoding::UTF8);
+        TempBlob.CreateInStream(ImageStream);
+        PdfHelper.ConvertPdfToImage(PdfInstream, ImageStream, ImageFormat::Png, 1);
+        Assert.AreNotEqual(0, TempBlob.Length(), LengthErr);
```
nikolakukrika: You could also load the expected image to the resources and compare 2 base64 strings to ensure that the output is correct.

src/System Application/App/Document Processor/src/PDFHelper.Codeunit.al
Groenbech96: Why is it relevant? This pr adds simple function to convert a pdf page to image, as it is the current requirements. There is lots of work behind in the platform to enable it, and also license requirements we have to uphold. For now we are not going there.

src/System Application/App/Document Processor/app.json
Groenbech96: No reason to overcomplicate with modules for each. I will leave the Document Processor as module name and rename CodeUnit to PDF Document Processor.

src/System Application/App/Document Processor/src/PDFHelper.Codeunit.al
Groenbech96: Resolved.

src/System Application/App/Document Processor/src/PDFHelper.Codeunit.al
Groenbech96: This function provides 1 page conversion.

src/System Application/App/Document Processor/src/PDFHelperImpl.Codeunit.al
Groenbech96: Length is ok to check.

src/System Application/App/Document Processor/src/PDFHelperImpl.Codeunit.al
Groenbech96: Fixed.

src/System Application/App/Document Processor/app.json
nikolakukrika: It does not have a clear purpose. Document Processor is incredibly generic term. 
If I need to edit the PDF or export images out of it, I would not connect the Document Processing with these 2 actions.

src/System Application/Test/Document Processor/src/PDFDocumentProcessorTest.Codeunit.al
```diff
@@ -0,0 +1,36 @@
...
+        PdfHelper: Codeunit "PDF Document Processor";
+        TempBlob: Codeunit "Temp Blob";
+        ImageFormat: Enum "Image Format";
+        PdfInstream, ImageStream : InStream;
+    begin
+        // Setup
+        NavApp.GetResource('test.pdf', PdfInstream, TextEncoding::UTF8);
+        TempBlob.CreateInStream(ImageStream);
+        PdfHelper.ConvertToImage(PdfInstream, ImageStream, ImageFormat::Png, 1);
+        Assert.AreNotEqual(0, TempBlob.Length(), LengthErr);
```
nikolakukrika: Please improve the test with check that the proper image is returned. You can compare 2 base64 strings.

src/System Application/App/Document Processor/src/PDFDocumentProcessor.Codeunit.al
```diff
@@ -0,0 +1,33 @@
...
+        PDFHelperImpl: Codeunit "PDF Document Processor Impl.";
+
+    /// <summary>
+    /// This procedure is used to convert a PDF file to an image.
+    /// </summary>
+    /// <param name="DocumentStream">Stream of the PDF file.</param>
+    /// <param name="ImageStream">Stream of the image file.</param>
+    /// <param name="ImageFormat">Image format to convert the PDF to.</param>
+    /// <param name="PageNumber">Page number to convert.</param>
+    procedure ConvertToImage(DocumentStream: InStream; var ImageStream: InStream; ImageFormat: Enum "Image Format"; PageNumber: Integer)
```
nikolakukrika: The current implementation does not make much sense.

We should at least return the number of pages that the document has. 
Consider adding a method that converts the first page, so page number is not needed.

src/System Application/App/Document Processor/src/PDFDocumentProcessorImpl.Codeunit.al
```diff
@@ -0,0 +1,68 @@
...
+    procedure ConvertToImage(DocumentStream: InStream; var ImageStream: InStream; ImageFormat: Enum "Image Format"; PageNumber: Integer): Boolean
+    var
+        PdfConverterInstance: DotNet PdfConverter;
+        PdfTargetDevice: DotNet PdfTargetDevice;
+        MemoryStream: DotNet MemoryStream;
+        ImageMemoryStream: DotNet MemoryStream;
+        SharedDocumentStream: InStream;
+    begin
+        // Empty stream, no actions possible on the stream so return immediately
+        if DocumentStream.Length() < 1 then
```
nikolakukrika: Related to previous comment, we have no such checks in the product as far as I can see. This is the first time we check if it is less than 1. I'm not sure if it is correct. It can be -1 if length cannot be determined, I'm not sure when it is raised.

src/System Application/App/Document Processor/app.json
Groenbech96: And PDF document is even more generic. It is not a class like structure we are building. It is helper functions to do things on pdfs. Such as convert to IMG, fetch xml embedded inside. 
Why we chose PDF helper in the beginning.

src/System Application/App/Document Processor/src/PDFDocumentProcessor.Codeunit.al
Groenbech96: I have a use case where this is what I need. Why is there a general need to more? We can always add stuff, it is not like we are finalizing things. I think you are weirdly gatekeeping. Why should i return the number of pages? Its not relevant to what i am doing? 

Someone else later can just add it they need it.

src/System Application/App/Document Processor/src/PDFDocumentProcessorImpl.Codeunit.al
Groenbech96: Im all ears. How else would you determine it? There sis no IsEmpty, there is no blob. This is the only method that says something about the content size.

nikolakukrika (approved):

ankosiorek (approved):

PR #3811, opened 2025-06-03
Add Visual Indication to the user for paid copilot features

rupeshdan:
#### Work Item(s) <https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/568080>

Fixes #
1. Add a new column in 'copilot and agent capabilities' page. 
Column Name - Billing Type
Possible Values - Not Billed, Microsoft Billed, Custom Billed, Undefined (internal use only for existing capabilities)
NOTE: Naming conventions are yet to be finalized, feedback welcomed.
2. Authorize the usage of AI resources based on selected billing type at registration of a capability

image

### DIFF
--- a/src/System Application/App/AI/src/Azure AI Document Intelligence/AzureDIImpl.Codeunit.al
+++ b/src/System Application/App/AI/src/Azure AI Document Intelligence/AzureDIImpl.Codeunit.al
@@ -36,7 +36,7 @@ codeunit 7779 "Azure DI Impl." implements "AI Service Name"
 
     procedure RegisterCopilotCapability(CopilotCapability: Enum "Copilot Capability"; CopilotAvailability: Enum "Copilot Availability"; LearnMoreUrl: Text[2048]; CallerModuleInfo: ModuleInfo)
     begin
-        CopilotCapabilityImpl.RegisterCapability(CopilotCapability, CopilotAvailability, Enum::"Azure AI Service Type"::"Azure Document Intelligence", LearnMoreUrl, CallerModuleInfo);
+        CopilotCapabilityImpl.RegisterCapability(CopilotCapability, CopilotAvailability, Enum::"Copilot Billing Type"::"Undefined", Enum::"Azure AI Service Type"::"Azure Document Intelligence", LearnMoreUrl, CallerModuleInfo);
     end;
 
     /// <summary>
--- a/src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
+++ b/src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
@@ -47,7 +47,7 @@ codeunit 7772 "Azure OpenAI Impl" implements "AI Service Name"
         TelemetryMetapromptRetrievalErr: Label 'Unable to retrieve metaprompt from Azure Key Vault.', Locked = true;
         TelemetryFunctionCallingFailedErr: Label 'Function calling failed for function: %1', Comment = '%1 is the name of the function', Locked = true;
         AzureOpenAiTxt: Label 'Azure OpenAI', Locked = true;
-        BillingTypeAuthorizationErr: Label 'Usage of resources not authorized with current billing type.', Locked = true;
+        BillingTypeAuthorizationErr: Label 'Usage of AI resources not authorized with chosen billing type. Please reach out to your administrator to resolve this issue.';
 
     procedure IsEnabled(Capability: Enum "Copilot Capability"; CallerModuleInfo: ModuleInfo): Boolean
     begin
--- a/src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
+++ b/src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
@@ -47,6 +47,7 @@ codeunit 7772 "Azure OpenAI Impl" implements "AI Service Name"
         TelemetryMetapromptRetrievalErr: Label 'Unable to retrieve metaprompt from Azure Key Vault.', Locked = true;
         TelemetryFunctionCallingFailedErr: Label 'Function calling failed for function: %1', Comment = '%1 is the name of the function', Locked = true;
         AzureOpenAiTxt: Label 'Azure OpenAI', Locked = true;
+        BillingTypeAuthorizationErr: Label 'Usage of resources not authorized with current billing type.', Locked = true;
 
     procedure IsEnabled(Capability: Enum "Copilot Capability"; CallerModuleInfo: ModuleInfo): Boolean
     begin
@@ -459,6 +460,10 @@ codeunit 7772 "Azure OpenAI Impl" implements "AI Service Name"
         EmptySecretText: SecretText;
     begin
         ClearLastError();
+
+        if not IsBillingTypeAuthorized(AOAIAuthorization, CallerModuleInfo) then
+            Error(BillingTypeAuthorizationErr);
+
         case AOAIAuthorization.GetResourceUtilization() of
             Enum::"AOAI Resource Utilization"::"Microsoft Managed":
                 ALCopilotAuthorization := ALCopilotAuthorization.Create(EmptySecretText, AOAIAuthorization.GetManagedResourceDeployment(), EmptySecretText);
@@ -602,4 +607,25 @@ codeunit 7772 "Azure OpenAI Impl" implements "AI Service Name"
         if not TempPrivacyNotice.Insert() then;
     end;
 
+    procedure IsBillingTypeAuthorized(AOAIAuthorization: Codeunit "AOAI Authorization"; CallerModuleInfo: ModuleInfo): Boolean
+    var
+        BillingType: Enum "Copilot Billing Type";
+    begin
+        BillingType := CopilotCapabilityImpl.GetCopilotBillingType();
+        if (CopilotCapabilityImpl.IsPublisherMicrosoft(CallerModuleInfo)) then begin
+            if (AOAIAuthorization.GetResourceUtilization() = Enum::"AOAI Resource Utilization"::"First Party") then
+                exit(BillingType <> Enum::"Copilot Billing Type"::"Custom Billed");
+        end else begin
+            case BillingType of
+                Enum::"Copilot Billing Type"::"Custom Billed":
+                    exit(AOAIAuthorization.GetResourceUtilization() = Enum::"AOAI Resource Utilization"::"Self-Managed");
+                Enum::"Copilot Billing Type"::"Microsoft Billed":
+                    exit(AOAIAuthorization.GetResourceUtilization() = Enum::"AOAI Resource Utilization"::"Microsoft Managed");
+                Enum::"Copilot Billing Type"::"Not Billed":
+                    exit(AOAIAuthorization.GetResourceUtilization() = Enum::"AOAI Resource Utilization"::"Self-Managed");
+                else
+                    exit(true);
+            end;
+        end;
+    end;
 }
--- a/src/System Application/App/AI/src/Copilot/CopilotCapability.Codeunit.al
+++ b/src/System Application/App/AI/src/Copilot/CopilotCapability.Codeunit.al
@@ -44,6 +44,21 @@ codeunit 7773 "Copilot Capability"
     end;
 
     /// <summary>
+    /// Register a capability.
+    /// </summary>
+    /// <param name="CopilotCapability">The capability.</param>
+    /// <param name="CopilotAvailability">The availability.</param>
+    /// <param name="CopilotBillingType">BillingType.</param>
+    /// <param name="LearnMoreUrl">The learn more url.</param>
+    procedure RegisterCapability(CopilotCapability: Enum "Copilot Capability"; CopilotAvailability: Enum "Copilot Availability"; CopilotBillingType: Enum "Copilot Billing Type"; LearnMoreUrl: Text[2048])
+    var
+        CallerModuleInfo: ModuleInfo;
+    begin
+        NavApp.GetCallerModuleInfo(CallerModuleInfo);
+        CopilotCapabilityImpl.RegisterCapability(CopilotCapability, CopilotAvailability, CopilotBillingType, LearnMoreUrl, CallerModuleInfo);
+    end;
+
+    /// <summary>
     /// Modify an existing capability.
     /// </summary>
     /// <param name="CopilotCapability">The capability.</param>
@@ -55,6 +70,21 @@ codeunit 7773 "Copilot Capability"
     begin
         NavApp.GetCallerModuleInfo(CallerModuleInfo);
         CopilotCapabilityImpl.ModifyCapability(CopilotCapability, CopilotAvailability, LearnMoreUrl, CallerModuleInfo);
+    end;
+
+    /// <summary>
+    /// Modify an existing capability.
+    /// </summary>
+    /// <param name="CopilotCapability">The capability.</param>
+    /// <param name="CopilotAvailability">The availability.</param>
+    /// <param name="CopilotBillingType">BillingType.</param>
+    /// <param name="LearnMoreUrl">The learn more url.</param>
+    procedure ModifyCapability(CopilotCapability: Enum "Copilot Capability"; CopilotAvailability: Enum "Copilot Availability"; CopilotBillingType: Enum "Copilot Billing Type"; LearnMoreUrl: Text[2048])
+    var
+        CallerModuleInfo: ModuleInfo;
+    begin
+        NavApp.GetCallerModuleInfo(CallerModuleInfo);
+        CopilotCapabilityImpl.ModifyCapability(CopilotCapability, CopilotAvailability, CopilotBillingType, LearnMoreUrl, CallerModuleInfo);
     end;
 
     /// <summary>
--- a/src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
+++ b/src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
@@ -52,10 +52,15 @@ codeunit 7774 "Copilot Capability Impl"
 
     procedure RegisterCapability(CopilotCapability: Enum "Copilot Capability"; CopilotAvailability: Enum "Copilot Availability"; LearnMoreUrl: Text[2048]; CallerModuleInfo: ModuleInfo)
     begin
-        RegisterCapability(CopilotCapability, CopilotAvailability, Enum::"Azure AI Service Type"::"Azure OpenAI", LearnMoreUrl, CallerModuleInfo);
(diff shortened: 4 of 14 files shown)


src/System Application/Test/AI Partner/src/AzureOpenAITestPartner.Codeunit.al
```diff
@@ -0,0 +1,161 @@
...
+        LibraryAssert.AreEqual(CurrentModuleInfo.Publisher(), CopilotSettingsTestLibrary.GetPublisher(), 'Publisher is different from the current module');
+        LibraryAssert.AreEqual(CurrentModuleInfo.Publisher(), 'Partner', 'CurrentModule Publisher is Microsoft');
+
+        Metaprompt := 'metaprompt';
+
+        // [WHEN] GenerateTextCompletion is called
+        AzureOpenAI.GenerateTextCompletion(Metaprompt, Any.AlphanumericText(10), AOAIOperationResponse);
+
+        // [THEN] GenerateTextCompletion returns an error [CAPI with Partner Billed - Not allowed for Partner published capabilities]
+        LibraryAssert.ExpectedError('Usage of AI resources not authorized with chosen billing type. Please reach out to your administrator to resolve this issue.');
```
encimita: We could also check that `AOAIOperationResponse.IsSuccess()` is false. As a stricter check than ExpectedError.

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
```diff
@@ -451,4 +466,15 @@ codeunit 7774 "Copilot Capability Impl"
         CopilotCapability := Enum::"Copilot Capability".FromInteger(Capability);
         IsEnabled := AzureOpenAI.IsEnabled(CopilotCapability, Silent, AppId);
     end;
+
+    procedure IsPublisherMicrosoft(CallerModuleInfo: ModuleInfo): Boolean
+    begin
+        exit(CallerModuleInfo.Publisher() = 'Microsoft');
```
encimita: Instead of doing this, we usually compare the publisher of the CallerModuleInfo to the publisher of the GetCurrentModuleInfo()

encimita: (just in case in the future we decide to rename our publisher, or similar)

src/System Application/App/AI/src/Copilot/CopilotCapability.Codeunit.al
```diff
@@ -43,6 +43,21 @@ codeunit 7773 "Copilot Capability"
         CopilotCapabilityImpl.RegisterCapability(CopilotCapability, CopilotAvailability, LearnMoreUrl, CallerModuleInfo);
     end;
 
+    /// <summary>
```
encimita, line 49: Do we need to obsolete one or more of the functions above? Otherwise the partners will still be able to use them to register capabilities with Undefined billing type.

encimita: This is because we have two (main) ways of assigning values in AL
```
Rec.SomeField := Value;
```
and
```
Rec.Validate(SomeField, Value);
```

The first one does not run validation on the field, meaning the ValuesAllowed property will not be considered from the table CopilotSettings. The second runs all validations.

You can change the assignment to a `Validate` in your code in `CopilotCapabilityImpl`, but if you do, it means installing old extensions in new environments will cause errors, because the partner code will still use the old overloads of the RegisterCapability, which in turn will try to use the Undefined billing type.

---

Otherwise, we can leave the code as is, but add a validation in the new overload of `RegisterCapability()` in this codeunit, and check that the Billing Type is not Undefined (you can just have an `if billing type == undefined then error`), and at the same time obsolete the other overloads of RegisterCapability. This way, old partner code will still work because it will use the old RegisterCapability overloads (that call directly into the Impl codeunit), but new partner code will receive compilation warning for the obsolete functions, and will eventually be forced to call the new RegisterCapability, where we will validate the billing type (as per the first part of this comment paragraph).

But please check with PM about this approach :)

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
rupeshdan: The purpose is to check whether Publisher in Microsoft or Partner. Based on this, enforcements differ.
If we compare "publisher of the CallerModuleInfo to the publisher of the GetCurrentModuleInfo()", then we cannot verify the publisher value.

encimita: Not sure I'm following!

`GetCurrentModuleInfo()` gets the information about the AL app that contains the current file. This file (`CopilotCapabilityImpl`) is in the System Application module, so, if you call `GetCurrentModuleInfo()` here you will get the AL app information of the System Application, and the publisher there is always Microsoft.

(There is a small print here, which is that I am not sure which publisher System Application has in [embed ISV scenarios](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/deployment/embed-app-overview), But this is a tiny minority of our customers and Toolkit resources are anyway limited or blocked completely in this setup.)

src/System Application/App/AI/src/Copilot/CopilotCapability.Codeunit.al
rupeshdan: We can add validation in the new overload of RegisterCapability() and return error if billing type == undefined. 
I believe we want the partner to continue using old code for now and enforce the billing type configuration in future.
@pborring Thoughts?

pborring: If I understand Enrico, we suggests an overload with enforced option, and then deprecating old over a time to allow partners to migrate to new overload. That makes sense to me, we are likely only impacting 20 beta partners to my knowledge

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
```diff
@@ -47,6 +47,7 @@ codeunit 7772 "Azure OpenAI Impl" implements "AI Service Name"
         TelemetryMetapromptRetrievalErr: Label 'Unable to retrieve metaprompt from Azure Key Vault.', Locked = true;
         TelemetryFunctionCallingFailedErr: Label 'Function calling failed for function: %1', Comment = '%1 is the name of the function', Locked = true;
         AzureOpenAiTxt: Label 'Azure OpenAI', Locked = true;
+        BillingTypeAuthorizationErr: Label 'Usage of AI resources not authorized with chosen billing type. Please reach out to your administrator to resolve this issue.';
```
darjoo: Add the billing type and the capability to the error. That will make it easier for the enduser/admin to understand the error. It'll be more easily actionable.

src/System Application/App/AI/src/Copilot/CopilotCapability.Codeunit.al
darjoo: Agreed.
1) Obsolete the old, giving partners time to migrate
2) The old will register with "Undefined"
3) The new will validate the enum to ensure no "Undefined"

src/System Application/App/AI/src/Copilot/CopilotCapability.Codeunit.al
```diff
@@ -57,6 +72,21 @@ codeunit 7773 "Copilot Capability"
...
+    /// Modify an existing capability.
+    /// </summary>
+    /// <param name="CopilotCapability">The capability.</param>
+    /// <param name="CopilotAvailability">The availability.</param>
+    /// <param name="CopilotBillingType">BillingType.</param>
+    /// <param name="LearnMoreUrl">The learn more url.</param>
+    procedure ModifyCapability(CopilotCapability: Enum "Copilot Capability"; CopilotAvailability: Enum "Copilot Availability"; CopilotBillingType: Enum "Copilot Billing Type"; LearnMoreUrl: Text[2048])
```
darjoo, line 88: Obsolete the old function

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
darjoo: ISV Embeds will be affected if we use `GetCurrentModuleInfo()` to compare against the `CallerModuleInfo`. They'll have to use the scenarios based on like they are Microsoft, which will likely break them when they stop being ISV Embeds in the not too distant future.

That said, they can bypass by using `Undefined` temporarily as they're ISV Embeds anyway.

So, I'll agree with Enrico on this and lets change it to use `GetCurrentModuleInfo()` instead of hard coding `'Microsoft'`.

Just FYI, I know there are certain usages in AI Module with the publisher check for `'Microsoft'` but that was specifically for ISV Embeds, and it is bad practice which we will get rid of.

src/System Application/App/AI/src/Copilot/CopilotSettings.Table.al
```diff
@@ -48,6 +48,11 @@ table 7775 "Copilot Settings"
         {
             DataClassification = SystemMetadata;
         }
+        field(8; "Billing Type"; Enum "Copilot Billing Type")
+        {
+            DataClassification = SystemMetadata;
+            ValuesAllowed = 0, 1, 2;
```
darjoo: Does this need to be values or can it be the enum names? I'd prefer if it could be explicit instead of `0, 1, 2` which is difficult to understand.

src/System Application/Test/AI Partner/src/AzureOpenAITestPartner.Codeunit.al
darjoo: Agreed, update to use the reponse

src/System Application/Test/AI Partner/src/AzureOpenAITestPartner.Codeunit.al
```diff
@@ -0,0 +1,161 @@
...
+
+using System.AI;
+using System.Privacy;
+using System.TestLibraries.AI;
+using System.TestLibraries.Environment;
+using System.TestLibraries.Utilities;
+
+codeunit 139025 "Azure OpenAI Test Partner"
+{
+    Subtype = Test;
```
darjoo: There is the HttpClient mock, please block all outbound requests and ensure they are handled.

`TestHttpRequestPolicy = BlockOutboundRequests`

https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-httpclient-mock-outbound-calls

darjoo: There should be no outbound requests that need to actually go to any server.
For AI resources, it is handled by platform and this will not block it afaik.

src/System Application/Test/AI Partner/src/AzureOpenAITestPartner.Codeunit.al
rupeshdan: Done

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
rupeshdan: Updated code as suggested

src/System Application/App/AI/src/Copilot/CopilotCapability.Codeunit.al
rupeshdan: Obsoleted old API as suggested

src/System Application/App/AI/src/Copilot/CopilotCapability.Codeunit.al
rupeshdan: Done

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
rupeshdan: Updated the error message

src/System Application/App/AI/src/Copilot/CopilotSettings.Table.al
rupeshdan: Replaced values with names

src/System Application/Test/AI Partner/src/AzureOpenAITestPartner.Codeunit.al
rupeshdan: Updated changes.

src/System Application/App/Entity Text/src/EntityTextAIUpgrade.Codeunit.al
```diff
@@ -29,7 +29,7 @@ codeunit 2015 "Entity Text AI Upgrade"
         if not UpgradeTag.HasUpgradeTag(GetRegisterMarketingTextCapabilityTag()) then begin
```
encimita, line 29: Ideally, we should also have some logic on the lines of (pseudocode)

```
if not HasUpgradeTag(GetModifyBillingTypeTag()) then
    if IsSaas() then
        if CopilotCapability.IsRegistered(EntityText) then
            CopilotCapability.ModifyCapability(..., "Not Billed", ...);
```

But I don't think this belongs in this PR.
Maybe you can create a separate bug to keep track of this?

src/System Application/App/AI/src/Copilot/CopilotCapabilityInstall.Codeunit.al
```diff
@@ -26,20 +26,20 @@ codeunit 7760 "Copilot Capability Install"
 
     internal procedure RegisterCapabilities()
     begin
-        RegisterSaaSCapability(Enum::"Copilot Capability"::"Analyze List", Enum::"Copilot Availability"::Preview, AnalyzeListLearnMoreLbl);
-        RegisterSaaSCapability(Enum::"Copilot Capability"::Autofill, Enum::"Copilot Availability"::Preview, AutofillLearnMoreLbl);
-        RegisterSaaSCapability(Enum::"Copilot Capability"::Chat, Enum::"Copilot Availability"::Preview, ChatLearnMoreLbl);
-        RegisterSaaSCapability(Enum::"Copilot Capability"::Summarize, Enum::"Copilot Availability"::Preview, SummarizeLearnMoreLbl);
+        RegisterSaaSCapability(Enum::"Copilot Capability"::"Analyze List", Enum::"Copilot Availability"::Preview, Enum::"Copilot Billing Type"::"Not Billed", AnalyzeListLearnMoreLbl);
+        RegisterSaaSCapability(Enum::"Copilot Capability"::Autofill, Enum::"Copilot Availability"::Preview, Enum::"Copilot Billing Type"::"Not Billed", AutofillLearnMoreLbl);
+        RegisterSaaSCapability(Enum::"Copilot Capability"::Chat, Enum::"Copilot Availability"::Preview, Enum::"Copilot Billing Type"::"Not Billed", ChatLearnMoreLbl);
```
encimita, line 31: Same comment about `ModifyCapability`. Let's open a separate bug "Add upgrade code to mark platform capability as non billed".

src/System Application/App/AI/src/Azure AI Document Intelligence/AzureDIImpl.Codeunit.al
```diff
@@ -36,7 +36,7 @@ codeunit 7779 "Azure DI Impl." implements "AI Service Name"
 
     procedure RegisterCopilotCapability(CopilotCapability: Enum "Copilot Capability"; CopilotAvailability: Enum "Copilot Availability"; LearnMoreUrl: Text[2048]; CallerModuleInfo: ModuleInfo)
     begin
-        CopilotCapabilityImpl.RegisterCapability(CopilotCapability, CopilotAvailability, Enum::"Azure AI Service Type"::"Azure Document Intelligence", LearnMoreUrl, CallerModuleInfo);
+        CopilotCapabilityImpl.RegisterCapability(CopilotCapability, CopilotAvailability, Enum::"Copilot Billing Type"::"Undefined", Enum::"Azure AI Service Type"::"Azure Document Intelligence", LearnMoreUrl, CallerModuleInfo);
```
encimita, line 39: We should also make sure this has an overload that allows specifying the billing type.

encimita (approved):

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
```diff
@@ -47,13 +47,16 @@ codeunit 7772 "Azure OpenAI Impl" implements "AI Service Name"
...
-    procedure IsEnabled(Capability: Enum "Copilot Capability"; CallerModuleInfo: ModuleInfo): Boolean
+    procedure IsEnabled(Capability: Enum "Copilot Capability";
+        CallerModuleInfo: ModuleInfo): Boolean
     begin
         exit(CopilotCapabilityImpl.IsCapabilityEnabled(Capability, CallerModuleInfo));
     end;
 
-    procedure IsEnabled(Capability: Enum "Copilot Capability"; Silent: Boolean; CallerModuleInfo: ModuleInfo): Boolean
+    procedure IsEnabled(Capability: Enum "Copilot Capability"; Silent: Boolean;
+                                        CallerModuleInfo: ModuleInfo): Boolean
```
darjoo: Revert these parameters to oneline please.

darjoo: Same for all the other procedures that are affected

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
```diff
@@ -459,6 +471,12 @@ codeunit 7772 "Azure OpenAI Impl" implements "AI Service Name"
         EmptySecretText: SecretText;
     begin
         ClearLastError();
+
+        if not IsBillingTypeAuthorized(AOAIAuthorization, CallerModuleInfo) then begin
+            Error := StrSubstNo(BillingTypeAuthorizationErr, CopilotCapabilityImpl.GetCapabilityName(), CopilotCapabilityImpl.GetCopilotBillingType());
+            Error(Error);
```
darjoo, line 466: No reason for setting into the variable Error first. 
Just use `Error(StrSub....);`

rupeshdan:
> @rupeshdan please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > ```
> 
> Contributor License Agreement

darjoo (approved):

aholstrup1 (approved):

PR #3813, opened 2025-06-03
TestRunner will be powered by CodeunitMetadata instead

haoranpb:
#### Summary 

`AllObjWithCaption` was used to power TestRunner probably because we don't have `Codeunit Metadata` until recently.

Switching to `Codeunit Metadata` will:
1. Speed things up (naturally, with less data in the table)
2. Laying down the foundation to uptake `TestType` property on codeunits

#### Breaking Changes

Checking Telemetry for Symbol Usage:
1. The 2 pages have no usage, using pragma to suppress warnings
2. There is usage for `procedure GetTestMethods`. Properly obsolete it instead. 

#### Work Item(s) 

### DIFF
--- a/src/Tools/Test Framework/Test Runner/src/SelectTestRunner.Page.al
+++ b/src/Tools/Test Framework/Test Runner/src/SelectTestRunner.Page.al
@@ -11,9 +11,8 @@ page 130454 "Select TestRunner"
 {
     Editable = false;
     PageType = List;
-    SourceTable = AllObjWithCaption;
-    SourceTableView = where("Object Type" = const(Codeunit),
-                            "Object Subtype" = const('TestRunner'));
+    SourceTable = "CodeUnit Metadata";
+    SourceTableView = where(Subtype = const(TestRunner));
 
     layout
     {
@@ -22,12 +21,12 @@ page 130454 "Select TestRunner"
             repeater(Control1)
             {
                 ShowCaption = false;
-                field("Object ID"; Rec."Object ID")
+                field("Object ID"; Rec.ID)
                 {
                     ApplicationArea = All;
                     ToolTip = 'Specifies the Object ID';
                 }
-                field("Object Name"; Rec."Object Name")
+                field("Object Name"; Rec.Name)
                 {
                     ApplicationArea = All;
                     ToolTip = 'Specifies the Object Name';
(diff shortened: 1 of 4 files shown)


src/Tools/Test Framework/Test Runner/src/SelectTestRunner.Page.al
```diff
@@ -11,9 +11,10 @@ page 130454 "Select TestRunner"
 {
     Editable = false;
     PageType = List;
-    SourceTable = AllObjWithCaption;
-    SourceTableView = where("Object Type" = const(Codeunit),
-                            "Object Subtype" = const('TestRunner'));
+#pragma warning disable AS0035
+    SourceTable = "CodeUnit Metadata";
```
stkillen, line 15: Breaking change?

src/Tools/Test Framework/Test Runner/src/SelectTests.Page.al
```diff
@@ -11,9 +11,10 @@ page 130453 "Select Tests"
 {
     Editable = false;
     PageType = List;
-    SourceTable = AllObjWithCaption;
-    SourceTableView = where("Object Type" = const(Codeunit),
-                            "Object Subtype" = const('Test'));
+#pragma warning disable AS0035
+    SourceTable = "CodeUnit Metadata";
```
stkillen, line 15: Breaking change?

src/Tools/Test Framework/Test Runner/src/SelectTestRunner.Page.al
nikolakukrika: Can we see in SaaS for symbol usage if there are any page extensions? If not lets change the tables for both pages.

haoranpb: Considered doing it in a non-breaking way, but it looks rather ugly... (two extra pages like you mentioned).

Double checked Telemetry, there are no SaaS usage for those two pages. Will document this decision in the work item.

src/Tools/Test Framework/Test Runner/src/SelectTests.Page.al
haoranpb: Double checked Telemetry, there are no SaaS usage for those two pages

PR #3821, opened 2025-06-04
Introducing TestType to TestRunner

haoranpb:
#### Summary 

Introducing `TestType` to TestRunner. We can now select which `TestType` to run from CLI.

Related to https://github.com/microsoft/navcontainerhelper/pull/3959

There will be a follow-up PR to add this capability in both `ALTestRunner.psm1` and [navcontainerhelper  Get-Tests](https://github.com/microsoft/navcontainerhelper/blob/df270f9e3ed96260221f918f2deb0fd7af941f07/AppHandling/PsTestFunctions.ps1#L328C10-L328C13)

#### Work Item(s) 

### DIFF
--- a/src/Tools/Test Framework/Test Runner/src/CommandLineTestTool.Page.al
+++ b/src/Tools/Test Framework/Test Runner/src/CommandLineTestTool.Page.al
@@ -101,6 +101,8 @@ page 130455 "Command Line Test Tool"
                 ApplicationArea = All;
                 Caption = 'Test Type';
                 ToolTip = 'Specifies the Test Type';
+                BlankZero = true;
+                MinValue = 1;
 
                 trigger OnValidate()
                 var
@@ -484,7 +486,7 @@ page 130455 "Command Line Test Tool"
         FullErrorMessage: Text;
         StackTrace: Text;
         ExtensionId: Text;
-        TestType: Text;
+        TestType: Integer;
         RemoveTestMethod: Text;
         TestResultsJSONText: Text;
         CCResultsCSVText: Text;
--- a/src/Tools/Test Framework/Test Runner/src/TestSuiteMgt.Codeunit.al
+++ b/src/Tools/Test Framework/Test Runner/src/TestSuiteMgt.Codeunit.al
@@ -26,7 +26,6 @@ codeunit 130456 "Test Suite Mgt."
         DefaultTestSuiteNameTxt: Label 'DEFAULT', Locked = true;
         DialogUpdatingTestsMsg: Label 'Updating Tests: \#1#\#2#', Comment = '1 = Object being processed, 2 = Progress', Locked = true;
         ErrorMessageWithCallStackErr: Label 'Error Message: %1 - Error Call Stack: ', Locked = true;
-        InvalidTestTypeErr: Label 'Invalid test type: %1', Locked = true;
 
     procedure IsTestMethodLine(FunctionName: Text[128]): Boolean
     begin
@@ -310,7 +309,7 @@ codeunit 130456 "Test Suite Mgt."
         GetTestMethods(ALTestSuite, CodeunitMetadata);
     end;
 
-    internal procedure SelectTestMethodsByExtensionAndTestType(var ALTestSuite: Record "AL Test Suite"; ExtensionID: Text; TestType: Text)
+    internal procedure SelectTestMethodsByExtensionAndTestType(var ALTestSuite: Record "AL Test Suite"; ExtensionID: Text; TestType: Integer)
     var
         CodeunitMetadata: Record "CodeUnit Metadata";
         AppExtensionId: Guid;
@@ -322,17 +321,8 @@ codeunit 130456 "Test Suite Mgt."
         end;
 
         CodeunitMetadata.SetRange(SubType, CodeunitMetadata.SubType::Test);
-
-        case TestType of
-            'UnitTest':
-                CodeunitMetadata.SetRange(TestType, CodeunitMetadata.TestType::UnitTest);
-            'IntegrationTest':
-                CodeunitMetadata.SetRange(TestType, CodeunitMetadata.TestType::IntegrationTest);
-            'Uncategorized':
-                CodeunitMetadata.SetRange(TestType, CodeunitMetadata.TestType::Uncategorized);
-            else
-                Error(InvalidTestTypeErr, TestType);
-        end;
+        // inexplicit conversion from Integer to Option
+        CodeunitMetadata.SetRange(TestType, TestType);
 
         GetTestMethods(ALTestSuite, CodeunitMetadata);
     end;


src/Tools/Test Framework/Test Runner/src/CommandLineTestTool.Page.al
```diff
@@ -96,6 +96,25 @@ page 130455 "Command Line Test Tool"
                     if Rec.FindFirst() then;
                 end;
             }
+            field(TestType; TestType)
```
nikolakukrika, line 99: This must be an integer. We must avoid being affected by the translations. See how the other types were handled.

src/Tools/Test Framework/Test Runner/src/TestMethodLine.Table.al
```diff
@@ -106,6 +106,18 @@ table 130450 "Test Method Line"
         {
             Editable = false;
         }
+        field(13; "Test Type"; Option)
```
nikolakukrika: Should we introduce an enum to the platform? Then we can reuse it.

src/Tools/Test Framework/Test Runner/src/TestMethodLine.Table.al
```diff
@@ -106,6 +106,18 @@ table 130450 "Test Method Line"
...
+        {
+            Editable = false;
+            OptionMembers = None,UnitTest,IntegrationTest,Uncategorized;
+            Description = 'Must be align with TestType field in CodeunitMetadata table.';
+
+            trigger OnValidate()
+            begin
+                if "Test Type" = "Test Type"::None then
+                    Error('Test Type cannot be None. None is reserved for non-test codeunits.');
```
nikolakukrika: No inline errors, it needs to be a label.

src/Tools/Test Framework/Test Runner/src/TestMethodLine.Table.al
nikolakukrika: Do we need this field? Can't we do a lookup and show a value, we do not need to set it for every line. 
Also showing it on the codeunit is sufficient, test methods could skip the value. 
The value is read only so we do not need to store it in the database, unless there is something that Im missing.

src/Tools/Test Framework/Test Runner/src/TestSuiteMgt.Codeunit.al
```diff
@@ -309,6 +309,34 @@ codeunit 130456 "Test Suite Mgt."
         GetTestMethods(ALTestSuite, CodeunitMetadata);
     end;
 
+    /// <summary>
+    /// Converts a text representation of a TestType to an option.
+    /// </summary>
+    /// <param name="TestType">Text representation of TestType, specified from CLI</param>
+    /// <returns>One of the options from TestType field in CodeunitMetadata table, except for None</returns>
+    internal procedure ConvertTestTypeToOption(TestType: Text) result: Option None,UnitTest,IntegrationTest,Uncategorized;
```
nikolakukrika: See the comment on integer. If we have this in the platform on the Codeunit metadata, we can easly convert instead of hardcoding.

src/Tools/Test Framework/Test Runner/src/CommandLineTestTool.Page.al
```diff
@@ -96,6 +96,25 @@ page 130455 "Command Line Test Tool"
...
+            {
+                ApplicationArea = All;
```
nikolakukrika, line 101: We need the same field on the UI page, so we can set the test type on the header. It can be additional importance.

src/Tools/Test Framework/Test Runner/src/TestSuiteMgt.Codeunit.al
```diff
@@ -585,6 +613,7 @@ codeunit 130456 "Test Suite Mgt."
         TestMethodLine."Test Codeunit" := CodeunitMetadata.ID;
         TestMethodLine.Validate("Line Type", TestMethodLine."Line Type"::Codeunit);
         TestMethodLine.Name := CodeunitMetadata.Name;
+        TestMethodLine.Validate("Test Type", CodeunitMetadata.TestType);
```
nikolakukrika: This does not seem correct. The type should be on the suite. 
We should not call AddTestMethod if the type is not matching the type on the suite.

src/Tools/Test Framework/Test Runner/src/CommandLineTestTool.Page.al
haoranpb: Decided to make it a `Text` field because it would be ~more readable in the build scripts~ (we could do a Text <=> Integer mapping in build script, but then we will have to maintain the mapping in 2 places).

Not sure if this will be impacted by translation tbh, since the value come from build script, and we convert it to Integer with `ConvertTestTypeToOption`

```
function Set-TestType
(
    [ValidateSet('UnitTest','IntegrationTest','Uncategorized')]
    [string] $TestType,
...
    $testTypeControl = $ClientContext.GetControlByName($Form, "TestType")
}
```

src/Tools/Test Framework/Test Runner/src/TestMethodLine.Table.al
haoranpb: Great point! Just like `ExtensionId`, we don't need it in DB. Fixing now.

src/Tools/Test Framework/Test Runner/src/TestMethodLine.Table.al
haoranpb: Good catch, moved to label

src/Tools/Test Framework/Test Runner/src/TestMethodLine.Table.al
haoranpb: Good question. Decided to go with `Option` because it will not be used anywhere else, and we will have the flexibility to easily change it.

Especially with the latest commit, this is only converted inside one procedure, it is probably ok.

What do you think?

src/Tools/Test Framework/Test Runner/src/TestSuiteMgt.Codeunit.al
haoranpb: Replied in the comment regarding Enum, let's use that thread for this discussion.

src/Tools/Test Framework/Test Runner/src/CommandLineTestTool.Page.al
haoranpb: Not sure I understand this part, can you elaborate? You mean `page 130451 "AL Test Tool"`?

It will not be used outside CLI context, but agree that it will probably be useful to see the `TestType` in UI.

src/Tools/Test Framework/Test Runner/src/TestSuiteMgt.Codeunit.al
haoranpb: With the latest change, this is outdated.

src/Tools/Test Framework/Test Runner/src/TestSuiteMgt.Codeunit.al
```diff
@@ -309,6 +310,33 @@ codeunit 130456 "Test Suite Mgt."
         GetTestMethods(ALTestSuite, CodeunitMetadata);
     end;
 
+    internal procedure SelectTestMethodsByExtensionAndTestType(var ALTestSuite: Record "AL Test Suite"; ExtensionID: Text; TestType: Text)
```
stkillen: Why is this not a guid? Then you don't need to evaluate it to a guid either. You can still check if it's a null guid

stkillen (commented):
Should we also expose the test type in the UI page? E.g. filtering to test type could be useful

src/Tools/Test Framework/Test Runner/src/TestSuiteMgt.Codeunit.al
haoranpb: This is because the value is set from [function Set-ExtensionId](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_git/NAV?path=/Eng/Core/Tools/ALTestRunner/Internal/ALTestRunnerInternal.psm1&version=GBmaster&line=428&lineEnd=428&lineStartColumn=10&lineEndColumn=25&lineStyle=plain&_a=contents) from build script, and looks like we only support `[string]` type in [SaveValue](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_git/NAV?path=/Eng/Core/Helpers/ClientContext.ps1&version=GBmaster&_a=contents&line=339&lineStyle=plain&lineEnd=339&lineStartColumn=5&lineEndColumn=14).

Mostly just followed existing `procedure SelectTestMethodsByExtension`

src/Tools/Test Framework/Test Runner/src/CommandLineTestTool.Page.al
haoranpb: Unifying the comment from Stig to this thread https://github.com/microsoft/BCApps/pull/3821#pullrequestreview-2896239008

Agree it could be helpful in UI page, but just to make sure we are all on the same page, we are all talking about `page 130451 "AL Test Tool"`?

haoranpb: Played around for a bit, this is rather easy to achieve, however we might want to wait for this change because all tests are `UnitTest` atm, could be a better idea to add this in a following PR?

image

haoranpb: After some thought, it's probably best that we wait before making it available in UI. All tests are `UnitTest` for the time being, not sure how much value to see this in UI.

Will make follow up PRs after we start Phase Two. Documenting this decision.

stkillen (approved):

src/Tools/Test Framework/Test Runner/src/CommandLineTestTool.Page.al
haoranpb: Discussed offline: let's follow the same patterns here as other options.

nikolakukrika (dismissed):

src/Tools/Test Framework/Test Runner/src/CommandLineTestTool.Page.al
haoranpb: Converted to Integer tpy

PR #3826, opened 2025-06-04
Uptake TestType property in the pipeline

haoranpb:
#### Summary

Successor of:
1. https://github.com/microsoft/BCApps/pull/3821
2. https://github.com/microsoft/navcontainerhelper/pull/3959

Only run `UnitTest` in the build of BCApps for both System Application and Business Foundation. It will be possible to define `IntegrationTest` in BCApps with [AB#581154](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/581154), but they will be run in NAV. (We only have `UnitTests` for now)

Tried to take a look at disabled tests, they are not `IntegrationTest`, will not fix them as part of the effort. Re-enabling some easy fixes.

#### Work Item(s) 


haoranpb:
Tested in https://github.com/microsoft/BCApps/pull/3826/commits/2d746c17c0931cb542d45c97bc152c7c2f823388, all tests are indeed running on empty companies (i.e. without any demo data).

PR #3889, opened 2025-06-10
Create Activity Log 1P module

Groenbech96:
#### Summary 

Add module to allow 1p apps to use Activity Log platform feature.

#### Work Item(s) 
Fixes [AB#578938](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/578938)

### DIFF
--- a/src/System Application/App/ActivityLog/src/ActivityLogBuilderImpl.Codeunit.al
+++ b/src/System Application/App/ActivityLog/src/ActivityLogBuilderImpl.Codeunit.al
@@ -15,6 +15,7 @@ codeunit 3112 "Activity Log Builder Impl."
         LogEntry: DotNet ActivityLogEntry;
         AttributeType: DotNet ActivityLogAttribute;
         GlobalFieldNo: Integer;
+        BuilderNotInitializedErr: Label 'Activity Log Builder not initialized.';
 
     procedure Init(TableNo: Integer; FieldNo: Integer; RecSystemId: Guid): Codeunit "Activity Log Builder Impl."
     begin
@@ -53,6 +54,8 @@ codeunit 3112 "Activity Log Builder Impl."
 
     procedure Log()
     begin
+        if GlobalFieldNo = 0 then
+            Error(BuilderNotInitializedErr);
         LogEntry.Log();
     end;
 


src/System Application/App/ActivityLog/app.json
```diff
@@ -0,0 +1,31 @@
+{
+    "id": "d3433b68-4901-445f-9547-fdfeca57575b",
```
nikolakukrika, line 2: I believe that you need to add this app to the BC apps build process. We build and test all apps in separation.

nikolakukrika, line 2: The same app needs to be added to the app.json in the system app folder

src/System Application/App/ActivityLog/src/ActivityLogBuilderImpl.Codeunit.al
```diff
@@ -0,0 +1,59 @@
...
+
+    procedure SetReferenceTitle(ReferenceTitle: Text): Codeunit "Activity Log Builder Impl."
+    begin
+        LogEntry.AddFieldAttribute(this.GlobalFieldNo, AttributeType::ReferenceTitle, ReferenceTitle);
+        exit(this);
+    end;
+
+    procedure Log()
+    begin
+        LogEntry.Log();
```
nikolakukrika, line 59: Should we add checks if it was intialized or not? E.g. check the GlobalFieldNo different than zero?

src/System Application/App/ActivityLog/src/ActivityLogBuilder.Codeunit.al
```diff
@@ -0,0 +1,80 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+namespace Microsoft.eServices.EDocument.Processing.Import;
+
+/// <summary>
+/// Codeunit to build and log activity entries for a specific table and field.
+/// Currently first party applications only to allow for future API changes. 
+/// </summary>
+codeunit 3111 "Activity Log Builder"
```
nikolakukrika, line 11: We have activity log in the product, so lets avoid name clashes.
Recommended name Entry Log Builder

nikolakukrika: These entries are not connected to the activity, but more to the record so activity is not the best name.

src/System Application/App/ActivityLog/app.json
Groenbech96: Not needed :)

src/System Application/App/ActivityLog/src/ActivityLogBuilder.Codeunit.al
Groenbech96: Since this will be removed once the naitive AL support is build, lets not dwell over it. For now i will keep it the same as platform names for simplicity.

jonasblunck (approved):

nikolakukrika (approved):

PR #3935, opened 2025-06-20
Respect feedback setting for Deactivation survey

haoranpb:
#### Summary 

Respect feedback setting for Deactivation survey.

Extra the logic into a procedure.

Convert Copilot Setting global variable to local, because it's needed to pass it in a parameter.

#### Work Item(s) 
Fixes AB#579115

### DIFF
--- a/src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
+++ b/src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
@@ -22,7 +22,6 @@ codeunit 7774 "Copilot Capability Impl"
     Permissions = tabledata "Copilot Settings" = rimd;
 
     var
-        CopilotSettings: Record "Copilot Settings";
         FeatureTelemetry: Codeunit "Feature Telemetry";
         Telemetry: Codeunit Telemetry;
         CopilotCategoryLbl: Label 'Copilot', Locked = true;
@@ -44,6 +43,7 @@ codeunit 7774 "Copilot Capability Impl"
         TelemetryUnregisteredCopilotCapabilityLbl: Label 'Copilot capability unregistered.', Locked = true;
         TelemetryActivatedCopilotCapabilityLbl: Label 'Copilot capability activated.', Locked = true;
         TelemetryDeactivatedCopilotCapabilityLbl: Label 'Copilot capability deactivated.', Locked = true;
+        CopilotFeatureDeactivatedLbl: Label 'The copilot/AI capability %1, App Id %2 has been deactivated by UserSecurityId %3.', Locked = true;
 
     procedure RegisterCapability(CopilotCapability: Enum "Copilot Capability"; LearnMoreUrl: Text[2048]; CallerModuleInfo: ModuleInfo)
     begin
@@ -57,6 +57,7 @@ codeunit 7774 "Copilot Capability Impl"
 
     procedure RegisterCapability(CopilotCapability: Enum "Copilot Capability"; CopilotAvailability: Enum "Copilot Availability"; AzureAIServiceType: Enum "Azure AI Service Type"; LearnMoreUrl: Text[2048]; CallerModuleInfo: ModuleInfo)
     var
+        CopilotSettings: Record "Copilot Settings";
         CustomDimensions: Dictionary of [Text, Text];
     begin
         if IsCapabilityRegistered(CopilotCapability, CallerModuleInfo) then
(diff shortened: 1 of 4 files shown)


src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
```diff
@@ -442,6 +443,28 @@ codeunit 7774 "Copilot Capability Impl"
...
+                CopilotSettingsLocal.Modify(true);
+
+                SendDeactivateTelemetry(CopilotSettingsLocal.Capability, CopilotSettingsLocal."App Id", CopilotDeactivate.GetReason());
+                Session.LogAuditMessage(StrSubstNo(CopilotFeatureDeactivatedLbl, CopilotSettingsLocal.Capability, CopilotSettingsLocal."App Id", UserSecurityId()), SecurityOperationResult::Success, AuditCategory::ApplicationManagement, 4, 0);
+            end;
+        end else begin
+            CopilotSettingsLocal.Status := CopilotSettingsLocal.Status::Inactive;
+            CopilotSettingsLocal.Modify(true);
+
+            Session.LogAuditMessage(StrSubstNo(CopilotFeatureDeactivatedLbl, CopilotSettingsLocal.Capability, CopilotSettingsLocal."App Id", UserSecurityId()), SecurityOperationResult::Success, AuditCategory::ApplicationManagement, 4, 0);
```
darjoo: Line 458 and this are the same. Move these to the end, so between 456/466 outside of the if/else scope

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
```diff
@@ -442,6 +443,28 @@ codeunit 7774 "Copilot Capability Impl"
...
+        CopilotDeactivate: Page "Copilot Deactivate Capability";
+        ALCopilotFunctions: DotNet ALCopilotFunctions;
+    begin
+        if ALCopilotFunctions.IsCopilotFeedbackEnabled() then begin
+            CopilotDeactivate.SetCaption(Format(CopilotSettingsLocal.Capability));
+            if CopilotDeactivate.RunModal() = Action::OK then begin
+                CopilotSettingsLocal.Status := CopilotSettingsLocal.Status::Inactive;
+                CopilotSettingsLocal.Modify(true);
+
+                SendDeactivateTelemetry(CopilotSettingsLocal.Capability, CopilotSettingsLocal."App Id", CopilotDeactivate.GetReason());
```
darjoo: Can we just send this irregardless of the feedback?
The reason being empty is fine, but we still want the telemetry that they were deactivated

haoranpb: Good catch. When feedback is disabled, we send telemetry but without reason

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
haoranpb: This is by design. The page could be closed without actually decativation

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
```diff
@@ -442,6 +461,29 @@ codeunit 7774 "Copilot Capability Impl"
             GuidedExperience.ResetAssistedSetup(ObjectType::Page, Page::"Copilot AI Capabilities");
     end;
 
+    procedure DeactivateCapability(var CopilotSettingsLocal: Record "Copilot Settings")
+    var
+        CopilotDeactivate: Page "Copilot Deactivate Capability";
+        ALCopilotFunctions: DotNet ALCopilotFunctions;
+    begin
+        if ALCopilotFunctions.IsCopilotFeedbackEnabled() then begin
```
msft-sam: Should this go in a try function?

If it fails, should we assume feedback is enabled?

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
```diff
@@ -442,6 +461,29 @@ codeunit 7774 "Copilot Capability Impl"
...
+            CopilotDeactivate.SetCaption(Format(CopilotSettingsLocal.Capability));
+            if CopilotDeactivate.RunModal() = Action::OK then begin
```
msft-sam: Since we do the same thing consider just doing `RunModal() <> Action::OK` and then exiting, that way we avoid the duplication

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
haoranpb: I'd vote against putting it in a try function here, if the procedure fails, we probably prefer to get a bug about it and fix the issue.

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
haoranpb: Good point, refactorer.

PR #3995, opened 2025-06-27
Add API to feature management

WaelAbuSeada:
Add API to feature management
Fixes [AB#568708](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/568708)


src/System Application/App/Feature Key/src/APIFeatureManagement.Page.al
```diff
@@ -0,0 +1,129 @@
...
+                {
+                    Caption = 'Description In English';
+                }
+            }
+        }
+    }
+
+    [ServiceEnabled]
+    [Scope('Cloud')]
+    procedure EnableFeature(UpdateInBackground: Boolean; StartDateTime: DateTime)
```
onbuyuka: NST won't publish this bound action. If you want this to work as a bound action from API, then you need to follow this syntax:

```
[ServiceEnabled]
[Scope('Cloud')]
procedure EnableFeature(var ActionContext: WebServiceActionContext)
```

Only procedures that follow this syntax will be published as api/microsoft/featureManagement/v1.0/companies(\<companyId>)/features(\<featureId>)/Microsoft.NAV.enableFeature
And in ActionContext you can set these 

```
ActionContext.SetObjectType
ActionContext.SetObjectId
ActionContext.AddEntityKey
ActionContext.SetResultCode
```

Whatever you set in the ActionContext will be returned in the header when bound action is invoked.

onbuyuka: If you need any input from the user for the bound action then you need to add those as fields in the API. That means they need to first make a PATCH request and then invoke the bound action

src/System Application/App/Feature Key/src/APIFeatureManagement.Page.al
```diff
@@ -0,0 +1,129 @@
...
+
+namespace System.Environment.Configuration;
+
+/// <summary>
+/// API Page that enables a user to manage available features.
+/// </summary>
+page 2613 "API - Feature Management"
+{
+    PageType = API;
+    Caption = 'feature', Locked = true;
```
onbuyuka: Captions are not used for API pages

src/System Application/App/Feature Key/src/APIFeatureManagement.Page.al
```diff
@@ -0,0 +1,129 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Environment.Configuration;
+
+/// <summary>
+/// API Page that enables a user to manage available features.
+/// </summary>
+page 2613 "API - Feature Management"
```
onbuyuka: Should we consider moving this to APIV2? We have automation APIs already there (the ones start with APIV2Aut): https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_git/NAV?path=/App/Apps/W1/APIV2/app/src/pages/APIV2AutCompanies.Page.al

This looks like it would be a good fit there. Up to discussion

src/System Application/App/Feature Key/src/APIFeatureManagement.Page.al
```diff
@@ -0,0 +1,129 @@
...
+    ModifyAllowed = false;
+    DeleteAllowed = false;
+
+    layout
+    {
+        area(Content)
+        {
+            repeater(GroupName)
+            {
+                field(id; Rec.ID)
```
onbuyuka: SystemId would be better here, ID is Text[50], then you need to use quotes in the request. Partners are used to GUIDs

src/System Application/App/Feature Key/src/APIFeatureManagement.Page.al
WaelAbuSeada: I will let Kennie decide on this one

src/System Application/App/Feature Key/src/APIFeatureManagement.Page.al
WaelAbuSeada: This seems like it works well now

src/System Application/App/Feature Key/src/APIFeatureManagement.Page.al
WaelAbuSeada: Sure thing

WaelAbuSeada: After investigating, we can't really use system ID here, so will leave the ID as the key

src/System Application/App/Feature Key/src/APIFeatureManagement.Page.al
```diff
@@ -0,0 +1,131 @@
...
+        FeatureManagementFacade.OnAfterFeatureEnableConfirmed(Rec);
+
+        // Update data if needed
+        FeatureManagementFacade.GetFeatureDataUpdateStatus(Rec, FeatureDataUpdateStatus);
+
+        if FeatureDataUpdateStatus."Data Update Required" then begin
+            FeatureDataUpdateStatus."Background Task" := UpdateInBackground;
+            FeatureDataUpdateStatus."Start Date/Time" := CurrentDateTime();
+            FeatureDataUpdateStatus.Confirmed := true;
+            FeatureDataUpdateStatus.Modify(true);
```
darjoo: Missing `m` permission at top of file for `"Feature Data Update Status"`. From the permission sets, only indirect permissions are given.

WaelAbuSeada: good catch

src/System Application/App/Feature Key/src/APIFeatureManagement.Page.al
WaelAbuSeada: Kennie confirmed to move it to automation API

onbuyuka (approved):

PR #4034, opened 2025-07-03
[Email] Disable email retry

gggdttt:
#### Work Item(s) 
Fixes AB#582424


darjoo (approved):
We're disabling this temporarily and will reenable in >=26.4

haoranpb (approved):

PR #4040, opened 2025-07-04
[Draft] Uptake TestType for first-party apps

haoranpb:
#### Summary 

I can see there are PRs start to add extensions to the new Apps project. Uptake the `TestType` property for this project.

For UnitTests, tests should be able to run on empty dataabase.

For IntegrationTests, tests should be able to run with Contoso Demo Data (Setup Data).

If tests can't run with those two configurations, please reach out.

#### Work Item(s) 


build/scripts/ImportTestDataInBcContainer.ps1
```diff
@@ -40,4 +40,4 @@ foreach ($app in (Get-BcContainerAppInfo -containerName $ContainerName -tenantSp
 }
 
 # Generate demo data in the container
-Invoke-ContosoDemoTool -ContainerName $parameters.ContainerName
\ No newline at end of file
+Invoke-ContosoDemoTool -ContainerName $parameters.ContainerName -SetupData $parameters.SetupData
```
aholstrup1: We don't seem to be using this switch internally - Should we be? 
https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_git/NAV?path=/Eng/Core/Scripts/BuildW1Database.ps1&version=GBmaster&_a=contents

We should try to match the way we run tests on BCApps with how we run them internally.

aholstrup1: Do you mind changing this parameter to a boolean as well? :) 
https://github.com/microsoft/BCApps/blob/3783db2c67d3dfa4f8c68727f9dbd8ebeb9529d1/build/scripts/ImportTestDataInBcContainer.ps1#L11

Alternatively, you can remove the default value.

haoranpb: Agreed! We actually working on this, expected to be ready for PR review in 1-2 weeks. See:
* https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_git/NAV/pullrequest/218675
* https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_git/NAV/pullrequest/218408

build/projects/Apps (W1)/.AL-Go/ImportTestDataInBcContainer.ps1
```diff
@@ -1,6 +0,0 @@
-Param(
```
aholstrup1: I think it would be better if we could keep this script (and therefore the demo data generation separate from the test execution). 

RunTestsInBcContainer.ps1 is run for every test app. ImportTestDataInBCContainer is only called once.

aholstrup1 (changes requested):

build/projects/Apps (W1)/.AL-Go/ImportTestDataInBcContainer.ps1
haoranpb: Agree that we should keep demo data generation separate from the test execution.

---

I assume we have a sequence of execution like this:

```
ImportTestDataInBCContainer <== RunTestsInBcContainer
```

Thinking on top of my head, we probably need a bigger re-design if we'd like to run UnitTests on empty DB, but IntegrationTests with Demo Data?

Or maybe only run UnitTests in BCApps, and we can run IntegrationTests in NAV?

haoranpb: On a different note: when we move more apps to BCApps, you might find tests that can't even run on Contoso Demo Data.

Those tests will have to run in NAV, so maybe it's ok to run IntegrationTests in NAV

PR #4042, opened 2025-07-04
[Languages] Show all Windows Languages if no Allowed Languages are configured

darjoo:
#### Summary 
The filters on "Localization Exist" and "Globally Enabled" limited the list of languages that were previously allowed to be selected from. This removed numerous languages resulting in incidents.

Removal of the filter resets this back to how it worked before 26.0 and only if you set up Allowed Languages, then the list of languages will be filtered.

#### Work Item(s) 
Fixes [AB#591234](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/591234)

### DIFF
--- a/src/System Application/App/Language/src/AllowedLanguage.Table.al
+++ b/src/System Application/App/Language/src/AllowedLanguage.Table.al
@@ -23,7 +23,7 @@ table 3563 "Allowed Language"
             Caption = 'Language Id';
             NotBlank = true;
             BlankZero = true;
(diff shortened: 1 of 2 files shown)


haoranpb:
Should we introduce a test here for our future selves? That sounds like a tricky problem to identify

WaelAbuSeada (approved):

PR #4050, opened 2025-07-04
[Performance Scheduler] Allow scheduling profilers for all user types except external

onbuyuka:
#### Summary 
Recent change to show user name instead of ID in the scheduler card regressed ability to schedule for external users. That is because the user lookup page filters out external users. Added an API to the facade to not filter out external users.

#### Work Item(s) 
Fixes [AB#591223](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/591223)

### DIFF
--- a/src/System Application/Test/User Selection/src/UserSelectionTest.Codeunit.al
+++ b/src/System Application/Test/User Selection/src/UserSelectionTest.Codeunit.al
@@ -174,9 +174,9 @@ codeunit 135035 "User Selection Test"
 
         EnvironmentInfoTestLibrary.SetTestabilitySoftwareAsAService(true);
 
-        // [WHEN] OpenWithAllUsers function is called
+        // [WHEN] OpenWithSystemUsers function is called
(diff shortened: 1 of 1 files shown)


darjoo (approved):

src/System Application/App/User Selection/src/UserLookup.Page.al
```diff
@@ -66,13 +66,17 @@ page 9843 "User Lookup"
     var
         EnvironmentInformation: Codeunit "Environment Information";
     begin
-        UserSelectionImpl.HideExternalUsers(Rec);
+        if HideOnlyExternalUsers then
+            UserSelectionImpl.HideOnlyExternalUsers(Rec)
```
grobyns, line 70: consider a better name :-)

grobyns (approved):

gggdttt (approved):

PR #4116, opened 2025-07-11
[Cryptography Management] Failing test

darjoo:
#### Summary 
The DecryptWithPKCS1PaddingTextEncryptedWithOAEPPadding is failing due to a slightly different expectederror message showing up.

Fixing it by making the expected error message a little more generic.

#### Work Item(s) 
Fixes AB#591915

PR #4119, opened 2025-07-11
[releases/24.x] [Cryptography Management] Failing test

darjoo:
This pull request backports #4116 to releases/24.x

Fixes AB#[**Insert Work Item Number Here**]

### DIFF
--- a/src/System Application/Test/Cryptography Management/src/RSACryptoServiceProviderTests.Codeunit.al
+++ b/src/System Application/Test/Cryptography Management/src/RSACryptoServiceProviderTests.Codeunit.al
@@ -200,8 +200,9 @@ codeunit 132613 RSACryptoServiceProviderTests
         DecryptingTempBlob.CreateOutStream(DecryptedOutStream);
         asserterror RSACryptoServiceProvider.Decrypt(PrivateKeyXmlStringSecret, EncryptedInStream, false, DecryptedOutStream);
 
-        // [THEN] Error occures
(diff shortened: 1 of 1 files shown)


src/System Application/Test/Cryptography Management/src/RSACryptoServiceProviderTests.Codeunit.al
```diff
@@ -200,8 +200,9 @@ codeunit 132613 RSACryptoServiceProviderTests
         DecryptingTempBlob.CreateOutStream(DecryptedOutStream);
         asserterror RSACryptoServiceProvider.Decrypt(PrivateKeyXmlStringSecret, EncryptedInStream, false, DecryptedOutStream);
 
-        // [THEN] Error occures
-        LibraryAssert.ExpectedError('A call to System.Security.Cryptography.RSACryptoServiceProvider.Decrypt failed with this message: The parameter is incorrect.');
+
+        // [THEN] Error occurs
+        LibraryAssert.ExpectedError('A call to System.Security.Cryptography.RSACryptoServiceProvider.Decrypt failed with this message:');
```
Aleyenda, line 205: ?
The error message is blank now? Isn't this a failure with the code that throws the error and not with the test?

darjoo: No, the test fails because the specific error is different. The test is meant to catch an error. The specific error has changed and that is likely due to the dotnet library being updated.

I didn't want to hardcode the new message resulting in a later test instability when we update dlls. As long as we get an error failing on that specific dotnet function, we're good.

Aleyenda (approved):

gggdttt (approved):

PR #4134, opened 2025-07-14
[26.x][Performance Scheduler] Allow scheduling profilers for all user types except external

onbuyuka:
#### Summary 
Recent change to show user name instead of ID in the scheduler card regressed ability to schedule for external users. That is because the user lookup page filters out external users. Added an API to the facade to not filter out external users.

#### Work Item(s) 
Fixes [AB#591952](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/591952)


AndersLarsenMicrosoft (approved):
please go ahead

haoranpb (approved):

PR #4137, opened 2025-07-14
Introducing RequiredTestIsolation to TestRunner

haoranpb:
#### Summary 

Introducing `RequiredTestIsolation` to Test Runner. Now, engineers can specify the required test isolation on a codeunit.

e.g. see changes in `src/System Application/Test/Retention Policy/src/RetentionPolicyLogTest.Codeunit.al`.

There will be a few follow-up PRs, most importantly:
1. https://github.com/microsoft/navcontainerhelper/pull/3980
2. [System Application Tests (No Isolation)](https://github.com/microsoft/BCApps/tree/main/build/projects/System%20Application%20Tests%20(No%20Isolation)) should be able to be removed

#### Work Item(s) 
Fixes [AB#581153](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/581153)

### DIFF
--- a/src/Tools/Test Framework/Test Runner/src/CommandLineTestTool.Page.al
+++ b/src/Tools/Test Framework/Test Runner/src/CommandLineTestTool.Page.al
@@ -96,6 +96,14 @@ page 130455 "Command Line Test Tool"
                     if Rec.FindFirst() then;
                 end;
             }
+            field(RequiredTestIsolation; RequiredTestIsolation)
+            {
+                ApplicationArea = All;
+                Caption = 'Required Test Isolation';
+                ToolTip = 'Specifies the Required Test Isolation';
+                BlankZero = true;
+                MinValue = 0;
+            }
             field(TestType; TestType)
             {
                 ApplicationArea = All;
@@ -109,7 +117,7 @@ page 130455 "Command Line Test Tool"
                     TestSuiteMgt: Codeunit "Test Suite Mgt.";
                 begin
                     TestSuiteMgt.DeleteAllMethods(GlobalALTestSuite);
-                    TestSuiteMgt.SelectTestMethodsByExtensionAndTestType(GlobalALTestSuite, ExtensionId, TestType);
+                    TestSuiteMgt.SelectTestMethodsByExtensionAndTestCategorization(GlobalALTestSuite, ExtensionId, TestType, RequiredTestIsolation);
                     if Rec.FindSet() then;
                 end;
             }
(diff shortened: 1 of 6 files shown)


stkillen (dismissed):

onbuyuka (dismissed):

src/Tools/Test Framework/Test Runner/src/CommandLineTestTool.Page.al
```diff
@@ -96,6 +96,14 @@ page 130455 "Command Line Test Tool"
                     if Rec.FindFirst() then;
                 end;
             }
+            field(RequiredTestIsolation; RequiredTestIsolation)
+            {
+                ApplicationArea = All;
+                Caption = 'Required Test Isolation';
+                ToolTip = 'Specifies the Required Test Isolation';
```
NKarolak: Could you please create a user-friendlier tooltip?
Something like (I don't know how it works, please fix): "Specifies the test isolation level in which the test codeunit will be run."

Shouldn't `RequiredTestIsolation` rather be an enum than an integer? That way, you could also get rid of the `MinValue` and `BlankZero` properties.

haoranpb: Updating the Tooltip

---

`RequiredTestIsolation` is not exposed as an Enum type because we don't expect it to be used anywhere else (it is of type Option in `table CodeUnit Metadata`). This field will mostly be set from build pipelines where string or integers are supported

src/Tools/Test Framework/Test Runner/src/CommandLineTestTool.Page.al
```diff
@@ -96,6 +96,14 @@ page 130455 "Command Line Test Tool"
                     if Rec.FindFirst() then;
                 end;
             }
+            field(RequiredTestIsolation; RequiredTestIsolation)
+            {
+                ApplicationArea = All;
+                Caption = 'Required Test Isolation';
+                ToolTip = 'Specifies the RequiredTestIsolation for test filtering, test methods will only be updated after TestType is set';
```
NKarolak:
```suggestion
                ToolTip = 'Specifies the RequiredTestIsolation for test filtering. Test methods will only be updated after TestType is set.';
```

onbuyuka (approved):

stkillen (approved):

PR #4175, opened 2025-07-17
[Draft] Add Test for Application Manifest Consistency

aholstrup1:
#### Summary 
Add Test for Application Manifest Consistency

#### Work Item(s) 
Fixes [AB#591348](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/591348)

### DIFF
--- a/.github/actions/TestApplicationManifestConsistency/action.ps1
+++ b/.github/actions/TestApplicationManifestConsistency/action.ps1
@@ -1,0 +1,53 @@
+function Test-ApplicationManifestConsistency
+{
+    $errors = @()
+
+    $bcAppsManifests = (Get-ChildItem "src" -Recurse -Include app.json).FullName
+    $expectedAppVersion = "27.0.0.0"
+    $expectedPlatVersion = "27.0.0.0"
+    foreach($path in $bcAppsManifests)
+    {
+        $manifest = Get-Content $path | ConvertFrom-Json
+        if ($expectedAppVersion -ne $manifest.version)
+        {
+            $errors += "ERROR: Wrong application version in manifest $path. Expected: $expectedAppVersion. Actual: $($manifest.version)"
+        }
+        foreach($item in $manifest.dependencies)
+        {
+            if ($expectedAppVersion -ne $item.version)
+            {
+                $errors += "ERROR: Wrong dependency version ($($item.name)) in manifest $path. Expected: $expectedAppVersion. Actual: $($item.version)"
+            }
+        }
+        if ($manifest.platform -and ($expectedPlatVersion -ne $manifest.platform))
+        {
+            # If path is under App\BCApps, it means the app in the BCApps repository. In that case, do not fail if platform version do not match, as platform version is uptaken in BCApps at a later stage.
+            if($path -like "$env:InetRoot\App\BCApps\*") {
+                Write-Host "Skip checking platform version in $path as it is part of the BCApps repository."
+            }
+            else {
+                $errors += "ERROR: Wrong platform version in manifest $path. Expected: $expectedPlatVersion. Actual: $($manifest.platform)"
+            }
+        }
+        if ($manifest.publisher -ne "Microsoft")
+        {
+            # Allow Partner publisher if path contains 'Partner Test' and is under App\BCApps
+            if ($path -like "*Partner Test*" -and $manifest.publisher -eq "Partner")
+            {
+                Write-Host "Allowing Partner publisher in $path as it contains 'Partner Test' and is under BCApps."
+            }
+            else
+            {
+                $errors += "ERROR: Wrong publisher in manifest $path. Expected: Microsoft. Actual: $($manifest.publisher)"
+            }
+        }
+    }
+
+    if ($errors)
+    {
+        $errors | Write-Log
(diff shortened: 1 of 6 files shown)


.github/actions/TestApplicationManifestConsistency/action.ps1
```diff
@@ -0,0 +1,100 @@
...
+                $errors += "ERROR: Wrong dependency version ($($item.name)) in manifest $path. Expected: $ExpectedAppVersion. Actual: $($item.version)"
+            }
+        }
+        if ($manifest.platform -and ($ExpectedPlatformVersion -ne $manifest.platform))
+        {
+            $errors += "ERROR: Wrong platform version in manifest $path. Expected: $ExpectedPlatformVersion. Actual: $($manifest.platform)"
+        }
+        if ($manifest.publisher -ne "Microsoft")
+        {
+            # Allow Partner publisher if path contains 'Partner Test' and is under App\BCApps
```
mazhelez:
```suggestion
            # Allow Partner publisher if path contains 'Partner Test'

File: .github/actions/TestApplicationManifestConsistency/action.ps1 [hunk 2]
```diff
@@ -0,0 +1,100 @@
...
+            if ($path -like "*Partner Test*" -and $manifest.publisher -eq "Partner")
+            {
+                Write-Host "Allowing Partner publisher in $path as it contains 'Partner Test' and is under BCApps."
```
mazhelez:
```suggestion
                Write-Host "Allowing Partner publisher in $path as it contains 'Partner Test'."
```

.github/actions/TestApplicationManifestConsistency/action.ps1
```diff
@@ -0,0 +1,100 @@
...
+            }
+            else
+            {
+                $errors += "ERROR: Wrong publisher in manifest $path. Expected: Microsoft. Actual: $($manifest.publisher)"
```
mazhelez, line 39: If the path contains "Partner Test", but the publisher isn't "Partner", this message will be wrong.

The check needs more attention.

.github/actions/TestApplicationManifestConsistency/action.ps1
```diff
@@ -0,0 +1,100 @@
...
+        $manifest = Get-Content $path | ConvertFrom-Json
+        if ($ExpectedAppVersion -ne $manifest.version)
+        {
+            $errors += "ERROR: Wrong application version in manifest $path. Expected: $ExpectedAppVersion. Actual: $($manifest.version)"
+        }
+        foreach($item in $manifest.dependencies)
+        {
+            if ($ExpectedAppVersion -ne $item.version)
+            {
+                $errors += "ERROR: Wrong dependency version ($($item.name)) in manifest $path. Expected: $ExpectedAppVersion. Actual: $($item.version)"
```
mazhelez, line 23: Wouldn't this be a problem when incrementing the version for add-on apps?

E.g. the dependency will need to be, say, "28.0.0.0", but the artifact will have a 27 version.

.github/actions/TestApplicationManifestConsistency/action.ps1
```diff
@@ -0,0 +1,100 @@
...
+}
+
+function GetApplicationVersion
+{
+    Import-Module (Join-Path $PSScriptRoot "..\..\..\build\scripts\EnlistmentHelperFunctions.psm1" -Resolve)
+    $repoVersion = Get-ConfigValue -Key "repoVersion" -ConfigType AL-GO
+    return "$repoVersion.0.0"
+}
+
+function GetPlatformVersion
```
mazhelez, line 58: Too much of an effort, if you ask me.

The _platform_ property indicates a _minumum_ required version, right?

Either check if the same as the app version, or one lower.
Or check if the major version of the _platform_ property is the same as the major version from the artifact version.

.github/actions/TestApplicationManifestConsistency/action.ps1
```diff
@@ -0,0 +1,100 @@
...
+{
+    param(
+        [Parameter(Mandatory = $false)]
+        [string]$ExpectedAppVersion = (GetApplicationVersion),
+        [Parameter(Mandatory = $false)]
+        [string]$ExpectedPlatformVersion = (GetPlatformVersion)
+    )
+    $errors = @()
+
+    $bcAppsManifests = (Get-ChildItem "src" -Recurse -Include app.json).FullName
```
mazhelez, line 11: How about parametrizing this script, so it can be re-used instead of copied?

.github/actions/TestApplicationManifestConsistency/action.ps1
aholstrup1: Idk... I'd say the expected publisher should always be Microsoft regardless of the path. In this one app we allowed it to be something else but that is an exception.

mazhelez: That's fine as well.

My understanding was the _publisher_ should be "Partner" for apps in the "Partner Test" folder.

PR #4185, opened 2025-07-19
Ready Test Runner for transition period and add column for Test Categorization

haoranpb:
#### Summary 

The changes are for Test Runner:
1. Add column for Test Categorization 
2. Add procedure for transition period

#### Work Item(s) 
Fixes [AB#581421](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/581421)

### DIFF
--- a/src/Tools/Test Framework/Test Runner/src/CommandLineTestTool.Page.al
+++ b/src/Tools/Test Framework/Test Runner/src/CommandLineTestTool.Page.al
@@ -117,7 +117,7 @@ page 130455 "Command Line Test Tool"
                     TestSuiteMgt: Codeunit "Test Suite Mgt.";
                 begin
                     TestSuiteMgt.DeleteAllMethods(GlobalALTestSuite);
-                    TestSuiteMgt.SelectTestMethodsByTestCategorization(GlobalALTestSuite, TestType, RequiredTestIsolation);
+                    TestSuiteMgt.SelectTestMethodsByExtensionAndTestCategorization(GlobalALTestSuite, ExtensionId, TestType, RequiredTestIsolation);
                     if Rec.FindSet() then;
                 end;
             }
--- a/src/Tools/Test Framework/Test Runner/src/TestSuiteMgt.Codeunit.al
+++ b/src/Tools/Test Framework/Test Runner/src/TestSuiteMgt.Codeunit.al
@@ -329,10 +329,17 @@ codeunit 130456 "Test Suite Mgt."
         GetTestMethods(ALTestSuite, CodeunitMetadata);
     end;
 
-    internal procedure SelectTestMethodsByTestCategorization(var ALTestSuite: Record "AL Test Suite"; TestType: Integer; RequiredTestIsolation: Integer)
+    internal procedure SelectTestMethodsByExtensionAndTestCategorization(var ALTestSuite: Record "AL Test Suite"; ExtensionID: Text; TestType: Integer; RequiredTestIsolation: Integer)
     var
         CodeunitMetadata: Record "CodeUnit Metadata";
+        AppExtensionId: Guid;
     begin
+        if ExtensionID <> '' then begin
+            Evaluate(AppExtensionId, ExtensionID);
+
+            CodeunitMetadata.SetRange("App ID", AppExtensionId);
+        end;
+
         CodeunitMetadata.SetRange(SubType, CodeunitMetadata.SubType::Test);
 
         // inexplicit conversion from Integer to Option


src/Tools/Test Framework/Test Runner/src/CommandLineTestTool.Page.al
```diff
@@ -49,9 +49,15 @@ page 130455 "Command Line Test Tool"
                 var
                     TestSuiteMgt: Codeunit "Test Suite Mgt.";
                 begin
-                    TestSuiteMgt.DeleteAllMethods(GlobalALTestSuite);
-                    TestSuiteMgt.SelectTestMethodsByRange(GlobalALTestSuite, TestCodeunitRangeFilter);
-                    if Rec.FindFirst() then;
+                    if TestType = 0 then begin
```
nikolakukrika: Nit - Maybe it is better to move the logic to the SelectTestMethodsByRange
The new method could be an overload of the existing and we could have a check on the testtype.
This is to have less logic on the page.

src/Tools/Test Framework/Test Runner/src/CommandLineTestTool.Page.al
```diff
@@ -49,9 +49,15 @@ page 130455 "Command Line Test Tool"
                 var
                     TestSuiteMgt: Codeunit "Test Suite Mgt.";
                 begin
```
nikolakukrika, line 51: Do we need the same change on the main page?

src/Tools/Test Framework/Test Runner/src/TestMethodLine.Table.al
```diff
@@ -133,6 +133,26 @@ table 130450 "Test Method Line"
...
+        {
+            Access = Internal;
+            Editable = false;
+            OptionMembers = None,UnitTest,IntegrationTest,Uncategorized;
+            Caption = 'Test Type';
+            ToolTip = 'Specifies the type of the test, indicating the context in which it runs.';
+            FieldClass = FlowField;
+            CalcFormula = lookup("Codeunit Metadata".TestType where(ID = field("Test Codeunit")));
+        }
+        field(58; "Required Test Isolation"; Option)
```
nikolakukrika: I believe this property should be settable on the header. I do not mind having it in the list, but it should be visible false.

src/Tools/Test Framework/Test Runner/src/TestSuiteMgt.Codeunit.al
```diff
@@ -284,6 +284,24 @@ codeunit 130456 "Test Suite Mgt."
         GetTestMethods(ALTestSuite, CodeunitMetadata);
     end;
 
+    /// <summary>
+    /// Selects test methods by range and test categorization (only the overlapping ones).
+    /// This procedure is mostly needed during the transition period when the test categorization is not yet fully implemented.
+    /// </summary>
+    internal procedure SelectTestMethodsByRangeAndTestCategorization(var ALTestSuite: Record "AL Test Suite"; TestCodeunitFilter: Text; TestType: Integer; RequiredTestIsolation: Integer)
```
nikolakukrika: We could have the overload of the existing method. SelectTestMethods adding 2 new parameters is intuitive enough.
There is a lot of text in the name of this one.

src/Tools/Test Framework/Test Runner/src/TestMethodLine.Table.al
```diff
@@ -133,6 +133,26 @@ table 130450 "Test Method Line"
             CalcFormula = count("Test Method Line" where("Test Suite" = field("Test Suite"), "Test Codeunit" = field("Test Codeunit"), "Line Type" = const(Function)));
             Editable = false;
         }
+        field(57; "Test Type"; Option)
```
nikolakukrika: The issue here is that the test type applies to the entire codeunit. We will get the data also reflected on the methods. 
I would add this option to the header in a similar way we add the test isolation. We could also add it to the Select dialogs so it is visible. I do not mind having it on the list, we can add it as visible false if it is adding too much data.

nikolakukrika (dismissed):

src/Tools/Test Framework/Test Runner/src/TestMethodLine.Table.al
haoranpb: Agreed! Moving it to `SelectTests.Page.al`

src/Tools/Test Framework/Test Runner/src/CommandLineTestTool.Page.al
haoranpb: Good question, I don't think so.

Don't think find test codeunit by range is a common scenario outside of CLI context, at least I never used it myself

src/Tools/Test Framework/Test Runner/src/TestMethodLine.Table.al
haoranpb: Moved to SelectTests Page, will not put it on the header yet. Because I can imagine people will try to call get test codeunits from different types during the transition period at least

WaelAbuSeada (dismissed):

nikolakukrika (approved):

stkillen (approved):

PR #4244, opened 2025-07-28
Get all app symbols when gathering baselines

aholstrup1:
#### Summary 
We ran into an issue where the baseline app (needed by AppSourceCop) had an dependency to an app that the current version of the app doesn't. 
`Error: AS0091 One or more dependencies of the previous version of the extension could not be found in the baseline package cache folder. Details: error AL1022: A package with publisher 'Microsoft', name 'E-Document Connector - Pagero', and a version compatible with '26.4.0.0' could not be found in the package cache folders: c:\shared\build\projects\Apps (W1)\.packages`

The proposed solution is to copy all baseline app symbols into the baselines folder so that those dependencies can be resolved. 

#### Work Item(s) 
Fixes [AB#593436](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/593436)

### DIFF
--- a/build/scripts/GuardingV2ExtensionsHelper.psm1
+++ b/build/scripts/GuardingV2ExtensionsHelper.psm1
@@ -101,6 +101,11 @@ function Restore-BaselinesFromArtifacts {
         $BaselineVersion = $baselinePackage.Version
     }
 
+    if (-not (Test-Path $TargetFolder)) {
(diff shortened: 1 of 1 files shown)


Groenbech96 (approved):
LGTM

ankosiorek (approved):

PR #4276, opened 2025-07-29
Reducing critical build time

aholstrup1:
#### Summary 
This PR reduces the build time with 25-30 minutes by reducing the build graph depth from 3 to 2. In order to minimize the build graph we will: 
* Compile the System Application, Business Foundation and their test library apps in the same job. 
* Compile the test apps and run the tests in the subsequent job 

This allows the "Build Apps (W1)" job to run at depth 2 instead of depth 3 which is what reduces the build time. Furthermore, it also moves up the compilation of the Business Foundation so if there are compilation issues in the Business Foundation the build will fail faster.  

Before https://github.com/microsoft/BCApps/actions/runs/16591734810 (1h 11m)

After: https://github.com/microsoft/BCApps/actions/runs/16592812687 (46m)

#### Work Item(s) 
Fixes [AB#593466](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/593466)


Groenbech96 (approved):
LGTM

jonasblunck (approved):

PR #4310, opened 2025-08-01
Update ABSFormatHelper.Codeunit.al

### LINKED ISSUE #4309: [Bug]: Evaluate Error when The time is not valid in the current time zone. (clock is adjusted forward) -> context Azure Blob Storage
### Describe the issue

There is a possibility that files(blobs) cannot be imported from Azure Blob Storage because the Created or Modified At timestamp cannot be evaluated.

The timestamps of the file/blobs are returned in the API using UTC time. The timestamps (created, modified) returned by the API are converted to DateTime as text using Evaluate.
Evaluate throws an error if the time does not exist in the region. (Error Message: The time '30.03.2025 02:00:01' is not valid in the current time zone. For example, when the clock is adjusted forward, any time in the period that is skipped is invalid.)
In my opinion, this can only happen in the context of the switch to summer saving time (e.g., from 2:00 a.m. to 3:00 a.m.). All times between 2 and 3 then do not exist.

However, this time does exist in UTC time, and it is possible to create a file in Azure Blob Storage during this time. Due to the Evaludate error, however, these files cannot be imported.

In my opinion, the Evaluate function should offer the option of performing a UTC validation. After all, the time entered is actually validated.

### Expected behavior

Blobs from Azure Blob Storage that were created during the time change should be able to be imported.

Evaluate should not throw an error if it is a Validate UTC input. Perhaps a UTC time parameter is needed.

### Steps to reproduce

Create a file in Azure Blob Storage during the time shift in your region and try to import it.

### Additional context

_No response_

MattiLE:
#### Work Item(s) 
Fixes #4309

### DIFF
--- a/src/System Application/App/Azure Blob Services API/src/Helper/ABSFormatHelper.Codeunit.al
+++ b/src/System Application/App/Azure Blob Services API/src/Helper/ABSFormatHelper.Codeunit.al
@@ -214,9 +214,26 @@ codeunit 9044 "ABS Format Helper"
         NewDateTime := 0DT;
         // PropertyValue is something like the following: 'Mon, 24 May 2021 12:25:27 GMT'
         // 'Evaluate' converts these correctly
-        if Evaluate(NewDateTime, PropertyValue) then;
-        exit(NewDateTime);
+        if Evaluate(NewDateTime, PropertyValue) then
+            exit(NewDateTime);
+        exit(EvaluateWithDotNetDateTime(PropertyValue));
     end;
+
+    local procedure EvaluateWithDotNetDateTime(var NewDateTime: DateTime; PropertyValue: Text)
+    var
+        DotNet_DateTime: Codeunit DotNet_DateTime;
+        DotNet_CultureInfo: Codeunit DotNet_CultureInfo;
+        DotNet_DateTimeStyles: Codeunit DotNet_DateTimeStyles;
+        DatePart: Date;
+    begin
+        if DotNet_DateTime.TryParse(DateTimeText, DotNet_CultureInfo, DotNet_DateTimeStyles) then
+            exit(DotNet_DateTime.ToDateTime());
+
+        Evaluate(DatePart, DateTimeText);
+        exit(CreateDateTime(DatePart, 0T));
+    end;
+
+
 
     procedure ConvertToInteger(PropertyValue: Text): Integer
     var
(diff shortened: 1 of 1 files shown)


src/System Application/App/Azure Blob Services API/src/Helper/ABSFormatHelper.Codeunit.al
```diff
@@ -214,7 +214,7 @@ codeunit 9044 "ABS Format Helper"
         NewDateTime := 0DT;
         // PropertyValue is something like the following: 'Mon, 24 May 2021 12:25:27 GMT'
         // 'Evaluate' converts these correctly
-        Evaluate(NewDateTime, PropertyValue);
+        if Evaluate(NewDateTime, PropertyValue) then;
```
JesperSchulz: I understand the problem, but is this the solution? So now, if it cannot convert, it returns 0DT. I guess that could be ok, but it somehow smells a little...

MattiLE: Hi @JesperSchulz, I understand what you mean. But I couldn't think of a really clean solution either. Maybe you could calculate an hour down based on the text or something like that...
I thought I'd suggest the simplest option first and then get feedback from you :)

From my point of view, however, the Evaluate function is actually the problem here. The time entered is actually correct. It's just not correct for the selected region. Do you think it would be possible to extend the Evaluate function so that you can specify that it is a UTC time? Perhaps with the 9 parameter?

nikolakukrika: This fix is not good - Hiding errors is dangerous. 
What is a consequence of returning a blank date? How do we troubleshoot these scenarios?

We should either change the evaluate to always use 9, if we are confident that this is safe.
If we need to use a different evaluation, then we either need a different method or a different event. 

We should avoid the constructs like if **_something_** then ; 
We need to implement the then part. If it is not possible, then it is the time to add a comment why.

MattiLE: Hi @nikolakukrika, I agree with you that this is not a good solution to the problem. However, the code as it currently stands causes real problems that can only be fixed manually at great effort. 
I only added this fix because I don't know if it's possible to extend the Evaluate function. If you show me an idea, I can expand/change the PR again.

@JesperSchulz fyi

nikolakukrika: @MattiLE I would write a more complex fix.

Clock shift errors happen very rarely. These errors are actually thrown by the .NET, so we cannot do much. These date times do not exist. 

Swallowing errors is not a good approach because we have no idea that we cannot parse the dates, it can be dangerous to swallow errors and get a bunch of 0DTs that may be hard to fix. Datetime could be in a weird format, this will happen much more often than the clock shift error. 

Currently the code throws an error. We can make this method as a try function and the behavior will be preserved.

In the calling code we can check for the error that is raised and if it is a clock error we could attempt to parse the date from the datetime and add a clock to the value that is supported, e.g. 00:00 or 30 minutes after midnight.

Aleyenda:
This does not appear to compile and it's stale. Closing. If this is still relevant, please submit a new issue.

PR #4335, opened 2025-08-04
Remove AOAI account validation

rupeshdan:
#### Summary 

Fixes [AB#582261](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/582261)


encimita (approved):
Looks good, let's wait for Preeti's confirmation!

encimita (approved):

src/System Application/App/AI/src/Azure OpenAI/AOAIAuthorization.Codeunit.al
```diff
@@ -95,6 +73,16 @@ codeunit 7767 "AOAI Authorization"
         ApiKey := NewApiKey;
         ManagedResourceDeployment := NewManagedResourceDeployment;
     end;
+#endif
+
+    [NonDebuggable]
```
darjoo, line 65: Remove the ApiKey/AOAIAccountName from the original function.
This codeunit is internal, so we can make those changes as they are non-breaking. I'd like to reduce the redundancy in the implementation codeunit.

Just point the facade's API to the right procedure now and it might need a pragma there.

darjoo (dismissed):

rupeshdan:
Comments addressed, please review.

PR #4366, opened 2025-08-05
Draft system app wrapper around CDS table extension generation

thloke:
#### Summary 

Adds a new system codeunit to handle generation of apps from CDS schema to allow for in-product extensions of CDS tables.

#### Work Item(s) 
Implements[AB#592771](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/592771/)

### DIFF
--- a/src/System Application/App/Extension Management/src/ExtensionGeneration.Codeunit.al
+++ b/src/System Application/App/Extension Management/src/ExtensionGeneration.Codeunit.al
@@ -1,0 +1,27 @@
+
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Apps;
+
+codeunit 2507 "Extension Generation"
+{
+    SingleInstance = true;
+
+    procedure StartCDSExtensionGeneration(OverwriteExisting: Boolean)
+    begin
+
+    end;
+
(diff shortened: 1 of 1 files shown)


SBalslev (approved):

src/System Application/App/Extension Management/src/ExtensionGeneration.Codeunit.al
```diff
@@ -0,0 +1,29 @@
...
+
+namespace System.Apps;
+
+codeunit 2507 "Extension Generation"
+{
+    SingleInstance = true;
+    InherentEntitlements = X;
+    InherentPermissions = X;
+
+    procedure StartCDSExtensionGeneration(OverwriteExisting: Boolean)
```
SBalslev, line 15: Long term do we want to keep the CDS prefix here or move it to a codeunit name?
If we're looking to create a builder API we can come up with better naming and put the purpose in the container.
ex. "Dataverse Table Builder" or "Dataverse Proxy Builder"

SBalslev: .BeginGeneration
.UpdateExistingTable
.CommitGeneration

Could be good procedure names

thloke: Good points, either the codeunit should be dataverse specific, or the methods should be more generic. Given the state of current discussions around extension generation, I think making this dataverse specific is better - so that we don't give off the wrong signals to partners on this.

onbuyuka:
Closing this one

PR #4385, opened 2025-08-06
Update delegated user plans on login

stkillen:
#### Summary 
Draft

**Problem**
Delegated plans are not updated on user sync. If they're ever changed, we do not update them.

**Solution**
Update delegated user plans on login.

#### Work Item(s) 
Fixes [AB#539137](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/539137/)

### DIFF
--- a/src/System Application/App/Azure AD Plan/src/AzureADPlanImpl.Codeunit.al
+++ b/src/System Application/App/Azure AD Plan/src/AzureADPlanImpl.Codeunit.al
@@ -77,6 +77,7 @@ codeunit 9018 "Azure AD Plan Impl."
     end;
 
     [NonDebuggable]
+    [InherentPermissions(PermissionObjectType::TableData, Database::"User Plan", 'r')]
     procedure IsPlanAssignedToUser(PlanGUID: Guid; UserGUID: Guid): Boolean
     var
         UserPlan: Record "User Plan";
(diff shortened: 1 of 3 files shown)


src/System Application/App/Azure AD Plan/src/AzureADPlanImpl.Codeunit.al
```diff
@@ -672,14 +676,32 @@ codeunit 9018 "Azure AD Plan Impl."
...
         UserPlan.Insert();
 
+        // Exit if user access should not be updated
+        if SkipUpdateUserAccess then
+            exit;
+
+        if not UserProperty.Get(UserSID) then
+            exit;
+
+        if DoesUserHavePlans(UserSID) then
```
darjoo: We don't want to update the user's plan if they have say Essentials already?
Would it make sense to still update with Delegated plan?

darjoo (dismissed):

src/System Application/App/Azure AD Plan/src/AzureADPlanImpl.Codeunit.al
stkillen: We do want to update the users plan, but we do that before this point. This part of the code is only reached when SkipUpdateUserAccess is false, e.g. when we want to update the user's permission sets. 

However, pointing this out made me realize I can't do the check here, as we've just assigned the plan. Will remove.

stkillen: Actually, it's a good safety measure to ensure we don't change permissions of a user that is already created. I will do the check before adding the plan

PR #4392, opened 2025-08-07
Add proper permission to email dispatcher

gggdttt:
#### Summary 
User cannot send email with permission error.

User can update the license to get rid of it.  But we should add the inherentpermission to force the permission. 

#### Work Item(s) 
Fixes AB#596403

### DIFF
--- a/src/System Application/App/Email/src/Email/EmailRetryImpl.Codeunit.al
+++ b/src/System Application/App/Email/src/Email/EmailRetryImpl.Codeunit.al
@@ -60,6 +60,7 @@ codeunit 8909 "Email Retry Impl."
         exit(true);
     end;
 
+    [InherentPermissions(PermissionObjectType::TableData, Database::"Email Retry", 'rd')]
     internal procedure CleanEmailRetry(MessageId: Guid)
     var
         EmailRetry: Record "Email Retry";


pri-kise (commented):
While I was looking at this PR I've checked the lines below and above the change.

is the `not` correct in the check 
```al
        if not EmailRetry.IsEmpty() then
            exit;
```
Shouldn't it be 
```al
        if EmailRetry.IsEmpty() then
            exit;
```

src/System Application/Test/Email/src/EmailRetryTest.Codeunit.al
```diff
@@ -472,6 +472,54 @@ codeunit 134703 "Email Retry Test"
...
+        EmailRetry.Insert();
+
+        EmailRetry.SetRange("Message Id", EmailMessage.GetId());
+        Assert.IsFalse(EmailRetry.IsEmpty, 'Email retry record should be empty');
+
+        Codeunit.Run(Codeunit::"Email Dispatcher", EmailOutbox);
+
+        // [Then] The email retry record should be clean
+        EmailRetry.SetRange("Message Id", EmailMessage.GetId());
+        Assert.IsTrue(EmailRetry.IsEmpty, 'Email retry record should be empty');
```
pri-kise, line 520: nit: Just as note.
The codeunit `Library - Assert` has a procdure `RecordIsNotEmpty` that could be used for this assertion instead. 
This method returns the current filters in the error message which sometimes can be useful. (In this case where we would only see a filter on a guid field, this wouldn't provide much more information).

Groenbech96 (approved):

PR #4413, opened 2025-08-11
Agents - Automatically resolve duplicates on User Name

nikolakukrika:
#### Summary 
Duplicates are possible on the User Name. We cannot remove the ability to specify the user name, we need to ensure that the error is not thrown, so we will append number if there are duplicate agents.

#### Work Item(s) 
Fixes [AB#580397](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/580397)

### DIFF
--- a/src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
+++ b/src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
@@ -22,8 +22,8 @@ codeunit 4301 "Agent Impl."
                   tabledata "Application User Settings" = rim,
                   tabledata User = r,
                   tabledata "User Personalization" = rim;
-    internal procedure CreateAgent(AgentMetadataProvider: Enum "Agent Metadata Provider"; var UserName: Code[50]; AgentUserDisplayName: Text[80]; var TempAgentAccessControl: Record "Agent Access Control" temporary): Guid
 
+    internal procedure CreateAgent(AgentMetadataProvider: Enum "Agent Metadata Provider"; var UserName: Code[50]; AgentUserDisplayName: Text[80]; var TempAgentAccessControl: Record "Agent Access Control" temporary): Guid
     var
         Agent: Record Agent;
     begin
@@ -436,7 +436,6 @@ codeunit 4301 "Agent Impl."
         AgentNumberSeparatorTok: Label '-', Locked = true;
     begin
         // Check if the user name is already unique
-        User.Ascending(false);
         User.SetRange("User Name", AgentUserName);
         User.ReadIsolation := IsolationLevel::ReadUncommitted;
         if User.IsEmpty() then


src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
```diff
@@ -425,6 +426,41 @@ codeunit 4301 "Agent Impl."
...
+    var
+        User: Record User;
+        UniqueUserName: Text[50];
+        AgentNamePrefix: Text[50];
+        NumberOfAgentDigits: Integer;
+        MaximumPrefixLength: Integer;
+        AgentNumberSeparatorTok: Label '-', Locked = true;
+    begin
+        // Check if the user name is already unique
+        User.Ascending(false);
```
darjoo: No reason for `User.Ascending(false)`, we are filtering on `"User Name"` and check `IsEmpty()`

src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
```diff
@@ -425,6 +426,41 @@ codeunit 4301 "Agent Impl."
...
+        User.SetRange("User Name", AgentUserName);
+        User.ReadIsolation := IsolationLevel::ReadUncommitted;
+        if User.IsEmpty() then
+            exit(AgentUserName);
+
+        // If not check if there is a user with digits at the end of the name
+        NumberOfAgentDigits := 2;
```
darjoo, line 445: Edge case: We have environments with more than 100 companies, this will result in the same issue if they enable the same agent on those companies.

nikolakukrika: The case will be even more rare. 
These duplicates should not occur, the developers should provide unique names. We are seeing errors in some strange cases, e.g. company was deleted and created with the same name. Even in these cases you can get 100 duplicates if your name is longer than 47 characters (which should be super rare). If name is less than 47 characters it will use all the possible space. In case we ever get issues we can HF and increase the number of digits to 3 - which is 1000.

src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
```diff
@@ -425,6 +426,40 @@ codeunit 4301 "Agent Impl."
...
+#pragma warning disable AA0139
+            AgentNamePrefix := AgentUserName + AgentNumberSeparatorTok
+#pragma warning restore AA0139
+        else
+            AgentNamePrefix := CopyStr(AgentUserName, 1, MaximumPrefixLength) + AgentNumberSeparatorTok;
+
+        // Generate a unique user name by appending digits
+        User.SetFilter("User Name", '%1', AgentNamePrefix + '*');
+#pragma warning disable AA0139
+        UniqueUserName := AgentNamePrefix + Format(User.Count() + 2);
```
mynjj, line 457: If agent users can be deleted then this approach to get the suffix number is not fail-safe (deleting messes up with the count). While this is even more an edge case, I'd feel more safe with FindLast+IncStr, but I'm also ok with this approach.

mynjj (approved):

src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
nikolakukrika: Users cannot be deleted if they have logged in. Also User is inaccessible if it is of type agent. 
We can HF if needed, this approach has a better performance.

PR #4435, opened 2025-08-12
[Shopify] "Shpfy Checklist Init. Test" test does not consider other extensions installed

onbuyuka:
#### Summary 
Now that we run tests on different dbs, assert on number of checklist records is unstable. We already verify Shopify checklist items so there's no need to check the number of records.

#### Work Item(s) 
Fixes [AB#596378](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/596378)


grobyns (approved):

BardurKnudsen (commented):
Looks good.

PR #4493, opened 2025-08-21
[Agent] Open Agent Setup Page without using an Agent record for RunModal

qutreson:
#### Summary 
Provide a method allowing to open up the agent setup page without dependency on the Agent record being the source table.

#### Work Item(s) 
Fixes [AB#593058](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/593058)


src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
```diff
@@ -460,6 +460,33 @@ codeunit 4301 "Agent Impl."
...
+        SourceRecordVariant: Variant;
+        SetupPageId: Integer;
+        UserSecurityIdTok: Label 'User Security ID', Locked = true;
+    begin
+        AgentMetadata := AgentMetadataProvider;
+        SetupPageId := AgentMetadata.GetSetupPageId(AgentUserSecurityID);
+
+        PageMetadata.Get(SetupPageId);
+        SetupPageRecordRef.Open(PageMetadata.SourceTable);
+
```
BazookaMusic, line 482: You can do now:
UserSecurityIdFieldRef  := SetupPageRecordRef.Field(UserSecurityIdTok);

src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
```diff
@@ -460,6 +460,33 @@ codeunit 4301 "Agent Impl."
...
+        FieldMetadata.SetRange(TableNo, PageMetadata.SourceTable);
+        FieldMetadata.SetRange(FieldName, UserSecurityIdTok);
+        FieldMetadata.FindFirst();
+
+        UserSecurityIdFieldRef := SetupPageRecordRef.Field(FieldMetadata."No.");
+        UserSecurityIdFieldRef.SetFilter(AgentUserSecurityID);
+        SourceRecordVariant := SetupPageRecordRef;
+        Page.RunModal(SetupPageId, SourceRecordVariant);
```
qutreson, line 494: Should we call CurrPage.Update(false); here?

This way the list is updated if the agent was disabled?

tsjmicrosoft: Notice that the agent card already calls CurrPage.Update(false) after using the setup page.

qutreson: Yes, this is why I'm asking. I'll add it

src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
```diff
@@ -460,6 +460,33 @@ codeunit 4301 "Agent Impl."
...
+        AgentMetadata: Interface IAgentMetadata;
+        SourceRecordVariant: Variant;
+        SetupPageId: Integer;
+        UserSecurityIdTok: Label 'User Security ID', Locked = true;
+    begin
+        AgentMetadata := AgentMetadataProvider;
+        SetupPageId := AgentMetadata.GetSetupPageId(AgentUserSecurityID);
+
+        PageMetadata.Get(SetupPageId);
+        SetupPageRecordRef.Open(PageMetadata.SourceTable);
```
nikolakukrika, line 481: We need to check if the page has the source table, and throw a message if it doesn't

src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
```diff
@@ -460,6 +460,33 @@ codeunit 4301 "Agent Impl."
...
+
+        FieldMetadata.SetRange(TableNo, PageMetadata.SourceTable);
+        FieldMetadata.SetRange(FieldName, UserSecurityIdTok);
+        FieldMetadata.FindFirst();
```
nikolakukrika: I would add if not FieldMetadata.FindFirst() then Error(SourceTableMustHaveUserIDFieldErr);
This is to make it more intuitive to the developer what is needed.

src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
```diff
@@ -460,6 +460,33 @@ codeunit 4301 "Agent Impl."
...
+
+        UserSecurityIdFieldRef := SetupPageRecordRef.Field(FieldMetadata."No.");
```
nikolakukrika, line 491: We should check if the type of the field is guid and throw an error if it is not.

src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
nikolakukrika: No need to do it in the codeunit, the callers should do this.

nikolakukrika: it is probably the wrong placement if it was referring to the list.

nikolakukrika (dismissed):

src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
qutreson: Yes, it was meant to be a comment on the list page!

src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
qutreson: Added!

src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
qutreson: I'm not sure I'm gaining much with the change since I need the FieldMetadata to perform the extra checks proposed by Nikola

src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
qutreson: Done!

src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
qutreson: Done!

PR #4495, opened 2025-08-21
External Storage - Document Attachments

### LINKED ISSUE #4494: [BC Idea]: Adjustments for External Storage module
### BC Idea Link

/

### Description

This is adaptation for new module External Storage.

StefanSosic:
#### Summary 
# Document Attachments External Storage for Microsoft Dynamics 365 Business Central

## Overview

The External Storage extension provides seamless integration between Microsoft Dynamics 365 Business Central and external storage systems such as Azure Blob Storage, SharePoint, and File Shares. This extension automatically manages document attachments by storing them in external storage systems while maintaining full functionality within Business Central.

### **Automatic Upload**
- Automatically uploads new document attachments to configured external storage
- Supports multiple storage connectors via the File Account framework
- Generates unique file names to prevent collisions
- Maintains original file metadata and associations

### **Flexible Deletion Policies**
- **Immediately**: Delete from internal storage right after external upload
- **1 Day**: Keep internally for 1 day before deletion
- **7 Days**: Keep internally for 7 days before deletion (default)
- **14 Days**: Keep internally for 14 days before deletion

### **Bulk Operations**
- Synchronize multiple files between internal and external storage
- Bulk upload to external storage
- Bulk download from external storage
- Progress tracking with detailed reporting

### Prerequisites
- Microsoft Dynamics 365 Business Central version 27.0 or later
- File Account module configured with external storage connector
- Appropriate permissions for file operations

### Installation Steps

1. **Configure File Account**
   - Open **File Accounts** page
   - Create a new File Account with your preferred connector:
     - Azure Blob Storage
     - SharePoint
     - File Share
   - Assign the account to **External Storage** scenario

2. **Configure External Storage**
   - Open **File Accounts** page
   - Select assigned **External Storage** scenario
   - Open **Additional Scenario Setup**
   - Configure settings:
     - **Auto Upload**: Enable automatic upload of new attachments
     - **Delete After**: Set retention policy for internal storage

#### Auto Upload Settings
- **Enabled**: New document attachments are automatically uploaded to external storage
- **Disabled**: Manual upload required via actions

### Automatic Mode
When Auto Upload is enabled:
1. User attaches a document to any Business Central record
2. System automatically uploads to external storage
3. File remains accessible through standard attachment functionality
4. Internal file is deleted based on configured retention policy

#### Individual File Operations
From **Document Attachment - External** page:
- **Upload to External Storage**: Upload selected file
- **Download from External Storage**: Download file for viewing
- **Download to Internal Storage**: Restore file to internal storage
- **Delete from External Storage**: Remove file from external storage
- **Delete from Internal Storage**: Remove file from internal storage

#### Bulk Operations
From **External Storage Synchronize** report:
- **To External Storage**: Upload multiple files to external storage
- **From External Storage**: Download multiple files from external storage
- **Delete Expired Files**: Clean up files based on retention policy

### File Access
- Files uploaded to external storage remain fully accessible through standard Business Central functionality
- Document preview, download, and management work seamlessly
- No change to end-user experience

**© 2025 Microsoft Corporation. All rights reserved.**

#### Work Item(s) 
Fixes [AB#592686](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/592686)

Fixes #4494

### DIFF
--- a/src/Apps/W1/External Storage - Document Attachments/app/src/AutomaticSync/DAExternalStorageMigration.Report.al
+++ b/src/Apps/W1/External Storage - Document Attachments/app/src/AutomaticSync/DAExternalStorageMigration.Report.al
@@ -17,7 +17,7 @@ report 8753 "DA External Storage Migration"
     ProcessingOnly = true;
     UseRequestPage = true;
     Extensible = false;
-    ApplicationArea = Basic, Suite;
+    ApplicationArea = All;
     UsageCategory = None;
     Permissions = tabledata "DA External Storage Setup" = r,
                   tabledata "Document Attachment" = rimd;
@@ -96,7 +96,7 @@ report 8753 "DA External Storage Migration"
                     Caption = 'General';
                     label(InfoLabel)
                     {
-                        ApplicationArea = Basic, Suite;
+                        ApplicationArea = All;
                         Caption = 'This report will migrate all document attachments from a previous environment or company folder to the current environment/company folder in external storage.';
                     }
                 }
--- a/src/Apps/W1/External Storage - Document Attachments/app/src/AutomaticSync/DAExternalStorageSync.Report.al
+++ b/src/Apps/W1/External Storage - Document Attachments/app/src/AutomaticSync/DAExternalStorageSync.Report.al
@@ -17,7 +17,7 @@ report 8752 "DA External Storage Sync"
     ProcessingOnly = true;
     UseRequestPage = true;
     Extensible = false;
-    ApplicationArea = Basic, Suite;
+    ApplicationArea = All;
     UsageCategory = None;
     Permissions = tabledata "DA External Storage Setup" = r,
                   tabledata "Document Attachment" = r;
@@ -115,21 +115,21 @@ report 8752 "DA External Storage Sync"
                     Caption = 'General';
                     field(SyncDirectionField; SyncDirection)
                     {
-                        ApplicationArea = Basic, Suite;
+                        ApplicationArea = All;
                         Caption = 'Sync Direction';
                         OptionCaption = 'To External Storage,To Internal Storage';
                         ToolTip = 'Specifies whether to sync to external storage or to internal storage.';
                     }
                     field(OperationField; Operation)
                     {
-                        ApplicationArea = Basic, Suite;
+                        ApplicationArea = All;
                         Caption = 'Operation';
                         OptionCaption = 'Copy,Move';
                         ToolTip = 'Specifies whether to copy files (leaving them in the source) or move them (deleting from the source after successful copy).';
                     }
                     field(MaxRecordsToProcessField; MaxRecordsToProcess)
                     {
-                        ApplicationArea = Basic, Suite;
+                        ApplicationArea = All;
                         Enabled = SyncDirection = SyncDirection::"To External Storage";
                         Caption = 'Maximum Records to Process';
                         ToolTip = 'Specifies the maximum number of records to process in one run. Leave 0 for unlimited.';
--- a/src/Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DAExternalStorageImpl.Codeunit.al
+++ b/src/Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DAExternalStorageImpl.Codeunit.al
@@ -28,7 +28,7 @@ codeunit 8751 "DA External Storage Impl." implements "File Scenario"
     /// <param name="Scenario">The ID of the file scenario.</param>
     /// <param name="Connector">The file storage connector.</param>
     /// <returns>True if the operation is allowed, otherwise false.</returns>
-    procedure BeforeAddOrModifyFileScenarioCheck(Scenario: Enum "File Scenario"; Connector: Enum System.ExternalFileStorage."Ext. File Storage Connector") SkipInsertOrModify: Boolean;
+    procedure BeforeAddOrModifyFileScenarioCheck(Scenario: Enum "File Scenario"; Connector: Enum "Ext. File Storage Connector"): Boolean;
     var
         ExternalStorageSetup: Record "DA External Storage Setup";
         FileAccount: Record "File Account";
@@ -48,16 +48,15 @@ codeunit 8751 "DA External Storage Impl." implements "File Scenario"
                 if ExternalStorageSetup.Enabled then begin
                     ExternalStorageSetup.CalcFields("Has Uploaded Files");
                     if ExternalStorageSetup."Has Uploaded Files" then begin
-                        SkipInsertOrModify := true;
                         Message(CannotReassignScenarioErr);
-                        exit;
+                        exit(true);
                     end;
                 end;
 
         // Open setup page for additional configuration automatically
         if ConfirmManagement.GetResponseOrDefault(ConfigureExternalStorageQst, true) then
             ExternalStorageSetupPage.Run();
-        SkipInsertOrModify := false;
+        exit(false);
     end;
 
     /// <summary>
@@ -66,7 +65,7 @@ codeunit 8751 "DA External Storage Impl." implements "File Scenario"
     /// <param name="Scenario">The ID of the file scenario.</param>
     /// <param name="Connector">The file storage connector.</param>
     /// <returns>True if additional setup is available, otherwise false.</returns>
-    procedure GetAdditionalScenarioSetup(Scenario: Enum "File Scenario"; Connector: Enum System.ExternalFileStorage."Ext. File Storage Connector") SetupExist: Boolean;
+    procedure GetAdditionalScenarioSetup(Scenario: Enum "File Scenario"; Connector: Enum "Ext. File Storage Connector"): Boolean;
     var
         ExternalStorageSetup: Page "DA External Storage Setup";
     begin
@@ -74,7 +73,7 @@ codeunit 8751 "DA External Storage Impl." implements "File Scenario"
             exit;
 
         ExternalStorageSetup.RunModal();
-        SetupExist := true;
+        exit(true);
     end;
 
     /// <summary>
@@ -83,7 +82,7 @@ codeunit 8751 "DA External Storage Impl." implements "File Scenario"
     /// <param name="Scenario">The ID of the file scenario.</param>
     /// <param name="Connector">The file storage connector.</param>
     /// <returns>True if the delete operation is handled and should not proceed, otherwise false.</returns>
-    procedure BeforeDeleteFileScenarioCheck(Scenario: Enum "File Scenario"; Connector: Enum System.ExternalFileStorage."Ext. File Storage Connector") SkipDelete: Boolean;
+    procedure BeforeDeleteFileScenarioCheck(Scenario: Enum "File Scenario"; Connector: Enum "Ext. File Storage Connector"): Boolean;
     var
         ExternalStorageSetup: Record "DA External Storage Setup";
         NotPossibleToUnassignScenarioMsg: Label 'External Storage scenario can not be unassigned when there are uploaded files.';
@@ -98,16 +97,16 @@ codeunit 8751 "DA External Storage Impl." implements "File Scenario"
         if not ExternalStorageSetup."Has Uploaded Files" then
             exit;
 
-        SkipDelete := true;
         Message(NotPossibleToUnassignScenarioMsg);
+        exit(true);
     end;
 
-    procedure BeforeReassignFileScenarioCheck(CurrentScenario: Enum "File Scenario") SkipReassign: Boolean
+    procedure BeforeReassignFileScenarioCheck(Scenario: Enum "File Scenario"): Boolean
     var
         ExternalStorageSetup: Record "DA External Storage Setup";
         NotPossibleToReassignScenarioMsg: Label 'External Storage scenario can not be reassigned when there are uploaded files.';
     begin
-        if not (CurrentScenario = Enum::"File Scenario"::"Doc. Attach. - External Storage") then
+        if not (Scenario = Enum::"File Scenario"::"Doc. Attach. - External Storage") then
             exit;
 
         if not ExternalStorageSetup.Get() then
@@ -117,8 +116,8 @@ codeunit 8751 "DA External Storage Impl." implements "File Scenario"
         if not ExternalStorageSetup."Has Uploaded Files" then
             exit;
 
-        SkipReassign := true;
         Message(NotPossibleToReassignScenarioMsg);
+        exit(true);
     end;
     #endregion
 
@@ -137,6 +136,7 @@ codeunit 8751 "DA External Storage Impl." implements "File Scenario"
         FileAccount: Record "File Account";
         ExternalFileStorage: Codeunit "External File Storage";
         FileScenarioCU: Codeunit "File Scenario";
+        DAFeatureTelemetry: Codeunit "DA Feature Telemetry";
         TempBlob: Codeunit "Temp Blob";
         FileScenario: Enum "File Scenario";
         InStream: InStream;
@@ -156,7 +156,7 @@ codeunit 8751 "DA External Storage Impl." implements "File Scenario"
             exit(false);
 
         // Telemetry logging for feature usage
-        LogFeatureUsedTelemetry();
+        DAFeatureTelemetry.LogFeatureUsed();
 
         // Get file content from document attachment
         TempBlob.CreateOutStream(OutStream);
@@ -179,7 +179,7 @@ codeunit 8751 "DA External Storage Impl." implements "File Scenario"
             DocumentAttachment."External File Path" := FileName;
             DocumentAttachment."Source Environment Hash" := GetCurrentEnvironmentHash();
             DocumentAttachment.Modify();
-            LogFileUploadedTelemetry(DocumentAttachment);
+            DAFeatureTelemetry.LogFileUploaded(DocumentAttachment);
             exit(true);
         end;
 
@@ -196,6 +196,7 @@ codeunit 8751 "DA External Storage Impl." implements "File Scenario"
         FileAccount: Record "File Account";
         ExternalFileStorage: Codeunit "External File Storage";
         FileScenarioCU: Codeunit "File Scenario";
+        DAFeatureTelemetry: Codeunit "DA Feature Telemetry";
         FileScenario: Enum "File Scenario";
         InStream: InStream;
         ExternalFilePath, FileName : Text;
@@ -225,7 +226,7 @@ codeunit 8751 "DA External Storage Impl." implements "File Scenario"
         ExternalFileStorage.GetFile(ExternalFilePath, InStream);
 
         if DownloadFromStream(InStream, '', '', '', FileName) then begin
-            LogFileDownloadedTelemetry(DocumentAttachment);
+            DAFeatureTelemetry.LogFileDownloaded(DocumentAttachment);
             exit(true);
         end;
 
@@ -367,6 +368,7 @@ codeunit 8751 "DA External Storage Impl." implements "File Scenario"
         FileAccount: Record "File Account";
         ExternalFileStorage: Codeunit "External File Storage";
         FileScenarioCU: Codeunit "File Scenario";
+        DAFeatureTelemetry: Codeunit "DA Feature Telemetry";
         FileScenario: Enum "File Scenario";
         ExternalFilePath: Text;
     begin
@@ -405,7 +407,7 @@ codeunit 8751 "DA External Storage Impl." implements "File Scenario"
         ExternalFileStorage.Initialize(FileScenario);
         if ExternalFileStorage.DeleteFile(ExternalFilePath) then begin
             DocumentAttachment.MarkAsNotUploadedToExternal();
-            LogFileDeletedTelemetry(DocumentAttachment);
+            DAFeatureTelemetry.LogFileDeleted(DocumentAttachment);
             exit(true);
         end;
 
@@ -511,186 +513,6 @@ codeunit 8751 "DA External Storage Impl." implements "File Scenario"
     end;
     #endregion
 
-    #region Document Attachment Handling
-    /// <summary>
-    /// Handles automatic upload of new document attachments to external storage upon insertion of the attachment record.
-    /// </summary>
-    /// <param name="Rec">The document attachment record.</param>
-    /// <param name="RunTrigger">Indicates if the trigger should run.</param>
-    [EventSubscriber(ObjectType::Table, Database::"Document Attachment", OnAfterInsertEvent, '', true, true)]
-    local procedure OnAfterInsertDocumentAttachment(var Rec: Record "Document Attachment"; RunTrigger: Boolean)
-    var
-        ExternalStorageSetup: Record "DA External Storage Setup";
-        ExternalStorageImpl: Codeunit "DA External Storage Impl.";
-    begin
-        // Exit early if trigger is not running
-        if not RunTrigger then
-            exit;
-
-        // Temporary records are not processed
-        if Rec.IsTemporary() then
-            exit;
-
-        // Check if auto upload is enabled
-        if not ExternalStorageSetup.Get() then
-            exit;
-
-        // Only process files with actual content
-        if not Rec."Document Reference ID".HasValue() then
-            exit;
-
-        // Upload to external storage
-        if not ExternalStorageImpl.UploadToExternalStorage(Rec) then
-            exit;
-
-        ExternalStorageImpl.DeleteFromInternalStorage(Rec);
-    end;
-
-    /// <summary>
-    /// Handles automatic deletion of document attachments from external storage upon deletion of the attachment record.
-    /// </summary>
-    /// <param name="Rec">The document attachment record.</param>
-    /// <param name="RunTrigger">Indicates if the trigger should run.</param>
-    [EventSubscriber(ObjectType::Table, Database::"Document Attachment", OnAfterDeleteEvent, '', true, true)]
-    local procedure OnAfterDeleteDocumentAttachment(var Rec: Record "Document Attachment"; RunTrigger: Boolean)
-    var
-        ExternalStorageSetup: Record "DA External Storage Setup";
-        ExternalStorageImpl: Codeunit "DA External Storage Impl.";
-    begin
-        // Exit early if trigger is not running
-        if not RunTrigger then
-            exit;
-
-        // Temporary records are not processed
-        if Rec.IsTemporary() then
-            exit;
-
-        // Check if auto upload is enabled
-        if not ExternalStorageSetup.Get() then
-            exit;
-
-        if not ExternalStorageSetup."Delete from External Storage" then
-            exit;
-
-        // Only process files that were uploaded to external storage
-        if not Rec."Stored Externally" then
-            exit;
-
-        // Delete from external storage
-        ExternalStorageImpl.DeleteFromExternalStorage(Rec);
-
-        if Rec."Skip Delete On Copy" then begin
-            Rec."Skip Delete On Copy" := false;
-            Rec.Modify();
-        end;
-    end;
-
-    /// <summary>
-    /// Handles exporting document attachment content to a stream for externally stored attachments.
-    /// </summary>
-    /// <param name="DocumentAttachment">The document attachment record.</param>
-    /// <param name="AttachmentOutStream">The output stream for the attachment content.</param>
-    /// <param name="IsHandled">Indicates if the event has been handled.</param>
-    [EventSubscriber(ObjectType::Table, Database::"Document Attachment", OnBeforeExportToStream, '', false, false)]
-    local procedure DocumentAttachment_OnBeforeExportToStream(var DocumentAttachment: Record "Document Attachment"; var AttachmentOutStream: OutStream; var IsHandled: Boolean)
-    var
-        ExternalStorageImpl: Codeunit "DA External Storage Impl.";
-    begin
-        // Only handle if file is uploaded externally and not available internally
-        if not ExternalStorageImpl.IsFileUploadedToExternalStorageAndDeletedInternally(DocumentAttachment) then
-            exit;
-
-        ExternalStorageImpl.DownloadFromExternalStorageToStream(DocumentAttachment."External File Path", AttachmentOutStream);
-        IsHandled := true;
-    end;
-
-    /// <summary>
-    /// Handles getting the temporary blob for externally stored document attachments.
-    /// </summary>
-    /// <param name="DocumentAttachment">The document attachment record.</param>
-    /// <param name="TempBlob">The temporary blob to be filled.</param>
-    /// <param name="IsHandled">Indicates if the event has been handled.</param>
-    [EventSubscriber(ObjectType::Table, Database::"Document Attachment", OnBeforeGetAsTempBlob, '', false, false)]
-    local procedure DocumentAttachment_OnBeforeGetAsTempBlob(var DocumentAttachment: Record "Document Attachment"; var TempBlob: Codeunit "Temp Blob"; var IsHandled: Boolean)
-    var
-        ExternalStorageImpl: Codeunit "DA External Storage Impl.";
-    begin
-        // Only handle if file is uploaded externally and not available internally
-        if not ExternalStorageImpl.IsFileUploadedToExternalStorageAndDeletedInternally(DocumentAttachment) then
-            exit;
-
-        ExternalStorageImpl.DownloadFromExternalStorageToTempBlob(DocumentAttachment."External File Path", TempBlob);
-        IsHandled := true;
-    end;
-
-    /// <summary>
-    /// Handles getting content type for externally stored document attachments.
-    /// </summary>
-    /// <param name="Rec">The document attachment record.</param>
-    /// <param name="ContentType">The content type to be set.</param>
-    /// <param name="IsHandled">Indicates if the event has been handled.</param>
-    [EventSubscriber(ObjectType::Table, Database::"Document Attachment", OnBeforeGetContentType, '', false, false)]
-    local procedure DocumentAttachment_OnBeforeGetContentType(var Rec: Record "Document Attachment"; var ContentType: Text[100]; var IsHandled: Boolean)
-    var
-        ExternalStorageImpl: Codeunit "DA External Storage Impl.";
-    begin
-        // Only handle if file is uploaded externally and not available internally
-        if not ExternalStorageImpl.IsFileUploadedToExternalStorageAndDeletedInternally(Rec) then
-            exit;
-
-        ExternalStorageImpl.FileExtensionToContentMimeType(Rec, ContentType);
-        IsHandled := true;
-    end;
-
-    /// <summary>
-    /// Handles checking if attachment content is available for externally stored document attachments.
-    /// </summary>
-    /// <param name="DocumentAttachment">The document attachment record.</param>
-    /// <param name="AttachmentIsAvailable">Indicates if the attachment is available.</param>
-    /// <param name="IsHandled">Indicates if the event has been handled.</param>
-    [EventSubscriber(ObjectType::Table, Database::"Document Attachment", OnBeforeHasContent, '', false, false)]
-    local procedure DocumentAttachment_OnBeforeHasContent(var DocumentAttachment: Record "Document Attachment"; var AttachmentIsAvailable: Boolean; var IsHandled: Boolean)
-    var
-        ExternalStorageImpl: Codeunit "DA External Storage Impl.";
-    begin
-        // Only handle if file is uploaded externally and not available internally
-        if not ExternalStorageImpl.IsFileUploadedToExternalStorageAndDeletedInternally(DocumentAttachment) then
-            exit;
-
-        AttachmentIsAvailable := ExternalStorageImpl.CheckIfFileExistInExternalStorage(DocumentAttachment."External File Path");
-        IsHandled := true;
-    end;
-
-    /// <summary>
-    /// Event handler for before checking Document Reference ID on insert.
-    /// </summary>
-    /// <param name="DocumentAttachment">The document attachment record.</param>
-    /// <param name="IsHandled">Indicates if the event has been handled.</param>
-    [EventSubscriber(ObjectType::Table, Database::"Document Attachment", OnInsertOnBeforeCheckDocRefID, '', false, false)]
-    local procedure "Document Attachment_OnInsertOnBeforeCheckDocRefID"(var DocumentAttachment: Record "Document Attachment"; var IsHandled: Boolean)
-    begin
-        if DocumentAttachment."Stored Externally" then
-            IsHandled := true;
-    end;
-
-    [EventSubscriber(ObjectType::Codeunit, Codeunit::"Document Attachment Mgmt", OnCopyAttachmentsOnAfterSetToDocumentFilters, '', false, false)]
-    local procedure "Document Attachment Mgmt_OnCopyAttachmentsOnAfterSetToDocumentFilters"(var ToDocumentAttachment: Record "Document Attachment"; ToRecRef: RecordRef; ToAttachmentDocumentType: Enum "Attachment Document Type"; ToNo: Code[20]; ToLineNo: Integer)
-    begin
-        ToDocumentAttachment."Skip Delete On Copy" := ToDocumentAttachment."Stored Externally";
-    end;
-
-    [EventSubscriber(ObjectType::Table, Database::"Document Attachment", OnBeforeOpenInOneDrive, '', false, false)]
-    local procedure "Document Attachment_OnBeforeOpenInOneDrive"(var Rec: Record "Document Attachment"; var IsHandled: Boolean)
-    var
-        NotSupportedMsg: Label 'Opening Document Attachments stored in External Storage via OneDrive is not supported.';
-    begin
-        if Rec."Stored Externally" then begin
-            IsHandled := true;
-            Message(NotSupportedMsg);
-        end;
-    end;
-    #endregion
-
     /// <summary>
     /// Opens a folder selection dialog for choosing the root folder.
     /// </summary>
@@ -720,9 +542,10 @@ codeunit 8751 "DA External Storage Impl." implements "File Scenario"
         TableNameFolder: Text[100];
         EnvironmentHashFolder: Text[32];
         FileNamePart: Text;
+        FileNameFormatLbl: Label '%1-%2.%3', Comment = '%1 = File Name, %2 = GUID, %3 = File Extension', Locked = true;
     begin
         // Generate unique filename to prevent collisions
-        FileNamePart := DocumentAttachment."File Name" + '-' + DelChr(Format(CreateGuid()), '=', '{}') + '.' + DocumentAttachment."File Extension";
+        FileNamePart := StrSubstNo(FileNameFormatLbl, DocumentAttachment."File Name", DelChr(Format(CreateGuid()), '=', '{}'), DocumentAttachment."File Extension");
 
         // Get table name folder (based on the source table of the attachment)
         TableNameFolder := GetTableNameFolder(DocumentAttachment."Table ID");
@@ -907,6 +730,7 @@ codeunit 8751 "DA External Storage Impl." implements "File Scenario"
     procedure RunCompanyMigration(): Integer
     var
         DocumentAttachment: Record "Document Attachment";
+        DAFeatureTelemetry: Codeunit "DA Feature Telemetry";
         MigratedCount: Integer;
         StartMigrationQst: Label 'This will migrate all document attachments from the previous company folder to the current company folder.\\Do you want to continue?';
         MigrationCompletedMsg: Label '%1 file(s) have been successfully migrated to the current company folder.', Comment = '%1 = Number of files';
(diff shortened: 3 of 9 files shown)


StefanSosic:
> @StefanSosic could maybe elaborate a little bit on your events. Personally I'm not a real fan of the IsHandled Events.
> 
> Couldn't we solve this by adding an well designed interface to the FileScenario.. We could have DefaultImplementation that does nothing. If an enum value implements this interface then we would be able to Call Methods for the relevant enum value.
> 
> Maybe something like the following
> 
> ```al
> interface FileScenario
> 
> procedure HasAdditionalSetup(): Boolean
> procedure OpenAdditionalSetup(TBDParameters)
> procedure CheckScenarioBeforeDeletion()
> ...
> ```
> 
> Maybe you could show some example code how you would like to subscribe to this in the Microsoft BaseApp. This could help finding a good design for the System Application.

The events created are providing possibility for additional control, which is not mandatory. That's why it's by design with events, not interface. This module is specific and we wanted to put disclaimer message before adding scenario and so on, but I think lot of scenarios won't use that feature. 

Setup action is using event as discovery method. Again, not mandatory to have and implement setup, that's why didn't went with interfaces. I did had a call to align with the concept, and events were mentioned.

Let me try to bring more context to it, implementation of File Accounts with Scenarios shouldn't be limited to Base App and internal development, but give partners possibility to adjust and work on their own scenarios.

GetAdditionalScenarioSetup - Discovery event if there is additional setup for a scenario, must not be existing
BeforeDeleteFileScenarioCheck - Is there any restriction when deleting scenario, additional checks, in this case if files are already uploaded you can't unassign until all files are back to internal storage
BeforeAddOrModifyFileScenarioCheck - When adding scenario, additional checks or so, in this case disclaimer message

For scenarios, we already have default one. In this case, this are optional features, which I see more that they won't end being used then they will, if they will it will be implemented one of these three for example. Don't want to implement interface and to use just one of the procedures, rest needs to be handled unnecessary. But I got your point, interface is nice and good, but just for this case I don't see it fitting, just because of type of procedures to be used. My scenario is for Document Attachments, but it could be just anything, Email, Sms, Backup, anything, in those case this interface procedure wouldn't make any sense. We need to look for a full picture and wide spectrum of implementation of File Scenarios and that could be anything. 

Module is generally done and PR's created, Jesper will take it for a spin on Monday.
Here are PR's:
System Application adjustments for External Storage Module:
[Extensibility adjustments to File Account module in order for External Storage module by StefanSosic · Pull Request #4495 · microsoft/BCApps](https://github.com/microsoft/BCApps/pull/4495)
 
Base Application adjustments for External Storage Module:
[Extensibility adjustments to File Account module in order for External Storage module by StefanSosic · Pull Request #1538 · microsoft/BusinessCentralApps](https://github.com/microsoft/BusinessCentralApps/pull/1538)
 
AL App Extension - Full Module:
[External Storage - Document Attachments by StefanSosic · Pull Request #29087 · microsoft/ALAppExtensions](https://github.com/microsoft/ALAppExtensions/pull/29087)

I am open for suggestions, keep in mind that this is first version, this is just handful of functionality, no need to introduce more and more, since we plan to push it to preview asap.

And as you can see, point is to provide this as separate extension, and this should be also an demo for partners, how to implement their own File Scenario for any need.

src/Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DAExternalStorageImpl.Codeunit.al
```diff
@@ -0,0 +1,74 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+codeunit 8751 "DA External Storage Impl." implements "File Scenario"
+{
+    Permissions = tabledata "DA External Storage Setup" = r;
```
pri-kise: this codeunit should be internal

StefanSosic:
@TKapitan @pri-kise @tonyabriccomeske 
Hi, we made a decision to go with interfaces. It's implemented at this point

src/System Application/App/External File Storage/src/Scenario/FileScenario.Interface.al
```diff
@@ -0,0 +1,31 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+interface "File Scenario"
+{
```
pri-kise, line 9: why are we providing the scenario as Integer and not the scenario enum in those methods?

src/Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DocumentAttachmentExtStor.TableExt.al
```diff
@@ -0,0 +1,78 @@
...
+        {
+            Caption = 'Uploaded Externally';
+            DataClassification = CustomerContent;
+            Editable = false;
+            ToolTip = 'Specifies if the file has been uploaded to external storage.';
+        }
+        field(8751; "External Upload Date"; DateTime)
+        {
+            Caption = 'External Upload Date';
+            DataClassification = CustomerContent;
```
pri-kise: Some fields here could have the DataClassification `SystemMetadata` according to the documentation.
(I would keep CustomerContent for the field `External File Path`)
https://github.com/MicrosoftDocs/dynamics365smb-devitpro-pb/blob/main/dev-itpro/developer/includes/data_classifications.md

src/System Application/App/app.json
```diff
@@ -4,7 +4,7 @@
   "publisher": "Microsoft",
   "brief": "Provides a standard set of capabilities that serve as a foundation for developing business apps.",
   "description": "Contains an expansive set of open source modules that make it easier to build, maintain, and easily upgrade on-premises and online apps. These modules let you focus on the business logic, and the needs of your users or customers.",
-  "version": "27.0.0.0",
+  "version": "27.0.90003.0",
```
pri-kise: This change should be reverted

src/Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DAExternalStorageImpl.Codeunit.al
StefanSosic: Changed, thanks!

src/System Application/App/External File Storage/src/Scenario/FileScenario.Interface.al
StefanSosic: Good question! 😂 I did it with Scenario field which is integer, will check

src/Apps/W1/External Storage - Document Attachments/app/src/AutomaticSync/DAExternalStorageSync.Report.al
```diff
@@ -0,0 +1,174 @@
...
+            {
+                group(General)
+                {
+                    Caption = 'General';
+                    field(SyncDirectionField; SyncDirection)
+                    {
+                        ApplicationArea = Basic, Suite;
+                        Caption = 'Sync Direction';
+                        OptionCaption = 'To External Storage,From External Storage';
+                        ToolTip = 'Specifies whether to sync to external storage, from external storage, or delete expired files.';
```
pri-kise: This ToolTip implies three values, but I can only see the in the OptionCaption?
Is there something missing?

src/Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DAExternalStorageImpl.Codeunit.al
```diff
@@ -0,0 +1,80 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace Microsoft.ExternalStorage.DocumentAttachments;
+
+using System.ExternalFileStorage;
+using System.Utilities;
+
+codeunit 8751 "DA External Storage Impl." implements "File Scenario"
+{
```
pri-kise, line 15: Implementation Codeunits are normally with InherentPermissions + InherentEntitlements.

pri-kise (changes requested):
We might need to condsider to add extensions for some Entitlements.

src/Apps/W1/External Storage - Document Attachments/app/src/AutomaticSync/DAExternalStorageSync.Report.al
StefanSosic: Yep, not good. It's now separate boolean for delete expired files. Will adjust the tooltip

src/Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DAExternalStorageImpl.Codeunit.al
StefanSosic: Added

src/System Application/App/External File Storage/src/Scenario/FileScenario.Interface.al
StefanSosic: File Account Scenario has Scenario which is integer:
image

So, all good here. I won't convert at this point integer back to an enum, back and forth, it would, in my opinion, just add unnecessary complexity to it

StefanSosic:
> After I worked with the E-Mail Scenario again today (The file scenario is inspired on the email scenario), I revisited this PR again today.
> 
> I understand what you're trying to achive with this change on the file account scenario, but I'm not sure if it's the right way.
> 
> e.g. when I want to get a file account for a scenario and no file account is setup for this scenario then I get the default file account. https://github.com/microsoft/BCApps/blob/main/src/System%20Application/App/External%20File%20Storage/src/Scenario/FileScenarioImpl.Codeunit.al
> 
> In the codeunit `"DA External Storage Processor"` the file account is retrieved with the scenario, but you could get the default file account.
> 
> I'm not sure if the current logic is really fail-save and I'm unsure if we're able to ensure that a customer won't delete a scenario or a file account for document attachments that are uploaded to this external storage.

I see the issue.
Well procedure GetFileAccount should be split into two,
1. GetFileAccount - get's onlt that file account
2. GetFileAccountOrDefault - get that file  account or default

Anything against this approach? @pri-kise

src/Apps/W1/External Storage - Document Attachments/app/src/Permissions/DAExtStorView.PermissionSet.al
```diff
@@ -0,0 +1,22 @@
...
+/// <summary>
+/// Permission set for External Storage functionality.
+/// Grants necessary permissions to use external storage features.
+/// </summary>
+permissionset 8750 "DA Ext. Stor. View"
+{
+    Assignable = true;
+    Caption = 'DA - External Storage View';
+    Permissions = tabledata "DA External Storage Setup" = R,
+        table "DA External Storage Setup" = X,
```
grobyns: You should create a separate permissionset for the Objects (execute permission) and include that in this permissionset

src/Apps/W1/External Storage - Document Attachments/app/src/Permissions/DAExtStorAdmin.PermissionSet.al
```diff
@@ -0,0 +1,24 @@
...
+/// <summary>
+/// Permission set for External Storage functionality.
+/// Grants necessary permissions to use external storage features.
+/// </summary>
+permissionset 8751 "DA Ext. Stor. Admin"
+{
+    Assignable = true;
+    Caption = 'DA - External Storage Admin';
+    Permissions = tabledata "DA External Storage Setup" = RIMD,
+        table "DA External Storage Setup" = X,
```
grobyns: instead of repeating permissions, you should include the View permission set and only list the additional permissions here

src/Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DAExternalStorageImpl.Codeunit.al
```diff
@@ -0,0 +1,82 @@
...
+    /// <summary>
+    /// Called before adding or modifying a file scenario.
+    /// </summary>
+    /// <param name="Scenario">The ID of the file scenario.</param>
+    /// <param name="Connector">The file storage connector.</param>
+    /// <returns>True if the operation is allowed, otherwise false.</returns>
+    procedure BeforeAddOrModifyFileScenarioCheck(Scenario: Integer; Connector: Enum System.ExternalFileStorage."Ext. File Storage Connector") SkipInsertOrModify: Boolean
+    var
+        ConfirmManagement: Codeunit "Confirm Management";
+        DisclaimerPart1Lbl: Label 'You are about to enable External Storage!!!';
```
grobyns: splitting text like this is a nightmare to translate. Please make 1 label.

src/Apps/W1/External Storage - Document Attachments/app/src/AutomaticSync/DAExternalStorageSync.Report.al
```diff
@@ -0,0 +1,174 @@
...
+                            if not DocumentAttachment."Document Reference ID".HasValue() then
+                                DocumentAttachment.Mark(false)
+                            else
+                                DocumentAttachment.Mark(true);
+                        until DocumentAttachment.Next() = 0;
+                        DocumentAttachment.MarkedOnly(true);
+                    end;
+                end;
+            SyncDirection::"From External Storage":
+
```
grobyns: why the blank line?

src/Apps/W1/External Storage - Document Attachments/app/src/AutomaticSync/DAExternalStorageSync.Report.al
```diff
@@ -0,0 +1,174 @@
...
+                        if not ExternalStorageProcessor.UploadToExternalStorage(DocumentAttachment) then
+                            FailedCount += 1;
+                    SyncDirection::"From External Storage":
+
+                        if not ExternalStorageProcessor.DownloadFromExternalStorage(DocumentAttachment) then
+                            FailedCount += 1;
+                end;
+                if DeleteExpiredFiles then
+                    if CalcDate('<+' + GetDateFormulaFromExternalStorageSetup() + '>', DocumentAttachment."External Upload Date".Date()) >= Today() then
```
grobyns: you have the formula split into 3 parts. why?
in the procedure, you get the int, concatenate with D, then here, concatenate again with <>.

Why not have the procedure simply return the number of Days as int, and then do CalcDate(StrSubstNo(DateFormulaLbl, NoOfDays) etc...

StefanSosic: Very true :)

src/Apps/W1/External Storage - Document Attachments/app/src/Permissions/DAExtStorAdmin.PermissionSet.al
```diff
@@ -0,0 +1,17 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+namespace Microsoft.ExternalStorage.DocumentAttachments;
+
+/// <summary>
+/// Permission set for External Storage functionality.
+/// Grants necessary permissions to use external storage features.
+/// </summary>
+permissionset 8751 "DA Ext. Stor. Admin"
```
grobyns, line 11: Was discussing the permissions with @JesperSchulz and we think it should be slightly different.
Basically, all permissions should be inherent, except for the tabledata "DA External Storage Setup" which only an admin should be able to RIMD.
Everyone else still needs little r, so that should be inherent and then you would need to also add little r in code wherever the table is read.

src/Apps/W1/External Storage - Document Attachments/app/src/DAExtStorageDeleteAfter.Enum.al
```diff
@@ -0,0 +1,31 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace Microsoft.ExternalStorage.DocumentAttachments;
+
+/// <summary>
+/// Defines when attachments should be deleted from internal storage after upload to external storage.
+/// </summary>
+enum 8750 "DA Ext. Storage - Delete After"
```
JesperSchulz: We don't want this. The "Delete After" should be a freely configurable dateformula.

src/Apps/W1/External Storage - Document Attachments/app/src/AutomaticSync/DAExternalStorageSync.Report.al
```diff
@@ -0,0 +1,173 @@
...
+    local procedure SetFilters()
+    begin
+        case SyncDirection of
+            SyncDirection::"To External Storage":
+                begin
+                    DocumentAttachment.SetRange("Uploaded Externally", false);
+                    if DocumentAttachment.FindSet() then begin
+                        repeat
+                            if not DocumentAttachment."Document Reference ID".HasValue() then
+                                DocumentAttachment.Mark(false)
```
grobyns: Marks get very slow after +- 200 rows. Maybe Marks is not the best solution for this...

src/Apps/W1/External Storage - Document Attachments/app/src/DAExternalStorageProcessor.Codeunit.al
```diff
@@ -0,0 +1,381 @@
...
+    Permissions = tabledata "Tenant Media" = rimd,
+                  tabledata "Document Attachment" = rimd,
+                  tabledata "DA External Storage Setup" = r;
+
+    /// <summary>
+    /// Uploads a document attachment to external storage.
+    /// </summary>
+    /// <param name="DocumentAttachment">The document attachment record to upload.</param>
+    /// <returns>True if upload was successful, false otherwise.</returns>
+    internal procedure UploadToExternalStorage(var DocumentAttachment: Record "Document Attachment"): Boolean
```
grobyns: we don't need double internals...

src/Apps/W1/External Storage - Document Attachments/app/src/DAExternalStorageProcessor.Codeunit.al
```diff
@@ -0,0 +1,381 @@
...
+using Microsoft.Foundation.Attachment;
+using System.Environment;
+using System.ExternalFileStorage;
+using System.Utilities;
+
+/// <summary>
+/// Provides functionality to manage document attachments in external storage systems.
+/// Handles upload, download, and deletion operations for Business Central attachments.
+/// </summary>
+codeunit 8750 "DA External Storage Processor"
```
grobyns: This looks like an implementation codeunit (it's internal, has all the logic) so please name it as such... (add Impl. to the name)

src/Apps/W1/External Storage - Document Attachments/app/src/DAExternalStorageProcessor.Codeunit.al
```diff
@@ -0,0 +1,381 @@
...
+
+        exit(ExternalStorageSetup."Delete After" = ExternalStorageSetup."Delete After"::Immediately);
+    end;
+
+    /// <summary>
+    /// Maps file extensions to their corresponding MIME types.
+    /// </summary>
+    /// <param name="Rec">The document attachment record.</param>
+    /// <param name="ContentType">The content type to set based on the file extension.</param>
+    internal procedure FileExtensionToContentMimeType(var Rec: Record "Document Attachment"; var ContentType: Text[100])
```
grobyns: this looks very similar to:
https://github.com/microsoft/BCApps/blob/9fd60bad4445a8a694fbf3c575e454307dd9f165/src/System%20Application/App/Email/src/Message/EmailMessageImpl.Codeunit.al#L231

which has already been copy pasted elsewhere... It's time for someone to extract it and provide a public method for this metdata...

src/Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DAExternalStorageSubs.Codeunit.al
```diff
@@ -0,0 +1,162 @@
...
+namespace Microsoft.ExternalStorage.DocumentAttachments;
+
+using Microsoft.Foundation.Attachment;
+using System.Utilities;
+
+/// <summary>
+/// Event subscribers for External Storage functionality.
+/// Handles automatic upload of new attachments and cleanup operations.
+/// </summary>
+codeunit 8752 "DA External Storage Subs."
```
grobyns: i would not have this codeunit. Just add the subscribers in "DA External Storage Impl." codeunit

src/Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DAExternalStorageSubs.Codeunit.al
```diff
@@ -0,0 +1,162 @@
...
+    /// </summary>
+    /// <param name="Rec">The document attachment record.</param>
+    /// <param name="RunTrigger">Indicates if the trigger should run.</param>
+    [EventSubscriber(ObjectType::Table, Database::"Document Attachment", OnAfterInsertEvent, '', true, true)]
+    local procedure OnAfterInsertDocumentAttachment(var Rec: Record "Document Attachment"; RunTrigger: Boolean)
+    var
+        ExternalStorageSetup: Record "DA External Storage Setup";
+        ExternalStorageProcessor: Codeunit "DA External Storage Processor";
+    begin
+        // Exit early if trigger is not running
```
grobyns: should the code here check for IsTemporary?

src/Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DocumentAttachmentExtStor.TableExt.al
```diff
@@ -0,0 +1,82 @@
...
+            Editable = false;
+            ToolTip = 'Specifies the value of the Deleted Internally field.';
+        }
+    }
+
+    /// <summary>
+    /// Marks the document attachment as uploaded to external storage.
+    /// </summary>
+    /// <param name="ExternalFilePath">The path to the file in external storage.</param>
+    internal procedure MarkAsUploadedToExternal(ExternalFilePath: Text[2048])
```
grobyns: is it really worth having a helper method when it is only called from 1 place...

src/Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DocumentAttachmentExternal.Page.al
```diff
@@ -0,0 +1,204 @@
...
+                {
+                    ToolTip = 'Specifies the date the file was attached.';
+                }
+                field("Attached By"; Rec."Attached By")
+                {
+                    ToolTip = 'Specifies the user who attached the file.';
+                }
+                field("Deleted Internally"; Rec."Deleted Internally")
+                {
+                }
```
grobyns, line 69: everything before this control has a Tooltip, everything after has a Caption... this one has neither... ? weird.

src/System Application/App/External File Storage/src/Scenario/FileScenario.Interface.al
```diff
@@ -0,0 +1,33 @@
...
+    /// <summary>
+    /// Called before adding or modifying a file scenario.
+    /// </summary>
+    /// <param name="Scenario">The ID of the file scenario.</param>
+    /// <param name="Connector">The file storage connector.</param>
+    /// <returns>True if the operation is allowed; otherwise false.</returns>
+    procedure BeforeAddOrModifyFileScenarioCheck(Scenario: Integer; Connector: Enum System.ExternalFileStorage."Ext. File Storage Connector") SkipInsertOrModify: Boolean;
```
grobyns: the naming sounds odd... it sounds more like events than like interface methods... maybe we can come up with something better.

src/Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DocumentAttachmentExternal.Page.al
pri-kise: For the ToolTip
We can keep the ToolTip on the TableExtension for new fields, but the other fields don't have a ToolTip specified on the table level yet.
Therefore the ToolTips needs to be duplicated from the original page in the BaseApp

src/Apps/W1/External Storage - Document Attachments/app/src/AutomaticSync/DAExternalStorageSync.Report.al
pri-kise: Do we really need those mark?
Normally all Document Attachments should have a value in the MediaSet field.
Couldn't we simple skip those that don't have a value in the processing?

src/Apps/W1/External Storage - Document Attachments/app/src/AutomaticSync/DAExternalStorageSync.Report.al
```diff
@@ -0,0 +1,175 @@
...
+
+                        if not ExternalStorageProcessor.DownloadFromExternalStorage(DocumentAttachment) then
+                            FailedCount += 1;
+                end;
+                if DeleteExpiredFiles then
+                    if CalcDate(StrSubstNo(DateFormulaLbl, GetDateFormulaFromExternalStorageSetup()), DocumentAttachment."External Upload Date".Date()) >= Today() then
+                        if ExternalStorageProcessor.DeleteFromInternalStorage(DocumentAttachment) then
+                            DeleteCount += 1
+                        else
+                            DeleteFailedCount += 1;
```
pri-kise: Should we maybe have an commit after each record?
Since we communicate to an external system it's very likely that we receive an error in a large batch.
The files will have been moved to the exteranl system, therefore we shouldn't lost the information about the changes on the document attachment.

src/Apps/W1/External Storage - Document Attachments/app/app.json
```diff
@@ -0,0 +1,35 @@
...
+  "publisher": "Microsoft",
+  "version": "27.0.0.0",
+  "brief": "External Storage processor for Business Central document attachments",
+  "description": "Processes document attachments from Business Central and stores them in configured External Storage connectors (Azure Blob, File Share, SharePoint)",
+  "privacyStatement": "https://go.microsoft.com/fwlink/?linkid=724009",
+  "EULA": "https://go.microsoft.com/fwlink/?linkid=2009120",
+  "help": "https://go.microsoft.com/fwlink/?linkid=2134520",
+  "url": "https://go.microsoft.com/fwlink/?linkid=724011",
+  "logo": "ExtensionLogo.png",
+  "application": "27.0.0.0",
```
JesperSchulz: Probably 28 now.

src/Apps/W1/External Storage - Document Attachments/app/src/DAExternalStorageProcessor.Codeunit.al
JesperSchulz: This isn't the mime type. This code doesn't exist anywhere else.

src/Apps/W1/External Storage - Document Attachments/app/src/Permissions/DAExtStorExec.PermissionSet.al
```diff
@@ -0,0 +1,22 @@
...
+// ------------------------------------------------------------------------------------------------
+namespace Microsoft.ExternalStorage.DocumentAttachments;
+
+/// <summary>
+/// Permission set for External Storage functionality.
+/// Grants necessary permissions to use external storage features.
+/// </summary>
+permissionset 8752 "DA Ext. Stor. Exec."
+{
+    Assignable = false;
```
JesperSchulz: Move all these to inherent permissions on object level. No one should need to be able to change these permissions.

TKapitan (changes requested):
We have a similar solution for our customers and I this change is missing one very important logic (that is probably the most complex to implement, especially in base app).

It's the environment and company copy. Once you have attachments in external storage, you must prevent someone from deleting, changing or manipulating with production attachments from sandbox envrionment. On the other hand, you must not block copying the references to newly created environments as it may break implementation processes. 

The other issue that seems not solved are shared references and problem of deleting file in shared reference. By shared reference I mean document attachments that target the same external file (e.g. when you have attachment configured to flow to sales, purchase etc. documents).



I like having this functionality in the base app, but this implementation is quite simple in comparison to what we found as minimum required scope when discussing with our customers. The problems above are quite critical as they impact production data (e.g. the sandbox environments).

I really recommend to review all usecases and copy features before merging this.

 @JesperSchulz  @grobyns @StefanSosic

JesperSchulz:
> We have a similar solution for our customers and I this change is missing one very important logic (that is probably the most complex to implement, especially in base app).
> 
> It's the environment and company copy. Once you have attachments in external storage, you must prevent someone from deleting, changing or manipulating with production attachments from sandbox envrionment. On the other hand, you must not block copying the references to newly created environments as it may break implementation processes.
> 
> The other issue that seems not solved are shared references and problem of deleting file in shared reference. By shared reference I mean document attachments that target the same external file (e.g. when you have attachment configured to flow to sales, purchase etc. documents).
> 
> I like having this functionality in the base app, but this implementation is quite simple in comparison to what we found as minimum required scope when discussing with our customers. The problems above are quite critical as they impact production data (e.g. the sandbox environments).
> 
> I really recommend to review all usecases and copy features before merging this.
> 
> @JesperSchulz @grobyns @StefanSosic

I agree that this iteration 1 is simple. And I also agree that we need to deal with the scenarios you mention! For this first version, I would accept that we ship with limited capabilities - it's not a feature enabled by default (the 1st party app won't even be published to upgraded environments). Generally "External File Connectors" should not get copied to sandboxes. Sandboxes should by definition be cut off the outside world - hence the name. So it makes no sense to copy those external connectors. You can create new ones within sandboxes for testing of course. That of course will break the links to the external documents, which now are on a share you no longer can access. But that's a limitation you will have to agree to, when enabling the feature in this version.

With regards to copying between production companies, I'm a bit more puzzled what to do. As long as the file is stored in the DB, you get a new copy of the attachment. To mimic that behavior, a copy of the file share / files should be made and all the references should get updated. That's again not a iteration 1 thing to do. So I guess we have to pick of two things: should the copy of the company point to the same external document, or should the ties be broken and the document attachments not be part of the copy? I tend towards the latter for this version.

What do you think?

src/Apps/W1/External Storage - Document Attachments/app/src/AutomaticSync/DAExternalStorageSync.Report.al
IceOnly: I see it like pri-kise. EMpty file are not normal. Just skip them.

src/Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DocumentAttachmentExtStor.TableExt.al
StefanSosic: Removed. Done more for future, but yeah :)

src/Apps/W1/External Storage - Document Attachments/app/src/AutomaticSync/DAExternalStorageSync.Report.al
StefanSosic: Makes sense

src/Apps/W1/External Storage - Document Attachments/app/src/Telemetry/DATelemetryLogger.Codeunit.al
```diff
@@ -0,0 +1,30 @@
...
+{
+    Access = Internal;
+
+    procedure LogMessage(EventId: Text; Message: Text; Verbosity: Verbosity; DataClassification: DataClassification; TelemetryScope: TelemetryScope; CustomDimensions: Dictionary of [Text, Text])
+    begin
+        Session.LogMessage(EventId, Message, Verbosity, DataClassification, TelemetryScope, CustomDimensions);
+    end;
+
+    // For the functionality to behave as expected, there must be exactly one implementation of the "Telemetry Logger" interface registered per app publisher
+    [EventSubscriber(ObjectType::Codeunit, Codeunit::"Telemetry Loggers", 'OnRegisterTelemetryLogger', '', true, true)]
```
pri-kise: The system application already provides the telmetry logger for the publisher microsoft.

StefanSosic: Hm, yeah, leftover from v1, deleted

src/Apps/W1/External Storage - Document Attachments/app/src/AutomaticSync/DAExternalStorageSync.Report.al
```diff
@@ -0,0 +1,180 @@
...
+
+                if GuiAllowed() then
+                    Dialog.Update(1, ProcessedCount);
+
+                case SyncDirection of
+                    SyncDirection::"To External Storage":
+                        if not ExternalStorageImpl.UploadToExternalStorage(DocumentAttachment) then
+                            FailedCount += 1;
+                    SyncDirection::"From External Storage":
+
```
nemanja899: For example SharePoint’s API enforces request rate limits and can return 429 (Too Many Requests) or 503 (Server Busy) responses when the threshold is exceeded.
Retry logic should be added  with exponential backoff (e.g., respect the Retry-After header) to prevent sync failures when for example SharePoint connector throttles the requests.

src/Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DAExternalStorageImpl.Codeunit.al
```diff
@@ -0,0 +1,915 @@
...
+        exit(CopyStr(FileName, 1, 2048));
+    end;
+
+    local procedure GetTableNameFolder(TableID: Integer): Text[100]
+    var
+        RecRef: RecordRef;
+        TableName: Text;
+    begin
+        // Open the RecordRef to get table metadata
+        RecRef.Open(TableID, false);
```
fvet: @StefanSosic 

Suppose scenario below:

- 3rd party app is installed
- 3rd party app adds new custom tables with document attachment support
- new records are added in these custom tables
- new document attachments are linked to records in these custom tables
- result : document attachment records with **custom** table id.
- 3rd party app is un-installed
- document attachment records are not deleted

If the DA External Storage app would be used to move the document attachments to an external storage, the app will process these 'orphaned' document attachment records too, but the _record.open_ would probably **fail**, since the table metadata does no longer exist.

Options:
- Tryfunction?
- Check table metadata presence? (slow)
- Fallback to 'tableid' for folder name?

src/Apps/W1/External Storage - Document Attachments/app/src/Setup/DAExternalStorageSetup.Page.al
```diff
@@ -0,0 +1,246 @@
...
+            {
+                Caption = 'Migrate Files';
+                Image = MoveToNextPeriod;
+                ToolTip = 'Migrate all document attachments from the previous environment/company folder to the current environment/company folder.';
+
+                trigger OnAction()
+                var
+                    DAExternalStorageImpl: Codeunit "DA External Storage Impl.";
+                begin
+                    DAExternalStorageImpl.RunCompanyMigration();
```
fvet: @StefanSosic  For long running tasks such as moving a large set of files, wouldn't it be better to launch a background task / schedule a job queue entry instead?

src/Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DAExternalStorageImpl.Codeunit.al
fvet, line 1: @StefanSosic 
There seems to be a **OpenInOneDrive** procedure in Document Attachment, which is using the "Document Reference ID".MediaId() to share/export the file contents to OneDrive.

If the document attachment would be stored externally, I think this feature will be broken, as the Base App expects the Media ID.

Please either fix (add support in Base App), or block with 'not supported error' in _OnBeforeOpenInOneDrive_

src/Apps/W1/External Storage - Document Attachments/app/src/AutomaticSync/DAExternalStorageSync.Report.al
fvet: +1 We encountered a similar issue when moving 5000+ files from internal database storage to SharePoint.

StefanSosic: Agree, but that should be an improvement on the SharePoint connector in first place, to improve handling of uploads. This module just lays on top of those connectors.

nemanja899: I disagree, all connectors have api rate limits. Gave SharePoint example because he has the lowest api limit.

StefanSosic: All connectors should be improved then :)

StefanSosic: All connectors should be improved then :)

src/Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DAExternalStorageImpl.Codeunit.al
```diff
@@ -0,0 +1,915 @@
...
+    /// </summary>
+    /// <param name="Scenario">The ID of the file scenario.</param>
+    /// <param name="Connector">The file storage connector.</param>
+    /// <returns>True if the operation is allowed, otherwise false.</returns>
+    procedure BeforeAddOrModifyFileScenarioCheck(Scenario: Enum "File Scenario"; Connector: Enum System.ExternalFileStorage."Ext. File Storage Connector") SkipInsertOrModify: Boolean;
+    var
+        FileAccount: Record "File Account";
+        ConfirmManagement: Codeunit "Confirm Management";
+        FileScenarioCU: Codeunit "File Scenario";
+        DisclaimerMsg: Label 'You are about to enable External Storage!!!\\This feature is provided as-is, and you use it at your own risk.\Microsoft is not responsible for any issues or data loss that may occur.\\Do you wish to continue?';
```
JesperSchulz:
```suggestion
        DisclaimerMsg: Label 'You are about to enable External Storage.\\When this feature is enabled, files will be stored outside the Business Central service boundary.\Microsoft does not manage, back up, or restore data stored in external storage.\\You are responsible for the configuration, security, compliance, backup, and recovery of all externally stored files.\This feature is provided as-is, and you enable it at your own risk.\\Do you want to continue?';
```

src/Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DAExternalStorageImpl.Codeunit.al
StefanSosic: Addressed in commit: https://github.com/microsoft/BCApps/pull/4495/commits/4f4c2077da9d68607baf60e47a3684083d229774

src/Apps/W1/External Storage - Document Attachments/app/src/Setup/DAExternalStorageSetup.Page.al
StefanSosic: Agree.
Done in:
https://github.com/microsoft/BCApps/pull/4495/commits/f12957b72cd9db0d02fc316459e712d79764ccce

image

I decided to use report and you could use Schedule if you don't want to run immediately

JesperSchulz:
> We should have some test coverage for this new functionality.
> 
> I don't know if we already have a mocking Storage in the System Application, but we could add this here.
> 
> For starters I'm thinking about the following test cases
> 
> ```al
> // [GIVEN] Mocked External Storage (Maybe via a internal table or single instance codeunit)
> // [GIVEN] Valid Setup for Document Attachments Exteranl Storage provided
> // [GIVEN] Document Attachment is created
> // [WHEN] Document Attachment sync runs
> // [THEN] Verify Document Attachment is updated correctly
> ```
> 
> The following test case is an example for one procedure of the document attachment.
> 
> ```al
> // [GIVEN] Mocked External Storage
> // [GIVEN] Valid Setup for Document Attachments Exteranl Storage provided
> // [GIVEN] Document Attachment is created
> // [GIVEN] Document Attachment synced
> // {WHEN] TemplBlob is retrieved via procedure from Document Attachment 
> // [THEN] Verify TempBlob
> ```
> 
> We should have tests for all the procedures that we handle.

Agents to the rescue. Those can be made in minutes.

src/Apps/W1/External Storage - Document Attachments/app/src/AutomaticSync/DAExternalStorageSync.Report.al
```diff
@@ -0,0 +1,179 @@
...
+                            SyncSuccess := ExternalStorageImpl.UploadToExternalStorage(DocumentAttachment);
+                            if SyncSuccess and (Operation = Operation::Move) then begin
+                                DeleteSuccess := ExternalStorageImpl.DeleteFromInternalStorage(DocumentAttachment);
+                                if not DeleteSuccess then
+                                    FailedCount += 1;
+                            end;
+                        end;
+                    SyncDirection::"To Internal Storage":
+                        begin
+                            SyncSuccess := ExternalStorageImpl.DownloadFromExternalStorageToInternal(DocumentAttachment);
```
myp-bcarmo, line 71: Hi @StefanSosic, great piece of work, currently we are looking forward to this feature.
Small question:
From what I am seeing (correct me if I'm wrong), we are not supporting the scenario where for some reason an user directly deletes the file on the external storage and that file was linked to a document attachment.

In that scenario, during the sync process shouldn't we delete the document attachment record itself ?
Because at this point we have a Document Attachment record that is pointing to an external file that no longer exists.

@JesperSchulz

src/Apps/W1/External Storage - Document Attachments/app/src/AutomaticSync/DAExternalStorageMigration.Report.al
```diff
@@ -0,0 +1,124 @@
...
+/// Report for migrating document attachments from previous environment/company folder to current environment/company folder.
+/// Can be scheduled via job queue for background processing.
+/// </summary>
+report 8753 "DA External Storage Migration"
+{
+    Caption = 'External Storage Migration';
+    ProcessingOnly = true;
+    UseRequestPage = true;
+    Extensible = false;
+    ApplicationArea = Basic, Suite;
```
darjoo: Is there a specific reason these objects are limited to Basic/Suite?

src/Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DAExternalStorageImpl.Codeunit.al
```diff
@@ -0,0 +1,981 @@
...
+                  tabledata "DA External Storage Setup" = r;
+
+    #region File Scenario Interface Implementation
+    /// <summary>
+    /// Called before adding or modifying a file scenario.
+    /// </summary>
+    /// <param name="Scenario">The ID of the file scenario.</param>
+    /// <param name="Connector">The file storage connector.</param>
+    /// <returns>True if the operation is allowed, otherwise false.</returns>
+    procedure BeforeAddOrModifyFileScenarioCheck(Scenario: Enum "File Scenario"; Connector: Enum System.ExternalFileStorage."Ext. File Storage Connector") SkipInsertOrModify: Boolean;
```
darjoo: I don't think we need to name the return variable. 

Just using `exit(true)` and `exit(false)` in the right places would be clearer.

darjoo: Same with the other procedures

src/Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DAExternalStorageImpl.Codeunit.al
```diff
@@ -0,0 +1,981 @@
...
+
+        if not ExternalStorageSetup.Get() then
+            exit;
+
+        ExternalStorageSetup.CalcFields("Has Uploaded Files");
+        if not ExternalStorageSetup."Has Uploaded Files" then
+            exit;
+
+        SkipDelete := true;
+        Message(NotPossibleToUnassignScenarioMsg);
```
darjoo, line 100: What is this message for? Debugging or is it supposed to be an error to user?

darjoo: Review other `Message(...)` as well.

src/Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DAExternalStorageImpl.Codeunit.al
```diff
@@ -0,0 +1,981 @@
...
+            IsHandled := true;
+    end;
+
+    [EventSubscriber(ObjectType::Codeunit, Codeunit::"Document Attachment Mgmt", OnCopyAttachmentsOnAfterSetToDocumentFilters, '', false, false)]
+    local procedure "Document Attachment Mgmt_OnCopyAttachmentsOnAfterSetToDocumentFilters"(var ToDocumentAttachment: Record "Document Attachment"; ToRecRef: RecordRef; ToAttachmentDocumentType: Enum "Attachment Document Type"; ToNo: Code[20]; ToLineNo: Integer)
+    begin
+        ToDocumentAttachment."Skip Delete On Copy" := ToDocumentAttachment."Stored Externally";
+    end;
+
+    [EventSubscriber(ObjectType::Table, Database::"Document Attachment", OnBeforeOpenInOneDrive, '', false, false)]
```
darjoo: Move the entire subscriber region to the end of file.

src/Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DAExternalStorageImpl.Codeunit.al
```diff
@@ -0,0 +1,981 @@
...
+    var
+        ExternalStorageSetup: Record "DA External Storage Setup";
+        FileName: Text;
+        RootFolder: Text;
+        TableNameFolder: Text[100];
+        EnvironmentHashFolder: Text[32];
+        FileNamePart: Text;
+    begin
+        // Generate unique filename to prevent collisions
+        FileNamePart := DocumentAttachment."File Name" + '-' + DelChr(Format(CreateGuid()), '=', '{}') + '.' + DocumentAttachment."File Extension";
```
darjoo: Create a label with the parts then combine with StrSubstNo

src/Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DAExternalStorageImpl.Codeunit.al
```diff
@@ -0,0 +1,981 @@
...
+
+    #region Telemetry Logging
+    local procedure LogFeatureUsedTelemetry()
+    var
+        DAFeatureTelemetry: Codeunit "DA Feature Telemetry";
+    begin
+        DAFeatureTelemetry.LogFeatureUsed();
+    end;
+
+    local procedure LogFileUploadedTelemetry(DocumentAttachment: Record "Document Attachment")
```
darjoo: Remove these procedures, just call `DAFeatureTelemetry.LogFileUploaded(DocumentAttachment)` directly where this procedure is used.

src/Apps/W1/External Storage - Document Attachments/app/src/Permissions/DAExtStorAdmin.PermissionSet.al
```diff
@@ -0,0 +1,16 @@
...
+{
+    Assignable = true;
+    Caption = 'DA - External Storage Admin';
+    Permissions = tabledata "DA External Storage Setup" = RIMD;
```
darjoo: It looks like you're giving indirect permissions on the various objects that use the table. I'd reduce this to just rimd then.

src/Apps/W1/External Storage - Document Attachments/app/src/Setup/DAExternalStorageSetup.Page.al
```diff
@@ -0,0 +1,152 @@
...
+{
+    PageType = Card;
+    SourceTable = "DA External Storage Setup";
+    Caption = 'External Storage Setup';
+    UsageCategory = None;
+    ApplicationArea = Basic, Suite;
+    InsertAllowed = false;
+    DeleteAllowed = false;
+    Extensible = false;
+    Permissions = tabledata "DA External Storage Setup" = rmid,
```
darjoo:
```suggestion
    Permissions = tabledata "DA External Storage Setup" = rimd,
```
NIT

src/System Application/App/External File Storage/src/Scenario/FileScenario.Interface.al
```diff
@@ -0,0 +1,40 @@
...
+    /// <param name="Connector">The file storage connector.</param>
+    /// <returns>True if the delete operation is handled and should not proceed; otherwise false.</returns>
+    procedure BeforeDeleteFileScenarioCheck(Scenario: Enum "File Scenario"; Connector: Enum System.ExternalFileStorage."Ext. File Storage Connector") SkipDelete: Boolean;
+
+    /// <summary>
+    /// Called before reassigning a file scenario from one to another.
+    /// </summary>
+    /// <param name="CurrentScenario">The ID of the old file scenario.</param>
+    /// <returns>True if the reassign operation is handled and should not proceed, otherwise false.</returns>
+    procedure BeforeReassignFileScenarioCheck(CurrentScenario: Enum "File Scenario") SkipReassign: Boolean
```
darjoo:
```suggestion
    procedure BeforeReassignFileScenarioCheck(Scenario: Enum "File Scenario") SkipReassign: Boolean
```

src/System Application/App/External File Storage/src/Scenario/FileScenario.Interface.al
```diff
@@ -0,0 +1,40 @@
...
+
+interface "File Scenario"
+{
+    /// <summary>
+    /// Called before adding or modifying a file scenario.
+    /// </summary>
+    /// <param name="Scenario">The ID of the file scenario.</param>
+    /// <param name="Connector">The file storage connector.</param>
+    /// <returns>True if the operation is allowed; otherwise false.</returns>
+    procedure BeforeAddOrModifyFileScenarioCheck(Scenario: Enum "File Scenario"; Connector: Enum System.ExternalFileStorage."Ext. File Storage Connector") SkipInsertOrModify: Boolean;
```
darjoo, line 16: Fix the Enum System.ExternalFileStorage.

Import that and remove it from these lines.

src/System Application/App/External File Storage/src/Scenario/FileScenarioSetup.Page.al
```diff
@@ -80,6 +80,25 @@ page 9452 "File Scenario Setup"
...
             group(Scenario)
             {
+                action(ScenarioSetup)
+                {
+                    Visible = (TypeOfEntry = TypeOfEntry::Scenario) and CanUserManageFileSetup;
+                    Caption = 'Additional Scenario Setup';
+                    ToolTip = 'Additional scenario setup for the selected scenario.';
+                    Image = Setup;
+                    Scope = Repeater;
+                    trigger OnAction()
```
darjoo, line 91:
```suggestion

                    trigger OnAction()
```

src/System Application/App/External File Storage/src/Scenario/FileScenariosFactBox.Page.al
```diff
@@ -35,8 +35,42 @@ page 9453 "File Scenarios FactBox"
...
+    {
+        area(Processing)
+        {
+            action(ScenarioSetup)
+            {
+                Caption = 'Additional Scenario Setup';
+                ToolTip = 'Additional scenario setup for the selected scenario.';
+                Image = Setup;
+                Scope = Repeater;
+                trigger OnAction()
```
darjoo, line 58:
```suggestion

                trigger OnAction()
```

src/Apps/W1/External Storage - Document Attachments/app/src/Permissions/DAExtStorAdmin.PermissionSet.al
```diff
@@ -0,0 +1,16 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+namespace Microsoft.ExternalStorage.DocumentAttachments;
+
+/// <summary>
+/// Permission set for External Storage functionality.
+/// Grants necessary permissions to use external storage features.
+/// </summary>
+permissionset 8751 "DA Ext. Stor. Admin"
```
darjoo, line 11: You are also missing an objects PS for the page.

src/Apps/W1/External Storage - Document Attachments/app/src/AutomaticSync/DAExternalStorageSync.Report.al
JesperSchulz: Uh, that's a can of worms you're opening there 🤣 This has been discussed at length! Ideally, you would not grant users delete permissions on the SharePoint folder where BC is storing its data. If you want to delete data, do it from BC.
Happy to discuss your solution as well, but it won't be for this iteration. Stefan's solution behaves as we discussed it.

myp-bcarmo: I agree with you.
I think an additional report would need to be created for this purpose, that would iterate through the "Document Attachments" records stored externally and check if the external file path actually points to an existing file. The result of this check can be delete the document attachment or other solution that you find more suitable.

src/Apps/W1/External Storage - Document Attachments/app/src/DocumentAttachmentIntegration/DAExternalStorageImpl.Codeunit.al
StefanSosic: It stops processing and gives info to user. Avoided Error, not to be too agresive

src/Apps/W1/External Storage - Document Attachments/app/src/AutomaticSync/DAExternalStorageMigration.Report.al
StefanSosic: Isn't this standard for this apps? Do you suggest to move to All or?

StefanSosic: Moved to All

src/Apps/W1/External Storage - Document Attachments/app/src/Permissions/DAExtStorAdmin.PermissionSet.al
```diff
@@ -12,5 +12,6 @@ permissionset 8751 "DA Ext. Stor. Admin"
 {
     Assignable = true;
     Caption = 'DA - External Storage Admin';
-    Permissions = tabledata "DA External Storage Setup" = RIMD;
+    Permissions = tabledata "DA External Storage Setup" = rimd,
+                  page "DA External Storage Setup" = x;
```
darjoo: The execute permission should go into a separate permission set.

See [EmailObjects.PermissionSet.al](https://github.com/microsoft/BCApps/blob/main/src/System%20Application/App/Email/permissions/EmailObjects.PermissionSet.al) for example and this PS will include that.

darjoo: Also give it X, which is direct, rather than x, indirect.

JesperSchulz: Or give the page inherent entitlements and permissions. Why would you want to control execute permissions of that page? The source table's permissions are controlled, which should suffice?

StefanSosic: Did with Inherent Permissions

src/Apps/W1/External Storage - Document Attachments/app/src/Setup/DAExternalStorageSetup.Page.al
```diff
@@ -24,6 +24,7 @@ page 8750 "DA External Storage Setup"
     Extensible = false;
     Permissions = tabledata "DA External Storage Setup" = rimd,
                   tabledata "Job Queue Entry" = r;
+    InherentPermissions = X;
```
JesperSchulz, line 27: InherentEntitlements missing.

darjoo (approved):

PR #4549, opened 2025-08-26
[Agent] Only show company in Agent Card permissions if permissions span multiple companies

qutreson:
#### Summary 
To reduce the risk of potential misconfiguration of agents, do not show the company when the agent has permissions in only one company. A toggle is added to show the company if needed.

image

#### Work Item(s) 
Fixes [AB#593056](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/593056)

### DIFF
--- a/src/System Application/App/Agent/Setup/AgentUserSubform.PageExt.al
+++ b/src/System Application/App/Agent/Setup/AgentUserSubform.PageExt.al
@@ -5,8 +5,10 @@
 
 namespace System.Agents;
 
-using System.Security.User;
 using System.Security.AccessControl;
+using System.Security.User;
+using System.Environment;
+using System.Environment.Configuration;
 
 pageextension 4318 "Agent User Subform" extends "User Subform"
 {
@@ -87,29 +89,17 @@ pageextension 4318 "Agent User Subform" extends "User Subform"
 
     local procedure AccessControlForSingleCompany(var SingleCompanyName: Text[30]): Boolean
     var
-        AccessControl: Record "Access Control";
+        TempCompany: Record Company temporary;
+        UserSettings: Codeunit "User Settings";
     begin
-        // var UserSettings: Codeunit "User Settings";
-        // UserSettings.GetAllowedCompaniesForCurrentUser(Rec);
-
         if not IsAgent then
             exit(false);
 
-        AccessControl.SetRange("User Security ID", Rec."User Security ID");
-        if not AccessControl.FindFirst() then
+        UserSettings.GetAllowedCompaniesForUser(Rec."User Security ID", TempCompany);
+        if TempCompany.Count() <> 1 then
             exit(false);
 
-        SingleCompanyName := AccessControl."Company Name";
-        if SingleCompanyName = '' then
-            // The agent has access to all companies.
-            exit(false);
-
-        while AccessControl.Next() <> 0 do
-            if SingleCompanyName <> AccessControl."Company Name" then begin
-                SingleCompanyName := '';
-                exit(false);
-            end;
-
+        SingleCompanyName := TempCompany.Name;
         exit(true);
     end;
 
--- a/src/System Application/App/User Settings/src/UserSettings.Codeunit.al
+++ b/src/System Application/App/User Settings/src/UserSettings.Codeunit.al
@@ -92,6 +92,18 @@ codeunit 9176 "User Settings"
     end;
 
     /// <summary>
+    /// Get the companies the specified user has access to.
+    /// </summary>
+    /// <param name="UserSecurityId">The user security ID.</param>
+    /// <param name="TempCompany">Companies the current user has access to.</param>
+    procedure GetAllowedCompaniesForUser(UserSecurityId: Guid; var TempCompany: Record Company temporary)
+    var
+        UserSettingsImpl: Codeunit "User Settings Impl.";
+    begin
+        UserSettingsImpl.GetAllowedCompaniesForUser(UserSecurityId, TempCompany);
+    end;
+
+    /// <summary>
     /// Allows the user to select the new profile for given User Settings
     /// </summary>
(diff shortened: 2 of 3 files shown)


src/System Application/App/Agent/Setup/AgentUserSubform.PageExt.al
```diff
@@ -0,0 +1,109 @@
...
+                    ShowCompanyOverride := true;
+                    ShowCompanyField := not ShowCompanyField;
+                end;
+            }
+        }
+    }
+
+    trigger OnAfterGetCurrRecord()
+    begin
+        UpdateGlobalVariables();
```
nikolakukrika: This is going to run too often. Do we need this trigger?
If we do, check if the User ID has changed, store the current on the global variables and exit fast if it has not changed.

src/System Application/App/Agent/Setup/AgentUserSubform.PageExt.al
```diff
@@ -0,0 +1,109 @@
...
+        AccessControl.SetRange("User Security ID", Rec."User Security ID");
+        if not AccessControl.FindFirst() then
+            exit(false);
+
+        SingleCompanyName := AccessControl."Company Name";
+        if SingleCompanyName = '' then
+            // The agent has access to all companies.
+            exit(false);
+
+        while AccessControl.Next() <> 0 do begin
```
nikolakukrika: You could do the count of the records, then set the filter to be different than the company name and compare the counts. It will be faster.

src/System Application/App/Agent/Setup/AgentUserSubform.PageExt.al
```diff
@@ -0,0 +1,109 @@
...
+            {
+                ApplicationArea = All;
+                Caption = 'Show/hide company';
+                Enabled = IsAgent;
+                Image = CompanyInformation;
+                Visible = IsAgent;
+
+                trigger OnAction()
+                begin
+                    ShowCompanyOverride := true;
```
nikolakukrika: We need a small confirm here explaining the consequences.

src/System Application/App/Agent/Setup/AgentUserSubform.PageExt.al
```diff
@@ -0,0 +1,109 @@
...
+                    ShowCompanyField := not ShowCompanyField;
```
nikolakukrika, line 45: We probably need CurrPage.Update(false) here, then we could remove the OnAfterGetCurrRecord trigger.

src/System Application/App/Agent/Setup/AgentUserSubform.PageExt.al
nikolakukrika: The simple code is SetRange on user ID and count. Then SetRange on the first company and count. If they match, then they are all the same.

src/System Application/App/Agent/Setup/AgentUserSubform.PageExt.al
```diff
@@ -0,0 +1,109 @@
...
+        end;
+
+        exit(true);
+    end;
+
+    var
+        IsAgent: Boolean;
+        ShowCompanyField: Boolean;
+        ShowCompanyOverride: Boolean;
+        GlobalSingleCompanyName: Text[30];
```
nikolakukrika, line 110: GlobalFirstCompanyName?

src/System Application/App/Agent/Setup/AgentUserSubform.PageExt.al
```diff
@@ -0,0 +1,109 @@
...
+            end;
+        end;
+
+        exit(true);
+    end;
+
+    var
+        IsAgent: Boolean;
+        ShowCompanyField: Boolean;
+        ShowCompanyOverride: Boolean;
```
nikolakukrika: Minor - ShowCompanyFieldOverride?

src/System Application/App/Agent/Setup/AgentUserSubform.PageExt.al
qutreson: This is just temporary code while exposing something from the server through NavUseAccountHelper to be sure it is in sync with the NST

src/System Application/App/Agent/Setup/AgentUserSubform.PageExt.al
qutreson: I'll add that

src/System Application/App/Agent/Setup/AgentUserSubform.PageExt.al
qutreson: Semantically, this method was doing a Single rather than a First since I want it to return an empty value when there are multiple ones.

src/System Application/App/Agent/Setup/AgentUserSubform.PageExt.al
qutreson: Makes sense for consistency

src/System Application/App/Agent/Setup/AgentUserSubform.PageExt.al
qutreson: I'll try this out. Thanks.

src/System Application/App/Agent/Setup/AgentUserSubform.PageExt.al
qutreson: Updated based on the other suggestion.

src/System Application/App/Agent/Setup/AgentUserSubform.PageExt.al
qutreson: Added

blrobl (approved):

darjoo (approved):

PR #4602, opened 2025-08-27
Bugs/578172

stkillen:
#### Summary 
**Problem**
Changing the test runner is not reflected, until the test suite is recreated. Additionally, `AssistEditTestRunner` uses the wrong record

**Solution**
Update test runner of AL test suite on each test run. 

#### Work Item(s) 
Fixes [AB#578172](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/578172/)


src/Tools/AI Test Toolkit/src/TestSuite/AITALTestSuiteMgt.Codeunit.al
```diff
@@ -141,15 +144,22 @@ codeunit 149037 "AIT AL Test Suite Mgt"
 
         ALTestSuite.Name := AITTestMethodLine."AL Test Suite";
         ALTestSuite.Description := CopyStr(AITTestMethodLine.Description, 1, MaxStrLen(ALTestSuite.Description));
-        AITTestSuite.ReadIsolation := IsolationLevel::ReadUncommitted;
-        AITTestSuite.SetLoadFields("Test Runner Id");
-        if AITTestSuite.Get(AITTestMethodLine."Test Suite Code") then
-            ALTestSuite."Test Runner Id" := AITTestSuite."Test Runner Id";
+        AssignTestRunnerToTestSuite(CopyStr(AITTestMethodLine."Test Suite Code", 1, 10), ALTestSuite);
```
haoranpb: instead of 10, consider MaxLen? 

When we make it larger, it will fail in the gate, we remember to update here

stkillen: You're right, updated

haoranpb (approved):

WaelAbuSeada (approved):

Issue #4653, opened 2025-09-02
[Bug]: There is no method to compare old and current password since SecretTexts cannot be compared in Cloud


bc4all:
### Describe the issue

When Microsoft obsoleted functions in Password Dialog (SystemApp module) to use Secret Texts instead of Text there is no possibility to check if the password typed by the user is the same as password stored. In AL there is no option to compare Secret Texts (SecretTextA <> SecretTextB is not allowed), also in Extensions for Cloud it is not possible to use UnWrap().

I will fix this bug adding new function to facade codeunit that would allow to compare the secret texts.

### Expected behavior

Ability to compare secret texts in a password module

### Steps to reproduce

No steps

### I will provide a fix for a bug

- [x] I will provide a fix for a bug

JesperSchulz:
The PR will have to undergo security review, but let's approve this issue and take a look at the actual code 🚀

bc4all:
Thank you @JesperSchulz  here is the pull request - https://github.com/microsoft/BCApps/pull/4654

PR #4654, opened 2025-09-02
#4653 There is no method to compare old and current password since SecretTexts cannot be compared in Cloud

bc4all:
#### Summary 
Adding the function to compare secret texts for Password module. The new function is needed to make sure that the two passwords match. There is already function ValidateOldPasswordMatch in Implementation however it cannot be used since:
1. Name is not correct
2. It is not used in any facade codeunit
3. it exists if the input is empty

Therefore for clarity the similar function has been added with the proper name. Also to facade codeunit the function has been added to be able to run it by developers.

### DIFF
--- a/src/System Application/App/Password/src/PasswordDialogImpl.Codeunit.al
+++ b/src/System Application/App/Password/src/PasswordDialogImpl.Codeunit.al
@@ -91,4 +91,11 @@ codeunit 9811 "Password Dialog Impl."
         if CurrentPasswordToCompare.Unwrap() = NewPassword.Unwrap() then
             Error(PasswordSameAsNewErr);
     end;
+
+    [NonDebuggable]
+    procedure ValidatedPasswordMatch(CurrentPasswordToCompare: SecretText; PasswordToCompare: SecretText)
+    begin
+        if CurrentPasswordToCompare.Unwrap() <> PasswordToCompare.Unwrap() then
+            Error(CurrentPasswordMismatchErr);
+    end;
 }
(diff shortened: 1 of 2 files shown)


JesperSchulz:
The PR did not pass the security review 🥺 @darjoo might be able to elaborate.

bc4all:
@JesperSchulz @darjoo  hank you for info but I need more info since the MS broke our code (an therefore whole solution that is used by customers). Moreover there is no alternative ways to do it in proper way just by bypassing MS system modules:

1. The code is exact copy of code you already use in the module - check functions ValidateOldPasswordMatch and ValidateNewPasswordUniqueness. How this procedure that is propose does not pass security validation and Microsoft code pass it?
2. How it is possible that nondebuggable code that is marked as onPrem only (run only on MS server) is not secure? 
3. If this code break the security how developers can in a secure way compare SecretTexts? Password is only one thing but there are other partners also waiting on this.
4. If there are any way to do it why you remove fuctions that cannot be replaced by other way that coping whole module?

I am really looking forward to get your answer on this since I need to take an action

StefanMaron (approved):
I dont see any issues with this

JesperSchulz:
@bc4all, your case has been reopened based on the detailed background your provided, which was most helpful. We're looking into options to unblock you.
When do you need those changes. Are these a must for 27.0, or can you wait until 27.1? Just so that I understand the urgency of this request?

bc4all:
In the worst case scenario we can wait until 27.1 I know how you all are busy right now before release.

darjoo (approved):
Rediscussed internally, approving.

PR #4661, opened 2025-09-02
Fix obsoletion in v27.0 for No Series

haoranpb:
#### Summary 

`"Allow Gaps in Nos."` field is obsoleted, but we are still referencing it in `Install` codeunit.

Moving the upgrade logic completely into `Upgrade` codeunit.

#### Work Item(s) 
Fixes [AB#599301](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/599301)

### DIFF
--- a/src/Business Foundation/App/NoSeries/src/Upgrade/NoSeriesUpgrade.Codeunit.al
+++ b/src/Business Foundation/App/NoSeries/src/Upgrade/NoSeriesUpgrade.Codeunit.al
@@ -1,9 +1,12 @@
+#if not CLEANSCHEMA27
 // ------------------------------------------------------------------------------------------------
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 // ------------------------------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


src/Business Foundation/App/NoSeries/src/Upgrade/NoSeriesUpgradeTags.Codeunit.al
```diff
@@ -1,3 +1,4 @@
+#if not CLEANSCHEMA27
```
darjoo, line 1:
```suggestion
#if not CLEAN27
```

CLEANSCHEMA is only used for tables.

src/Business Foundation/App/NoSeries/src/Upgrade/NoSeriesUpgrade.Codeunit.al
```diff
@@ -1,10 +1,13 @@
+#if not CLEANSCHEMA27
```
darjoo, line 1:
```suggestion
#if not CLEAN27
```
Same here

darjoo (approved):

PR #4710, opened 2025-09-08
[27.x] Add DOTNET module for Power Platform API Wrapper

dawidwoz:
#### Summary
We add a new Power Platform API Wrapper DOTNET module to be able to use it in Power Automate Environment Picker.

#### Work Item(s)
Fixes AB#593052

### DIFF
--- a/src/System Application/App/DotNet Aliases/src/dotnet.al
+++ b/src/System Application/App/DotNet Aliases/src/dotnet.al
@@ -2213,4 +2213,13 @@ dotnet
         {
         }
     }
+    assembly("Microsoft.Dynamics.Nav.PowerPlatform.Api")
+    {
+        Culture = 'neutral';
+        PublicKeyToken = '31bf3856ad364e35';
+
+        type("Microsoft.Dynamics.Nav.PowerPlatform.Api.PowerPlatformApiWrapper"; "PowerPlatformApiWrapper")
+        {
+        }
+    }
 }


src/System Application/App/Environment Information/src/EnvironmentInformationImpl.Codeunit.al
```diff
@@ -162,4 +162,3 @@ codeunit 3702 "Environment Information Impl."
         // Do not use this event in a production environment. This should be subscribed to only in tests.
     end;
 }
-
```
WaelAbuSeada: No need for this

src/System Application/App/Environment Information/app.json
```diff
@@ -22,6 +22,12 @@
       "name": "Server Settings",
       "publisher": "Microsoft",
       "version": "27.1.0.0"
+    },
```
WaelAbuSeada: Do you need this?

src/System Application/App/Environment Information/src/EnvironmentInformationImpl.Codeunit.al
dawidwoz: Re-added

src/System Application/App/Environment Information/app.json
dawidwoz: No, I removed it.

PR #4736, opened 2025-09-09
First attempt: Move batch of 1st party apps from ADO to GitHub/BCApps

JesperSchulz:
Abandoned

### DIFF
--- a/src/Apps/W1/EssentialBusinessHeadlines/test/src/codeunits/TestEssentialBusHeadlines.Codeunit.al
+++ b/src/Apps/W1/EssentialBusinessHeadlines/test/src/codeunits/TestEssentialBusHeadlines.Codeunit.al
@@ -381,33 +381,6 @@ codeunit 139600 "Test Essential Bus. Headlines"
     end;
 
     [Test]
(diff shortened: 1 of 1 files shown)


src/Apps/W1/DataArchive/test/TestDataArchiveImpl.codeunit.al
```diff
@@ -0,0 +1,161 @@
+// ------------------------------------------------------------------------------------------------
```
aholstrup1, line 1: Can we rename the _app_ folders to _App_ and _test_ folders to _Test_ (so.. capitalize the first letter) so it matches with the other apps already in the BCApps repo.

JesperSchulz: Sure. Can do 🙂

JesperSchulz: It looks like changing the case on a folder isn't considered a change in git 😮 I will have to abandon this PR and create a new one with the correct casing...

PR #4743, opened 2025-09-10
Added methods in X509Certificate2 to export the certificate public key

### LINKED ISSUE #4238: [Bug]: X509Certificate2.GetCertificatePublicKey returns public key parameters (modulus and exponent) instead of the key value, and there is no way to obtain the generated public key value from an X509 certificate
### Describe the issue

Internally, `X509Certificate2.GetCertificatePublicKey` invokes `AsymmetricAlgorithm.ToXmlString(false)` which returns an XML representation of the crypto service provider. In case of the RSA provider, this is an XML containing the modulus and the exponent values:
```
<RSAKeyValue>
   <Modulus>…</Modulus>
   <Exponent>…</Exponent>
</RSAKeyValue>
```
[RSA.ToXmlString(Boolean) Method](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.rsa.toxmlstring?view=net-9.0#system-security-cryptography-rsa-toxmlstring(system-boolean))

I can use the returned XML to initialize a signature key  instance, but can't obtain the public key value that can be sent to a third party to verify a signature. Neither does the `SignatureKey` codeunit have any methods to extract the public key.

### Expected behavior

Codeunit `X509Certificate2` should have a method similar to `CertificateManagement.GetPublicKeyAsBase64String` which is a wrapper around .Net `X509Certificate2.GetRawCertData()`.

### Steps to reproduce

For now, the only way to obtain the public key from a certificate is to upload it into the isolated storage and use the "Certificate Management" codeunit, but no method to get the key when the certificate comes from an external source.

### Additional context

_No response_

adrogin:
#### Summary
Added methods in X509Certificate2 to export the certificate public key:
- GetCertificatePublicKeyAsBase64String
- GetRawCertDataAsBase64String

Both methods are wrapper around respective .Net methods of the `X509Certificate2` class, with the output encoded in Base64 string.

#### 
Fixes #4238

### DIFF
--- a/src/System Application/App/Cryptography Management/src/X509Certificate2Impl.Codeunit.al
+++ b/src/System Application/App/Cryptography Management/src/X509Certificate2Impl.Codeunit.al
@@ -137,6 +137,27 @@ codeunit 1285 "X509Certificate2 Impl."
     end;
 
     [NonDebuggable]
+    procedure GetCertificatePublicKeyAsBase64String(CertBase64Value: Text; Password: SecretText): Text
+    var
+        X509Certificate2: DotNet X509Certificate2;
+        AsymmetricAlgorithm: DotNet AsymmetricAlgorithm;
+        Convert: DotNet Convert;
+    begin
+        InitializeX509Certificate(CertBase64Value, Password, X509Certificate2);
+        exit(Convert.ToBase64String(X509Certificate2.GetPublicKey()));
+    end;
+
+    [NonDebuggable]
+    procedure GetRawCertDataAsBase64String(CertBase64Value: Text; Password: SecretText): Text
+    var
+        X509Certificate2: DotNet X509Certificate2;
+        Convert: DotNet Convert;
+    begin
+        InitializeX509Certificate(CertBase64Value, Password, X509Certificate2);
+        exit(Convert.ToBase64String(X509Certificate2.GetRawCertData()));
+    end;
+
+    [NonDebuggable]
     [TryFunction]
     local procedure TryInitializeCertificate(CertBase64Value: Text; Password: SecretText; var X509Certificate2: DotNet X509Certificate2)
     var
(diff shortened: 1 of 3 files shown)


WaelAbuSeada:
> @darjoo @WaelAbuSeada As suggested in the linked issue discussion (#4238), requesting your review. Looks like I don't have permission to explicitly add specific reviewers in the PR and can only tag you in comments.

Please get this to pass the build system and I will approve it. It mostly looks fine

darjoo:
Hey @adrogin,

Seems like it is due to an unused variable. 
[Build System Application, Business Foundation and Tools (Clean) / System Application, Business Foundation and Tools (Clean): ./src/System Application/App/Cryptography Management/src/X509Certificate2Impl.Codeunit.al#L143](https://github.com/microsoft/BCApps/pull/4743/files#annotation_39739657861)
AA0137 Variable 'AsymmetricAlgorithm' is unused in 'GetCertificatePublicKeyAsBase64String'.

You can find the errors more convinently located by doing the following:
1. Click on the failing test
2. It goes to a page with the logs
3. On the left navigation bar, click the summary navigation
4. Scroll down to see the errors, typically they're right at the top-ish.

src/System Application/App/Cryptography Management/src/X509Certificate2Impl.Codeunit.al
```diff
@@ -136,6 +136,27 @@ codeunit 1285 "X509Certificate2 Impl."
         exit(AsymmetricAlgorithm.ToXmlString(false));
     end;
 
+    [NonDebuggable]
+    procedure GetCertificatePublicKeyAsBase64String(CertBase64Value: Text; Password: SecretText): Text
+    var
+        X509Certificate2: DotNet X509Certificate2;
+        AsymmetricAlgorithm: DotNet AsymmetricAlgorithm;
```
WaelAbuSeada: unused

src/System Application/App/Cryptography Management/src/X509Certificate2Impl.Codeunit.al
```diff
@@ -136,6 +136,27 @@ codeunit 1285 "X509Certificate2 Impl."
...
+        Convert: DotNet Convert;
```
JesperSchulz, line 143:
```suggestion
        X509Certificate2: DotNet X509Certificate2;
        Convert: DotNet Convert;
```

JesperSchulz (approved):

PR #4749, opened 2025-09-11
New method to add KeyInfo clause to SignedXml from X509 certificate

### LINKED ISSUE #4135: [Bug]: SignedXml codeunit is missing a method to add the X509Data node from an RSA key
### Describe the issue

XML signature syntax defines the X509Data element which contains identifiers of keys or X509 certificates: [https://www.w3.org/TR/xmldsig-core/#sec-X509Data](https://www.w3.org/TR/xmldsig-core/#sec-X509Data)

.Net KeyInfo class allows to add this information from an RSA key value without the need to explicitly format the XML node for the certificate info.

For example, I can add the key info with the following statement in PowerShell
```
$X509Certificate2 = [System.Security.Cryptography.X509Certificates.X509Certificate2]::new($cert, $password, [System.Security.Cryptography.X509Certificates.X509KeyStorageFlags]::Exportable)

$keyInfo = [System.Security.Cryptography.Xml.KeyInfo]::new()
$keyInfo.AddClause([System.Security.Cryptography.Xml.KeyInfoX509Data]::new($X509Certificate2));
$signedXml.KeyInfo = $keyInfo;
```

In this case, the signed XML will contain the KeyInfo element with embedded X509Certificate information, like the following:
```
 <Signature xmlns="http://www.w3.org/2000/09/xmldsig#">
     <SignedInfo>
         <CanonicalizationMethod Algorithm="http://www.w3.org/TR/2001/REC-xml-c14n-20010315"/>
         <SignatureMethod Algorithm="http://www.w3.org/2000/09/xmldsig#rsa-sha1" />
         <Reference URI="#NFe31060243816719000108550000000010001234567897">
         <Transforms>
             <Transform Algorithm="http://www.w3.org/2000/09/xmldsig#enveloped-signature"/>
             <Transform Algorithm="http://www.w3.org/TR/2001/REC-xml-c14n-20010315"/>
         </Transforms>
         <DigestMethod Algorithm="http://www.w3.org/2000/09/xmldsig#sha1"/>
         <DigestValue>vFL68WETQ+mvj1aJAMDx+oVi928=</DigestValue>
         </Reference>
     </SignedInfo>
     <SignatureValue>IhXNhbdL1F9UGb2ydVc5v/gTB/y6r0KIFaf5evUi1i ...</SignatureValue>
     <KeyInfo>
         <X509Data>
             <X509Certificate>MIIFazCCBFOgAwIBAgIQaHEfNaxSeOEvZGlVDANB ... </X509Certificate>
         </X509Data>
      </KeyInfo>
 </Signature>
```

[See also examples in this MS Learn article](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.xml.keyinfo.addclause?view=windowsdesktop-9.0#examples).

Business Central's SignedXml codeunit only allows adding the key info clause as an XML element, which requires explicit formatting of the XML structure for the key info instead of leveraging the available .Net functionality.

### Expected behavior

Codeunit SignedXml should expose an overload for the AddClause method accepting an instance of the "Signature Key" codeunit as its argument to allow initialization of the key info from a certificate.
I expect to be able to write code similar to the PowerShell example above in BC to obtain a signature with the certificate info.

```
  var
      SignedXml: Codeunit SignedXml;
      SignatureKey: Codeunit "Signature Key";
  begin
      ...
      SignedXml.InitializeKeyInfo();
      SignedXml.AddClause(SignatureKey);
      SignedXml.ComputeSignature();
      ...
  end;
```

### Steps to reproduce

Method `SignedXml.AddClause(KeyInfoNodeXmlElement: XmlElement)` requires the developer to explicitly format the key info element. Calling `SignedXml.ComputeSignature` without setting the key info clause will produce a signature without the KeyInfo node.

### Additional context

_No response_

adrogin:
#### Summary
New method in the SignedXml codeunit to add KeyInfo clause to the signature from an X509 certificate.
Key info contains the **X509Data** element conforming to the [W3C XML signature syntax](https://www.w3.org/TR/xmldsig-core/#sec-X509Data).

#### Work Item(s)
Resolves #4135 

Fixes [AB#608915](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/608915)

### DIFF
--- a/src/System Application/Test/Cryptography Management/src/Xml/SignedXmlModuleTest.Codeunit.al
+++ b/src/System Application/Test/Cryptography Management/src/Xml/SignedXmlModuleTest.Codeunit.al
@@ -115,6 +115,7 @@ codeunit 132612 "Signed Xml Module Test"
         LibraryAssert.AreEqual(C14NTransformUriTok, Node.AsXmlAttribute().Value, 'Incorrect transform was applied.');
     end;
 
+    [Test]
     procedure SignXmlDocumentWithSigningKeyAsSecretText()
     var
         XmlToSign: XmlDocument;
@@ -157,7 +158,6 @@ codeunit 132612 "Signed Xml Module Test"
     [Test]
     procedure SignXmlDocumentAddKeyInfoInSignature()
     var
-        SignedXml: Codeunit SignedXml;
         SignatureKey: Codeunit "Signature Key";
         XmlDoc: XmlDocument;
         DataElement: XmlElement;


src/System Application/Test/Cryptography Management/src/Xml/SignedXmlModuleTest.Codeunit.al
```diff
@@ -154,6 +154,47 @@ codeunit 132612 "Signed Xml Module Test"
         LibraryAssert.AreEqual('Signature', SignedXmlElement.LocalName, 'Signature was not computed.');
     end;
 
+    [Test]
+    procedure SignXmlDocumentAddKeyInfoInSignature()
+    var
+        SignedXml: Codeunit SignedXml;
```
WaelAbuSeada: SignedXml is a global var. was a bit hidden:
  Error: AA0198 The name of the local variable 'SignedXml' is identical to a global variable.

WaelAbuSeada:
> > > @WaelAbuSeada Thanks a lot! I'm not sure though, why the build failed - the error message does not look very informative. I updated the branch and pushed again, but now the PR waiting for the approval to start the workflow.
> > 
> > 
> > I'll reinitiate the build :-)
> 
> Thanks a lot! But it's failing again. Can we see the actual compiler error? All I see now is a reference to BCContainerHelper which doesn't help me to pinpoint the issue. And of course I can build the app and run the tests in my container locally.

Ping me when you fix it

WaelAbuSeada (approved):

PR #4755, opened 2025-09-11
[Shopify] Company mapping test are instable

onbuyuka:
#### Summary 
Company mapping test are instable

#### Work Item(s) 
Fixes [AB#602442](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/602442)

### DIFF
--- a/src/Apps/W1/Shopify/Test/Companies/ShpfyCompanyMappingTest.Codeunit.al
+++ b/src/Apps/W1/Shopify/Test/Companies/ShpfyCompanyMappingTest.Codeunit.al
@@ -447,7 +447,7 @@ codeunit 134245 "Shpfy Company Mapping Test"
     begin
         Id := Any.IntegerInRange(10000, 99999);
         ShopifyCompany.SetRange(Id, Id);
-        ShopifyCompany.Delete();
+        ShopifyCompany.DeleteAll();
         ShopifyCompany.Init();
         ShopifyCompany.Id := Id;
         ShopifyCompany."Shop Code" := Shop."Code";
@@ -485,7 +485,7 @@ codeunit 134245 "Shpfy Company Mapping Test"
     begin
         Id := Any.IntegerInRange(10000, 99999);
         ShopifyCustomer.SetRange(Id, Id);
-        ShopifyCustomer.Delete();
+        ShopifyCustomer.DeleteAll();
         ShopifyCustomer.Init();
         ShopifyCustomer.Id := Id;
         ShopifyCustomer.Insert(false);


src/Apps/W1/Shopify/Test/Companies/ShpfyCompanyMappingTest.Codeunit.al
```diff
@@ -478,11 +483,12 @@ codeunit 134245 "Shpfy Company Mapping Test"
 
     local procedure CreateShopifyCustomer(var ShopifyCustomer: Record "Shpfy Customer"; Id: BigInteger)
```
grobyns: The 1 place where this method is called with id <> 0, you get the random int, same as if it was 0. Basically, you have 1 place where you need to know the id, so just return it. If you don't need it, ignore the return value.

instead of doing an if get then delete, you could set a filter and delete. You don't want the record, just want to make sure the id is not used.

src/Apps/W1/Shopify/Test/Companies/ShpfyCompanyMappingTest.Codeunit.al
```diff
@@ -478,11 +483,12 @@ codeunit 134245 "Shpfy Company Mapping Test"
...
+        if Id = 0 then
+            Id := Any.IntegerInRange(10000, 99999);
+        if ShopifyCustomer.Get(Id) then
+            ShopifyCustomer.Delete();
         ShopifyCustomer.Init();
-        if Id <> 0 then
-            ShopifyCustomer.Id := Id
-        else
-            ShopifyCustomer.Id := Any.IntegerInRange(10000, 99999);
+        ShopifyCustomer.Id := Id;
```
grobyns, line 490:
```suggestion
        local procedure CreateShopifyCustomer(var ShopifyCustomer: Record "Shpfy Customer");
    begin
        Id := Any.IntegerInRange(10000, 99999);
        ShopifyCustomer.SetRange(Id, Id);
        ShopifyCustomer.Delete();
        ShopifyCustomer.Init();
        ShopifyCustomer.Id := Id;
        ShopifyCustomer.Insert(false);
```

src/Apps/W1/Shopify/Test/Companies/ShpfyCompanyMappingTest.Codeunit.al
grobyns: don't even need to return it, just get it from the record...

WaelAbuSeada (approved):

WaelAbuSeada (dismissed):

grobyns (dismissed):

PR #4757, opened 2025-09-12
Moving the Power BI apps to BCApps

mynjj:
#### Summary 

Moving the Power BI apps to the BCApps repository.

#### Work Item(s) 
Fixes [AB#602534](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/602534)

PR #4760, opened 2025-09-12
Keep apps installed during Apps build and add RunNumberOffset to build numbers

aholstrup1:
#### Summary 
During the Apps (W1) build where we compile, publish and test the add-on apps we start out by creating the container and then we uninstall all apps and unpublish a certain set of apps. The set of apps we unpublish contains the apps under the Apps/W1 folder here in BCApps. 

Right now, the unpublish step fails in #4757 due to the fact that there are apps that take dependency on the Power BI app. To avoid having to unpublish a lot of apps, we can instead just uninstall all apps but keep the apps published. 

There is one slight challenge with this though. When we want to publish a new version of App A that takes dependency on App B. If App B is already published to the container in two different versions, it will use the version with the highest version number. Right now, the apps with the highest version numbers are the apps that come with the BC Artifact. To get around this, we can add a runNumberOffset to ensure that the apps produced in BCApps have higher version numbers than the ones on the BC Artifact. 

#### Work Item(s) 
Fixes [AB#602818](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/602818)

### DIFF
--- a/build/projects/Apps (W1)/.AL-Go/NewBcContainer.ps1
+++ b/build/projects/Apps (W1)/.AL-Go/NewBcContainer.ps1
@@ -4,4 +4,4 @@ Param(
 Import-Module (Join-Path $PSScriptRoot "../../../scripts/EnlistmentHelperFunctions.psm1" -Resolve)
 
 $script = Join-Path $PSScriptRoot "../../../scripts/NewBcContainer.ps1" -Resolve
-. $script -parameters $parameters -AppsToUnpublish (Get-AppsInFolder "src/Apps/W1")
+. $script -parameters $parameters -AppsToUnpublish @("None")
--- a/.github/AL-Go-Settings.json
+++ b/.github/AL-Go-Settings.json
@@ -98,5 +98,6 @@
     "pullRequestLabels": [
       "Automation"
     ]
-  }
+  },
+  "runNumberOffset": 60000
 }


build/projects/Apps (W1)/.AL-Go/NewBcContainer.ps1
```diff
@@ -4,4 +4,4 @@ Param(
 Import-Module (Join-Path $PSScriptRoot "../../../scripts/EnlistmentHelperFunctions.psm1" -Resolve)
 
 $script = Join-Path $PSScriptRoot "../../../scripts/NewBcContainer.ps1" -Resolve
-. $script -parameters $parameters -AppsToUnpublish (Get-AppsInFolder "src/Apps/W1")
\ No newline at end of file
+. $script -parameters $parameters -AppsToUnpublish @("None")
```
mazhelez, line 7: Why don't you make it a switch, e.g. `UnpublishInstalledApps` or something like that?

aholstrup1: I think it used to be a switch before I changed the mechanism due to a bug (related to the issue in this PR). I'll keep it like this for now I case we need to do something similar again in the future

PR #4820, opened 2025-09-19
[Email] Add filter on time when counting email concurrency to avoid fake "Processing" email block email sending

gggdttt:
#### Summary 
Now the email sending could be blocked because the status of the emails' updating failed. For example, the email sending already completed, but for permission issue or any other issue, the email status of the email in email outbox is not updated correctly. Then if more than 10 emails stuck there with status "processing", it will block the whole email sending.

To solve this issue, we consider to add a time range when consider the concurrency of email. 

#### Work Item(s) 
Fixes [AB#603459](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/603459)


BardurKnudsen (approved):
looks good

haoranpb (approved):

PR #4830, opened 2025-09-22
[Shopify] Introduce new setting to determine whether to create invoice or order from fulfilled order

onbuyuka:
#### Summary 
Introduce new setting to determine whether to create invoice or order from fulfilled order

#### Work Item(s) 
Fixes [AB#604148](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/604148)

### DIFF
--- a/src/Apps/W1/Shopify/App/src/Order handling/Codeunits/ShpfyProcessOrder.Codeunit.al
+++ b/src/Apps/W1/Shopify/App/src/Order handling/Codeunits/ShpfyProcessOrder.Codeunit.al
@@ -75,7 +75,7 @@ codeunit 30166 "Shpfy Process Order"
             ShopifyOrderHeader.TestField("Sell-to Customer No.");
             SalesHeader.Init();
             SalesHeader.SetHideValidationDialog(true);
-            if ShopifyOrderHeader."Fulfillment Status" = ShopifyOrderHeader."Fulfillment Status"::Fulfilled then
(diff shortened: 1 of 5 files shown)


src/Apps/W1/Shopify/App/src/Order handling/Codeunits/ShpfyProcessOrder.Codeunit.al
```diff
@@ -75,7 +75,7 @@ codeunit 30166 "Shpfy Process Order"
             ShopifyOrderHeader.TestField("Sell-to Customer No.");
             SalesHeader.Init();
             SalesHeader.SetHideValidationDialog(true);
-            if ShopifyOrderHeader."Fulfillment Status" = ShopifyOrderHeader."Fulfillment Status"::Fulfilled then
+            if (ShopifyOrderHeader."Fulfillment Status" = ShopifyOrderHeader."Fulfillment Status"::Fulfilled) and ShopifyShop."Create Invoices From Orders" then
```
mynjj, line 78: just a thought: the field name doesn't tell me what would it happen if the value is false, I have to find this code and read the else branch. Also by being a bool it's making the schema hard to evolve (I don't know how applicable is this for future scenarios)

This could be avoided by having a field that returns the action to take when processing the order and an enum instead.

OTOH this is also minor, so feel free to won't fix.

onbuyuka: you only have two options: order and invoice. if not invoice then order, there will be no future scenario

PR #4867, opened 2025-09-24
Add gate for duplicate app and object ids 

aholstrup1:
#### Summary 
From time to time we see failures in the uptake due to issues with the object ids or app ids. Internally, we require that
* App IDs are unique 
* All test objects fall within a certain range 
* Object ids are unique 

This PR adds a gate for that in BCApps. The last-mentioned gate currently only validates this within the code in BCApps. In the future, we could look into downloading e.g. the base application source code prior to running the check so we can also validate that we're not reusing object ids used in the base application. 

#### Work Item(s) 
Fixes [AB#591348](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/591348)


.github/workflows/VerifyAppChanges.yaml
```diff
@@ -22,3 +22,5 @@ jobs:
           ref: ${{ github.sha }}
 
       - uses: microsoft/BCApps/.github/actions/TestPreprocessorSymbols@main
+
+      - uses: microsoft/BCApps/.github/actions/TestObjectIdsAndManifests@private/aholstrup/objectidtests
```
mazhelez: Don't forget to:

```suggestion
      - uses: microsoft/BCApps/.github/actions/TestObjectIdsAndManifests@main
```

build/scripts/AppObjectValidation.psm1
```diff
@@ -0,0 +1,178 @@
...
+    foreach ($TestObject in $ApplicationObjects.TestObjects) {
+        $ObjectID = GetObjectId $TestObject
+        if (($ObjectID -lt $MinTestObjectId) -or ($ObjectID -gt $MaxTestObjectId)) {
+            $offendingObjects += $ObjectID
+        }
+    }
+
+    if ($IntroducedDuplicates.Count -gt 0) {
+        Write-Host "##[error]Clashing object IDs detected: $($IntroducedDuplicates -join ',')"
+        throw "Clashing object IDs detected. When adding new objects, ensure that introduced object IDs are not currently in use."
```
mazhelez: Minor: consider accumulating the errors.

PR #4891, opened 2025-09-25
Slice 573373: [World-Class Service] Early Preview Program - Show Notification

JesperSchulz:
#### Summary 
Adding "Early Access Preview" notification capabilities to environment information.

#### Work Item(s) 
Fixes AB#573373

### DIFF
--- a/src/System Application/App/Environment Information/src/EnvInfNotificationImpl.Codeunit.al
+++ b/src/System Application/App/Environment Information/src/EnvInfNotificationImpl.Codeunit.al
@@ -19,12 +19,12 @@ codeunit 3706 "Env. Inf. Notification Impl."
 
     procedure ShowEarlyPreviewNotification()
     var
-        EnvironmentInformation: Codeunit "Environment Information";
+        EnvironmentInformationImpl: Codeunit "Environment Information Impl.";
     begin
         if not GuiAllowed then
             exit;
 
-        if not EnvironmentInformation.IsEarlyPreview() then
+        if not EnvironmentInformationImpl.IsEarlyPreview() then
             exit;
 
         ShowNotification();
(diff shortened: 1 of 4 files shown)


src/System Application/App/Environment Information/src/EnvInfNotificationImpl.Codeunit.al
```diff
@@ -0,0 +1,53 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+namespace System.Environment;
+
+codeunit 3706 "Env. Inf. Notification Impl."
+{
+    Access = Internal;
+    SingleInstance = true;
```
grobyns, line 10: why single instance?

src/System Application/App/Environment Information/src/EnvInfNotificationImpl.Codeunit.al
```diff
@@ -0,0 +1,53 @@
...
+        LearnMoreUrlTxt: Label ' https://go.microsoft.com/fwlink/?linkid=2336818';
+
+    procedure ShowEarlyPreviewNotification()
+    var
+        EnvironmentInformationImpl: Codeunit "Environment Information Impl.";
+    begin
+        if not GuiAllowed then
+            exit;
+
+        if not EnvironmentInformationImpl.IsEarlyPreview() then
```
gggdttt, line 27: Is there any way that we can avoid preview logic affecting the loading time of real PROD? 

If not it's fine, but according to the performance lab, all the page's performance growing around 200% since 23.0

JesperSchulz: I don't see how, I'm afraid :-( This particular call should be cheap, but you're right, that many tiny chances of this kind would add up.

src/System Application/App/Environment Information/src/EnvInfNotificationImpl.Codeunit.al
JesperSchulz: Because the others are. Both the other impl. and facade. I guess it doesn't really matter here?

WaelAbuSeada (commented):
lgtm

PR #4901, opened 2025-09-26
Remove calcfield calculation from "Client Type"

BazookaMusic:
#### Summary 
Client type in performance profile list is no longer a flowfield, so we shouldn't try to calculate it.

#### Work Item(s)
Fixes [AB#606011](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/606011)


src/System Application/App/Performance Profiler/src/PerformanceProfileList.Page.al
```diff
@@ -268,7 +267,7 @@ page 1931 "Performance Profile List"
             exit(PerformanceProfileScheduler.Description);
     end;
 
-local procedure ComputePlatformCallDuration(): Duration
+    local procedure ComputePlatformCallDuration(): Duration
```
BazookaMusic, line 270: autoformatted, it looks ok

kalberes (dismissed):

nikolakukrika (dismissed):

nikolakukrika (approved):

PR #4903, opened 2025-09-26
Add AL support for OCV feedback API

edwardUL99:
#### Summary 
This PR adds the following API to allow AL code request submission of feedback to OCV:
- **RequestFeedback**: Request general feedback for a feature, (or copilot feature). Also allows specifying an OCV sub-area (featureArea parameter). Allows the user to select if it is like, dislike, or a suggestion
- **RequestLikeFeedback**: Requests feedback for a feature, (or copilot feature) when the user likes the feature. Also allows specifying an OCV sub-area (featureArea parameter).
**RequestDislikeFeedback**: Requests feedback for a feature, (or copilot feature) when the user dislikes the feature. Also allows specifying an OCV sub-area (featureArea parameter).
**SurveyTimerActivity**: OCV provides a concept surrounding usage time, for example if a user has used a feature for 300 seconds. They provide a trigger name associated with the survey as the activityName parameter, and start whether to start or end the timer
**SurveyTriggerActivity**: OCV also provides a non-timer based trigger, which could be for example, a signal that a user pressed a button. This takes the activityName parameter for the trigger associated with the OCV survey.

#### Work Item(s) 
Fixes #605545

### DIFF
--- a/src/System Application/App/Feedback/app.json
+++ b/src/System Application/App/Feedback/app.json
@@ -1,0 +1,33 @@
+{
+  "id": "f7964d32-7685-400f-8297-4bc17d0aab0e",
+  "name": "Feedback",
+  "publisher": "Microsoft",
+  "version": "28.0.0.0",
+  "brief": "Functionality for collecting user feedback",
+  "description": "Functionality to allow for AL developers to collect user feedback from the application.",
+  "privacyStatement": "https://go.microsoft.com/fwlink/?LinkId=724009",
+  "EULA": "https://go.microsoft.com/fwlink/?linkid=2009120",
+  "help": "",
+  "url": "https://go.microsoft.com/fwlink/?LinkId=724011",
+  "logo": "",
+  "dependencies": [],
+  "screenshots": [],
+  "platform": "28.0.0.0",
+  "application": "28.0.0.0",
+  "resourceExposurePolicy": {
+    "allowDebugging": true,
+    "allowDownloadingSource": true,
+    "includeSourceInSymbolFile": true
+  },
+  "runtime": "17.0",
+  "features": [
+    "NoImplicitWith"
+  ],
+  "target": "OnPrem",
+  "idRanges": [
+    {
+      "from": 1595,
+      "to": 1599
+    }
+  ]
+}
--- a/src/System Application/App/DotNet Aliases/src/dotnet.al
+++ b/src/System Application/App/DotNet Aliases/src/dotnet.al
@@ -711,6 +711,10 @@ dotnet
         type("Microsoft.Dynamics.Nav.Runtime.ActivityLogAttribute"; "ActivityLogAttribute")
         {
         }
+
+        type("Microsoft.Dynamics.Nav.Runtime.ALFeedback"; ALFeedback)
+        {
+        }
     }
 
     assembly("Microsoft.Dynamics.Nav.OAuth")
--- a/src/System Application/App/Feedback/README.md
+++ b/src/System Application/App/Feedback/README.md
@@ -1,0 +1,120 @@
+# Feedback API Documentation
+
+This document describes the APIs provided by the `Feedback` codeunit (ID: 1599) in the `System.Feedback` namespace. These APIs enable user feedback collection and survey management for features in Business Central.
+
+## Overview
+The `Feedback` codeunit provides procedures to request feedback (general, like, dislike) for features, including Copilot features and feature areas, as well as triggering surveys.
+
+## API Reference
+
+---
+
+### RequestFeedback
+Requests general feedback for a feature, optionally specifying if it is a Copilot feature and its area.
+
+**Signature:**
+```al
+procedure RequestFeedback(FeatureName: Text[255]; IsCopilotFeature: Boolean; FeatureArea: Text[255])
+```
+**Parameters:**
+- `FeatureName`: The name of the feature for which feedback is requested.
+- `IsCopilotFeature`: Specifies if the feature is a Copilot feature.
+- `FeatureArea`: The area or sub-area of the feature.
+
+---
+
+### RequestLikeFeedback
+Requests a 'like' (positive) feedback for a feature, optionally specifying if it is a Copilot feature and its area.
+
+**Signature:**
+```al
+procedure RequestLikeFeedback(FeatureName: Text[255]; IsCopilotFeature: Boolean; FeatureArea: Text[255])
+```
+**Parameters:**
+- `FeatureName`: The name of the feature for which like feedback is requested.
+- `IsCopilotFeature`: Specifies if the feature is a Copilot feature.
+- `FeatureArea`: The area or sub-area of the feature.
+
+---
+
+### RequestDislikeFeedback
+Requests a 'dislike' (negative) feedback for a feature, optionally specifying if it is a Copilot feature and its area.
+
+**Signature:**
+```al
+procedure RequestDislikeFeedback(FeatureName: Text[255]; IsCopilotFeature: Boolean; FeatureArea: Text[255])
+```
+**Parameters:**
+- `FeatureName`: The name of the feature for which dislike feedback is requested.
+- `IsCopilotFeature`: Specifies if the feature is a Copilot feature.
+- `FeatureArea`: The area or sub-area of the feature.
+
+---
+
+### SurveyTimerActivity
+Starts or stops a survey timer activity. This is used to start a timer to count up user usage times, which can then trigger a survey prompt after a certain threshold is reached.
+
+For example: 5 minutes of usage time from timer start to timer end.
+
+**Signature:**
+```al
+procedure SurveyTimerActivity(ActivityName: Text[255]; Start: Boolean)
+```
+**Parameters:**
+- `ActivityName`: The name of the activity for which the timer is started or stopped.
+- `Start`: If true, starts the timer; if false, stops the timer.
+
+---
+
+### SurveyTriggerActivity
+Sends a one-time trigger event based on a specific activity name. The event could be, for example, a user clicking a button.
+
+**Signature:**
+```al
+procedure SurveyTriggerActivity(ActivityName: Text[255])
+```
+**Parameters:**
+- `ActivityName`: The name of the activity that triggers the survey.
+
+---
+
+## Usage Examples
+
+### Basic Feedback Request
+```al
+var
+    Feedback: Codeunit Feedback;
+begin
+    // Request general feedback for a feature
+    Feedback.RequestFeedback('MyFeature', false, 'MainArea');
+    
+    // Request positive feedback for a Copilot feature
+    Feedback.RequestLikeFeedback('CopilotAssist', true, 'AI');
+    
+    // Request negative feedback for troubleshooting
+    Feedback.RequestDislikeFeedback('DataSync', false, 'Integration');
+end;
+```
+
+### Survey Activity Tracking
+```al
+var
+    Feedback: Codeunit Feedback;
+begin
+    // Start timing user activity
+    Feedback.SurveyTimerActivity('DocumentProcessing', true);
+    
+    // ... user performs document processing tasks ...
+    
+    // Stop timing when task is complete
+    Feedback.SurveyTimerActivity('DocumentProcessing', false);
+    
+    // Trigger survey based on specific user action
+    Feedback.SurveyTriggerActivity('ReportGeneration');
+end;
+```
+
+---
+
+## License
+Licensed under the MIT License. See License.txt in the project root for license information.
--- a/src/System Application/App/Feedback/src/Feedback.Codeunit.al
+++ b/src/System Application/App/Feedback/src/Feedback.Codeunit.al
@@ -1,0 +1,76 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Feedback;
+
+/// <summary>
+/// Codeunit for providing user feedback.
+/// </summary>
+codeunit 1599 Feedback
+{
+    /// <summary>
+    /// Requests general feedback for a feature, optionally specifying if it is a Copilot feature and its area.
+    /// </summary>
+    /// <param name="FeatureName">The name of the feature for which feedback is requested.</param>
+    /// <param name="IsCopilotFeature">Specifies if the feature is a Copilot feature.</param>
+    /// <param name="FeatureArea">The area or sub-area of the feature.</param>
+    procedure RequestFeedback(FeatureName: Text[255]; IsCopilotFeature: Boolean; FeatureArea: Text[255])
+    var
+        ALFeedback: DotNet ALFeedback;
+    begin
+        ALFeedback.RequestFeedback(FeatureName, IsCopilotFeature, FeatureArea);
+    end;
+
+    /// <summary>
+    /// Requests a 'like' (positive) feedback for a feature, optionally specifying if it is a Copilot feature and its area.
+    /// </summary>
+    /// <param name="FeatureName">The name of the feature for which like feedback is requested.</param>
+    /// <param name="IsCopilotFeature">Specifies if the feature is a Copilot feature.</param>
+    /// <param name="FeatureArea">The area or sub-area of the feature.</param>
+    procedure RequestLikeFeedback(FeatureName: Text[255]; IsCopilotFeature: Boolean; FeatureArea: Text[255])
+    var
+        ALFeedback: DotNet ALFeedback;
+    begin
+        ALFeedback.RequestLikeFeedback(FeatureName, IsCopilotFeature, FeatureArea);
+    end;
+
+    /// <summary>
+    /// Requests a 'dislike' (negative) feedback for a feature, optionally specifying if it is a Copilot feature and its area.
+    /// </summary>
+    /// <param name="FeatureName">The name of the feature for which dislike feedback is requested.</param>
+    /// <param name="IsCopilotFeature">Specifies if the feature is a Copilot feature.</param>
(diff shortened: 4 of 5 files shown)


src/System Application/App/Feedback/src/MicrosoftFeedback.Codeunit.al
```diff
@@ -0,0 +1,163 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace SystemFeedback;
```
JesperSchulz:
```suggestion
namespace System.Feedback;
```

Depending on the module name, this should get adjusted. I tend towards System.UserFeedback...

src/System Application/App/Feedback/src/MicrosoftFeedback.Codeunit.al
```diff
@@ -0,0 +1,163 @@
...
+
+/// <summary>
+/// Codeunit for providing feedback to Microsoft. To be used by internal Microsoft apps only.
+/// </summary>
+codeunit 1599 MicrosoftFeedback
```
JesperSchulz: If we want to stick to that name, please make the following change:
```suggestion
codeunit 1599 "Microsoft Feedback"
```

I wonder if we should be more specific though. Again, "User Feedback" seems right to me.

src/System Application/App/Microsoft User Feedback/src/MicrosoftUserFeedback.Codeunit.al
```diff
@@ -0,0 +1,163 @@
...
+    end;
+
+    /// <summary>
+    /// Requests general feedback for a feature, optionally specifying if it is a Copilot feature and its area.
+    /// </summary>
+    /// <param name="FeatureName">The name of the feature for which feedback is requested.</param>
+    /// <param name="IsAIFeature">Specifies if the feature is an AI feature.</param>
+    /// <param name="FeatureArea">The area or sub-area of the feature. ID on OCV.</param>
+    /// <param name="FeatureAreaDisplayName">The display name of the feature area.</param>
+    procedure RequestFeedback(FeatureName: Text; IsAIFeature: Boolean; FeatureArea: Text; FeatureAreaDisplayName: Text)
```
JesperSchulz: We're using the [facade pattern](https://alguidelines.dev/docs/patterns/facade-pattern/) in System Application modules. Please move all the logic to an implementation codeunit, which is "internal". This public facade then calls into the procedures of the implementation codeunit (look at any other module - most of them implement that pattern).

src/System Application/App/Feedback/app.json
```diff
@@ -0,0 +1,33 @@
+{
+  "id": "f7964d32-7685-400f-8297-4bc17d0aab0e",
+  "name": "Feedback",
```
JesperSchulz: Feedback is probably a too generic name. "Microsoft Feedback" also doesn't really hit the nail on the head. Maybe "User Feedback"?

src/System Application/App/Feedback/app.json
```diff
@@ -0,0 +1,33 @@
...
+  "dependencies": [],
+  "screenshots": [],
+  "platform": "28.0.0.0",
+  "application": "28.0.0.0",
+  "resourceExposurePolicy": {
+    "allowDebugging": true,
+    "allowDownloadingSource": true,
+    "includeSourceInSymbolFile": true
+  },
+  "runtime": "17.0",
```
JesperSchulz: No need to specify.

JesperSchulz (changes requested):
Let's start with some basics, like naming and general architecture. When that is settled, we can look into the specifics.

src/System Application/App/Feedback/src/MicrosoftFeedback.Codeunit.al
edwardUL99: @JesperSchulz The concern here is that we want this feedback to be only used within Microsoft only, and not a general purpose one for everyone to use. So I wanted to name to reflect that.

So, if you have any other suggestions

src/System Application/App/Feedback/app.json
edwardUL99: See above comment about the naming.

This module is for feedback sent from users to microsoft, so wanted to reflect that

src/System Application/App/Feedback/src/MicrosoftFeedback.Codeunit.al
JesperSchulz: OK, how about "User Feedback for Microsoft"? Or "Microsoft User Feedback". That's pretty specific.

src/System Application/App/Microsoft User Feedback/src/MicrosoftUserFeedbackImpl.Codeunit.al
```diff
@@ -0,0 +1,85 @@
...
+    /// <summary>
+    /// Requests general feedback for a feature, optionally specifying if it is a Copilot feature and its area.
+    /// </summary>
+    /// <param name="FeatureName">The name of the feature for which feedback is requested.</param>
+    /// <param name="IsAIFeature">Specifies if the feature is an AI feature.</param>
+    /// <param name="FeatureArea">The area or sub-area of the feature. ID on OCV.</param>
+    /// <param name="FeatureAreaDisplayName">The display name of the feature area.</param>
+    /// <param name="ContextFiles">Map of filename to base64 file to attach to the feedback. Must contain the filename in the extension.</param>
+    /// <param name="ContextProperties">Additional data to pass properties to the feedback mechanism.</param>
+    procedure RequestFeedback(FeatureName: Text; IsAIFeature: Boolean; FeatureArea: Text; FeatureAreaDisplayName: Text; ContextFiles: Dictionary of [Text, Text]; ContextProperties: Dictionary of [Text, Text])
```
darjoo: Since this is purely for Microsoft, please use the callermoduleinfo vs currentmoduleinfo check pattern. See my configuration management pr, https://github.com/microsoft/BCApps/pull/5084.

I know onprem cannot be used as it requires uptake apps to be onprem which we cannot guarantee for our 1st party apps and we want to move more to cloud.

src/System Application/App/Microsoft User Feedback/src/MicrosoftUserFeedbackImpl.Codeunit.al
```diff
@@ -0,0 +1,85 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Microsoft.UserFeedback;
+
+/// <summary>
+/// Implementation codeunit for providing feedback to Microsoft. To be used by internal Microsoft apps only.
+/// </summary>
+codeunit 1599 "Microsoft User Feedback Impl"
+{
```
darjoo, line 12: Missing access=internal and inherententitlements/permissions

src/System Application/App/Microsoft User Feedback/src/MicrosoftUserFeedback.Codeunit.al
```diff
@@ -0,0 +1,157 @@
...
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Microsoft.UserFeedback;
+
+/// <summary>
+/// Codeunit for providing feedback to Microsoft. To be used by internal Microsoft apps only.
+/// </summary>
+codeunit 1600 "Microsoft User Feedback"
+{
+    Access = Public;
```
darjoo, line 13: Missing inherententitlements/permissions. I assume we do not want to have to deal with permission issues for a feedback related CU.

src/System Application/App/Microsoft User Feedback/src/MicrosoftUserFeedbackImpl.Codeunit.al
```diff
@@ -0,0 +1,85 @@
...
+    /// The event could be, for example, a user clicking a button
+    /// </summary>
+    /// <param name="ActivityName">The name of the activity that triggers the survey.</param>
+    procedure SurveyTriggerActivity(ActivityName: Text)
+    begin
+        Feedback.SurveyTriggerActivity(ActivityName);
+    end;
+
+    var
+        Feedback: Codeunit System.Feedback.Feedback;
```
onbuyuka: Let's not use the fully qualified name and intead add using System.Feedback statement to top of the codeunit

src/System Application/App/Microsoft User Feedback/src/MicrosoftUserFeedback.Codeunit.al
```diff
@@ -0,0 +1,157 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Microsoft.UserFeedback;
+
+/// <summary>
+/// Codeunit for providing feedback to Microsoft. To be used by internal Microsoft apps only.
+/// </summary>
+codeunit 1590 "Microsoft User Feedback"
```
onbuyuka, line 11: As of now, this codeunit is open to third party apps but README mentions this is for internal use only?

src/System Application/App/Microsoft User Feedback/src/MicrosoftUserFeedbackImpl.Codeunit.al
darjoo: I'm not a fan of having `IsAIFeature` in the parameter list. I'd rather have a setter/builder pattern for that. For others to comment whether it's worth it.

For example
```
UserFeedback.IsAIFeature()
            .RequestFeedback(...);
```

src/System Application/App/Microsoft User Feedback/src/MicrosoftUserFeedbackImpl.Codeunit.al
```diff
@@ -0,0 +1,85 @@
...
```
darjoo: Put the namespace at the top import.

src/System Application/App/Microsoft User Feedback/src/MicrosoftUserFeedback.Codeunit.al
```diff
@@ -0,0 +1,157 @@
...
+    /// <summary>
+    /// Requests a 'like' (positive) feedback for a feature, optionally specifying if it is a Copilot feature and its area.
+    /// </summary>
+    /// <param name="FeatureName">The name of the feature for which like feedback is requested.</param>
+    /// <param name="IsAIFeature">Specifies if the feature is an AI feature.</param>
+    procedure RequestLikeFeedback(FeatureName: Text; IsAIFeature: Boolean)
+    begin
+        this.RequestLikeFeedback(FeatureName, IsAIFeature, '', '');
+    end;
+
```
darjoo, line 76: Empty line

src/System Application/App/Microsoft User Feedback/src/MicrosoftUserFeedback.Codeunit.al
edwardUL99: @onbuyuka, not sure how to close it off to 3rd party apps

src/System Application/App/Microsoft User Feedback/src/MicrosoftUserFeedbackImpl.Codeunit.al
darjoo: If you go this route, check with platform or whoever implemented the OCV backend that ISV tenants do not cause this to also send feedback. As they can bypass any of these checks.

edwardUL99: @darjoo can you elaborate on the ISV part and why they can bypass these checks?

darjoo: @edwardUL99 ISVs can customize baseapp and they can remove these checks. They will then be able to call these functions, which I do not know if we want to actually support. The only way to block that is on the backend (platform) which triggers the OCV calls from actually triggering.

src/System Application/App/Microsoft User Feedback/src/MicrosoftUserFeedback.Codeunit.al
edwardUL99: I've added checks that caller is microsoft only

src/System Application/App/Microsoft User Feedback/src/MicrosoftUserFeedbackImpl.Codeunit.al
edwardUL99: Good point, can create a follow up in platform for that

edwardUL99: @darjoo do you know what differentiates an ISV tenant? Are they what we host on the *ISOLATED components

src/System Application/App/Microsoft User Feedback/src/MicrosoftUserFeedbackImpl.Codeunit.al
```diff
@@ -0,0 +1,122 @@
...
+        this.CheckFeedbackCollectionAllowed();
+        Feedback.SurveyTriggerActivity(ActivityName);
+    end;
+
+    local procedure CheckFeedbackCollectionAllowed()
+    var
+        CallerModuleInfo: ModuleInfo;
+        CurrentModuleInfo: ModuleInfo;
+    begin
+        NavApp.GetCallerModuleInfo(CallerModuleInfo);
```
darjoo: This does not work, this will always be the same.

`NavApp.GetCallerModuleInfo(CallerModuleInfo);` needs to be in the facade and passed through to the implementation layer

src/System Application/App/Microsoft User Feedback/src/MicrosoftUserFeedback.Codeunit.al
```diff
@@ -0,0 +1,160 @@
...
+        this.RequestDislikeFeedback(FeatureName, '', '');
+    end;
+
+    /// <summary>
+    /// Requests a 'dislike' (negative) feedback for a feature, optionally specifying if it is a Copilot feature and its area.
+    /// </summary>
+    /// <param name="FeatureName">The name of the feature for which dislike feedback is requested.</param>
+    /// <param name="FeatureArea">The area or sub-area of the feature. ID of the sub-area on OCV.</param>
+    /// <param name="FeatureAreaDisplayName">The display name of the feature area.</param>
+
```
darjoo:
```suggestion
```

src/System Application/App/Microsoft User Feedback/src/MicrosoftUserFeedback.Codeunit.al
```diff
@@ -0,0 +1,160 @@
...
+    procedure RequestLikeFeedback(FeatureName: Text; FeatureArea: Text; FeatureAreaDisplayName: Text; ContextFiles: Dictionary of [Text, Text]; ContextProperties: Dictionary of [Text, Text])
+    begin
+        this.FeedbackImpl.RequestLikeFeedback(FeatureName, FeatureArea, FeatureAreaDisplayName, ContextFiles, ContextProperties);
+    end;
+
+    /// <summary>
+    /// Requests a 'dislike' (negative) feedback for a feature, optionally specifying if it is a Copilot feature and its area.
+    /// </summary>
+    /// <param name="FeatureName">The name of the feature for which dislike feedback is requested.</param>
+
```
darjoo:
```suggestion
```

src/System Application/App/Microsoft User Feedback/src/MicrosoftUserFeedback.Codeunit.al
```diff
@@ -0,0 +1,193 @@
...
+    /// Requests general feedback for a feature, optionally specifying if it is a Copilot feature and its area.
+    /// </summary>
+    /// <param name="FeatureName">The name of the feature for which feedback is requested.</param>
+    procedure RequestFeedback(FeatureName: Text)
+    var
+        CallerModuleInfo: ModuleInfo;
+        EmptyContextFiles: Dictionary of [Text, Text];
+        EmptyContextProperties: Dictionary of [Text, Text];
+    begin
+        this.FeedbackImpl.RequestFeedback(FeatureName, '', '', EmptyContextFiles, EmptyContextProperties, CallerModuleInfo);
```
darjoo, line 28: Missing `NavApp.GetCallerModuleInfo(CallerModuleInfo);`

src/System Application/App/Microsoft User Feedback/app.json
```diff
@@ -0,0 +1,33 @@
...
+  "version": "28.0.0.0",
+  "brief": "Provides functionality for collecting user feedback which is sent to Microsoft.",
+  "description": "Functionality to allow for AL developers to collect user feedback from the application. Since this module sends feedback to Microsoft, it should be used by Microsoft internal only.",
+  "privacyStatement": "https://go.microsoft.com/fwlink/?LinkId=724009",
+  "EULA": "https://go.microsoft.com/fwlink/?linkid=2009120",
+  "help": "",
+  "url": "https://go.microsoft.com/fwlink/?LinkId=724011",
+  "logo": "",
+  "dependencies": [
+    {
```
darjoo: You don't call any dotnet, this dependency is redundant.

src/System Application/App/Microsoft User Feedback/src/MicrosoftUserFeedbackImpl.Codeunit.al
```diff
@@ -0,0 +1,126 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Microsoft.UserFeedback;
```
darjoo: Lets use the same namespace as System.Feedback.

PR #4931, opened 2025-09-30
[E-Document][Peppol] - Move PEPPOL BIS 3.0 from BaseApp to the 1st-party App

GMatuleviciute:

**Implementation**

Created new PEPPOL and Tests-PEPPOL apps to move PEPPOL BIS 3.0 format out of Base app to a 1P app. Replaced the extensibility points with interfaces instead of events.

This initial implementation includes:

- Core codeunits for managing and validating PEPPOL documents 
- XMLports for exporting Sales Invoices and Credit Memos
- New enums and interfaces to provide extensibility
- Service support for handling Service Invoices and Credit Memos
- The implementation provides a framework for creating PEPPOL-compliant documents and ensures data integrity through validation rules.

You can find new app extensibility examples in extensibility_examples.md file (part of this pull request)

> [!IMPORTANT]
> This Pull request related with  https://github.com/microsoft/BusinessCentralApps/pull/1595

Fixes #

### DIFF
--- a/src/Apps/W1/PEPPOL/App/src/Management/Codeunits/PEPPOL30Management.Codeunit.al
+++ b/src/Apps/W1/PEPPOL/App/src/Management/Codeunits/PEPPOL30Management.Codeunit.al
@@ -218,9 +218,13 @@ codeunit 37200 "PEPPOL30 Management" implements "PEPPOL Attachment Handler"
     /// <param name="CompanyID">Returns the company VAT registration ID.</param>
     /// <param name="CompanyIDSchemeID">Returns the company ID scheme ID.</param>
     /// <param name="TaxSchemeID">Returns the tax scheme ID (e.g., VAT).</param>
+#if not CLEAN25
 #pragma warning disable AL0432
+#endif
     procedure GetAccountingSupplierPartyTaxSchemeBIS(var VATAmtLine: Record "VAT Amount Line"; var CompanyID: Text; var CompanyIDSchemeID: Text; var TaxSchemeID: Text)
+#if not CLEAN25
 #pragma warning restore AL0432
+#endif
     begin
         PEPPOLManagementImpl.GetAccountingSupplierPartyTaxSchemeBIS(VATAmtLine, CompanyID, CompanyIDSchemeID, TaxSchemeID);
     end;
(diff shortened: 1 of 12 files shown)


src/Apps/W1/PEPPOL/App/src/Setup/CompanyInformation.TableExt.al
```diff
@@ -0,0 +1,24 @@
...
+
+/// <summary>
+/// Table extension for Company Information to add PEPPOL e-document format selection.
+/// Extends the Company Information table with fields for configuring electronic document formats.
+/// </summary>
+tableextension 37200 "Company Information" extends "Company Information"
+{
+    fields
+    {
+        field(37200; "E-Document Format"; Enum "E-Document Format")
```
DanielGoehler, line 17: For what use case is this field intended?

Are there any plans to support multiple formats for a Business Central company?

**Context:**
From the perspective of our customers in Germany, two formats are relevant (EN 16931):

* **ZUGFeRD** (essentially [CII](https://en.wikipedia.org/wiki/UN/CEFACT)/XML format embedded in a PDF), widely used in German industry.
* **XRechnung** (a PEPPOL variant in [UBL](https://en.wikipedia.org/wiki/Universal_Business_Language)/XML format), required by government entities.

Many customer will need both. And for them with customers outside Germany but within the EU, these formats may be uncommon. An exception is **Factur-X** in France, which is the same format as ZUGFeRD. For broader EU invoicing, **PEPPOL** could become a good option for sending invoices to customers, even if—at least as far as I know—there are currently no near-term plans to make it mandatory. Still, it seems like a sensible approach.

CII

Groenbech96:
> Nice. Are there any plans to remove the old PepPol logic in the document sending profile?
> 
> image

Yes. That is indeed the intention here. 
To come :)

src/Apps/W1/PEPPOL/App/src/Setup/CompanyInformation.TableExt.al
Groenbech96: In this iteration it is picked up by the XML ports to determine what implementation to use for certain parts of the export.

Are there any plans to support multiple formats for a Business Central company?
- I would say we this already when you setup multiple services in a e-document workflow. Or how do you think on multiple formats?

DanielGoehler: Yes, I can create several workflows for each format, each connected to a different e-document service. I’m just wondering what this field is actually used for.

Groenbech96: Im not sure the field on CI is the right place to put this, but the enum allows partial/full override of the behavior when creating PEPPOL documents. 

Since XML ports are from baseapp is run using xmlport.run, then the impl would need to live somewhere, and allowed to be overwritten. Thats the idea anway.

DanielGoehler: @Groenbech96 I'm sorry, but I don't understand your response.
My question is: why is this field set up on *Company Information*?

My reasoning is: If you want to support multiple PEPPOL-based formats (such as XRechnung and PEPPOL 3.0) side by side, wouldn't it make more sense to place this on the *E-Document Service* table where the format is selected?

Groenbech96: @DanielGoehler great we look at it the same. 

Yes, i was thinking to add some API on the codeunits that call the XMLPorts, that would allow you to set that as you call it. So you could implement your own format that allow you to overwrite it with custom code.

I plan to update the PR with said changes soon.

src/Apps/W1/PEPPOL/App/src/Management/Codeunits/PEPPOL30Management.Codeunit.al
```diff
@@ -218,7 +218,9 @@ codeunit 37200 "PEPPOL30 Management" implements "PEPPOL Attachment Handler"
     /// <param name="CompanyID">Returns the company VAT registration ID.</param>
     /// <param name="CompanyIDSchemeID">Returns the company ID scheme ID.</param>
     /// <param name="TaxSchemeID">Returns the tax scheme ID (e.g., VAT).</param>
+#pragma warning disable AL0432
```
grobyns, line 222: you should add preprocessor symbols to remove the pragma when no longer needed.

JesperSchulz:
Re-triggering a build to check status.

PR #5075, opened 2025-10-07
[27.x] Changing 2 functions from internal to global

AydinKelky:
#### Summary
Backport of #5090.

Changing the following functions from internal to global

1. InitFromServiceCommitment in 8001 "Sub. Contract Renewal Line"
2. CreateCustomerContractLinesFromServiceCommitments in table 8052 "Customer Subscription Contract"

from internal to global

#### Work Item(s) 
Fixes [AB#610233](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/610233)


src/Apps/W1/Subscription Billing/App/Contract Renewal/Tables/SubContractRenewalLine.Table.al
```diff
@@ -194,7 +194,12 @@ table 8001 "Sub. Contract Renewal Line"
         OnAfterRefreshContractInfo(Rec, ServiceCommitment);
     end;
 
-    internal procedure InitFromServiceCommitment(var ServiceCommitment: Record "Subscription Line"): Boolean
+    /// <summary>
+    /// Initiates the subscription contract renewal line from subscription line
+    /// </summary>
+    /// <param name="ServiceCommitment">VAR Record "Subscription Line".</param>
```
pri-kise, line 200: `Return value of type Boolean.` doesn't provide any help.
I believe that a xml summary for a return value should provide information about what does the boolean indicates.
e.g. something like: _Returns true if the Rec was succesful initialized; Returns fals when a (planned) renewal line already exists_

Maybe you could revisit the other xml comments, too.

AydinKelky: Done :)

PR #5076, opened 2025-10-07
Update PowerBI Reports test with AL0432

Alexander-Ya:
#### Work Item(s) 
Fixes [AB#609027](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/609027)

### DIFF
--- a/src/Apps/W1/PowerBIReports/Test/E2EPowerBIManufacturingTest.Codeunit.al
+++ b/src/Apps/W1/PowerBIReports/Test/E2EPowerBIManufacturingTest.Codeunit.al
@@ -1,5 +1,6 @@
 #if not CLEAN28
 #pragma warning disable AL0801
+#pragma warning disable AL0801
 #endif
 // ------------------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


ventselartur (dismissed):

src/Apps/W1/PowerBIReports/Test/E2EPowerBIManufacturingTest.Codeunit.al
```diff
@@ -1,5 +1,6 @@
 #if not CLEAN28
 #pragma warning disable AL0801
+#pragma warning disable AL0801
```
mynjj: it was already disabled above, it's probably some other diagnostic missing

mynjj (changes requested):

src/Apps/W1/PowerBIReports/Test/E2EPowerBIManufacturingTest.Codeunit.al
Alexander-Ya: Updated

PR #5077, opened 2025-10-07
[E-Doc] Improve QR code render logic 

gggdttt:
#### Summary 

Test locally, we don't need to use AL to control the length of rendering.  The UI will automatically match it. So remove useless logic here.

#### Work Item(s) 
Fixes [AB#605043](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/605043)


gggdttt:
> Will you backport fix? @gggdttt

Yes, I shall backport it to 27.x

PR #5084, opened 2025-10-08
Configuration management module

darjoo:
#### Summary 
Addition of a new module, Configuration Management for retrieval of configuration values.
Limited to Microsoft usage at this time.

#### Work Item(s) 
Fixes [AB#602235](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/602235)

### DIFF
--- a/src/System Application/App/Configuration Management/app.json
+++ b/src/System Application/App/Configuration Management/app.json
@@ -10,7 +10,14 @@
   "help": "https://go.microsoft.com/fwlink/?linkid=2103698",
   "url": "https://go.microsoft.com/fwlink/?linkid=724011",
   "logo": "",
-  "dependencies": [],
+  "dependencies": [
+    {
+      "id": "7e3b999e-1182-45d2-8b82-d5127ddba9b2",
+      "publisher": "Microsoft",
+      "name": "DotNet Aliases",
+      "version": "28.0.0.0"
+    }
+  ],
   "internalsVisibleTo": [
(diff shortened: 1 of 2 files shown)


src/System Application/App/Configuration Management/src/ConfigurationManagement.Codeunit.al
```diff
@@ -0,0 +1,35 @@
...
+
+    var
+        ConfigurationManagementImpl: Codeunit "Configuration Management Impl.";
+
+    /// <summary>
+    /// Gets the configuration value for the specified key.
+    /// </summary>
+    /// <param name="ConfigKey">The configuration key to look up.</param>
+    /// <returns>The configuration value associated with the specified key.</returns>
+    procedure GetConfiguration(ConfigKey: Text): Text
```
Groenbech96, line 26: Since this is Microsoft only, why not use onprem scope? 

I get the extension publisher, but i don't think it is needed. 

Look at activity log module for reference

mazhelez (approved):

src/System Application/App/Configuration Management/src/ConfigurationManagement.Codeunit.al
darjoo: If we make it onprem, non-onprem scoped apps won't be able to use this.

PR #5091, opened 2025-10-08
AOAI Policy configuration

dchourasia30:
#### Summary 
Add support for Policy configuration via AOAI Policy Params.
Support for enabling/disabling XPIA detection.

#### Work Item(s) 
Fixes [AB#561651](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/561651)

Some regressions noted in PA-Harms test for UPIA datasets (4 datasets). I have manually reviewed 5 failing cases from each of the failing dataset and it looks to be False positive. So we should be good to go with this.


src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIChatMessagesImpl.Codeunit.al
```diff
@@ -258,6 +258,12 @@ codeunit 7764 "AOAI Chat Messages Impl"
         MessagesTokenCount := AOAIToken.GetGPT4TokenCount(TotalMessages);
     end;
 
+    [NonDebuggable]
+    procedure EnforceXPIADetection(var Input: Text)
```
qutreson: We should move this into the platform symbols rather than here.

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicyParams.Codeunit.al
```diff
@@ -0,0 +1,54 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+namespace System.AI;
+
+/// <summary>
+/// Represents the AOAI Policy management for combining Harms Severity and XPIA Detection settings.
+/// </summary>
+codeunit 7787 "AOAI Policy Params"
+{
+    Access = Public;
```
qutreson: Access = Public is the default value, no need to mention it

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicyHarmsSeverity.Enum.al
```diff
@@ -0,0 +1,13 @@
+namespace System.AI;
+
+enum 7787 "AOAI Policy Harms Severity"
+{
+    Extensible = true;
```
qutreson: I don't think we have an extensibility story defined that would allow partners to define their own level of harm severity. We should make this not extensible IMO.

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicyHarmsSeverity.Enum.al
```diff
@@ -0,0 +1,13 @@
...
+
+    value(1; Low)
```
qutreson, line 11: You should add comments explaining what Low & Medium means

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicyParams.Codeunit.al
```diff
@@ -0,0 +1,54 @@
...
+    procedure SetHarmsSeverity(HarmsSeverity: Enum "AOAI Policy Harms Severity")
+    begin
+        AOAIPolicyParamsImpl.SetHarmsSeverity(HarmsSeverity);
+    end;
+
+    /// <summary>
+    /// Sets the AOAI Policy XPIA Detection.
+    /// </summary>
+    /// <param name="XPIADetection">The AOAI Policy XPIA Detection to set.</param>
+    procedure SetXPIADetection(XPIADetection: Enum "AOAI Policy XPIA Detection")
```
qutreson: Do we need an enum here? Enabled/Disabled could be replaced with a boolean instead

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicyParams.Codeunit.al
```diff
@@ -0,0 +1,54 @@
...
+    /// </summary>
+    /// <param name="HarmsSeverity">The AOAI Policy Harms Severity to set.</param>
+    procedure SetHarmsSeverity(HarmsSeverity: Enum "AOAI Policy Harms Severity")
+    begin
+        AOAIPolicyParamsImpl.SetHarmsSeverity(HarmsSeverity);
+    end;
+
+    /// <summary>
+    /// Sets the AOAI Policy XPIA Detection.
+    /// </summary>
```
qutreson, line 47: I would mention here in the docs that messages must be marked for XPIA detection and link to the new method EnforceXPIADetection in the AOAI Chat Message codeunit.

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicyParamsMapping.Codeunit.al
```diff
@@ -0,0 +1,29 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.AI;
+
+/// <summary>
+/// Interface for mapping AOAI policy parameters to AOAI policy enums.
+/// </summary>
+codeunit 7777 "AOAI Policy Params Mapping"
```
qutreson: Do we need to expose this codeunit and the enum? Can't it be an internal method within the AOAIPolicyParams codeunit? I guess our main recommendation will be to use the AOAIPolicyParam to define the type of validation, so we don't really have a need to expose the enum. The enum seems more like an internal concept not something that needs to be exposed.

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicy.Enum.al
```diff
@@ -0,0 +1,38 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+namespace System.AI;
+
+enum 7789 "AOAI Policy"
+{
+    Extensible = false;
```
qutreson, line 9: See my other comment, I wouldn't expose these, they could be internal only.

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIChatMessagesImpl.Codeunit.al
dchourasia30: As discussed we can keep it here

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicyParams.Codeunit.al
dchourasia30: Made the changes

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicyHarmsSeverity.Enum.al
dchourasia30: Made it non-extensible

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicyParams.Codeunit.al
dchourasia30: As discussed, made the changes

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicyParamsMapping.Codeunit.al
dchourasia30: As discussed, made the changes

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicy.Enum.al
dchourasia30: Made the changes

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicyHarmsSeverity.Enum.al
dchourasia30: Added the same

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicyHarmsSeverity.Enum.al
```diff
@@ -0,0 +1,20 @@
+namespace System.AI;
+
+enum 7788 "AOAI Policy Harms Severity"
+{
+    Extensible = false;
+
+    /// <summary>
+    /// Applies the strictest policy controls.
```
qutreson: Maybe mentions that this is the recommended approach

dchourasia30: made the changes

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIChatMessages.Codeunit.al
```diff
@@ -403,6 +403,15 @@ codeunit 7763 "AOAI Chat Messages"
         exit(AOAIToolsImpl.GetToolInvokePreference());
     end;
 
+    /// <summary>
+    /// Enforce XPIA Detection for an Input text.
+    /// </summary>
+    [NonDebuggable]
+    procedure EnforceXPIADetection(var Input: Text)
```
encimita: What does Enforce mean? This function by itself doesn't enforce anything. Consider naming it "AddXPIADetectionTags" or "PrepareTextForXPIADetection" or something else that makes it clear that it acts on the text rather than directly enforcing anything.

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicy.Enum.al
```diff
@@ -0,0 +1,39 @@
...
+
+    /// <summary>
+    /// Medium harms severity with XPIA detection enabled
+    /// </summary>
+    value(3; "MediumWithXPIA")
+    {
+    }
+
+    /// <summary>
+    /// Medium harms severity with XPIA detection disabled
```
encimita: The default value to my understanding is passed on directly on all the way. So I would rename this comment and say "Let the system decide the default policy to apply"

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
```diff
@@ -281,6 +282,10 @@ codeunit 7772 "Azure OpenAI Impl" implements "AI Service Name"
         CopilotCapabilityImpl.AddTelemetryCustomDimensions(CustomDimensions, CallerModuleInfo);
 
         AOAIChatCompletionParams.AddChatCompletionsParametersToPayload(Payload);
+
+        AzureOpenAIPolicy := Payload.GetText('azureOpenAIPolicy');
```
msft-sam: It would be cleaner to add a getter method to the params rather than extracting and removing it from the payload, i.e.
`AzureOpenAIPolicy := AOAIChatCompletionParams.GetAOAIPolicyParams().GetAOAIPolicy()`

I would avoid adding it to the payload as it is technically not part of the AOAI payload so it is a bit hacky to add and then remove it

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicyParams.Codeunit.al
```diff
@@ -0,0 +1,78 @@
...
+    /// <param name="IsEnabled">Enable/Disable AOAI Policy XPIA Detection</param>
+    /// <remarks>When XPIA detection is enabled, use <see cref="AOAI Chat Messages.EnforceXPIADetection"/> to mark messages for XPIA detection.</remarks>
+    procedure SetXPIADetection(IsEnabled: Boolean)
+    begin
+        AOAIPolicyParamsImpl.SetXPIADetection(IsEnabled);
+    end;
+
+    /// <summary>
+    /// Gets a Custom AOAI policy.
```
encimita: Gets the value for the custom AOAI policy, if set using SetCustomAOAIPolicy

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIChatComplParamsImpl.Codeunit.al
```diff
@@ -158,5 +179,10 @@ codeunit 7762 "AOAI Chat Compl Params Impl"
         SetMaxTokens(0);
         SetMaxHistory(10);
         SetJsonMode(false);
+
+        DefaultPolicyParams.SetHarmsSeverity("AOAI Policy Harms Severity"::Low);
+        DefaultPolicyParams.SetXPIADetection(true);
```
msft-sam: Given that we don't do this today, it would be a breaking change for all existing consumers and could affect their prompt behavior - is that intentional? (perhaps we want to mark something as obsolete to inform partners that they should move to a new default)

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIChatComplParamsImpl.Codeunit.al
```diff
@@ -158,5 +179,10 @@ codeunit 7762 "AOAI Chat Compl Params Impl"
         SetMaxTokens(0);
         SetMaxHistory(10);
         SetJsonMode(false);
+
+        DefaultPolicyParams.SetHarmsSeverity("AOAI Policy Harms Severity"::Low);
```
msft-sam: Minor: it could be nice to move the defaulting of policy params to within the codeunit

then here you would just call `DefaultPolicyParams.InitializeDefaults()`, that way if you are just making a change to the `AOAI Policy Params` codeunit you only need to remember to update the procedure there, rather than remembering you need to update this file as well

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIChatMessages.Codeunit.al
msft-sam: +1 to `AddXPIADetectionTags`

Or even to just `AddDocumentTags` and explain in the doc comment this is used for XPIA detection (maybe explain what XPIA is)

Would be nice to add some remarks to remember to configure the appropriate AOAI policy to actually apply the XPIA policy to the request (could we emit a warning/error if this is not done?), and also mention that this does not add it to a message, you still have to do that yourself

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIChatMessagesImpl.Codeunit.al
```diff
@@ -258,6 +258,12 @@ codeunit 7764 "AOAI Chat Messages Impl"
...
+    begin
+        Input := '"""<documents>' + Input + '</documents>""" End';
```
msft-sam, line 263: Is the `End` required here?

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicy.Enum.al
```diff
@@ -0,0 +1,39 @@
...
+
+    /// <summary>
+    /// Low harms severity with XPIA detection enabled
+    /// </summary>
+    value(1; "ConservativeWithXPIA")
+    {
+    }
+
+    /// <summary>
+    /// Low harms severity with XPIA detection disabled
```
msft-sam: `Low harms severity` is a bit ambiguous

Do you mean the harms severity is low, i.e. we have a high tolerance for harms, or that the harms with a low severity are blocked?

I would use a few more words just to clarify, i.e. `Requests containing harms with a low severity or higher are blocked`, and only mention the XPIA detection on the policies that include it

Consider also linking to the AOAI documentation that includes the definition of a `low harm`:
https://learn.microsoft.com/en-us/azure/ai-foundry/openai/concepts/content-filter-severity-levels?tabs=definitions#hate-and-fairness-severity-levels

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicyHarmsSeverity.Enum.al
```diff
@@ -0,0 +1,20 @@
+namespace System.AI;
+
+enum 7788 "AOAI Policy Harms Severity"
+{
+    Extensible = false;
+
+    /// <summary>
+    /// Recommended. Applies the strictest policy controls.
```
msft-sam: Same as previous comment, link out to the docs that defines what these are:
https://learn.microsoft.com/en-us/azure/ai-foundry/openai/concepts/content-filter-severity-levels?tabs=definitions#hate-and-fairness-severity-levels

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicyParamsImpl.Codeunit.al
```diff
@@ -0,0 +1,106 @@
...
+    var
+        AOAIPolicyHarmsSeverity: Enum "AOAI Policy Harms Severity";
+        AOAIPolicyXPIADetection: Boolean;
+        CombinationKey: Text;
+    begin
+        AOAIPolicyHarmsSeverity := GetHarmsSeverity();
+        AOAIPolicyXPIADetection := GetXPIADetection();
+
+        // Create readable combination key
+        CombinationKey := 'Harms' + Format(AOAIPolicyHarmsSeverity) + '|XPIA' + (AOAIPolicyXPIADetection = true ? 'Enabled' : 'Disabled');
```
msft-sam, line 89: It would be good to add a test to ensure this is appropriately resolved, does this code still work if the severity is localized?

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicyHarmsSeverity.Enum.al
dchourasia30: made the changes

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicy.Enum.al
dchourasia30: Users won't have access to the AOAI policy enums directly. They are supposed to work with the AOAI Policy Params and internally we map it to AOAI policy enum. Have updated the comments anyways.

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIChatMessagesImpl.Codeunit.al
dchourasia30: It is not required but added it on Quentin's suggestion. Anyways this bug would be fixed in Dec release and we can omit putting trailing characters entirely after that

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIChatComplParamsImpl.Codeunit.al
dchourasia30: Made the Changes

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
dchourasia30: Made the changes

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicy.Enum.al
dchourasia30: The AOAI policy enum  is not exposed to the end user and instead mapped internally based on AOAI Policy Params.
Also this **Default** enum is different from the default AOAI policy that we recommend. Have updated the comments to reflect that

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIChatMessages.Codeunit.al
dchourasia30: I don't see a way to emit a warning for this. meanwhile I have added required comments/remarks

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicy.Enum.al
```diff
@@ -0,0 +1,41 @@
...
+    /// <summary>
+    /// Requests containing harms with a medium severity are blocked and XPIA detection is enabled.
+    /// </summary>
+    value(3; "MediumWithXPIA")
+    {
+    }
+
+    /// <summary>
+    /// Requests containing harms with a medium severity are blocked.
+    /// This is not the default policy that we recommend but rather this is the Default AOAI policy (Policy ID : 0 ) in CAPI
```
dchourasia30: review the comment

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicy.Enum.al
```diff
@@ -0,0 +1,41 @@
...
+enum 7787 "AOAI Policy"
+{
+    Extensible = false;
+    Access = Internal;
+
+    /// <summary>
+    /// Requests containing harms with a low severity are blocked and XPIA detection is enabled.
+    /// </summary>
+    value(1; "ConservativeWithXPIA")
+    {
```
dchourasia30, line 16: lock it

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicyParamsImpl.Codeunit.al
```diff
@@ -0,0 +1,106 @@
...
+                exit("AOAI Policy"::"Default");
+        end;
+    end;
+
+    procedure InitializeDefaults()
+    begin
+        Initialized := true;
+        HarmsSeverity := "AOAI Policy Harms Severity"::Low;
+        IsXPIADetectionEnabled := true;
+        CustomAOAIPolicy := '';
```
dchourasia30, line 111: maybe change it to medium and false

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicyParamsImpl.Codeunit.al
dchourasia30: as discussed , this wont be necessary after locking the enums

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicy.Enum.al
```diff
@@ -0,0 +1,45 @@
...
+    value(3; "MediumWithXPIA")
+    {
+        Caption = 'MediumWithXPIA', Locked = true;
+    }
+
+    /// <summary>
+    /// Requests containing harms with a medium severity are blocked.
+    /// This is not the default policy that we recommend but rather this is the Default AOAI policy in CAPI
+    /// </summary>
+    value(4; "Default")
```
msft-sam: Consider calling the property "MediumWithoutXPIA", while keeping the "Default" caption, this will make it clear from the code what this value does

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicy.Enum.al
```diff
@@ -0,0 +1,45 @@
...
+    /// Requests containing harms with a medium severity are blocked and XPIA detection is enabled.
+    /// </summary>
+    value(3; "MediumWithXPIA")
+    {
+        Caption = 'MediumWithXPIA', Locked = true;
+    }
+
+    /// <summary>
+    /// Requests containing harms with a medium severity are blocked.
+    /// This is not the default policy that we recommend but rather this is the Default AOAI policy in CAPI
```
msft-sam: Don't think this comment is needed if we rename the property

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicyParamsImpl.Codeunit.al
```diff
@@ -0,0 +1,106 @@
...
+
+        case CombinationKey of
+            'HarmsLow|XPIAEnabled':
+                exit("AOAI Policy"::"ConservativeWithXPIA");
+            'HarmsLow|XPIADisabled':
+                exit("AOAI Policy"::"ConservativeWithoutXPIA");
+            'HarmsMedium|XPIAEnabled':
+                exit("AOAI Policy"::"MediumWithXPIA");
+            else
+                exit("AOAI Policy"::"Default");
```
msft-sam: Should we cover all the cases here explicitly (i.e. add `HarmsMedium|XPIADisabled`) and for the `else` case emit some telemetry and fallback to the default?

This way if a new combination is added but it is not updated here we will observe that through telemetry, but also provide "medium" harms detection by default

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicy.Enum.al
dchourasia30: good point, made the changes

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicy.Enum.al
dchourasia30: removed the extra comment

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicyParamsImpl.Codeunit.al
dchourasia30: Handled `HarmsMedium|XPIADisabled` explicitly. For the else case:  I am falling back to 'ConservativeWithXPIA' (the default policy nowonwards).
I am not sure about the telemetry thing, how to do that?

dchourasia30: Added telemetry as Normal

msft-sam (dismissed):

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIPolicyParams.Codeunit.al
```diff
@@ -0,0 +1,83 @@
...
+
+    /// <summary>
+    /// Sets a Custom AOAI policy.
+    /// </summary>
+    procedure SetCustomAOAIPolicy(CustomAOAIPolicyParams: Text)
+    begin
+        AOAIPolicyParamsImpl.SetCustomAOAIPolicy(CustomAOAIPolicyParams);
+    end;
+
+    procedure InitializeDefaults()
```
qutreson, line 74: NIT: missing function header

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIChatMessagesImpl.Codeunit.al
```diff
@@ -258,6 +258,12 @@ codeunit 7764 "AOAI Chat Messages Impl"
         MessagesTokenCount := AOAIToken.GetGPT4TokenCount(TotalMessages);
     end;
 
+    [NonDebuggable]
```
qutreson: NIT: I'm wondering if we really need the NonDebuggable? It's easy to guess what this method did by just inspecting the input before/after.

qutreson (dismissed):

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIChatMessagesImpl.Codeunit.al
dchourasia30: Right, we don't need it now

aholstrup1 (approved):

src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIChatComplParamsImpl.Codeunit.al
```diff
@@ -11,13 +11,15 @@ codeunit 7762 "AOAI Chat Compl Params Impl"
...
     var
+        AOAIPolicyParams: Codeunit "AOAI Policy Params";
         Initialized: Boolean;
         Temperature: Decimal;
         MaxTokens: Integer;
         JsonMode: Boolean;
         MaxHistory: Integer;
         PresencePenalty: Decimal;
         FrequencyPenalty: Decimal;
+
```
darjoo:
```suggestion
```

darjoo (dismissed):

darjoo (approved):

PR #5119, opened 2025-10-10
Deprecate old NSAT surveys

onbuyuka:
#### Summary 
Deprecate old NSAT surveys

#### Work Item(s) 
Fixes [AB#609746](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/609746)


src/System Application/App/ControlAddIns/src/SatisfactionSurveyAsync.ControlAddIn.al
```diff
@@ -1,4 +1,4 @@
-// ------------------------------------------------------------------------------------------------
+#if not CLEAN28// ------------------------------------------------------------------------------------------------
```
JesperSchulz: Don't mess with my OCD 🤣
```suggestion
#if not CLEAN28
// ------------------------------------------------------------------------------------------------
```

darjoo (approved):

PR #5146, opened 2025-10-13
[MCP] Allow microsoft publisher API pages as tools from code

onbuyuka:
#### Summary 
Allow microsoft publisher API pages as tools from code

#### Work Item(s) 
Fixes [AB#609863](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/609863)

### DIFF
--- a/src/System Application/App/MCP/src/Configuration/Codeunits/MCPConfig.Codeunit.al
+++ b/src/System Application/App/MCP/src/Configuration/Codeunits/MCPConfig.Codeunit.al
@@ -55,17 +55,7 @@ codeunit 8350 "MCP Config"
     /// <param name="Allow">True to allow create, update and delete tools, false to disallow.</param>
     procedure AllowCreateUpdateDeleteTools(ConfigId: Guid; Allow: Boolean)
     begin
-        MCPConfigImplementation.AllowProdChanges(ConfigId, Allow);
-    end;
-
-    /// <summary>
-    /// Allows production changes for the specified MCP configuration.
-    /// </summary>
-    /// <param name="ConfigId">The SystemId (GUID) of the configuration.</param>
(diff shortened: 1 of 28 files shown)


src/System Application/App/MCP/src/Configuration/Codeunits/MCPConfig.Codeunit.al
```diff
@@ -116,7 +116,18 @@ codeunit 8350 "MCP Config"
...
+        exit(MCPConfigImplementation.CreateAPITool(ConfigId, APIPageId, true));
+    end;
+
+    /// <summary>
+    /// Creates a new API tool for the specified configuration and API page. It does not validate the publisher of the API page.
+    /// </summary>
+    /// <param name="ConfigId">The SystemId (GUID) of the configuration.</param>
+    /// <param name="APIPageId">The ID of the API page.</param>
+    /// <returns>The SystemId (GUID) of the created tool.</returns>
+    procedure CreateAPIToolWithoutAPIPublisherValidation(ConfigId: Guid; APIPageId: Integer): Guid
```
darjoo: Why do we need this? Is there a valid reason for bypassing validation?

Why block above and this one unblocks?
If we want to allow both and the default is to validate, do an overload `CreateAPITool(ConfigId: Guid; APIPageId: Integer; ValidateAPIPublisher: Boolean)`

If this is something only internal needs, consider blocking via OnPrem or app publisher check. But do consider if it must be internal.

onbuyuka: very very fair point. added overload

JesperSchulz (approved):

PR #5155, opened 2025-10-14
[Transaction Storage] Replace dotnet usage with AL implementation.

darjoo:
#### Summary 
Transaction storage uses dotnet `Convert` to convert hexadecimal into decimal. 
Replacing this dotnet usage with an AL implementation.

Tested the implementation against the dotnet `Convert.ToInt32(Input, 16)` with a local test.

Test not included in PR as it's a local method and I don't want to make it internal just to write a test. 
Testing the `CalcTenantExportStartTime` method isn't an option as it requires the aadtenantid to be set and it cannot be controlled.

The testing:
```
    procedure Test()
    var
        alphanumericLbl: Label '0123456789abcdef', Locked = true;
        twochars: Text[2];
        idx1, idx2 : Integer;
    begin
        for idx1 := 1 to StrLen(alphanumericLbl) do
            for idx2 := 1 to StrLen(alphanumericLbl) do begin
                twochars := CopyStr(alphanumericLbl, idx1, 1) + CopyStr(alphanumericLbl, idx2, 1);
                if HexToDecimal(twochars) <> DotNetConvert(twochars) then
                    Error('Mismatch for %1: %2 vs %3', twochars, HexToDecimal(twochars), DotNetConvert(twochars));
            end;
        Message('All tests passed');
    end;

    local procedure DotNetConvert(HexString: Text[2]): Integer
    var
        Convert: DotNet Convert;
    begin
        exit(Convert.ToInt32(HexString, 16));
    end;
```

#### Work Item(s) 
Fixes [AB#605002](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/605002)

### DIFF
--- a/src/Apps/W1/TransactionStorage/App/src/TransactStorageExport.Codeunit.al
+++ b/src/Apps/W1/TransactionStorage/App/src/TransactStorageExport.Codeunit.al
@@ -149,7 +149,7 @@ codeunit 6203 "Transact. Storage Export"
         exit(true);
     end;
 
-    local procedure HexToDecimal(HexString: Text[2]) DecimalValue: Integer
+    local procedure HexToDecimal(HexString: Text) DecimalValue: Integer
     var
         CharValue: Integer;
         i: Integer;


src/Apps/W1/TransactionStorage/App/src/TransactStorageExport.Codeunit.al
```diff
@@ -151,6 +149,33 @@ codeunit 6203 "Transact. Storage Export"
...
+        for i := 1 to StrLen(HexString) do begin
+            CurrentChar := HexString[i];
+
+            case CurrentChar of
+                '0' .. '9':
+                    CharValue := CurrentChar - 48; // ASCII value of '0' is 48
+                'a' .. 'f':
+                    CharValue := CurrentChar - 87; // ASCII value of 'a' is 97, we want 10 for 'a'
+                else
+                    CharValue := 0;
```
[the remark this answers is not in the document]
darjoo: Input is validated prior to parsing in `IsHexString`. Won't fix.

src/Apps/W1/TransactionStorage/App/src/TransactStorageExport.Codeunit.al
```diff
@@ -151,6 +149,33 @@ codeunit 6203 "Transact. Storage Export"
         exit(true);
     end;
 
+    local procedure HexToDecimal(HexString: Text) DecimalValue: Integer
```
JesperSchulz, line 152: Strictly speaking such helper procedures should be placed in some public module, for other modules and apps to consume. But I guess we can always refactor when another app needs this.

darjoo: We don't have a module like this as of yet and I'm not entirely sure what we'd like to call it. Maybe `Convert` would be it, but for a broader discussion when we need to build it. 
Will leave as is and refactor into the new module when we so decide we need it.

PR #5169, opened 2025-10-15
[Cryptography Management] Fix instability test

darjoo:
#### Summary 
When encrypting with OAEP, PKCS should not be able to decrypt it. However the padding can randomly be correct.

Updated the test to ensure we validate the output and ensure it is garbage data decrypted when it is decrypting with a different padding.

#### Work Item(s) 
Fixes [AB#609358](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/609358)

### DIFF
--- a/src/System Application/Test/Cryptography Management/src/RSATest.Codeunit.al
+++ b/src/System Application/Test/Cryptography Management/src/RSATest.Codeunit.al
@@ -242,14 +242,15 @@ codeunit 132617 "RSA Test"
 
         // [WHEN] Decrypt encrypted text stream using OAEP Padding
         DecryptingTempBlob.CreateOutStream(DecryptedOutStream);
-        DecryptionFailed := not TryDecrypt(RSA, PrivateKeyXmlStringSecret, EncryptedInStream, true, DecryptedOutStream);
+        DecryptionFailed := not TryDecryptWithOaepPadding(RSA, PrivateKeyXmlStringSecret, EncryptedInStream, DecryptedOutStream);
 
         // [THEN] Either decryption fails with an exception, or the decrypted text is garbage (not equal to plaintext)
         if not DecryptionFailed then begin
             DecryptingTempBlob.CreateInStream(DecryptedInStream);
             DecryptedText := Base64Convert.FromBase64(Base64Convert.ToBase64(DecryptedInStream));
-            LibraryAssert.AreNotEqual(PlainText, DecryptedText, 'Decryption with wrong padding should fail or return garbage data.');
-        end;
+            LibraryAssert.AreNotEqual(PlainText, DecryptedText, 'Decryption failed with garbage data.');
+        end else
+            LibraryAssert.IsTrue(DecryptionFailed, 'Decryption failed with wrong padding.');
     end;
 
     [Test]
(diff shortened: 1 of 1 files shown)


Groenbech96 (approved):

onbuyuka (approved):

src/System Application/Test/Cryptography Management/src/RSATest.Codeunit.al
```diff
@@ -251,22 +263,40 @@ codeunit 132617 "RSA Test"
...
+        // [THEN] Either decryption fails with an exception, or the decrypted text is garbage (not equal to plaintext)
+        if not DecryptionFailed then begin
+            DecryptingTempBlob.CreateInStream(DecryptedInStream);
+            DecryptedText := Base64Convert.FromBase64(Base64Convert.ToBase64(DecryptedInStream));
+            LibraryAssert.AreNotEqual(PlainText, DecryptedText, 'Decryption with wrong padding should fail or return garbage data.');
+        end;
+    end;
+
+    [TryFunction]
+    local procedure TryDecrypt(RSA: Codeunit RSA; XmlString: SecretText; EncryptedInStream: InStream; OaepPadding: Boolean; DecryptedOutStream: OutStream)
```
grobyns: it is better to have 2 named helpers than have a boolean parameter

src/System Application/Test/Cryptography Management/src/RSATest.Codeunit.al
```diff
@@ -222,22 +222,34 @@ codeunit 132617 "RSA Test"
...
         RSA.Encrypt(PrivateKeyXmlStringSecret, EncryptingInStream, false, EncryptedOutStream);
         EncryptedTempBlob.CreateInStream(EncryptedInStream);
 
         // [WHEN] Decrypt encrypted text stream using OAEP Padding
         DecryptingTempBlob.CreateOutStream(DecryptedOutStream);
-        asserterror RSA.Decrypt(PrivateKeyXmlStringSecret, EncryptedInStream, true, DecryptedOutStream);
+        DecryptionFailed := not TryDecrypt(RSA, PrivateKeyXmlStringSecret, EncryptedInStream, true, DecryptedOutStream);
+
+        // [THEN] Either decryption fails with an exception, or the decrypted text is garbage (not equal to plaintext)
+        if not DecryptionFailed then begin
```
grobyns, line 248: there's no verification if encryption fails, no check whether the error is the expected one...

src/System Application/Test/Cryptography Management/src/RSATest.Codeunit.al
```diff
@@ -222,22 +222,34 @@ codeunit 132617 "RSA Test"
...
+            DecryptingTempBlob.CreateInStream(DecryptedInStream);
+            DecryptedText := Base64Convert.FromBase64(Base64Convert.ToBase64(DecryptedInStream));
+            LibraryAssert.AreNotEqual(PlainText, DecryptedText, 'Decryption with wrong padding should fail or return garbage data.');
```
grobyns: your message here should only be about garbage data no? if encryption fails you will never get to this assert

haoranpb (approved):

JesperSchulz (approved):

PR #5208, opened 2025-10-17
[Shopify] Change fulfillment service callback URL and update existing one on open shop

onbuyuka:
#### Summary 
Change fulfillment service callback URL and update existing ones on open shop

#### Work Item(s) 
Fixes [AB#609960](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/609960)

### DIFF
--- a/src/Apps/W1/Shopify/App/src/Base/Tables/ShpfyShop.Table.al
+++ b/src/Apps/W1/Shopify/App/src/Base/Tables/ShpfyShop.Table.al
@@ -805,15 +805,20 @@ table 30102 "Shpfy Shop"
             ToolTip = 'Specifies if fully fulfilled Shopify orders should be created as sales invoices.';
             InitValue = true;
         }
-#if not CLEANSCHEMA28
+#if not CLEANSCHEMA31
         field(206; "Fulfillment Service Updated"; Boolean)
         {
             Caption = 'Fulfillment Service Updated';
             DataClassification = SystemMetadata;
             Description = 'Indicates whether the Shopify Fulfillment Service has been updated to the latest version.';
             ObsoleteReason = 'This field is no longer used.';
+#if CLEAN28
+            ObsoleteState = Removed;
+            ObsoleteTag = '31.0';
+#else
             ObsoleteState = Pending;
             ObsoleteTag = '28.0';
+#endif
         }
 #endif
     }


src/Apps/W1/Shopify/App/src/GraphQL/Codeunits/ShpfyGQLLocation.Codeunit.al
```diff
@@ -0,0 +1,33 @@
...
+/// Implements the IGraphQL interface for retrieving Shopify location using GraphQL.
+/// </summary>
+codeunit 30411 "Shpfy GQL Location" implements "Shpfy IGraphQL"
+{
+    Access = Internal;
+
+    /// <summary>
+    /// GetGraphQL.
+    /// </summary>
+    /// <returns>Return value of type Text.</returns>
```
darjoo, line 19: These docs are useless

src/Apps/W1/Shopify/App/src/Inventory/Tables/ShpfyShopLocation.Table.al
```diff
@@ -147,6 +147,20 @@ table 30113 "Shpfy Shop Location"
...
             end;
         }
+        field(13; "Fulfillment Service Id"; BigInteger)
+        {
+            Caption = 'Fulfillment Service Id';
+            DataClassification = SystemMetadata;
+            Editable = false;
+            Description = 'The Id of the fulfillment service in Shopify.';
+        }
+        field(14; "Fulfillment Srv. Callback Url"; Text[500])
```
darjoo, line 157: Can we make this internal? 
If this is external, someone else can modify this url. Or do we have validation checks for what this can actually be.

darjoo: Unless this url does not matter.

onbuyuka: This is only a representation, changing this will not change the actual URL in Shopify. That is controlled by the codeunit

src/Apps/W1/Shopify/App/src/GraphQL/Codeunits/ShpfyGQLLocation.Codeunit.al
onbuyuka: Yeah I have a cleanup task to do

src/Apps/W1/Shopify/App/src/Inventory/Tables/ShpfyShopLocation.Table.al
WaelAbuSeada: What does this mean? Is the url consumed in our codeunits at all? Or how are we using the value form that field?

onbuyuka: it is used for pulling only. we show list of locations from Shopify in BC. but we do not sync it back to Shopify. Only time we sync this Shopify is not from this field, we have a constant value in a codeunit for our callback url.

WaelAbuSeada: We are trying to move away from adding urls as fields in the table. Is that the full path url? does it need to be populated by the user? do we know about it from somewhere else and can build it ourselves?

onbuyuka: It is full url. It is not populated by the user. It is not exposed to the user. It is only pulled from Shopify, that's it. No other usage. Even if it is changed it has no effect.

PR #5215, opened 2025-10-19
Bug 609291: Invoice discount fails to finalize

ventselartur:
Setting the "Allow Inv. Disc" to true for G/L accounts if total discount is set in the header. For other types like items and fixed assets i suppose that users need to set the proper value in the master data, while for g/l accounts it is always false.

[AB#609291](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/609291)


Groenbech96:
Maybe as AI to write a test for the scenario :)

mynjj (approved):

PR #5218, opened 2025-10-20
[Config Mgt][Test] Add test library for controlling control or treatment allocation in tests

darjoo:
#### Summary 
In tests, for now specifically AI tests, we are unable to control the allocation which makes it difficult to run AI Tests on both control & treatment.

This gives AI Tests the possibility of controlling which allocation they want.

#### Work Item(s) 
Fixes [AB#609197](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/609197)

### DIFF
--- a/.github/AL-Go-Settings.json
+++ b/.github/AL-Go-Settings.json
@@ -6,7 +6,7 @@
   "runs-on": "windows-latest",
   "cacheImageName": "",
   "UsePsSession": false,
-  "artifact": "bcinsider/Sandbox/28.0.41125.0//latest",
+  "artifact": "bcinsider/Sandbox/28.0.41171.0//latest",
   "country": "base",
   "useProjectDependencies": true,
   "repoVersion": "28.0",


src/System Application/Test Library/Configuration Management/app.json
```diff
@@ -0,0 +1,36 @@
+{
+  "id": "d4dcc3bc-265b-48fc-99e2-3262a6130236",
+  "name": "Configuration Management Test Library",
```
grobyns, line 3: I think this name is too vague... configuration of what? Configuration Management could be anything... sounds like you should rename that module too

darjoo: It's meant to be a configuration management of anything, that we specifically control, unlike feature management that customers decide if they want to enable/disable. A fixed config value (true/false/A/B/whatever), A/B experimentation values, etc.

Groenbech96: Control Treatment Test Library?

darjoo: It's the test library for the module, and I'll leave it as such.
I'll take it offline with Gert regarding the sys app module name, and if it will be changed, I'll make the change to this at the same time.

PR #5303, opened 2025-10-27
Adding configuration in PO matching to receive at posting of the invoice

mynjj:
# Summary 
Implementation of configuration settings for what to do when getting a new invoice and the underlying matched order has not yet been marked as received. There are 5 different configurations a user can select, found in `EDocPOMConfiguration.Enum.al`:
- "Always ask"
- "Always receive at posting"
- "Never receive at posting"
- "Receive at posting only for certain vendors"
- "Receive at posting except for certain vendors"

Each of these affect the process of processing an invoice in the following ways:
- When first trying to match
- When deciding whether or not we should warn the user
- When finalizing to create the final invoice.

Additionally an orthogonal setting to allow for receiving lines when they are of type G/L Account is also considered.

### When first trying to match

When trying to match, we will prevent the match with an error if:
- The configuration is set to "Never receive"
- The configuration is set to "Receive at posting except for certain vendors", and the vendor of the invoice is selected

This happens in the procedure `MatchPOLinesToEDocumentLine`.

### When deciding whether or not we should warn the user

We want to warn the user if the order has not been received only when the configuration is set to:
- Always ask
- Never receive
- Receive at posting only for certain vendors (and one of the non-specified vendors is selected for this invoice)
- Receive at posting except for certain vendors (and one of the specified vendors is selected for this invoice)

The logic for this is in the procedure `AppendPOMatchWarnings`.

### When finalizing to create the final invoice

We want to ask the user if they want to continue in the following cases:
- Configuration set to "Always ask"
- Configuration set to "Receive at posting only for certain vendors" and the vendor of the invoice is not one of the configured vendors.

We want to error out in the following cases:
- Configuration set to "Never receive"
- Configuration set to "Receive at posting except for certain vendors" and one of the selected vendors is the vendor of the invoice

For the rest of the scenarios, we allow finalizing without user confirmation.

This happens in `EDocCreatePurchaseInvoice.Codeunit.al`.

image

#### Work Item(s) 
Fixes [AB#599135](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/599135)

### DIFF
--- a/src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMConfiguration.Enum.al
+++ b/src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMConfiguration.Enum.al
@@ -6,6 +6,8 @@ namespace Microsoft.eServices.EDocument.Processing.Import.Purchase;
 
 enum 6113 "E-Doc. PO M. Configuration"
 {
+    Access = Internal;
+    Extensible = false;
     value(0; "Always ask")
     {
         Caption = 'Always ask';
--- a/src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatching.Codeunit.al
+++ b/src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatching.Codeunit.al
@@ -584,7 +584,7 @@ codeunit 6196 "E-Doc. PO Matching"
         CalculatePOMatchWarnings(EDocumentPurchaseHeader, TempPOMatchWarnings);
         TempPOMatchWarnings.SetRange("Warning Type", "E-Doc PO Match Warning"::NotYetReceived);
         // For each line that has a Not Yet Received warning, we check if it can be matched without receipt
-        if not TempPOMatchWarnings.IsEmpty() then
+        if TempPOMatchWarnings.FindSet() then
             repeat
                 EDocumentPurchaseLine.GetBySystemId(TempPOMatchWarnings."E-Doc. Purchase Line SystemId");
                 LoadPOLinesMatchedToEDocumentLine(EDocumentPurchaseLine, TempPurchaseLine);
--- a/src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMConfigReceipt.Enum.al
+++ b/src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMConfigReceipt.Enum.al
@@ -7,6 +7,7 @@ namespace Microsoft.eServices.EDocument.Processing.Import.Purchase;
 enum 6112 "E-Doc. PO M. Config. Receipt"
 {
     Access = Internal;
+    Extensible = false;
 
     value(0; "Always ask")
     {


src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMConfiguration.Enum.al
```diff
@@ -0,0 +1,29 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+namespace Microsoft.eServices.EDocument.Processing.Import.Purchase;
+
+enum 6113 "E-Doc. PO M. Configuration"
+{
+    value(0; "Always ask")
```
Groenbech96, line 11: Internal? Or why is it not? Evt extensible false

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatching.Codeunit.al
```diff
@@ -478,4 +496,152 @@ codeunit 6196 "E-Doc. PO Matching"
...
+    procedure VerifyEDocumentMatchedLinesAreValidMatches(EDocumentPurchaseHeader: Record "E-Document Purchase Header"): Boolean
+    var
+        EDocumentPurchaseLine: Record "E-Document Purchase Line";
+        TempPurchaseLine: Record "Purchase Line" temporary;
+        TempPOMatchWarnings: Record "E-Doc PO Match Warning" temporary;
+    begin
+        CalculatePOMatchWarnings(EDocumentPurchaseHeader, TempPOMatchWarnings);
+        TempPOMatchWarnings.SetRange("Warning Type", "E-Doc PO Match Warning"::NotYetReceived);
+        // For each line that has a Not Yet Received warning, we check if it can be matched without receipt
+        if not TempPOMatchWarnings.IsEmpty() then
```
Groenbech96: Why not findset?

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatchingSetup.Table.al
```diff
@@ -0,0 +1,73 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+namespace Microsoft.eServices.EDocument.Processing.Import.Purchase;
+
+using Microsoft.Purchases.Vendor;
+
+table 6116 "E-Doc. PO Matching Setup"
```
Groenbech96, line 9: I would think this should be living on the vendor page somehow. 
So far we have added the configuration there. I guess a separate table is okay, the field on the vendor page could always be a flow field?

Not sure where the UI for this is.

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatchingSetup.Table.al
```diff
@@ -0,0 +1,73 @@
...
+{
+    Access = Internal;
+    InherentEntitlements = RID;
```
Groenbech96, line 12: Why not RIMD?

mynjj: the current model is that whenever it's configured it gets recreated, so M wasn't required

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatching.Codeunit.al
mynjj: thanks!

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatchingSetup.Table.al
mynjj: Yes, currently it will only be surfaced in the Payables Agent setup page. This is one of the points to bring up to the UX review with Jacob :) The configuration is not only per-vendor, it can also be set globally, so having a separate table is nicer.

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMConfiguration.Enum.al
mynjj: Good point. I think by default enums are extensible false (?), but it doesn't hurt to be explicit.

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatchingSetup.Table.al
mynjj: image

Jose-agg (approved):
Left some comments

PR #5337, opened 2025-10-28
Extensibility Interface for File Scenario

### LINKED ISSUE #4494: [BC Idea]: Adjustments for External Storage module

StefanSosic:
#### Summary 

The extensibility part was separated from PR #4495 

#### Work Item(s) 
Fixes #4494

Fixes [AB#611443](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/611443)

### DIFF
--- a/src/System Application/App/External File Storage/src/Scenario/FileScenario.Codeunit.al
+++ b/src/System Application/App/External File Storage/src/Scenario/FileScenario.Codeunit.al
@@ -17,7 +17,7 @@ codeunit 9452 "File Scenario"
     /// <returns>True if an account for the default scenario was found; otherwise - false.</returns>
     procedure GetDefaultFileAccount(var TempFileAccount: Record "File Account" temporary): Boolean
     begin
-        exit(FileScenarioImpl.GetFileAccount(Enum::"File Scenario"::Default, TempFileAccount));
+        exit(FileScenarioImpl.GetFileAccountOrDefault(Enum::"File Scenario"::Default, TempFileAccount));
     end;
 
     /// <summary>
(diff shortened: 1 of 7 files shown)


src/System Application/App/External File Storage/src/Scenario/FileScenario.Codeunit.al
```diff
@@ -17,7 +17,7 @@ codeunit 9452 "File Scenario"
     /// <returns>True if an account for the default scenario was found; otherwise - false.</returns>
     procedure GetDefaultFileAccount(var TempFileAccount: Record "File Account" temporary): Boolean
     begin
-        exit(FileScenarioImpl.GetFileAccount(Enum::"File Scenario"::Default, TempFileAccount));
+        exit(FileScenarioImpl.GetFileAccountOrDefault(Enum::"File Scenario"::Default, TempFileAccount));
     end;
 
     /// <summary>
```
pri-kise, line 23: @StefanSosic you changed the behaviour of the method `GetFileAccount`.
```al
    /// <summary>
    /// Gets the file account used by the given file scenario.
    /// If the no account is defined for the provided scenario, the default account (if defined) will be returned.
    /// </summary>
    /// <param name="Scenario">The scenario to look for.</param>
    /// <param name="TempFileAccount">Out parameter holding information about the file account.</param>
    /// <returns>True if an account for the specified scenario was found; otherwise - false.</returns>
    procedure GetFileAccount(Scenario: Enum "File Scenario"; var TempFileAccount: Record "File Account" temporary): Boolean
    begin
        exit(FileScenarioImpl.GetFileAccount(Scenario, TempFileAccount));
        // Needs to be changed to:
        // exit(FileScenarioImpl.GetFileAccountOrDefault(Scenario, TempFileAccount));
    end;
```

Now this method won't return the Default File Account.
To be non breaking I would suggest to call the new method `GetFileAccountOrDefault`.

Furthermore I would suggest to add a new procedure in the public facade:

```al
    /// Gets the file account used by the given file scenario.
    /// </summary>
    /// <param name="Scenario">The scenario to look for.</param>
    /// <param name="TempFileAccount">Out parameter holding information about the file account.</param>
    /// <returns>True if an account for the specified scenario was found; otherwise - false.</returns>
    procedure GetSpecificFileAccount(Scenario: Enum "File Scenario"; var TempFileAccount: Record "File Account" temporary): Boolean
    begin
        exit(FileScenarioImpl.GetFileAccount(Scenario, TempFileAccount));
    end;
```

StefanSosic: It should not return default at first place, so in this case it's by design

pri-kise: Then you should update the documentation.

StefanSosic: Since tests are complaining to this point likewise
I think we will do here a bit refactoring, something like you mentioned in first comment. Naming will still be wrong, so thinking still whats best to do, maybe even adjusting tests, but let's see
Thanks for pointing it out 🚀

StefanSosic:
At the end, no need for a split into two PR. Will go with https://github.com/microsoft/BCApps/pull/4495

Comments from here will be addressed there :)

PR #5340, opened 2025-10-28
[master] Error occurs when attempting to save an Item Reference longer than 20…

### LINKED ISSUE #5339: [Bug]:  [master] Error occurs when attempting to save an Item Reference longer than 20 characters - table 6105 "E-Document Line Mapping"
### Describe the issue

Error occurs when attempting to save an Item Reference longer than 20 characters: "Cannot write the value [number] to the field [BC] Item Reference No. ... because the value is either too long or the content is invalid."

### Expected behavior

.

### Steps to reproduce

.

### Additional context

_No response_

djukicmilica:
… characters: "Cannot write the value [number] to the field [BC] Item Reference No. ... because the value is either too long or the content is invalid."

#### Work Item(s) 
Fixes #5339
[AB#611317](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/611317)

### DIFF
--- a/src/Apps/W1/EDocument/App/src/Processing/Import/EDocumentLineMapping.Table.al
+++ b/src/Apps/W1/EDocument/App/src/Processing/Import/EDocumentLineMapping.Table.al
@@ -95,7 +95,13 @@ table 6105 "E-Document Line Mapping"
             TableRelation = "Dimension Value".Code where("Global Dimension No." = const(2),
                                                           Blocked = const(false));
         }
-        field(10; "Item Reference No."; Code[20])
+#if not CLEAN27
+#pragma warning disable AS0086
+#endif
+        field(10; "Item Reference No."; Code[50])
+#if not CLEAN27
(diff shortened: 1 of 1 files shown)


src/Apps/W1/EDocument/App/src/Processing/Import/EDocumentLineMapping.Table.al
```diff
@@ -95,7 +95,13 @@ table 6105 "E-Document Line Mapping"
             TableRelation = "Dimension Value".Code where("Global Dimension No." = const(2),
```
mynjj: I thought this table was no longer existing - afair this one got merged into the EDocPurchaseLine table (with the `[BC]` fields), but we never got around deprecating it?

Should we do that instead? This fix should not really be used or needed.

cc: @Groenbech96

PR #5382, opened 2025-10-31
[27.x] FORNAV connector tests uptake in private NAV - fix CZ VAT

Groenbech96:
#### Summary 
When tests are run in CZ, then VAT Date check is there. Solution is to disable VAT check.

#### Work Item(s) 
Fixes [AB#611626](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/611626)

[AB#608829](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/608829)

### DIFF
--- a/src/Apps/W1/EDocumentConnectors/Avalara/Test/src/IntegrationTests.Codeunit.al
+++ b/src/Apps/W1/EDocumentConnectors/Avalara/Test/src/IntegrationTests.Codeunit.al
@@ -9,6 +9,8 @@ using Microsoft.Sales.Customer;
 using Microsoft.Purchases.Document;
 using Microsoft.Foundation.Company;
 using Microsoft.Purchases.Vendor;
+using Microsoft.Finance.GeneralLedger.Setup;
+using Microsoft.Finance.VAT.Setup;
 using System.Threading;
 using Microsoft.eServices.EDocument.Integration;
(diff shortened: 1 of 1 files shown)


src/Apps/W1/EDocumentConnectors/Avalara/Test/src/IntegrationTests.Codeunit.al
```diff
@@ -532,10 +534,17 @@ codeunit 148191 "Integration Tests"
...
         CompanyInformation: Record "Company Information";
+        GeneralLedgerSetup: Record "General Ledger Setup";
         AvalaraAuth: Codeunit Authenticator;
         KeyGuid: Guid;
     begin
         LibraryPermission.SetOutsideO365Scope();
+
+        GeneralLedgerSetup.Get();
+        PrevVATReportingDateValue := GeneralLedgerSetup."VAT Reporting Date Usage";
+        GeneralLedgerSetup."VAT Reporting Date Usage" := Enum::"VAT Reporting Date Usage"::Disabled;
```
haoranpb, line 545: There are many scenarios (or maybe the same but different branches), you might want to add a test lib for that

PR #5431, opened 2025-11-05
Rijndael encrypt decrypt binary data for 27.x

### LINKED ISSUE #5338: [Bug]: Support binary-safe AES encryption/decryption in AL
### Describe the issue

We need to introduce support for AES encryption and decryption in AL that works directly with binary data.
Currently, the built-in Rijndael Cryptography codeunit only handles text-based operations. This makes it unsuitable for scenarios where files are encrypted or decrypted in binary form, such as .zip or .pdf files.
When encrypting or decrypting binary files using the existing Rijndael Cryptography codeunit, the Decrypt method outputs a Text value.
As a result, decrypted binary data (e.g., .zip contents) is assigned to a Text variable, corrupting the output file and making it unusable.

### Expected behavior

Supporting AES operations that work with binary data (via Base64 or TempBlob) would allow developers to safely encrypt and decrypt files of any type without data loss.

### Steps to reproduce

Use the Rijndael Cryptography codeunit in AL.
Call the Decrypt method to decrypt a file that was encrypted in binary form (for example, a .zip or .pdf file).
Observe that the procedure returns the decrypted content as Text.
Write the result to a file and try to open it — the file will be corrupted.

### Proposed solution

Add new methods in the Rijndael Cryptography and Cryptography Management Impl. codeunits (or introduce a new AES Cryptography codeunit) that:

Accept and return Base64-encoded strings or TempBlob data
Support AES encryption and decryption
Handle binary data safely without conversion to Text

_No response_

AlbertSanocki:
#### Summary 
Adds methods to support encryption and decryption of binary files.
The new methods enable secure processing of binary data represented in Base64 format, ensuring compatibility with file encryption workflows.
#### Work Item(s) 
Fixes #5338 
Fixes [AB#612438](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/612438)

### DIFF
--- a/src/System Application/Test/Cryptography Management/src/RijndaelCryptographyTest.Codeunit.al
+++ b/src/System Application/Test/Cryptography Management/src/RijndaelCryptographyTest.Codeunit.al
@@ -113,6 +113,65 @@ codeunit 132575 "Rijndael Cryptography Test"
         LibraryAssert.AreEqual(PlainText, RijndaelCryptography2.Decrypt(CryptedText), 'Set Encryption Datay and Decrypt failed');
     end;
 
+    [Test]
+    procedure VerifyEncryptBinaryDataProducesExpectedResult()
+    var
+        RijndaelCryptography: Codeunit "Rijndael Cryptography";
+        BinaryDataAsBase64: Text;
+        EncryptedBinaryDataAsBase64: Text;
+        ExpectedEncryptedBinaryDataAsBase64: Text;
+    begin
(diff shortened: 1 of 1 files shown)


darjoo:
@AlbertSanocki I'll approve the PR. Will sort out the work item validation as I can't find it in ADO.

Can you also create a PR for the main branch.

darjoo (approved):

WaelAbuSeada (approved):

PR #5489, opened 2025-11-12
[E-Document] Introduce eligibility interface to allow control of when to export

Groenbech96:
#### Summary 

Moving code from GetTypeFromPostedSourceDocument to single function GetTypeFromSourceDocument, that now converts both posted and non posted types.

IsDocumentTypeSupported is now called IsDocumentSupported, and checks first the type using the GetTypeFromSourceDocument, and then also checks the ShouldExport impl. This works for both check and create methods. 
Only check when you want to export.

Add default enum impl on service as a field, that can be set for a given service record, allowing control for partners to determine if documents should be exported or not.

Fixes [AB#612844](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/612844)

### DIFF
--- a/src/Apps/W1/EDocument/App/src/Processing/EDocExport.Codeunit.al
+++ b/src/Apps/W1/EDocument/App/src/Processing/EDocExport.Codeunit.al
@@ -48,7 +48,7 @@ codeunit 6102 "E-Doc. Export"
             if EDocumentService.FindSet() then
                 repeat
                     Type := EDocumentProcessing.GetTypeFromSourceDocument(EDocSourceRecRef);
-                    if IsDocumentSupported(EDocSourceRecRef, EDocumentService, Type) then begin
+                    if IsDocumentSupported(EDocumentService, EDocSourceRecRef, Type) then begin
                         EDocumentInterface := EDocumentService."Document Format";
                         EDocumentInterface.Check(EDocSourceRecRef, EDocumentService, EDocumentProcessingPhase);
                     end;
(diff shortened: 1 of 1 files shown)


nikolakukrika (approved):

src/Apps/W1/EDocument/App/src/Processing/EDocExport.Codeunit.al
```diff
@@ -463,113 +465,39 @@ codeunit 6102 "E-Doc. Export"
...
-                end;
-            Database::"Purch. Inv. Header":
-                EDocSourceType := EDocSourceType::"Purchase Invoice";
-            Database::"Purch. Cr. Memo Hdr.":
-                EDocSourceType := EDocSourceType::"Purchase Credit Memo";
+        case EDocumentType of
+            EDocumentType::"Sales Order":
+                exit(EDocServiceSupportedType.Get(EDocService.Code, EDocSourceType::"Sales Order") or EDocServiceSupportedType.Get(EDocService.Code, EDocSourceType::"Sales Invoice"));
+            EDocumentType::"Sales Return Order":
+                exit(EDocServiceSupportedType.Get(EDocService.Code, EDocSourceType::"Sales Return Order") or EDocServiceSupportedType.Get(EDocService.Code, EDocSourceType::"Sales Credit Memo"));
```
gggdttt, line 477: Not sure if  Sales Credit Memo belongs to "Sales Return Order". Feel free to resolve this comment if you are 100% sure it's correct.

gggdttt (approved):
Looks good. Really like your new design

kasperdj:
Please note that Jesper has previously promised this to be released in BC27.1 on https://github.com/microsoft/ALAppExtensions/pull/28877. I can see that the issue is now labeled with 28.0! I expect this to be part of 27.2. Thx

PR #5492, opened 2025-11-13
[Shopify] Synchronize product collections

onbuyuka:
#### Summary 
Contains new functionality where user can choose custom Product Collections to which new products will be added to on creation.

Solution contains new page and table - Product Collections available from Shop. On the page user can import custom product collections available for shop and choose to which new products should be added to by selecting product collection(s) as default.

Pulled product collections are updated in BC - in case the collection was removed in Shopify it's removed from list in BC also.

- GQL queries to import the custom product collections
- New codeunit Shpfy Product Collection API to handle product collection import
- In Product API, added procedure to handle new way of publishing new products, to add product to collections to join
- Product Collections page and table
- Tests

#### Work Item(s) 
Fixes [AB#443907](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/443907)

### DIFF
--- a/src/Apps/W1/Shopify/App/src/Products/Tables/ShpfyShopCollectionMap.Table.al
+++ b/src/Apps/W1/Shopify/App/src/Products/Tables/ShpfyShopCollectionMap.Table.al
@@ -95,7 +95,7 @@ table 30128 "Shpfy Shop Collection Map"
         {
             Caption = 'Default for Export';
             DataClassification = CustomerContent;
-
+#if not CLEAN28
             trigger OnValidate()
             var
                 Collection: Record "Shpfy Shop Collection Map";
@@ -110,6 +110,7 @@ table 30128 "Shpfy Shop Collection Map"
                     if "Default for Export" then
                         Collection.ModifyAll("Default for Export", false, false);
             end;
+#endif
         }
         field(9; Version; BigInteger)
         {


src/Apps/W1/Shopify/App/src/Products/Codeunits/ShpfyProductCollectionAPI.Codeunit.al
```diff
@@ -0,0 +1,92 @@
...
+        CurrentCollections := RetrieveCollections(ShopCode);
+
+        CommunicationMgt.SetShop(ShopCode);
+        GraphQLType := GraphQLType::GetCustomProductCollections;
+
+        repeat
+            JResponse := CommunicationMgt.ExecuteGraphQL(GraphQLType, Parameters);
+            if JsonHelper.GetJsonArray(JResponse, JPublications, 'data.collections.edges') then begin
+                ExtractProductCollections(JPublications, ShopCode, CurrentCollections, Cursor);
+                if Parameters.ContainsKey('After') then
```
darjoo, line 38: I don't think there's a need for checking `ContainsKey`, `Set` should work with and without the key existing.

darjoo (approved):

aholstrup1 (approved):

PR #5495, opened 2025-11-13
Private/nhsejth/memory telemetry

nhsejth:
#### Summary 
Add telemetry to the Base64 converter and Json functions that that emits the source length when then input exceeds a given limit. These functions are often seen in call stacks when the NST gets OutOfMemory exceptions and we need to see how large elements are causing this.
There are no functional changes in this PR.

#### Work Item(s) 
Fixes [AB#612984](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/612984)

### DIFF
--- a/src/System Application/App/Base64 Convert/src/Base64ConvertImpl.Codeunit.al
+++ b/src/System Application/App/Base64 Convert/src/Base64ConvertImpl.Codeunit.al
@@ -15,23 +15,14 @@ codeunit 4111 "Base64 Convert Impl."
     InherentPermissions = X;
 
     var
-        IsInitialized: Boolean;
         SourceWarningLength: Integer;
         TextLengtWarningTxt: Label 'The input string length (%1) exceeds the maximum suggested length (%2) for Base64 conversion.', Locked = true;
         StreamLengtWarningTxt: Label 'The input stream length (%1) exceeds the maximum suggested length (%2) for Base64 conversion.', Locked = true;
 
-    internal procedure Initialize()
-    begin
-        if not IsInitialized then begin
-            SourceWarningLength := 10 * 1024 * 1024; // 10 MB
-            IsInitialized := true;
-        end;
-    end;
-
     internal procedure EmitLengthWarning(SourceLength: Integer; tag: Text; FormatString: Text)
     begin
-        if not IsInitialized then
-            this.Initialize();
+        if SourceWarningLength <= 0 then
+            SourceWarningLength := 10485760; // 10 * 1024 * 1024 = 10 MB
 
         if SourceLength > SourceWarningLength then
             Session.LogMessage(tag, StrSubstNo(FormatString, SourceLength, SourceWarningLength), Verbosity::Warning, DataClassification::SystemMetadata, TelemetryScope::All, 'resources', 'memory');
--- a/src/System Application/App/Json/src/JsonImpl.Codeunit.al
+++ b/src/System Application/App/Json/src/JsonImpl.Codeunit.al
@@ -16,25 +16,15 @@ codeunit 5461 "Json Impl."
     InherentPermissions = X;
 
     var
-        IsInitialized: Boolean;
         SourceWarningLength: Integer;
         JsonArrayDotNet: DotNet JArray;
         JsonObjectDotNet: DotNet JObject;
         LogLimitWarningTxt: Label 'The JSON input length (%1) exceeds the maximum suggested length (%2) for JSON processing.', Locked = true;
 
-
-    internal procedure Initialize()
-    begin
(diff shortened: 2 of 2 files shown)


src/System Application/App/Base64 Convert/src/Base64ConvertImpl.Codeunit.al
```diff
@@ -14,6 +14,16 @@ codeunit 4111 "Base64 Convert Impl."
...
     InherentPermissions = X;
 
+    var
+        SourceWarningLength: Integer;
+        TextLengtWarningTxt: Label 'The input string length (%1) exceeds the maximum suggested length (%2) for Base64 conversion.';
+        StreamLengtWarningTxt: Label 'The input stream length (%1) exceeds the maximum suggested length (%2) for Base64 conversion.';
+
+    trigger OnRun()
+    begin
+        SourceWarningLength := 10 * 1024 * 1024; // 10 MB
```
pri-kise: Is this a feature that isn't documented?
Or since wenn is the OnRun Trigger called when we simple call a method of a codeunit?

We have the OnInit Trigger on Reports that is called when the object is used, but I didn't know that this happens for the OnRun Trigger of a codeunit.

darjoo: OnRun doesn't run unless this CU is called with `Run`, `SourceWarningLength` will stay 0 when being used otherwise.

nhsejth: Yep, it was my bad memory. Updated and simplified.

src/System Application/App/Json/src/JsonImpl.Codeunit.al
```diff
@@ -18,22 +18,35 @@ codeunit 5461 "Json Impl."
     var
         JsonArrayDotNet: DotNet JArray;
         JsonObjectDotNet: DotNet JObject;
+        LogLimitWarningTxt: Label 'The JSON input length (%1) exceeds the maximum suggested length (%2) for JSON processing.';
+        SourceWarningLength: Integer
```
JesperSchulz: Missing ;

```suggestion
        SourceWarningLength: Integer;
```

src/System Application/App/Base64 Convert/src/Base64ConvertImpl.Codeunit.al
pri-kise: @nhsejth will you update the other codeunit (json) as well?

nhsejth: yep, it's updated now

src/System Application/App/Json/src/JsonImpl.Codeunit.al
nhsejth: fixed

src/System Application/App/Base64 Convert/src/Base64ConvertImpl.Codeunit.al
```diff
@@ -14,6 +14,29 @@ codeunit 4111 "Base64 Convert Impl."
     InherentEntitlements = X;
     InherentPermissions = X;
 
+    var
+        IsInitialized: Boolean;
+        SourceWarningLength: Integer;
+        TextLengtWarningTxt: Label 'The input string length (%1) exceeds the maximum suggested length (%2) for Base64 conversion.', Locked = true;
+        StreamLengtWarningTxt: Label 'The input stream length (%1) exceeds the maximum suggested length (%2) for Base64 conversion.', Locked = true;
+
+    internal procedure Initialize()
```
Groenbech96: Remove the initialize, and just calculate the length on the fly. 
Less and more readable code.

Groenbech96 (changes requested):

src/System Application/App/Base64 Convert/src/Base64ConvertImpl.Codeunit.al
JesperSchulz: Matter of taste, I guess.

JesperSchulz (dismissed):

src/System Application/App/Base64 Convert/src/Base64ConvertImpl.Codeunit.al
darjoo: Or write 10485760 to avoid any calculation

nhsejth: Fixed - simplified

PR #5591, opened 2025-11-22
PO matching in E-Docs for a single receipt line

mynjj:
#### Summary 
Scoping as agreed for 27.3: 
- an invoice line can only be linked to a single receipt line
- if only an order line is assigned, a covering receipt is autosuggested
- the created invoice line affects the PO when posted
- default configured to avoid creating an invoice when it hasn't been received

#### Work Item(s)
Fixes [AB#599135](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/599135)

### DIFF
--- a/src/Apps/W1/EDocument/App/src/Processing/Import/FinishDraft/EDocCreatePurchaseInvoice.Codeunit.al
+++ b/src/Apps/W1/EDocument/App/src/Processing/Import/FinishDraft/EDocCreatePurchaseInvoice.Codeunit.al
@@ -236,7 +236,7 @@ codeunit 6117 "E-Doc. Create Purchase Invoice" implements IEDocumentFinishDraft,
         EDocRecordLink."Target Table No." := Database::"Purchase Header";
         EDocRecordLink."Target SystemId" := PurchaseHeader.SystemId;
         EDocRecordLink."E-Document Entry No." := EDocumentPurchaseHeader."E-Document Entry No.";
-        if EDocRecordLink.Insert() then;
+        EDocRecordLink.Insert();
     end;
 
     local procedure LinkEDocumentLineToPurchaseLine(EDocumentPurchaseLine: Record "E-Document Purchase Line"; PurchaseLine: Record "Purchase Line")
@@ -248,7 +248,7 @@ codeunit 6117 "E-Doc. Create Purchase Invoice" implements IEDocumentFinishDraft,
         EDocRecordLink."Target Table No." := Database::"Purchase Line";
         EDocRecordLink."Target SystemId" := PurchaseLine.SystemId;
         EDocRecordLink."E-Document Entry No." := EDocumentPurchaseLine."E-Document Entry No.";
-        if EDocRecordLink.Insert() then;
+        EDocRecordLink.Insert();
     end;
 
 }
--- a/src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/EDocPurchaseDraftSubform.Page.al
+++ b/src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/EDocPurchaseDraftSubform.Page.al
@@ -166,7 +166,7 @@ page 6183 "E-Doc. Purchase Draft Subform"
                 group("Order Matching")
                 {
                     Caption = 'Order matching';
-                    action(MatchToOrderLines)
+                    action(MatchToOrderLine)
                     {
                         ApplicationArea = All;
                         Caption = 'Match to order line';


src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatching.Codeunit.al
```diff
@@ -545,6 +573,127 @@ codeunit 6196 "E-Doc. PO Matching"
...
+    var
+        EDocumentPurchaseLine: Record "E-Document Purchase Line";
+        EDocPurchaseLinePOMatch: Record "E-Doc. Purchase Line PO Match";
+        PurchaseOrderLine: Record "Purchase Line";
+        PurchaseReceiptLine: Record "Purch. Rcpt. Line";
+        TempPurchaseReceiptLine: Record "Purch. Rcpt. Line" temporary;
+        EDocLineQuantity: Decimal;
+        NullGuid: Guid;
+    begin
+        EDocumentPurchaseLine.SetRange("E-Document Entry No.", EDocumentPurchaseHeader."E-Document Entry No.");
```
ventselartur, line 586: Could we add SetLoadFields? Looks like you need only certain fields from the table.

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatching.Codeunit.al
```diff
@@ -545,6 +573,127 @@ codeunit 6196 "E-Doc. PO Matching"
...
+        if EDocumentPurchaseLine.FindSet() then
+            repeat
+                Clear(EDocPurchaseLinePOMatch);
+                EDocPurchaseLinePOMatch.SetRange("E-Doc. Purchase Line SystemId", EDocumentPurchaseLine.SystemId);
+                EDocPurchaseLinePOMatch.SetRange("Receipt Line SystemId", NullGuid);
+                if not EDocPurchaseLinePOMatch.FindFirst() then
+                    continue; // No PO lines matched, so no receipt can be suggested
+                if not PurchaseOrderLine.GetBySystemId(EDocPurchaseLinePOMatch."Purchase Line SystemId") then
```
ventselartur, line 594: SetLoadFields?

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatching.Codeunit.al
```diff
@@ -545,6 +573,127 @@ codeunit 6196 "E-Doc. PO Matching"
...
+                    continue; // Should not happen, but we skip in case it does, this procedure doesn't error out
+                EDocPurchaseLinePOMatch.SetRange("Purchase Line SystemId", PurchaseOrderLine.SystemId);
+                EDocPurchaseLinePOMatch.SetFilter("Receipt Line SystemId", '<> %1', NullGuid);
+                if not EDocPurchaseLinePOMatch.IsEmpty() then
+                    continue; // There's already at least one receipt line matched, so no suggestion is needed
+                PurchaseReceiptLine.SetRange("Order No.", PurchaseOrderLine."Document No.");
+                PurchaseReceiptLine.SetRange("Order Line No.", PurchaseOrderLine."Line No.");
+                PurchaseReceiptLine.SetFilter(Quantity, '> 0');
```
ventselartur, line 603: I am not sure it is fine to filter on Decimal fields. Would that perform better if you remove the filter, but inside the loop do "If Quantity = 0 then exit"?

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatching.Codeunit.al
```diff
@@ -545,6 +573,127 @@ codeunit 6196 "E-Doc. PO Matching"
...
+                EDocPurchaseLinePOMatch.SetRange("E-Doc. Purchase Line SystemId", EDocumentPurchaseLine.SystemId);
+                EDocPurchaseLinePOMatch.SetRange("Receipt Line SystemId", NullGuid);
+                if not EDocPurchaseLinePOMatch.FindFirst() then
+                    continue; // No PO lines matched, so no receipt can be suggested
+                if not PurchaseOrderLine.GetBySystemId(EDocPurchaseLinePOMatch."Purchase Line SystemId") then
+                    continue; // Should not happen, but we skip in case it does, this procedure doesn't error out
+                EDocPurchaseLinePOMatch.SetRange("Purchase Line SystemId", PurchaseOrderLine.SystemId);
+                EDocPurchaseLinePOMatch.SetFilter("Receipt Line SystemId", '<> %1', NullGuid);
+                if not EDocPurchaseLinePOMatch.IsEmpty() then
+                    continue; // There's already at least one receipt line matched, so no suggestion is needed
```
ventselartur, line 599: It is going to be hard to troubleshoot these continue calls. Does it make sense to add telemetry?

src/Apps/W1/EDocument/Test/src/Matching/EDocPOMatchingUnitTests.Codeunit.al
```diff
@@ -2309,6 +2023,499 @@ codeunit 133508 "E-Doc. PO Matching Unit Tests"
...
+        CreateMockReceiptLine(PurchaseReceiptLine1, PurchaseReceiptHeader, Item1."No.", 10, PurchaseOrderLine1);
+        CreateMockReceiptLine(PurchaseReceiptLine2, PurchaseReceiptHeader, Item2."No.", 15, PurchaseOrderLine2);
+        CreateMockReceiptLine(PurchaseReceiptLine3, PurchaseReceiptHeader, Item3."No.", 20, PurchaseOrderLine3);
+
+        // Create E-Document lines and match them
+        EDocumentPurchaseLine1 := LibraryEDocument.InsertPurchaseDraftLine(EDocument);
+        EDocumentPurchaseLine1.Quantity := 10;
+        EDocumentPurchaseLine1.Modify();
+        MatchEDocumentLineToPOLine(EDocumentPurchaseLine1, PurchaseOrderLine1);
+        MatchEDocumentLineToReceiptLine(EDocumentPurchaseLine1, PurchaseReceiptLine1);
```
ventselartur: Any chance to extract this to the function to keep be able to see the body of the test case without scrolling?

ventselartur: Same for other test cases

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatching.Codeunit.al
mynjj: It's an internal table, not meant for high volume, with 3 fields. I don't think it's needed in this case.

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatching.Codeunit.al
mynjj: I rather not to, the record is passed to GetEDocumentLineQuantityInBaseUoM and MatchReceiptLinesToEDocumentLine, which they only use certain fields. But it's not easy to see from the caller which fields will be needed.

What I want to avoid is some future code-churn that assume in some of these procedures that certain record fields are loaded and use them, only for later to figure out that they weren't loaded because of the SetLoadFields.

In this case, the table is internal, so there's no concern of JOINs being added, and there is an index for the filter used, so hopefully this won't cause performance problems.

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatching.Codeunit.al
mynjj: Good point, I'm adding telemetry to two cases I think could care about

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatching.Codeunit.al
mynjj: It is true that table 121 has no index covering Quantity, but I still think it's faster to let SQL do it, the query planner will probably use the primary key for the first two and then filter on the small result set

Anyway, a single receipt shouldn't have over 10k rows, so I don't think the concern is very big either.

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatching.Codeunit.al
```diff
@@ -545,6 +573,127 @@ codeunit 6196 "E-Doc. PO Matching"
...
+        EDocumentPurchaseLine.SetRange("E-Document Entry No.", EDocumentPurchaseHeader."E-Document Entry No.");
+        if EDocumentPurchaseLine.FindSet() then
+            repeat
+                Clear(EDocPurchaseLinePOMatch);
+                EDocPurchaseLinePOMatch.SetRange("E-Doc. Purchase Line SystemId", EDocumentPurchaseLine.SystemId);
+                EDocPurchaseLinePOMatch.SetRange("Receipt Line SystemId", NullGuid);
+                if not EDocPurchaseLinePOMatch.FindFirst() then
+                    continue; // No PO lines matched, so no receipt can be suggested
+                if not PurchaseOrderLine.GetBySystemId(EDocPurchaseLinePOMatch."Purchase Line SystemId") then
+                    continue; // Should not happen, but we skip in case it does, this procedure doesn't error out
```
Jose-agg, line 595: Would it be usefull to trigger telemetry in case this happens? As a way for us to understand if there are undiscovered scenarios?

src/Apps/W1/EDocument/App/src/Processing/Import/FinishDraft/EDocCreatePurchaseInvoice.Codeunit.al
```diff
@@ -142,16 +146,17 @@ codeunit 6117 "E-Doc. Create Purchase Invoice" implements IEDocumentFinishDraft,
...
+        PurchaseLineNo := GetLastLineNumberOnPurchaseInvoice(PurchaseHeader."No."); // We get the last line number, even if this is a new document since recurrent lines get inserted on the header's creation
         EDocumentPurchaseLine.SetRange("E-Document Entry No.", EDocument."Entry No");
         if EDocumentPurchaseLine.FindSet() then
             repeat
+                Clear(PurchaseLine);
                 PurchaseLine."Document Type" := PurchaseHeader."Document Type";
                 PurchaseLine."Document No." := PurchaseHeader."No.";
-                PurchaseLine."Line No." += 10000;
+                PurchaseLineNo += 10000;
+                PurchaseLine."Line No." := PurchaseLineNo;
```
Jose-agg, line 158: Shouldn't we validate this values to make sure that whatever we assigned here is valid? In case this fails the error message wouldn't be recoverable by an user (or the agent). 
The same decision should apply to lines 156 and 157

src/Apps/W1/EDocument/App/src/Processing/Import/FinishDraft/EDocCreatePurchaseInvoice.Codeunit.al
```diff
@@ -226,4 +228,27 @@ codeunit 6117 "E-Doc. Create Purchase Invoice" implements IEDocumentFinishDraft,
...
+
+        SourceRecordRef.GetTable(SourceRecord);
+        TargetRecordRef.GetTable(TargetRecord);
+
+        EDocRecordLink."Source Table No." := SourceRecordRef.Number();
+        EDocRecordLink."Source SystemId" := SourceRecordRef.Field(SourceRecordRef.SystemIdNo).Value();
+        EDocRecordLink."Target Table No." := TargetRecordRef.Number();
+        EDocRecordLink."Target SystemId" := TargetRecordRef.Field(TargetRecordRef.SystemIdNo).Value();
+        EDocRecordLink."E-Document Entry No." := EDocument."Entry No";
+        if EDocRecordLink.Insert() then;
```
Jose-agg: Why are we swalloging this? Do we expect this insert to fail?

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatching.Codeunit.al
```diff
@@ -139,22 +140,38 @@ codeunit 6196 "E-Doc. PO Matching"
...
+        TempPurchaseReceiptLine: Record "Purch. Rcpt. Line" temporary;
+    begin
+        Clear(TempPurchaseReceiptHeader);
+        TempPurchaseReceiptHeader.DeleteAll();
+        LoadReceiptLinesMatchedToEDocumentLine(EDocumentPurchaseLine, TempPurchaseReceiptLine);
+        if TempPurchaseReceiptLine.FindSet() then
+            repeat
+                if PurchaseReceiptHeader.Get(TempPurchaseReceiptLine."Document No.") then begin
+                    Clear(TempPurchaseReceiptHeader);
+                    TempPurchaseReceiptHeader := PurchaseReceiptHeader;
```
Jose-agg, line 152: Minor: consider using TransferFields or Copy for transferring fields in between records

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatching.Codeunit.al
Jose-agg, line 1: A general suggestion for this and other files: if you don't have a need to use a procedure outside the current codeunit, mark is as local, or internal

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatching.Codeunit.al
```diff
@@ -139,22 +140,38 @@ codeunit 6196 "E-Doc. PO Matching"
...
+    procedure LoadReceiptLinesMatchedToEDocumentLine(EDocumentPurchaseLine: Record "E-Document Purchase Line"; var TempPurchaseReceiptLine: Record "Purch. Rcpt. Line" temporary)
+    var
         PurchaseReceiptLine: Record "Purch. Rcpt. Line";
         EDocPurchaseLinePOMatch: Record "E-Doc. Purchase Line PO Match";
         NullGuid: Guid;
     begin
-        Clear(TempPurchaseReceiptHeader);
-        TempPurchaseReceiptHeader.DeleteAll();
+        Clear(TempPurchaseReceiptLine);
+        TempPurchaseReceiptLine.DeleteAll();
```
Jose-agg, line 165: Minor: I don't see any initiallization of this temp record in the 2 functions that call this procedure - so it will be empty by default. We can do a check if it's empty before calling the deleteall

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatching.Codeunit.al
```diff
@@ -139,22 +140,38 @@ codeunit 6196 "E-Doc. PO Matching"
...
             repeat
-                if PurchaseReceiptLine.GetBySystemId(EDocPurchaseLinePOMatch."Receipt Line SystemId") then
-                    if PurchaseReceiptHeader.Get(PurchaseReceiptLine."Document No.") then begin
-                        Clear(TempPurchaseReceiptHeader);
-                        TempPurchaseReceiptHeader := PurchaseReceiptHeader;
-                        if TempPurchaseReceiptHeader.Insert() then;
-                    end;
+                if PurchaseReceiptLine.GetBySystemId(EDocPurchaseLinePOMatch."Receipt Line SystemId") then begin
+                    Clear(TempPurchaseReceiptLine);
+                    TempPurchaseReceiptLine := PurchaseReceiptLine;
```
Jose-agg, line 172: Same as before

src/Apps/W1/EDocument/App/src/Processing/Import/FinishDraft/EDocCreatePurchaseInvoice.Codeunit.al
mynjj: It's more like we don't want to fail in this procedure, for example when linking already linked records.

In general, we don't expect it, but if it happens, it's not critical and it shouldn't block any process, so the "contract" of the procedure is that it will ensure that the records are linked, not that it will verify anything.

That being said, the error on the top is a bit weird.. I would not have added it, instead I would have been more explicit on the procedure signature even if sacrificing generality. I'm pretty sure Magnus did it 😛 that's kinda his style, hard errors for dev errors. I really just moved this procedure from the history codeunit, because the linking became relevant not only for history, and that error came along.

Maybe I'll remove that one 😄

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatching.Codeunit.al
mynjj: This codeunit is internal in general, and all the non-local procedures are meant for consumption by other files, or to which are you referring to? @Jose-agg

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatching.Codeunit.al
```diff
@@ -545,6 +573,127 @@ codeunit 6196 "E-Doc. PO Matching"
...
+                EDocPurchaseLinePOMatch.SetFilter("Receipt Line SystemId", '<> %1', NullGuid);
+                if not EDocPurchaseLinePOMatch.IsEmpty() then
+                    continue; // There's already at least one receipt line matched, so no suggestion is needed
+                PurchaseReceiptLine.SetRange("Order No.", PurchaseOrderLine."Document No.");
+                PurchaseReceiptLine.SetRange("Order Line No.", PurchaseOrderLine."Line No.");
+                PurchaseReceiptLine.SetFilter(Quantity, '> 0');
+                if PurchaseReceiptLine.FindSet() then
+                    repeat
+                        if GetEDocumentLineQuantityInBaseUoM(EDocumentPurchaseLine, EDocLineQuantity) then
+                            if PurchaseReceiptLine.Quantity >= EDocLineQuantity then begin
```
Jose-agg, line 607: What is the flow if the EDocument has a larger quantity that the one specify in the Purchase Receipt? Is this a scenario we consider?

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatching.Codeunit.al
mynjj: It shouldn't be very expensive to do a DeleteAll of an empty temporary record. In general, this ensures that the returned temporary record has only what was loaded by the call. I won't fix this one, it adds an extra condition that is handled just as well by the default handling.

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatching.Codeunit.al
```diff
@@ -545,6 +573,127 @@ codeunit 6196 "E-Doc. PO Matching"
...
+        PurchaseOrderLine: Record "Purchase Line";
+        PurchaseReceiptLine: Record "Purch. Rcpt. Line";
+        TempPOLineToMatch: Record "Purchase Line" temporary;
+        TempReceiptLineToMatch: Record "Purch. Rcpt. Line" temporary;
+    begin
+        PurchaseInvoiceLine.SetRange("Document Type", PurchaseHeader."Document Type");
+        PurchaseInvoiceLine.SetRange("Document No.", PurchaseHeader."No.");
+        PurchaseInvoiceLine.SetFilter("Receipt No.", '<>%1', '');
+        PurchaseInvoiceLine.SetFilter("Receipt Line No.", '<>%1', 0);
+        if PurchaseInvoiceLine.FindSet() then
```
Jose-agg: Consider exiting if you don't find any entry. Specially since you are doing a 2 ModifyAll at the end of the procedure.

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatching.Codeunit.al
mynjj: It will not have a receipt suggested. Which means that at finalization it will fail since no receipt is selected. In the scope of 27.3 that leaves the user with the options of either remove the match or assign a covering receipt to the line instead.

In the scope planned for v28.0 this can be expanded to suggest and assign multiple receipts that cover the line.

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatching.Codeunit.al
```diff
@@ -545,6 +573,127 @@ codeunit 6196 "E-Doc. PO Matching"
...
+        PurchaseLine: Record "Purchase Line";
+        TempPurchaseReceiptLine: Record "Purch. Rcpt. Line" temporary;
+    begin
+        EDocumentPurchaseLine.SetRange("E-Document Entry No.", EDocument."Entry No");
+        if EDocumentPurchaseLine.FindSet() then
+            repeat
+                LoadReceiptLinesMatchedToEDocumentLine(EDocumentPurchaseLine, TempPurchaseReceiptLine);
+                if not TempPurchaseReceiptLine.FindFirst() then // We only support a single receipt line match in BaseApp
+                    continue;
+                PurchaseLine := EDocumentPurchaseLine.GetLinkedPurchaseLine();
```
Jose-agg, line 654: Is it possible that at this point we don't have a linked Purchase Line? If so, we will fail in line 655 when trying to modify a record that is not initialized.

Jose-agg (dismissed):

src/Apps/W1/EDocument/App/src/Processing/Import/FinishDraft/EDocCreatePurchaseInvoice.Codeunit.al
mynjj: Changing this behavior is a bit risky, and it's not really what this PR changed. This PR just maintained the existing logic while refactoring to be able to clear the PurchaseLine to avoid assigning trailing values to the rest of the lines

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatching.Codeunit.al
mynjj: It's not possible: this is called at finalization (currently) and each pair of lines get linked, it would only not have a linked line if it comes from some other trigger, for example recurrent lines, but in this case, the `continue` above should ensure that a receipt is also assigned to that line, which is only possible right now for explicit draft lines. 

But this is actually a great point, it's not explicitly handled and it should, I'll add something here.

ventselartur (dismissed):

src/Apps/W1/EDocument/App/src/Processing/Import/FinishDraft/EDocCreatePurchaseInvoice.Codeunit.al
```diff
@@ -38,24 +38,25 @@ codeunit 6117 "E-Doc. Create Purchase Invoice" implements IEDocumentFinishDraft,
...
         if not EDocPOMatching.VerifyEDocumentMatchedLinesAreValidMatches(EDocumentPurchaseHeader) then
             Error(YourMatchedLinesAreNotValidErr);
 
+        EDocPOMatching.SuggestReceiptsForMatchedOrderLines(EDocumentPurchaseHeader);
         EDocPOMatching.CalculatePOMatchWarnings(EDocumentPurchaseHeader, TempPOMatchWarnings);
         TempPOMatchWarnings.SetRange("Warning Type", "E-Doc PO Match Warning"::NotYetReceived);
         if not TempPOMatchWarnings.IsEmpty() then
-            if not ConfirmManagement.GetResponse(SomeLinesNotYetReceivedMsg) then
-                Error(''); // User cancelled the operation
+            Error(SomeLinesNotYetReceivedErr);
```
Groenbech96, line 54: Make this an error action. Makes it easy to remove matches. Massive win for UX

src/Apps/W1/EDocument/App/src/Processing/Import/FinishDraft/EDocCreatePurchaseInvoice.Codeunit.al
Groenbech96: If you remove it, then let's at least have record ref. 

If someone were to pass in a non record, and we did not check if it is record, the get record would fail. Just make the error a bit more soft than whatever platform throws. 

So regardless of style, this procedure can fail if variant is input. I think in latest platform we got Record -> recref implicit conversion, so maybe call site can stay the same if you change variant to recref 🙏

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/History/EDocPurchaseHistMapping.Codeunit.al
```diff
@@ -204,28 +203,6 @@ codeunit 6120 "E-Doc. Purchase Hist. Mapping"
         EDocActivityLogSession.Set(ActivityLogSessionToken, ActivityLog);
     end;
 
-    /// <summary>
-    /// Track header and line mapping between source and target records.
-    /// </summary>
-    procedure TrackRecord(EDocument: Record "E-Document"; SourceRecord: Variant; TargetRecord: Variant)
```
Groenbech96, line 210: Something seems ok it removing it here seems wrong. Like codeunit was a class.

Groenbech96 (dismissed):

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatching.Codeunit.al
mynjj: I've added telemetry around this part, not specifically to this case

src/Apps/W1/EDocument/App/src/Processing/Import/FinishDraft/EDocCreatePurchaseInvoice.Codeunit.al
mynjj: It may be that the matches are completely valid and the lines haven't been received, in which case make it easy to remove the matches is not desired. Debarati also had the idea of suggesting to receive some of these lines. I have a meeting later this week with Debarati to scope the UX for 27.3 where we will decide on these ideas and also how it will look on baseapp - I'll add you as optional so you can watch the recording when you're back.

For now I won't apply this comment.

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/History/EDocPurchaseHistMapping.Codeunit.al
mynjj: I've tried to read this comment for a while and I don't get it 😆

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatching.Codeunit.al
mynjj: Resolving

src/Apps/W1/EDocument/App/src/Processing/Import/FinishDraft/EDocCreatePurchaseInvoice.Codeunit.al
mynjj: Removed in favor of explicitness, no need of handling variant cases and what not when the DRY benefit is so small. In any case, it's something very small 😁

src/Apps/W1/EDocument/App/src/Processing/Import/FinishDraft/EDocCreatePurchaseInvoice.Codeunit.al
```diff
@@ -226,4 +227,28 @@ codeunit 6117 "E-Doc. Create Purchase Invoice" implements IEDocumentFinishDraft,
...
+    local procedure LinkEDocumentHeaderToPurchaseHeader(EDocumentPurchaseHeader: Record "E-Document Purchase Header"; PurchaseHeader: Record "Purchase Header")
+    var
+        EDocRecordLink: Record "E-Doc. Record Link";
+    begin
+        EDocRecordLink."Source Table No." := Database::"E-Document Purchase Header";
+        EDocRecordLink."Source SystemId" := EDocumentPurchaseHeader.SystemId;
+        EDocRecordLink."Target Table No." := Database::"Purchase Header";
+        EDocRecordLink."Target SystemId" := PurchaseHeader.SystemId;
+        EDocRecordLink."E-Document Entry No." := EDocumentPurchaseHeader."E-Document Entry No.";
+        if EDocRecordLink.Insert() then;
```
onbuyuka: would be nice if you check whether the record exists before trying to insert. also, could modify be an option? meaning record exists but some values have changed

src/Apps/W1/EDocument/App/src/Processing/Import/FinishDraft/EDocCreatePurchaseInvoice.Codeunit.al
onbuyuka: would be nice if you check whether the record exists before trying to insert. also, could modify be an option? meaning record exists but some values have changed

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/EDocPurchaseDraftSubform.Page.al
```diff
@@ -169,9 +169,9 @@ page 6183 "E-Doc. Purchase Draft Subform"
                     action(MatchToOrderLines)
```
onbuyuka: I know it is a hustle to change and this is the most nit comment you can imagine, but seeing the action name not matching the action - I had to comment

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/EDocumentPurchaseLine.Table.al
```diff
@@ -336,6 +337,31 @@ table 6101 "E-Document Purchase Line"
         exit(OldDimSetID <> "[BC] Dimension Set ID");
     end;
 
+    internal procedure GetFromLinkedPurchaseLine(PurchaseLine: Record "Purchase Line"): Boolean
+    var
+        EDocumentRecordLink: Record "E-Doc. Record Link";
+    begin
+        Clear(Rec);
```
onbuyuka, line 344: why do you need this for? you are not doing anything with the rec other than retrieving it conditionally. if the idea is to clear in case this is used in a loop and you might end up with a wrong rec, it is better to explicitly clear there.

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatching.Codeunit.al
```diff
@@ -470,17 +481,22 @@ codeunit 6196 "E-Doc. PO Matching"
...
+            Error(MatchesToMultipleReceiptLinesNotSupportedErr);
+
         // Remove existing receipt line matches
         EDocPurchaseLinePOMatch.SetRange("E-Doc. Purchase Line SystemId", EDocumentPurchaseLine.SystemId);
         EDocPurchaseLinePOMatch.SetFilter("Receipt Line SystemId", '<>%1', NullGuid);
         EDocPurchaseLinePOMatch.DeleteAll();
 
         // Create new matches
         LoadPOLinesMatchedToEDocumentLine(EDocumentPurchaseLine, TempMatchedPurchaseLines);
+        SelectedReceiptLines.FindSet();
```
onbuyuka, line 499: you already know there is only one record here, you exit if empty and error if count>1. you could change this to findfirst

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/EDocPurchaseDraftSubform.Page.al
mynjj: it's actually not a big deal, this is currently only in `main` - changed

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/EDocumentPurchaseLine.Table.al
mynjj: I disagree, the "contract" of this procedure is that when you call it, you will end up with the linked record in `Rec` (which means an empty record when there's no record linked). I don't see any benefit of moving the responsibility of clearing the record to the caller other than maybe the flexibility of preserving the value of `Rec` if not found, which would be a weird guarantee too

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/PurchaseOrderMatching/EDocPOMatching.Codeunit.al
mynjj: Yes, that's true, but this code was meant originally to handle 1-M matches (count > 1), for 27.3 we are scoping it by adding hard validations in the entry point while the changes in BaseApp get uptaken and 1-M matches become possible in v28. Once we have that, we can remove this validation and the remainder of the code will handle it appropriately as before.

src/Apps/W1/EDocument/App/src/Processing/Import/FinishDraft/EDocCreatePurchaseInvoice.Codeunit.al
mynjj: I'll just remove the condition, it's not possible for the record to be duplicated (primary key autoincrement)

onbuyuka (approved):

PR #5593, opened 2025-11-24
[Profiler Schedule] Do not allow empty user name

onbuyuka:
#### Summary 
Do not allow empty user name
#### Work Item(s) 
Fixes [AB#612543](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/612543)

### DIFF
--- a/src/System Application/App/Performance Profiler/src/PerfProfilerScheduleCard.Page.al
+++ b/src/System Application/App/Performance Profiler/src/PerfProfilerScheduleCard.Page.al
@@ -195,9 +195,12 @@ page 1932 "Perf. Profiler Schedule Card"
                     var
                         User: Record User;
                     begin
+                        if UserName = '' then
+                            Error(UserNameMandatoryErr);
+
                         User.SetRange("User Name", UserName);
-                        if User.FindFirst() then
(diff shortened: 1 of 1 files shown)


src/System Application/App/Performance Profiler/src/PerfProfilerScheduleCard.Page.al
```diff
@@ -195,9 +195,12 @@ page 1932 "Perf. Profiler Schedule Card"
                     var
                         User: Record User;
                     begin
+                        if UserName = '' then
+                            Error(UserNameMandatoryErr);
+
                         User.SetRange("User Name", UserName);
-                        if User.FindFirst() then
-                            ScheduledPerfProfiler.ValidateScheduleCreationPermissions(UserSecurityId(), User."User Security ID");
+                        User.FindFirst();
```
WaelAbuSeada, line 202: Why you remove the if? what if findfirst returned empty based on the filter?

WaelAbuSeada (approved):

src/System Application/App/Performance Profiler/src/PerfProfilerScheduleCard.Page.al
onbuyuka: to make sure user is valid. with the if then you might end up with a invalid user

PR #5639, opened 2025-11-28
[Shopify] Introduce item filter to collections

onbuyuka:
#### Summary 
Introduce item filter to collections
#### Work Item(s) 
Fixes [AB#614352](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/614352)
Fixes [AB#614350](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/614350)

### DIFF
--- a/src/Apps/W1/Shopify/App/src/Products/Codeunits/ShpfyProductAPI.Codeunit.al
+++ b/src/Apps/W1/Shopify/App/src/Products/Codeunits/ShpfyProductAPI.Codeunit.al
@@ -136,9 +136,11 @@ codeunit 30176 "Shpfy Product API"
                 if Item.IsEmpty() then
                     continue;
             end;
+            if CollectionsBuilder.Length > 0 then
+                CollectionsBuilder.Append(', ');
             CollectionsBuilder.Append('\"gid://shopify/Collection/');
             CollectionsBuilder.Append(Format(ProductCollection.Id));
-            CollectionsBuilder.Append('\", ');
+            CollectionsBuilder.Append('\"');
         until ProductCollection.Next() = 0;
 
         if CollectionsBuilder.Length = 0 then
@@ -146,7 +148,6 @@ codeunit 30176 "Shpfy Product API"
 
         GraphQuery.Append(', collectionsToJoin: [');
         GraphQuery.Append(CollectionsBuilder.ToText());
-        GraphQuery.Remove(GraphQuery.Length - 1, 2);
         GraphQuery.Append(']');
     end;
 


src/Apps/W1/Shopify/App/src/Products/Tables/ShpfyProductCollection.Table.al
```diff
@@ -45,4 +50,21 @@ table 30163 "Shpfy Product Collection"
             Clustered = true;
         }
     }
+
+    procedure SetItemFilter(ItemFilter: Text)
+    var
+        OutStream: OutStream;
+    begin
+        "Item Filter".CreateOutStream(OutStream);
```
duichwer: Should the Item Filter maybe be saved with the explicit Encoding UTF8?


Additionally have you tried to save at first large filter and afterwards an smaller filter? I would suggest to clear the blob field at first before you create the Out stream

src/Apps/W1/Shopify/App/src/Products/Tables/ShpfyProductCollection.Table.al
```diff
@@ -45,4 +50,21 @@ table 30163 "Shpfy Product Collection"
...
+        OutStream.Write(ItemFilter);
+    end;
+
+    procedure GetItemFilter() ItemFilter: Text
+    var
+        InStream: InStream;
+    begin
+        CalcFields("Item Filter");
```
duichwer, line 67: If the Blob field has no value then you should simple return an empty text.

src/Apps/W1/Shopify/App/src/Products/Tables/ShpfyProductCollection.Table.al
duichwer: You didn't change here anything.
So the OutStream will be created with the Encoding MsDos 
https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/methods-auto/blob/blob-createoutstream-method

duichwer: Could you please update the CreateInStream method two to include the text Encoding?

duichwer (approved):

src/Apps/W1/Shopify/App/src/Products/Codeunits/ShpfyProductAPI.Codeunit.al
```diff
@@ -117,19 +117,35 @@ codeunit 30176 "Shpfy Product API"
...
+            CollectionsBuilder.Append(Format(ProductCollection.Id));
+            CollectionsBuilder.Append('\", ');
         until ProductCollection.Next() = 0;
+
+        if CollectionsBuilder.Length = 0 then
+            exit;
+
+        GraphQuery.Append(', collectionsToJoin: [');
+        GraphQuery.Append(CollectionsBuilder.ToText());
         GraphQuery.Remove(GraphQuery.Length - 1, 2);
```
darjoo: You can avoid this by using changing the above loop.
I don't like the remove just because it makes it difficult to understand what it is removing exactly.

Copy line 141 to 139, change to `if CollectionsBuilder.length > 0 then CollectionsBuilder.Append(', ')`
Change 141 to `CollectionsBuilder.Append('\"')`

src/Apps/W1/Shopify/App/src/Products/Pages/ShpfyProductCollections.Page.al
```diff
@@ -24,6 +25,25 @@ page 30175 "Shpfy Product Collections"
...
+                {
+                    Caption = 'Item Filter';
+                    ToolTip = 'Specifies the filter criteria for the items in the product collection.';
+
+                    trigger OnAssistEdit()
+                    var
+                        Item: Record Item;
+                        FilterPageBuilder: FilterPageBuilder;
+                    begin
+                        FilterPageBuilder.AddTable(Item.TableCaption(), Database::Item);
```
nikolakukrika, line 38: Minor -  I would recommend to always save / set the filters in the same global language. Alternative is to store in which language the filter is stored.
Otherwise you may run into problems if user A uses one language and user B uses a different one

src/Apps/W1/Shopify/Test/Products/ShpfyProductCollectionTest.Codeunit.al
```diff
@@ -172,6 +192,16 @@ codeunit 139556 "Shpfy Product Collection Test"
         Product.Insert(false);
     end;
 
+    local procedure CreateItem(var Item: Record Item; var Product: Record "Shpfy Product")
+    begin
+        Item.Init();
```
nikolakukrika, line 197: Minor - No inits - use clear or nothing : D

nikolakukrika (approved):

src/Apps/W1/Shopify/Test/Products/ShpfyProductCollectionTest.Codeunit.al
onbuyuka: Test

PR #5651, opened 2025-12-01
Update the Base64 Converter app to use the native Baser64Convert

nhsejth:
#### Summary 
Update the Base64 Converter app to uptake the platform native Base64 converter system codeunit and remove dependencies on DotNet interop using the Converter class. This is part of the platform push to lower the risk of out-of-memory exceptions and migrate expensive DotNet operations to platform code in native codeunits.

#### Work Item(s) 
Fixes [AB#615005](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/615005)

### DIFF
--- a/src/Apps/W1/Subscription Billing/Test/DisabledTests/ServiceObjectTest.json
+++ b/src/Apps/W1/Subscription Billing/Test/DisabledTests/ServiceObjectTest.json
@@ -1,0 +1,8 @@
+[
+    {
+        "bug": "615089",
+        "codeunitId": 148157,
+        "CodeunitName": "Service Object Test",
+        "Method": "UT_UpdateServiceDatesDoesNotCalculateCancellationPossibleUntilAndTermUntilWhenNoticePeriodIsEmpty"
+    }
+]


src/System Application/App/Base64 Convert/src/Base64ConvertImpl.Codeunit.al
```diff
@@ -30,110 +32,81 @@ codeunit 4111 "Base64 Convert Impl."
...
-    var
-        Convert: DotNet Convert;
-        Encoding: DotNet Encoding;
-        Base64FormattingOptions: DotNet Base64FormattingOptions;
     begin
         if SecretString.IsEmpty() then
             exit;
-        Base64FormattingOptions := Base64FormattingOptions.None;
-        exit(Convert.ToBase64String(Encoding.UTF8().GetBytes(SecretString.Unwrap()), Base64FormattingOptions));
+        exit(SystemNativeBase64Converter.ToBase64(SecretString.Unwrap(), false, TextEncoding::UTF8, 0));
```
darjoo, line 96: I would prefer if this didn't need to unwrap. But won't block PR.

darjoo (approved):

src/System Application/App/Base64 Convert/src/Base64ConvertImpl.Codeunit.al
nhsejth: We could add an extra overload in the system codeunit with a SecretText parameter so the platform will deal with the unwrapping. This is require a new platform first. Let me know if that is desired (will track that in a followup bug)

PR #5689, opened 2025-12-02
Agent Task Messages - Download multiple attachments

BazookaMusic:
#### Summary
When trying to download multiple attachments, only the last one was downloaded. This is a limitation in BC.

Solution:
Make a zip out of them if there's more than one, like the Email module does.

#### Work Item(s)
Fixes [AB#615056](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/615056)

### DIFF
--- a/src/System Application/App/Agent/app.json
+++ b/src/System Application/App/Agent/app.json
@@ -55,19 +55,19 @@
       "id": "7e3b999e-1182-45d2-8b82-d5127ddba9b2",
       "name": "DotNet Aliases",
       "publisher": "Microsoft",
-      "version": "27.2.0.0"
+      "version": "28.0.0.0"
     },
     {
       "id": "e31ad830-3d46-472e-afeb-1d3d35247943",
       "name": "BLOB Storage",
       "publisher": "Microsoft",
-      "version": "27.2.0.0"
+      "version": "28.0.0.0"
     },
     {
       "id": "008c9419-0c3b-4d08-b03e-84e3adca689f",
       "name": "Data Compression",
       "publisher": "Microsoft",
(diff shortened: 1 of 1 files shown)


src/System Application/App/Agent/Interaction/Internal/AgentMessageImpl.Codeunit.al
```diff
@@ -94,15 +98,47 @@ codeunit 4308 "Agent Message Impl."
...
+        if AttachmentCount = 1 then begin
             ShowOrDownloadAttachment(AgentTaskMessageAttachment."Task ID", AgentTaskMessageAttachment."File ID", true);
+            exit;
+        end;
+
+        // If multiple files, create a zip
+        DataCompression.CreateZipArchive();
+        repeat
+            if AgentTaskFile.Get(AgentTaskMessageAttachment."Task ID", AgentTaskMessageAttachment."File ID") then begin
+                AgentTaskFile.CalcFields(Content);
```
duichwer, line 130: I would recommend to Clear the InStream variable in the loop. Otherwise you might get problems

BazookaMusic: It's not needed because CreateInStream actually creates an entirely new InStream object behind the scenes (according to the platform code which is unfortunately not visible here)

BazookaMusic: But thank you very much for looking at the PR!

qutreson (approved):

PR #5692, opened 2025-12-02
5638, AB#614485, changed setup page so that assembly is not in the production group.

### LINKED ISSUE #5638: [Vendor] [Quality Management] Separate support for Assembly and Production scenarios in Quality Management
### Describe the issue

Linked to [AB#614485](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/614485)

Separate Assembly and Production in Quality Management

 

Current, Assembly and Production is set up together and the implementation is very tightly coupled with each other.

You see them together in the Quality Management Setup page
Image

You can create generation rule for either Production Orders OR Assembly Orders
Image

This creates issues for Essential users where the UI seems a bit odd. It exposes bits of production to the user who does not have premium license.

 
Image

Image

 
Image

This tight coupling in the UI, translates into code with code blocks like  

Image

and

Image

The ask here is to separate the Quality Management to support both Assembly and Production separately so that both Essential and Premium users get cleaner experience with Quality Management.

### Expected behavior

Support for Assembly and Production is setup and executed separately.

### Steps to reproduce

Go through Quality Managemnt Setup page or the wizard, you can see Assembly and Production goes together and for Essential User, the user experience is a bit odd

### Additional context

_No response_

- [ ] I will provide a fix for a bug

insightworks007:
Replaced with PR https://github.com/microsoft/BCApps/pull/6071 to handle recent object renames.

#### Summary 
Changed the setup page to separate the assembly setting into it's own group, out of the produciton group.
As clarified in teams, did not make changes to the wizard since there is a re-design of the wizard soon.
For a similar reason I did not add a link to start the wizard under the same presumption.

#### Work Item(s) 
Fixes #5638 (  [AB#614485](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/614485) ).


PredragMaricic (approved):

ChethanT (commented):
'Quality Management Setup' looks like the following for a Essential user.
image

ChethanT:
> > @insightworks007, setting the test generation rule for assembly output image
> > and doing a post preview on Assembly Order gives a bad used experience. 'Quality Inspection Test' window opens behind preview window with empty values image
> > Secondly, there are no 'Create Assembly Rule' nor 'Edit Assembly Rule' on the 'Quality Inspection Test Generation Rules' page
> 
> @ChethanT The post preview issue is going to require someone from Microsoft to make a change in codeunit 900 "Assembly-Post" as a pre-requisite such that the "PreviewMode" variable is accessible via the posting event, or another way of making it accessible. This flag in that codeunit is set via SetPreviewMode, but there is no way to retrieve it during the Codeunit::"Assembly-Post", 'OnAfterPost' event. Is there an other option with the existing codebase that I could use instead?

I have opened bug AB#616805 to track this issue.

ChethanT:
> > 'Quality Management Setup' looks like the following for a Essential user. image
> 
> @ChethanT This doesn't seem as simple as removing 'Assembly' from the application area on the field with assembly inside of the production group. The only way I can find for the production group to be completely gone is to remove the "InstructionalText" from group(SettingsForProductionAutomation) and group(SettingsForProduction) in conjunction with also removing 'Assembly' from the field(ChooseCreateNewRule_Production) link.
> 
> Approach 1 is I could remove the instructional text for those groups in addition to removing the assembly application area in the one field. Approach 2 is we just remove the assembly application area (the instructional text will still show). Is there a 3rd option to consider?

Apart from moving Assembly to its own group, you can also hide the Production by controlling the visibility of the group. This is how it would look for Essential user
image

PR #5769, opened 2025-12-09
[Shopify] Do not create Shopify Customer for the company if it exists

onbuyuka:
#### Summary 
Do not create Shopify Customer for the company if it exists
Also removed an internal ishandled event :)

#### Work Item(s) 
Fixes [AB#615837](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/615837)
Fixes [AB#615851](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/615851)

### DIFF
--- a/src/Apps/W1/Shopify/App/src/Customers/Codeunits/ShpfyCreateCustomer.Codeunit.al
+++ b/src/Apps/W1/Shopify/App/src/Customers/Codeunits/ShpfyCreateCustomer.Codeunit.al
@@ -172,8 +172,7 @@ codeunit 30110 "Shpfy Create Customer"
 
         Customer.Modify();
 
-        ShopifyCustomer.SetRange(Id, TempShopifyCustomer.Id);
-        if ShopifyCustomer.IsEmpty() then begin
+        if not ShopifyCustomer.Get(TempShopifyCustomer.Id) then begin
             ShopifyCustomer.Copy(TempShopifyCustomer);
             ShopifyCustomer."Customer SystemId" := Customer.SystemId;
             ShopifyCustomer.Insert();


src/Apps/W1/Shopify/App/src/Customers/Codeunits/ShpfyCreateCustomer.Codeunit.al
```diff
@@ -172,9 +172,12 @@ codeunit 30110 "Shpfy Create Customer"
 
         Customer.Modify();
 
-        ShopifyCustomer.Copy(TempShopifyCustomer);
-        ShopifyCustomer."Customer SystemId" := Customer.SystemId;
-        ShopifyCustomer.Insert();
+        ShopifyCustomer.SetRange(Id, TempShopifyCustomer.Id);
+        if ShopifyCustomer.IsEmpty() then begin
```
darjoo: This is insufficient, on line 183 it takes `ShopifyCustomer.Id` but you don't do a get here, as such, `ShopifyCustomer.Id` will be default value. This only works if the customer doesn't exist.

You probably want `if not ShopifyCustomer.Get(TempShopifyCustomer.Id)`. I didn't check the primary key, so validate before putting this code :D

darjoo (changes requested):

src/Apps/W1/Shopify/App/src/Customers/Codeunits/ShpfyCreateCustomer.Codeunit.al
onbuyuka: great catch

PR #5811, opened 2025-12-11
[27.x] ADI V4  

Groenbech96:
#### Summary 

Enable version 4 of ADI.

#### Work Item(s) 
Fixes
[AB#616060](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/616060)


darjoo (approved):

VolodySyn (approved):

Jose-agg (changes requested):
What is the change in accuracy with this change? Is it running faster or slower than before? Last update is much slower (we went from an average of 10 to 30 seconds to now taking between 40 seconds to 2 minutes), so if the performance is impacted and the accuracy not so much we might not want to update yet

Jose-agg (approved):

PR #5866, opened 2025-12-16
[Agent] Handle Company field in Agent Access Controls

qutreson:
#### Work Item(s) 
Fixes [AB#614168](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/614168)
Fixes [AB#617141](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/617141)

### DIFF
--- a/src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
+++ b/src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
@@ -521,6 +521,13 @@ codeunit 4301 "Agent Impl."
         UserSettings: Codeunit "User Settings";
     begin
         UserSettings.GetAllowedCompaniesForUser(AgentUserSecurityID, TempCompany);
+        if TempCompany.IsEmpty() then begin
+#pragma warning disable AA0139
+            SingleCompanyName := CompanyName();
+#pragma warning restore AA0139
+            exit(true);
+        end;
+
         if TempCompany.Count() <> 1 then
             exit(false);
 


src/System Application/App/Agent/Setup/AgentUserSubform.PageExt.al
```diff
@@ -89,28 +87,16 @@ pageextension 4318 "Agent User Subform" extends "User Subform"
             IsAgent := false;
 
         if not ShowCompanyFieldOverride then begin
-            ShowCompanyField := not AccessControlForSingleCompany(GlobalSingleCompanyName);
+            if IsAgent then
```
qutreson: Remove the if IsAgent

src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
```diff
@@ -408,20 +410,28 @@ codeunit 4301 "Agent Impl."
...
+                repeat
+                    AgentAccessControl.Delete()
+                until AgentAccessControl.Next() = 0;
             exit;
         end;
 
-        if AgentAccessControl.Get(Agent."User Security ID", UserSecurityId()) then begin
-            AgentAccessControl.TransferFields(TempAgentAccessControl, true);
-            AgentAccessControl."Agent User Security ID" := Agent."User Security ID";
+        // TODO(qutreson) - This needs to loop on records.
```
qutreson: TODO + add tests

src/System Application/App/Agent/Setup/AccessControls/SelectAgentPermissions.Page.al
```diff
@@ -0,0 +1,76 @@
...
+    end;
+
+    internal procedure GetTempAccessControlBuffer(var TempAccessControlBuffer: Record "Access Control Buffer" temporary)
+    begin
+        TempAccessControlBuffer.Copy(Rec, true);
+    end;
+
+    local procedure BackupAccessControlBuffer()
+    begin
+        TempBackupAccessControlBuffer.Copy(Rec, true);
```
nikolakukrika: I'm not sure if this will work. It may restore the first row, but will it restore all other rows if it is by reference?

nikolakukrika: Test with updating the third row, deleting the second and inserting a new row. See will it rollback the changes. If not we need a deep copy.

src/System Application/App/Agent/Setup/AccessControls/SelectAgentPermissionsPart.Page.al
```diff
@@ -0,0 +1,273 @@
...
+                            PermissionSetLookupRecord.SetRecFilter();
+                            exit(true);
+                        end;
+                    end;
+
+                    trigger OnValidate()
+                    begin
+                        PermissionSetLookupRecord.SetRange("Role ID", Rec."Role ID");
+                        if PermissionSetLookupRecord.FindFirst() then begin
+                            if PermissionSetLookupRecord.Count > 1 then
```
nikolakukrika: Nit - Count()
Enable all rules so we make sure we are not introducing warnings

src/System Application/App/Agent/Setup/AccessControls/SelectAgentPermissionsPart.Page.al
```diff
@@ -0,0 +1,273 @@
...
+            PermissionScope := '';
+        end;
+
+        if not ShowCompanyFieldOverride then begin
+            ShowCompanyField := not TryGetAccessControlForSingleCompany(GlobalSingleCompanyName);
+            CurrPage.Update(false);
+        end;
+    end;
+
+    local procedure AssignMyPermissions()
```
nikolakukrika: This should not be on the page. We can move it to one of the codeunits, maybe Agent Impl. or create a new one to handle permissions

src/System Application/App/Agent/Setup/AccessControls/ViewAgentPermissions.Page.al
```diff
@@ -133,29 +140,60 @@ page 4334 "View Agent Permissions"
...
+    end;
 
-        SingleCompanyName := TempCompany.Name;
-        exit(true);
+    local procedure SaveAccessControlBuffer(UserSecurityID: Guid; var TempAccessControlBuffer: Record "Access Control Buffer" temporary)
+    var
+        AccessControl: Record "Access Control";
+    begin
+        // Delete all existing access control records for the agent
+        AccessControl.SetRange("User Security ID", UserSecurityID);
```
nikolakukrika: I'm not sure that this is the best thing to do. Maybe we can follow the same pattern of deleting row by row to avoid locking out some scenarios. Row by row update would be less disrupting. 
The flow would be the same, first delete the sets that do not exist or insert new ones. There are few records here so we can afford spending a bit more time on this

src/System Application/App/Agent/Setup/AgentAccessControls/SelectAgentAccessControl.Page.al
```diff
@@ -0,0 +1,70 @@
...
+    end;
+
+    internal procedure GetTempAgentAccessControl(var TempAgentAccessControl: Record "Agent Access Control" temporary)
+    begin
+        TempAgentAccessControl.Copy(Rec, true);
+    end;
+
+    local procedure BackupAgentAccessControl()
+    begin
+        TempBackupAgentAccessControl.Copy(Rec, true);
```
nikolakukrika: Same comment here, check if it is a deep copy or not

src/System Application/App/Agent/Setup/AgentAccessControls/SelectAgentAccessCtrlPart.Page.al
```diff
@@ -0,0 +1,219 @@
...
+            exit;
+
+        // Check if there's at least one other owner in the temp table
+        TempAgentAccessControl.Copy(Rec, true);
+        TempAgentAccessControl.SetRange("Can Configure Agent", true);
+        TempAgentAccessControl.SetFilter("User Security ID", '<>%1', Rec."User Security ID");
+        TempAgentAccessControl.SetRange("Agent User Security ID", AgentUserSecurityID);
+
+        if TempAgentAccessControl.IsEmpty() then
+            Error('One owner must be defined for the agent.');
```
nikolakukrika: Inline errors will not be transleted, use label

src/System Application/App/Agent/Setup/AgentAccessControls/SelectAgentAccessCtrlPart.Page.al
```diff
@@ -0,0 +1,219 @@
...
+        if Evaluate(UserGuid, NewUserName) then begin
+            if not User.Get(UserGuid) then
+                exit(false);
+            UserSecurityID := User."User Security ID";
+            exit(true);
+        end;
+
+        User.SetRange("User Name", NewUserName);
+        if not User.FindFirst() then begin
+            User.SetFilter("User Name", '@*''''' + NewUserName + '''''*');
```
nikolakukrika, line 198: Please double check this code, it looks wrong to me. 
This filter means that we will find a substring
e.g. if the user name is Sotiris in the system
if I type Iris I will get Sotiris

I would expect that autocomplete should use a prefix not any part of the string.

src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
```diff
@@ -542,6 +521,19 @@ codeunit 4301 "Agent Impl."
         Page.RunModal(SetupPageId, SourceRecordVariant);
     end;
 
+    procedure TryGetAccessControlForSingleCompany(AgentUserSecurityID: Guid; var SingleCompanyName: Text[30]): Boolean
```
nikolakukrika: Nit - I would not call this one with try as it is not trying. GetAccessControlForSIngleCompany

nikolakukrika (approved):

src/System Application/App/Agent/Setup/AccessControls/SelectAgentPermissionsPart.Page.al
qutreson: Adding the parenthesis - we don't have a rule for this AFAIK

src/System Application/App/Agent/Setup/AgentAccessControls/SelectAgentAccessCtrlPart.Page.al
qutreson: This filter was already in the Agent Access Control part before my changes, I won't change. We should take it separately if that's not the behavior we want.

src/System Application/App/Agent/Setup/AccessControls/SelectAgentPermissions.Page.al
qutreson: It seems to work, I am using this for the Update and Cancel paths from the part and it preserved more than one record.

src/System Application/App/Agent/Setup/AccessControls/ViewAgentPermissions.Page.al
qutreson: I'll re-use the utility I am introducing in https://github.com/microsoft/BCApps/pull/5885/changes#diff-48700ee20e3f98a43ff4f22d4bfa9526920e1f4e47545920fae19a4f086aefcf

src/System Application/App/Agent/Setup/AgentAccessControls/SelectAgentAccessControl.Page.al
qutreson: Seems to work!

src/System Application/App/Agent/Setup/AccessControls/SelectAgentPermissionsPart.Page.al
qutreson: I actually should remove this, it should be here only for the playground agents and isn't needed for the generic Agent Card scenario.

src/System Application/App/Agent/Setup/AgentAccessControls/SelectAgentAccessCtrlPart.Page.al
qutreson: Fixed

src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
qutreson: Renamed

src/System Application/App/Agent/Setup/AgentAccessControls/SelectAgentAccessControl.Page.al
qutreson: Actually, you were right, I reverted now!

src/System Application/App/Agent/Setup/AgentAccessControls/SelectAgentAccessCtrlPart.Page.al
```diff
@@ -0,0 +1,245 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Agents;
+
+using System.Security.AccessControl;
+
+page 4325 "Select Agent Access Ctrl Part"
```
qutreson: Dup - need another one

src/System Application/App/Agent/Setup/AccessControls/AgentUserSubform.PageExt.al
```diff
@@ -0,0 +1,66 @@
...
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Agents;
+
+using System.Security.AccessControl;
+using System.Security.User;
+
+pageextension 4318 "Agent User Subform" extends "User Subform"
+{
+    ObsoleteState = Pending;
```
nikolakukrika, line 14: Why do we need a new object in state pending?
We could delete the controls and pragma suppress the warning

src/System Application/App/Agent/Setup/AccessControls/SelectAgentPermissions.Page.al
```diff
@@ -0,0 +1,96 @@
...
+
+using System.Security.AccessControl;
+
+page 4336 "Select Agent Permissions"
+{
+    PageType = StandardDialog;
+    ApplicationArea = All;
+    SourceTable = "Access Control Buffer";
+    SourceTableTemporary = true;
+    Caption = 'Edit Agent Permissions (Preview)';
```
nikolakukrika: Do we need preview in the caption?
I would not include it as we do not have it in other places.

src/System Application/App/Agent/Setup/AccessControls/SelectAgentPermissionsPart.Page.al
```diff
@@ -0,0 +1,220 @@
...
+            {
+                Caption = 'Agent permissions';
+
+                field(PermissionSet; Rec."Role ID")
+                {
+                    ApplicationArea = Basic, Suite;
+                    Caption = 'Permission Set';
+                    ToolTip = 'Specifies the ID of a security role that has been assigned to this Windows login in the current database.';
+                    Style = Unfavorable;
+                    StyleExpr = PermissionSetNotFound;
```
nikolakukrika, line 37: Nit - there is a simpler way of setting these, you can set the StyleExp as style, you can check the other places
E.g. StyleExpr = 'Unfavorable'

src/System Application/App/Agent/Setup/AccessControls/SelectAgentPermissionsPart.Page.al
```diff
@@ -0,0 +1,220 @@
...
+
+                    trigger OnLookup(var Text: Text): Boolean
+                    var
+                        LookupPermissionSet: Page "Lookup Permission Set";
+                    begin
+                        LookupPermissionSet.LookupMode := true;
+                        if LookupPermissionSet.RunModal() = ACTION::LookupOK then begin
+                            LookupPermissionSet.GetRecord(PermissionSetLookupRecord);
+                            Text := PermissionSetLookupRecord."Role ID";
+                            PermissionSetLookupRecord.SetRecFilter();
```
nikolakukrika, line 47: This is not how we usually do it. We do not need the global lookup record, it is simple to use the rec. You could call the same way of validating from both lookup and OnValidate trigger.

src/System Application/App/Agent/Setup/AccessControls/SelectAgentPermissionsPart.Page.al
```diff
@@ -0,0 +1,220 @@
...
+                            exit(true);
+                        end;
+                    end;
+
+                    trigger OnValidate()
+                    begin
+                        PermissionSetLookupRecord.SetRange("Role ID", Rec."Role ID");
```
nikolakukrika, line 54: Nit - we usually copy the record and copy it back after the operation.

src/System Application/App/Agent/Setup/AccessControls/SelectAgentPermissionsPart.Page.al
```diff
@@ -0,0 +1,220 @@
...
+                        if PermissionSetLookupRecord.FindFirst() then begin
+                            if PermissionSetLookupRecord.Count() > 1 then
+                                Error(MultipleRoleIDErr, Rec."Role ID");
```
nikolakukrika, line 57: I'm not sure what will happen to global variables, error will revert all of the database changes.

src/System Application/App/Agent/Setup/AccessControls/SelectAgentPermissionsPart.Page.al
```diff
@@ -0,0 +1,220 @@
...
+        AgentSingleCompany: Boolean;
+    begin
+        AgentUserSecurityID := NewAgentUserSecurityID;
+        Rec.Copy(TempAccessControlBuffer, true);
+
+        AgentSingleCompany := GetAccessControlForSingleCompany(GlobalSingleCompanyName);
+        if not ShowCompanyFieldOverride then
+            ShowCompanyField := not AgentSingleCompany;
+
+        CurrPage.Update(false);
```
nikolakukrika: This can be problematic, it is better to call Update in a separate call to avoid loops and unexpected currpage.update

nikolakukrika: Also no update should be needed if we initialize, different name should be used if we expect to use this method in update

src/System Application/App/Agent/Setup/AgentAccessControls/SelectAgentAccessCtrlPart.Page.al
```diff
@@ -0,0 +1,245 @@
...
+        if Rec."Can Configure Agent" then
+            VerifyOwnerExists();
+        exit(true);
+    end;
+
+    trigger OnInsertRecord(BelowxRec: Boolean): Boolean
+    var
+        UserSecurityID: Guid;
+    begin
+        // If UserName is already populated (e.g., when duplicating a row), validate it to set User Security ID.
```
nikolakukrika, line 133: Nit - it is much better to have a named method instead of coding within the trigger

src/System Application/App/Agent/Setup/AccessControls/AgentUserSubform.PageExt.al
qutreson: It is not a new object, I just moved it under a new folder, and simplified the logic within it, and the diff tool from Github doesn't see it as a Rename/Edit but as a Delete/New.

src/System Application/App/Agent/Setup/AgentAccessControls/SelectAgentAccessCtrlPart.Page.al
```diff
@@ -0,0 +1,245 @@
...
+        AgentSingleCompany: Boolean;
+    begin
+        AgentUserSecurityID := NewAgentUserSecurityId;
+        Rec.Copy(TempAgentAccessControl, true);
+
+        AgentSingleCompany := AgentImpl.GetAccessControlForSingleCompany(AgentUserSecurityID, GlobalSingleCompanyName);
+        if not ShowCompanyFieldOverride then
+            ShowCompanyField := not AgentSingleCompany;
+
+        CurrPage.Update(false);
```
nikolakukrika: Same comment on initialize

src/System Application/App/Agent/Setup/AgentAccessControls/SelectAgentAccessCtrlPart.Page.al
```diff
@@ -0,0 +1,245 @@
...
+            User.SetFilter("User Name", '@*''''' + NewUserName + '''''*');
+            if not User.FindFirst() then
+                exit(false);
+        end;
+
+        UserSecurityID := User."User Security ID";
+        exit(true);
+    end;
+
+    local procedure VerifyOwnerExists()
```
nikolakukrika, line 207: Nit - this could be moved outside of the page

src/System Application/App/Agent/Setup/AgentAccessControls/SelectAgentAccessCtrlPart.Page.al
```diff
@@ -0,0 +1,245 @@
...
+
+        OwnerFound := false;
+        if Rec.FindSet() then
+            repeat
+                if not ((Rec."User Security ID" = CurrentUserSecurityID) and (Rec."Company Name" = CurrentCompanyName)) then
+                    OwnerFound := true;
+            until (Rec.Next() = 0) or OwnerFound;
+
+        if not OwnerFound then begin
+            Rec.Copy(TempAgentAccessControl);
```
nikolakukrika: nit - duplicated rec.copy. It can be moved outside of if and we can have a single call

src/System Application/App/Agent/Setup/AccessControls/SelectAgentPermissions.Page.al
qutreson: Removing it.

src/System Application/App/Agent/Setup/AgentAccessControls/SelectAgentAccessCtrlPart.Page.al
qutreson: Updated

nikolakukrika (approved):

src/System Application/App/Agent/Setup/AgentAccessControls/SelectAgentAccessCtrlPart.Page.al
qutreson: Updating - duplicate line, but called only once

src/System Application/App/Agent/Setup/AccessControls/SelectAgentPermissionsPart.Page.al
qutreson: Updating

src/System Application/App/Agent/Setup/AccessControls/SelectAgentPermissionsPart.Page.al
qutreson: This logic used to be in the playground and we didn't get bugs logged for it. Also, there shouldn't be database changes since it's all based on temporary records

src/System Application/App/Agent/Setup/AccessControls/SelectAgentPermissionsPart.Page.al
qutreson: I'll leave the code as it is. This is code moved from the playground and was inspired from the UserSubform page which does the same.

src/System Application/App/Agent/Setup/AccessControls/SelectAgentPermissionsPart.Page.al
qutreson: I'll leave the code as it is. This is code moved from the playground and was inspired from the UserSubform page which does the same.

src/System Application/App/Agent/Setup/AccessControls/SelectAgentPermissionsPart.Page.al
qutreson: I'll leave the code as it is. This is code moved from the playground and was inspired from the UserSubform page which does the same.

src/System Application/App/Agent/Setup/AgentAccessControls/SelectAgentAccessCtrlPart.Page.al
qutreson: Updating

src/System Application/App/Agent/Setup/AgentAccessControls/SelectAgentAccessCtrlPart.Page.al
qutreson: We used to have it within the AgentImpl codeunit when it worked on actual records. Now that this all works based on temporary records, it is less likely to be re-used. It adds a bit of code on the page, but keeping the AgentImpl codeunit with useful utilities is also nice

src/System Application/App/Agent/Setup/AccessControls/SelectAgentPermissions.Page.al
qutreson: Done!

src/System Application/App/Agent/Setup/AccessControls/SelectAgentPermissionsPart.Page.al
qutreson: Done!

src/System Application/App/Agent/Setup/AgentAccessControls/SelectAgentAccessCtrlPart.Page.al
qutreson: Done

nikolakukrika (approved):

src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
```diff
@@ -551,8 +515,20 @@ codeunit 4301 "Agent Impl."
         Page.RunModal(SetupPageId, SourceRecordVariant);
     end;
 
+    procedure GetAccessControlForSingleCompany(AgentUserSecurityID: Guid; var SingleCompanyName: Text[30]): Boolean
+    var
+        TempCompany: Record Company temporary;
+        UserSettings: Codeunit "User Settings";
+    begin
+        UserSettings.GetAllowedCompaniesForUser(AgentUserSecurityID, TempCompany);
+        if TempCompany.Count() <> 1 then
```
qutreson, line 531: Check scenario where agent has no company (i.e. first time setup).

src/System Application/App/Agent/Setup/AgentAccessControls/ViewAgentAccessControl.Page.al
```diff
@@ -0,0 +1,151 @@
...
+page 4320 "View Agent Access Control"
+{
+#pragma warning restore AS0125
+
+    Caption = 'Agent Access Control';
+    PageType = ListPart;
+    ApplicationArea = All;
+    SourceTable = "Agent Access Control";
+    Editable = false;
+    InsertAllowed = false;
```
qutreson, line 21: Extensible = false

src/System Application/App/Agent/Setup/AccessControls/ViewAgentPermissions.Page.al
qutreson, line 16: Extensible = false

src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
```diff
@@ -374,68 +359,47 @@ codeunit 4301 "Agent Impl."
     local procedure UpdateAgentAccessControl(var TempAgentAccessControl: Record "Agent Access Control" temporary; var Agent: Record Agent)
     begin
         // We must delete or update the user doing the change the last to avoid removing permissions that are needed to commit the change
-        UpdateUsersOtherThanMainUser(TempAgentAccessControl, Agent);
+        UpdateAgentAccessControlForUsers(TempAgentAccessControl, Agent, '<>%1', UserSecurityId());
```
darjoo, line 362: I don't think we should have a parameter for inputting the UserSecurityIdFilter. It's more likely for future errors if we need to put in the filter. 

At least I think it should be hardcoded in another procedure that'll call `UpdateAgentAccessContorlForUsers` but it'll be more explicit with a proper procedure name. 

Maybe even separate the two functions from this, for delete and update

qutreson: Discussed, it's all local functions and the functions and it's not strictly a two step process with delete and update, but rather two distinct update steps (first for other users, then for the current one).

BazookaMusic (approved):

PR #5885, opened 2025-12-18
[Agent SDK] Additional validation on task creation

qutreson:
#### Summary 
Additional validation on task creation:
1. Check that a message is included to ensure a user intervention is raised before the task runs
2. Check that the agent has at least some permissions in the company where the task is created

#### Work Item(s) 
Fixes [AB#616256](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/616256)
Fixes [AB#616649](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/616649)

### DIFF
--- a/src/System Application/App/Agent/Interaction/Internal/AgentTaskBuilderImpl.Codeunit.al
+++ b/src/System Application/App/Agent/Interaction/Internal/AgentTaskBuilderImpl.Codeunit.al
@@ -36,7 +36,7 @@ codeunit 4310 "Agent Task Builder Impl."
         AgentTaskImpl: Codeunit "Agent Task Impl.";
     begin
         VerifyMandatoryFieldsSet();
-        VerifyTaskCanBeCreated(GlobalAgentUserSecurityId, RequiresMessage);
+        VerifyTaskCanBeCreated(RequiresMessage);
 
         AgentTaskImpl.CreateTask(GlobalAgentUserSecurityId, GlobalTaskTitle, GlobalExternalID, AgentTaskRecord);
         if MessageSet then begin
@@ -109,7 +109,7 @@ codeunit 4310 "Agent Task Builder Impl."
         Error(CodingErrorInfo);
     end;
 
-    local procedure VerifyTaskCanBeCreated(GlobalAgentUserSecurityId: Guid; RequiresMessage: Boolean)
+    local procedure VerifyTaskCanBeCreated(RequiresMessage: Boolean)
     var
         AllowedCompanies: Text;
     begin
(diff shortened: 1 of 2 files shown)


src/System Application/App/Agent/Interaction/Internal/AgentTaskImpl.Codeunit.al
```diff
@@ -186,6 +194,26 @@ codeunit 4300 "Agent Task Impl."
...
+            // 1. Tasks without messages cannot be set to ready to enforce a user intervention to approve the message/task.
+            // This temporary restriction ensures that the user reviews the message/task at least once.
+            AgentTaskMessage.SetRange("Task ID", AgentTask.ID);
+            if AgentTaskMessage.IsEmpty() then
+                Error(TaskCannotBeSetToReadyWithoutMessagesErr);
+        end;
+
+        // 2. Tasks cannot be set to ready if the agent has 0 access controls in the current company.
+        AccessControl.SetRange("User Security ID", AgentTask."Agent User Security ID");
+        AccessControl.SetRange("Company Name", CompanyName());
```
BazookaMusic: Company name or empty string

BazookaMusic: Also is this for the SDK only or better at platform level? If platform, there's utilities to check if a user can login

src/System Application/App/Agent/Interaction/AgentTaskBuilder.Codeunit.al
```diff
@@ -35,34 +35,38 @@ codeunit 4315 "Agent Task Builder"
...
-    /// <remarks>
-    /// The builder keeps the state, do not reuse the same instance of the builder to create multiple tasks. 
-    /// </remarks>
+    /// <returns>Agent task that was created.</returns>
+    /// <remarks>The builder keeps the state, do not reuse the same instance of the builder to create multiple tasks.</remarks>
     procedure Create(): Record "Agent Task"
     begin
         FeatureAccessManagement.AgentTaskManagementPreviewEnabled(true);
-        exit(AgentTaskBuilderImpl.Create(true));
+        exit(AgentTaskBuilderImpl.Create(true, true));
```
BazookaMusic, line 43: So we enforce this for the playground too?

qutreson: No, I am updating the playground to use the OnPrem overload below. As discussed, since the task creation from the playground is done "manually", it counts as a pre-review of the message.

src/System Application/App/Agent/Interaction/AgentTaskBuilder.Codeunit.al
```diff
@@ -35,34 +35,38 @@ codeunit 4315 "Agent Task Builder"
...
+
+    /// <summary>
+    /// Create a new task for the agent.
+    /// </summary>
+    /// <param name="SetTaskStatusToReady">Specifies if the task status should be set to ready after creation.</param>
+    /// <param name="RequiresMessage">Specifies whether a message is required, default is true.</param>
+    /// <returns>Agent task that was created.</returns>
+    /// <remarks>The builder keeps the state, do not reuse the same instance of the builder to create multiple tasks.</remarks>
+    [Scope('OnPrem')]
+    procedure Create(SetTaskStatusToReady: Boolean; RequiresMessage: Boolean): Record "Agent Task"
```
nikolakukrika, line 66: We should not design the builder like this. We have message builder, if it is set we append a message, if not no message is created. We should reuse the message builder. There is no need for an new overload method

nikolakukrika: If we want to allow the creation of the tasks without a message we could have a temporary method that is called AllowCreatingTasksWithoutMessages, that is OnPrem and hopefully temporary

src/System Application/App/Agent/Interaction/Internal/AgentTaskBuilderImpl.Codeunit.al
```diff
@@ -40,7 +40,7 @@ codeunit 4310 "Agent Task Builder Impl."
         end;
 
         if SetTaskStatusToReady then
-            AgentTaskImpl.SetTaskStatusToReadyIfPossible(AgentTaskRecord);
```
nikolakukrika, line 43: Can't we have a caller verification to see if it is Microsoft owned app and throw an error if no message builder is set on the lines above and throw an error if the message is not set? Shouldn't this be the simplest implementation?

src/System Application/App/Agent/Interaction/Internal/AgentTaskImpl.Codeunit.al
qutreson: We have utilities in AL to know the companies a user has access to, I'll change the logic to re-use that, it would cover the empty company case, good catch

src/System Application/App/Agent/Interaction/AgentTaskBuilder.Codeunit.al
qutreson: Updated as discussed

src/System Application/App/Agent/Interaction/Internal/AgentTaskImpl.Codeunit.al
qutreson: Fixed

src/System Application/App/Agent/Interaction/Internal/AgentTaskBuilderImpl.Codeunit.al
qutreson: Using the new overload is achieving a similar result and allows for easier testing

PR #5886, opened 2025-12-18
Agents - Uptake of reusable agent setup configuration page 

nikolakukrika:
#### Summary
Changes needed to be able to uptake the part on the Sales Order agent and custom agent

#### Work Item(s) 
Fixes [AB#597811](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/597811)


src/System Application/App/Agent/Setup/AgentSetupImpl.Codeunit.al
```diff
@@ -139,6 +139,32 @@ codeunit 4325 "Agent Setup Impl."
...
+    var
+        UserSettings: Record "User Settings";
+        Language: Codeunit Language;
+        NewSummaryText: Text;
+    begin
+        if SummaryText.Contains(LanguageUsedLbl) then
+            exit;
+
+        AgentSetupBuffer.GetUserSettings(UserSettings);
+        NewSummaryText := SummaryText + '\\' + StrSubstNo(LanguageUsedLbl, CopyStr(Language.GetWindowsLanguageName(UserSettings."Language ID"), 1, MaxStrLen(AgentSetupBuffer."Language Used")));
```
qutreson: For languages that read from right to left, should we make this a formattable label with placeholders instead of just concatenating strings? This would allow for translation apps to swap around the placeholders if needed.

nikolakukrika: Fixed, thanks

PR #5927, opened 2025-12-23
[Quality Management] Quality Inspection - Grade Code field doesn't support "open list" action

tomasevicst:
#### Summary 
Added to DrillDown Results list from Result Code

#### Work Item(s) 
Fixes [AB#615771](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/615771)

### DIFF
--- a/src/Apps/W1/Quality Management/app/src/Document/QltyInspectionList.Page.al
+++ b/src/Apps/W1/Quality Management/app/src/Document/QltyInspectionList.Page.al
@@ -58,8 +58,6 @@ page 20408 "Qlty. Inspection List"
                 }
                 field("Result Code"; Rec."Result Code")
                 {
-                    Visible = false;
-
                     trigger OnDrillDown()
                     begin
                         Rec.DrillDownResultCode();


src/Apps/W1/Quality Management/app/src/Document/QltyInspection.Page.al
```diff
@@ -110,9 +111,14 @@ page 20406 "Qlty. Inspection"
                     field("Result Code"; Rec."Result Code")
                     {
                         Importance = Additional;
-                    }
-                    field("Result Description"; Rec."Result Description")
-                    {
+
+                        trigger OnDrillDown()
```
JakovljevicDusan, line 114: Suggestion: Move to table-level, so improvement also encompasses list page.

src/Apps/W1/Quality Management/app/src/Document/QltyInspection.Page.al
```diff
@@ -110,9 +110,11 @@ page 20406 "Qlty. Inspection"
                     field("Result Code"; Rec."Result Code")
```
pri-kise, line 110: Shouldn't it be sufficent to set the LookupPageId and the DrillDownPageId on the Table `"Qlty. Inspection Result"`?
Since the table Relation is set this should lead to the plattform  drildown/lookup behaviour.

tomasevicst: I think no, as we want to filter on "Result Code".

PR #6012, opened 2026-01-02
[Agents] Handle empty rows when adding permissions or access controls

qutreson:
#### Work Item(s) 
Fixes [AB#615833](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/615833)


src/System Application/App/Agent/Setup/AccessControls/SelectAgentPermissionsPart.Page.al
```diff
@@ -211,6 +216,13 @@ page 4340 "Select Agent Permissions Part"
         exit(AgentImpl.GetAccessControlForSingleCompany(AgentUserSecurityID, SingleCompanyName));
     end;
 
+    local procedure DeleteEmptyRecords()
+    begin
+        Rec.SetRange("Role ID", '');
```
BazookaMusic: Can you instead set NotBlank = true on the role ID field? That should give a red star and force you to fill it in

BazookaMusic (approved):

darjoo (approved):

PR #6040, opened 2026-01-06
[Agent] Show reasoning of agents in Agent Task Log Entry List

qutreson:
#### Summary 
**Content:**
- Exposing the reason in the Agent Task Log Entry card
- The feedback support and disclaimer are defined in AL via a label and an action
- The action is only enabled for 1P and user-created agents since we don't support feedback for 3P agents yet
- Adding a new utility codeunit around feedback which re-uses the same known values as the client for the timeline

#### Work Item(s) 
Fixes [AB#617282](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/617282)

### DIFF
--- a/src/System Application/App/Agent/app.json
+++ b/src/System Application/App/Agent/app.json
@@ -68,6 +68,12 @@
       "name": "Data Compression",
       "publisher": "Microsoft",
       "version": "28.0.0.0"
+    },
+    {
+      "id": "f7964d32-7685-400f-8297-4bc17d0aab0e",
+      "name": "Microsoft User Feedback",
+      "publisher": "Microsoft",
+      "version": "28.0.0.0"
     }
   ],
   "propagateDependencies": true,


BazookaMusic (approved):

src/System Application/App/Agent/Interaction/Internal/AgentTaskImpl.Codeunit.al
```diff
@@ -206,6 +206,17 @@ codeunit 4300 "Agent Task Impl."
         exit((AgentTask.Status = AgentTask.Status::"Stopped by User") or (AgentTask.Status = AgentTask.Status::"Stopped by System"));
     end;
 
+    internal procedure TryGetAgentRecordFromTaskId(TaskId: Integer; var Agent: Record Agent): Boolean
+    var
+        AgentTask: Record "Agent Task";
+    begin
+        if AgentTask.Get(TaskId) then
```
nikolakukrika, line 213: Should we check on permissions, or do we want to throw an error if the user does not have the permissions?
I believe platform does not swallow permission errors.

nikolakukrika (approved):

src/System Application/App/Agent/Interaction/Internal/AgentTaskImpl.Codeunit.al
BazookaMusic: I think it's by design that we do not eat permission exceptions, so we should not hide it here.

It will not be very actionable. A lot of other things will fail if they don't have access to the agent/task

qutreson: The platform will through if the user has no permissions for the Agent Task table in general.
The records from the Agent Task and Agent table would be filtered based on agent access controls too.
I'd expect that it won't be an issue though. This method would generally not be called if the user is not already interacting with agent tasks. We also had the same logic in the agent playground - I just moved it here for reusability.

PR #6052, opened 2026-01-07
[Shopify] Align extensibility between order and refund creation

onbuyuka:
#### Summary 
Align extensibility between order and refund creation

#### Work Item(s) 
Fixes [AB#617361](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/617361)

### DIFF
--- a/src/Apps/W1/Shopify/App/src/Order Return Refund Processing/Codeunits/ShpfyRefundProcessEvents.Codeunit.al
+++ b/src/Apps/W1/Shopify/App/src/Order Return Refund Processing/Codeunits/ShpfyRefundProcessEvents.Codeunit.al
@@ -30,7 +30,7 @@ codeunit 30247 "Shpfy Refund Process Events"
     begin
     end;
 
-    [InternalEvent(false)]
+    [IntegrationEvent(false, false)]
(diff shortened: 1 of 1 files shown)


src/Apps/W1/Shopify/App/src/Order Return Refund Processing/Codeunits/ShpfyRefundProcessEvents.Codeunit.al
```diff
@@ -30,7 +30,7 @@ codeunit 30247 "Shpfy Refund Process Events"
     begin
     end;
```
AndreiPanko, line 32: Don't we need something like that: 
 
    /// <summary> 
    /// Raised Before Create Item Sales Line.
    /// </summary>
    /// <param name="RefundHeader">Parameter of type Record "Shpfy Refund Header".</param>
    /// <param name="RefundLine">Parameter of type Record "Shpfy RefundLine".</param>
    /// <param name="SalesHeader">Parameter of type Record "Sales Header".</param>
    /// <param name="SalesLine">Parameter of type Record "Sales Line".</param>
    /// <param name= NextLineNo>Parameter of type Integer.</param>
    /// <param name="Handled">Parameter of type Boolean.</param>

onbuyuka: we don't, it does not really add any value

JesperSchulz (approved):

PR #6082, opened 2026-01-08
[Agent] Additional changes to support Agent - Admin or Configure Agent Access Control without SUPER scenarios

qutreson:
#### Summary 
1. Only enable the Setup action from the Agent List and Agent Card if the user has rights to manage this agent (via permissions or via Agent Access Control)
2. Removing the logic to insert the current owner on agent creation if none is found because this is handled via the server during agent creation already. Also, this code was inserting the Agent Access Control for all companies while the user might have had agent creation permissions in one company.
3. Use a new server utility when inserting Access Control for an agent. This allows users with only Configure Agent Access Control and without SUPER, SECURITY or AGENT - ADMIN to do such changes without explicitly having Access Control modification permissions. They can only do this on the agent they manage (and cannot do this on other agents or other user types).

#### Work Item(s) 
Fixes [AB#617434](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/617434)


src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
```diff
@@ -54,47 +55,25 @@ codeunit 4301 "Agent Impl."
         ChangeAgentState(AgentUserSecurityID, false);
     end;
 
-    procedure InsertCurrentOwnerIfNoOwnersDefined(var Agent: Record Agent; var AgentAccessControl: Record "Agent Access Control")
```
qutreson, line 57: Delegating this to the server now.

src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
```diff
@@ -54,47 +55,25 @@ codeunit 4301 "Agent Impl."
...
         Agent: Record Agent;
     begin
         GetAgent(Agent, AgentUserSecurityID);
 
-        GetUserAccess(Agent, TempAgentAccessControl, false);
+        GetUserAccess(Agent, TempAgentAccessControl);
     end;
 
-    local procedure GetUserAccess(var Agent: Record Agent; var TempAgentAccessControl: Record "Agent Access Control" temporary; InsertCurrentUserAsOwner: Boolean)
+    local procedure GetUserAccess(var Agent: Record Agent; var TempAgentAccessControl: Record "Agent Access Control" temporary)
```
nikolakukrika, line 67: I liked the previous solution better as it is visible what is happening. We have comment now, but it is obscure a bit. 
I hope that the server will insert the user explicitly so it is visible that the user is in the list.

nikolakukrika (approved):

src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
qutreson: The server does the same (and was already doing it). Keeping it in app wasn't sustainable anyway since some users might not have the right permissions for this.

dchourasia30 (approved):

PR #6137, opened 2026-01-13
Fixes for various payables agent bugs

ventselartur:
Current pull requests contain changes for the following bugs:
- 'View Pdf" option does not work correctly. The fix contains two parts:
1. Different implementation of the HasPDFSource variable that is responsible for the visibility of the 'View pdf' action on the draft page
2. Adding the '.pdf' extension to the file name of the sample invoice which is previously resulted in the "not supported" file when calling this line of code:
`3. File.ViewFromStream(TempBlob.CreateInStream(), FileName, true)`

Fixes [AB#617937](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/617937/)

### DIFF
--- a/src/Apps/W1/EDocument/App/app.json
+++ b/src/Apps/W1/EDocument/App/app.json
@@ -37,11 +37,6 @@
       "id": "967eceac-106e-4102-b76a-fe5dc3d799e5",
       "name": "Payables Agent Tests",
       "publisher": "Microsoft"
-    },
-    {
-      "id": "fdeb586e-beff-49d8-947d-1e73ce980b35",
-      "name": "E-Document for Italy",
-      "publisher": "Microsoft"
     }
   ],
   "screenshots": [],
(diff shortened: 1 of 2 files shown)


src/Apps/W1/EDocument/App/app.json
```diff
@@ -37,6 +37,11 @@
       "id": "967eceac-106e-4102-b76a-fe5dc3d799e5",
       "name": "Payables Agent Tests",
       "publisher": "Microsoft"
+    },
+    {
```
Groenbech96: Why is internal needed?

src/Apps/W1/EDocument/App/src/Processing/Import/FinishDraft/EDocCreatePurchaseInvoice.Codeunit.al
```diff
@@ -265,4 +266,9 @@ codeunit 6117 "E-Doc. Create Purchase Invoice" implements IEDocumentFinishDraft,
             exit(PurchaseLine."Line No.");
     end;
 
+    [IntegrationEvent(false, false)]
+    local procedure OnBeforeModifyPurchHeaderWhenApplyDraftToBC(var PurchaseHeader: Record "Purchase Header"; EDocumentPurchaseHeader: Record "E-Document Purchase Header")
```
Groenbech96: OnBeforeModifyPurchHeader would be enough. Rest of context comes from Codeunit.

src/Apps/W1/EDocument/App/app.json
ventselartur: Because the codeunit i add the integration event to is internal. I cannot access it from the new E-Document IT app without adding it to InternalsVisibleTo.

chrishdmicrosoftcom (approved):

PR #6197, opened 2026-01-19
Agents - Port features and bugfixes

nikolakukrika:
#### Summary 
Port Agent fixes for 27.4

#### Work Item(s) 

[AB#597811](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/597811)
[AB#617715](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/617715)

### DIFF
--- a/src/System Application/App/Agent/Interaction/Internal/AgentTaskContextPart.Page.al
+++ b/src/System Application/App/Agent/Interaction/Internal/AgentTaskContextPart.Page.al
@@ -38,7 +38,6 @@ page 4343 "Agent Task Context Part"
                 Caption = 'Task ID';
                 ToolTip = 'Specifies the ID of the task that was executed.';
                 Editable = false;
-                ExtendedDatatype = Task;
 
                 trigger OnDrillDown()
                 var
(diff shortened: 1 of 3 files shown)


src/System Application/App/Agent/Interaction/Internal/AgentTaskContextPart.Page.al
```diff
@@ -0,0 +1,63 @@
...
+                begin
+                    TaskPane.ShowAgent(Rec."Agent User Security ID");
+                end;
+            }
+            field(TaskID; Rec.ID)
+            {
+                Caption = 'Task ID';
+                ToolTip = 'Specifies the ID of the task that was executed.';
+                Editable = false;
+                ExtendedDatatype = Task;
```
qutreson: This requires updating the system application runtime version to 16.1.

It should be fine as we are shipping this in a new minor which would be, I believe, all using a very recent platform. I don't recall exactly when this feature was shipped, but it is recent, so releasing such a change as an App HF could have a risk of being faster than the platform HF and then failing to deploy System App.

Hopefully, that's fine.

nikolakukrika: I have removed the extendeddatatatype from the page. We can add it as a separate PR.

PR #6241, opened 2026-01-21
Fix InitPath procedure to extract and prepend SharePoint site path to…

### LINKED ISSUE #3872: [Bug]: Can't Open or Download File from External File Storage (SharePoint Online)
### Describe the issue

The issue I’m facing is that, after successfully configuring the connection and confirming that it works, I consistently receive an error when trying to open any file (regardless of type) either from the BC user interface or via code when attempting to load it into memory using an InStream.

The configuration is as follows: under "External File Accounts", we’ve created the following entry:

Image

As you can see, it points to our SharePoint Online, specifically to a folder within the “FBRICA” library, which does exist and contains files:

Image

In Microsoft Entra ID, we created an application registration with FullControl permissions on the entire SharePoint site collection, along with a ClientSecret used to configure the connection from BC.

The connection seems to work, as we can browse the SharePoint storage from BC and see all the files:

Image

Image

However, when clicking on any file (we tested PDF, TXT, DOCX, XLSX, etc.), it fails with a non-descriptive error:

Image

Image

When attempting the same operation programmatically, the same error is triggered on the marked line:

Image

After reviewing the code, I believe I found the root of the problem, but I can’t fix it since it’s Microsoft-owned code.

In Codeunit 9455 “External File Storage Impl.”, the method DownloadFileContentByServerRelativeUrl calls the endpoint GetFileByServerRelativeUrl, which according to Microsoft’s official SharePoint Client documentation – and as noted in a comment in the Codeunit – should start with a forward slash ("/").

Image

However, the method SetMethod in the Codeunit "SharePoint Uri Builder" does not prepend the slash.

Image

Image

We also can’t pass the slash manually as a parameter, since doing so triggers a validation error in Codeunit 9455 "External File Storage Impl.", which is specifically designed to prevent this.

Image

In summary, we are currently unable to use Microsoft’s official “External File Storage” module in BC to read the contents of files stored in our SharePoint Online.

### Expected behavior

After select any file from the "Browse Storage" page using BC web client, the file must be dowloaded, also the method GetFile should load on the InStream variable the file content.

### Steps to reproduce

1. Create on a SharePoint Online a document Library and upload some files, no matter wich type.
2. Create on Microsoft Entra ID an App registration with SharePoint Full control premissions, and a client secret key
3. Assign to SharePoint Online full control permission to the App registration
4. Setup on BC the External File Storage to SharePoint Online using App registration and client secret key
5. Verify on BC you can browse the SharePoint document Library porpertly, creating a folder or even uploading new files.
6. On BC using browse storage page try to download any file to get the error
7. From VSCode create a report to load on an InStream variable the content of any file => the same error occurs.

### Additional context

_No response_

StefanSosic:
ΓÇª file paths

#### Summary 

The Problem:
The SharePoint connector was failing with a 400 Bad Request error when trying to download files because the paths being sent to SharePoint's REST API were incomplete.

Root Cause:
SharePoint REST API requires server-relative URLs that include the full site path. 

For example:
Wrong: /Shared Documents/TEST/Design.png
Correct: /sites/mysite/Shared Documents/TEST/Design.png

The original implementation only combined the base folder path with the file path, but didn't extract and prepend the site path from the SharePoint URL.

#### Work Item(s) 
Fixes #3872 

Fixes [AB#619102](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/619102)

### DIFF
--- a/src/Apps/W1/External File Storage - SharePoint Connector/App/src/ExtSharePointConnectorImpl.Codeunit.al
+++ b/src/Apps/W1/External File Storage - SharePoint Connector/App/src/ExtSharePointConnectorImpl.Codeunit.al
@@ -443,9 +443,38 @@ codeunit 4580 "Ext. SharePoint Connector Impl" implements "External File Storage
     local procedure InitPath(AccountId: Guid; var Path: Text)
     var
         SharePointAccount: Record "Ext. SharePoint Account";
+        SitePath: Text;
     begin
         SharePointAccount.Get(AccountId);
+
+        // Extract site path from SharePoint URL
+        SitePath := GetSitePathFromUrl(SharePointAccount."SharePoint Url");
+
+        // Combine base folder path with the file path
         Path := CombinePath(SharePointAccount."Base Relative Folder Path", Path);
+
+        // Ensure path starts with forward slash
+        if not Path.StartsWith('/') then
+            Path := '/' + Path;
+
+        // Prepend site path if it exists and path doesn't already include it
+        if (SitePath <> '') and (not Path.StartsWith(SitePath)) then
+            Path := SitePath + Path;
+    end;
+
+    local procedure GetSitePathFromUrl(SharePointUrl: Text): Text
+    var
+        Uri: Codeunit Uri;
+        PathSegment: Text;
+    begin
+        Uri.Init(SharePointUrl);
+        PathSegment := Uri.GetAbsolutePath();
+
+        // Remove trailing slash if present
+        if PathSegment.EndsWith('/') then
+            PathSegment := PathSegment.TrimEnd('/');
+
+        exit(PathSegment);
     end;
 
     local procedure CombinePath(Parent: Text; Child: Text): Text


src/Apps/W1/External File Storage - SharePoint Connector/App/src/ExtSharePointConnectorImpl.Codeunit.al
```diff
@@ -443,9 +443,38 @@ codeunit 4580 "Ext. SharePoint Connector Impl" implements "External File Storage
...
         SharePointAccount.Get(AccountId);
+
+        // Extract site path from SharePoint URL
+        SitePath := GetSitePathFromUrl(SharePointAccount."SharePoint Url");
+
+        // Combine base folder path with the file path
         Path := CombinePath(SharePointAccount."Base Relative Folder Path", Path);
+
+        // Ensure path starts with forward slash
+        if not Path.StartsWith('/') then
```
JesperSchulz, line 457: Ran this by an agent, who replied:
If SitePath is /sites/mysite and Path is /Shared Documents, the result would be /sites/mysite/Shared Documents - which is correct. However, if the Base Relative Folder Path is already configured with the site path prefix, you could get duplicate paths.

Recommendation: Consider validating that the resulting path doesn't contain the site path twice, or document clearly that Base Relative Folder Path should NOT include the site path.

haoranpb (approved):

src/Apps/W1/External File Storage - SharePoint Connector/App/src/ExtSharePointConnectorImpl.Codeunit.al
```diff
@@ -443,9 +443,38 @@ codeunit 4580 "Ext. SharePoint Connector Impl" implements "External File Storage
...
+            Path := '/' + Path;
+
+        // Prepend site path if it exists and path doesn't already include it
+        if (SitePath <> '') and (not Path.StartsWith(SitePath)) then
```
JesperSchulz, line 461: One more:
Missing edge case handling for root site URLs
If the SharePoint URL is a root site (e.g., https://contoso.sharepoint.com/) without /sites/xxx, then GetAbsolutePath() would return / or empty. The code handles empty SitePath, but returning just / and prepending it would create //Shared Documents.

Recommendation: Add handling for when SitePath is just /:
if (SitePath = '/') or (SitePath = '') then
    SitePath := '';

src/Apps/W1/External File Storage - SharePoint Connector/App/src/ExtSharePointConnectorImpl.Codeunit.al
StefanSosic: I did test it with that also
image

Handled that too

src/Apps/W1/External File Storage - SharePoint Connector/App/src/ExtSharePointConnectorImpl.Codeunit.al
StefanSosic: Not needed, had that code before:
image

Returns empty, it should:
image

Final path includes the full path to that file, then:
image

StefanSosic: With different path setup:
image


image


And one more:
image


image


And final:
image


image

src/Apps/W1/External File Storage - SharePoint Connector/App/src/ExtSharePointConnectorImpl.Codeunit.al
StefanSosic: Btw. "document clearly that Base Relative Folder Path should NOT include the site path."
It could, but it should not be doubled definitely. That wouldn't work also before this changes.

PR #6264, opened 2026-01-23
Agents - Add the attachments to the page

nikolakukrika:
#### Summary 
Adding attachments part to the generic review page

#### Work Item(s) 
Fixes [AB#615278](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/615278)

### DIFF
--- a/src/System Application/App/Agent/Interaction/Internal/AgentTaskMessageAttachments.Page.al
+++ b/src/System Application/App/Agent/Interaction/Internal/AgentTaskMessageAttachments.Page.al
@@ -83,15 +83,12 @@ page 4309 "Agent Task Message Attachments"
         AgentTaskFile.SetAutoCalcFields(Content);
 
         repeat
-            if not AgentTaskFile.Get(AgentTaskMessageAttachment."Task ID", AgentTaskMessageAttachment."File ID") then
-                exit;
-
-            if not Rec.Get(AgentTaskMessageAttachment."Task ID", AgentTaskMessageAttachment."File ID") then begin
-                Rec.Init();
-                Rec.TransferFields(AgentTaskFile, true);
-                Rec.Content := AgentTaskFile.Content;
-                Rec.Insert();
-            end;
+            if not Rec.Get(AgentTaskMessageAttachment."Task ID", AgentTaskMessageAttachment."File ID") then
+                if AgentTaskFile.Get(AgentTaskMessageAttachment."Task ID", AgentTaskMessageAttachment."File ID") then begin
+                    Rec.TransferFields(AgentTaskFile, true);
+                    Rec.Content := AgentTaskFile.Content;
+                    Rec.Insert();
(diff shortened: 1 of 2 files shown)


src/System Application/App/Agent/Interaction/Internal/AgentTaskMessageCard.Page.al
```diff
@@ -15,7 +15,6 @@ page 4308 "Agent Task Message Card"
     DeleteAllowed = false;
     Caption = 'Agent Task Message';
     DataCaptionExpression = '';
-    Extensible = false;
```
qutreson, line 18: What's the reason for this change?

src/System Application/App/Agent/Interaction/Internal/AgentTaskMessageAttachments.Page.al
```diff
@@ -0,0 +1,134 @@
...
+                    Caption = 'File Size';
+                    ToolTip = 'Specifies the size of the attachment';
+                }
+            }
+        }
+    }
+
+    trigger OnAfterGetRecord()
+    begin
+        AttachmentFileSize := FormatFileSize(Rec.Content.Length());
```
BazookaMusic, line 52: Don't you need to do CalcField(Rec.Content) for the length to be correct?

nikolakukrika: I do not believe, we are calculating and loading it on the LoadRecords

src/System Application/App/Agent/Interaction/Internal/AgentTaskMessageCard.Page.al
nikolakukrika: As talked offline, returning it back. I wanted to allow partners to create their own tooling, but it can work.

src/System Application/App/Agent/Interaction/Internal/AgentTaskMessageAttachments.Page.al
```diff
@@ -0,0 +1,134 @@
...
+        if not AgentTaskMessageAttachment.FindSet() then
+            exit;
+
+        AgentTaskFile.SetAutoCalcFields(Content);
+
+        repeat
+            if not AgentTaskFile.Get(AgentTaskMessageAttachment."Task ID", AgentTaskMessageAttachment."File ID") then
+                exit;
+
+            if not Rec.Get(AgentTaskMessageAttachment."Task ID", AgentTaskMessageAttachment."File ID") then begin
```
BazookaMusic: Maybe move this check first? No reason to read the AgentTaskFile if we are going to ignore it because it's already added. Although I don't see how the Rec would already have it, since they have the same primary keys. This check might be redundant

PR #6327, opened 2026-01-27
Issue 6272,manually changing the source table or intent does not chan…

### LINKED ISSUE #6272: [Vendor] [QM] Manually created inspection generation rules get Activation trigger value from previous line and do not respect the intent
### Describe the issue

 Manually created inspection generation rules get Activation trigger value from previous line and do not respect the intent.

Could you please investigate and fix it.

### Expected behavior

-

### Steps to reproduce

-

### Additional context

-

insightworks007:
…ge default triggers

#### Summary 
Request to change the behavior to resets the default action each time the intent changes on a generation rule record.

#### Work Item(s) 
Fixes #6272 
Fixes [AB#619131](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/619131)

### DIFF
--- a/src/Apps/W1/Quality Management/app/src/Configuration/GenerationRule/QltyInspectionGenRule.Table.al
+++ b/src/Apps/W1/Quality Management/app/src/Configuration/GenerationRule/QltyInspectionGenRule.Table.al
@@ -421,7 +421,14 @@ table 20404 "Qlty. Inspection Gen. Rule"
         if Certainty = Certainty::Yes then begin
             Rec.Intent := InferredIntent;
             SetDefaultTriggerValuesToNoTrigger();
-            if Rec."Activation Trigger" in [Rec."Activation Trigger"::"Manual or Automatic", Rec."Activation Trigger"::"Automatic only"] then
+            if Rec."Activation Trigger" in [Rec."Activation Trigger"::"Manual or Automatic", Rec."Activation Trigger"::"Automatic only"] then begin
+                Rec."Assembly Trigger" := Rec."Assembly Trigger"::NoTrigger;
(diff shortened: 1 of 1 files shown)


src/Apps/W1/Quality Management/app/src/Configuration/GenerationRule/QltyInspectionGenRule.Table.al
```diff
@@ -421,7 +421,14 @@ table 20404 "Qlty. Inspection Gen. Rule"
         if Certainty = Certainty::Yes then begin
             Rec.Intent := InferredIntent;
             SetDefaultTriggerValuesToNoTrigger();
-            if Rec."Activation Trigger" in [Rec."Activation Trigger"::"Manual or Automatic", Rec."Activation Trigger"::"Automatic only"] then
+            if Rec."Activation Trigger" in [Rec."Activation Trigger"::"Manual or Automatic", Rec."Activation Trigger"::"Automatic only"] then begin
+                Rec."Assembly Trigger" := Rec."Assembly Trigger"::NoTrigger;
```
JakovljevicDusan, line 425: Just to check whether it would be better to move this cleanup code outside the conditional logic, to ensure it always executes.

insightworks007: We don't want this reset to always execute, we really just want it to execute in this circumstance.
This particular code change is to improve usability for an edge case on what is an atypical flow in practice (at least based on current installs).

PR #6377, opened 2026-01-30
[AI SDK] Surface billing setup errors and handle default AOAI policy for self managed resources

qutreson:
#### Summary 
Issue 1:
We don't surface billing setup issues because an Error is raised within the try function and the operation result is never sent on the response codeunit passed by var. So the error message is never surfaced. Changing the SendRequest to return the error first.

image

Issue 2:
AOAI policies would default to ConveservativeWithXpia which generally does not exist for self managed resources since that's something dedicated in our copilot service. Changing the SendChatCompletionRequest logic to not pass any XPIA policy for self managed resources if the policy are still the default one.

#### Work Item(s) 
Fixes [AB#620032](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/620032), [AB#615189](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/615189)

### DIFF
--- a/src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
+++ b/src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
@@ -266,6 +266,7 @@ codeunit 7772 "Azure OpenAI Impl" implements "AI Service Name"
     [NonDebuggable]
     procedure GenerateChatCompletion(var ChatMessages: Codeunit "AOAI Chat Messages"; AOAIChatCompletionParams: Codeunit "AOAI Chat Completion Params"; var AOAIOperationResponse: Codeunit "AOAI Operation Response"; CallerModuleInfo: ModuleInfo)
     var
+        AOAIPolicyParams: Codeunit "AOAI Policy Params";
         CustomDimensions: Dictionary of [Text, Text];
         Payload, ToolChoicePayload : JsonObject;
         ToolsPayload: JsonArray;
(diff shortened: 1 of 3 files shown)


src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
```diff
@@ -283,7 +284,10 @@ codeunit 7772 "Azure OpenAI Impl" implements "AI Service Name"
 
         AOAIChatCompletionParams.AddChatCompletionsParametersToPayload(Payload);
 
-        AzureOpenAIPolicy := AOAIChatCompletionParams.GetAOAIPolicyParams().GetAOAIPolicy();
+        AOAIPolicyParams := AOAIChatCompletionParams.GetAOAIPolicyParams();
+        AzureOpenAIPolicy := (ChatCompletionsAOAIAuthorization.GetResourceUtilization() = Enum::"AOAI Resource Utilization"::"Self-Managed") and AOAIPolicyParams.IsDefaultPolicy()
+            ? '' // The default value is generally not compatible with self managed resources, yet we allow them to specify their own policy and must honor that.
```
msft-sam, line 289: Would be good to confirm that a blank policy (by default) will still apply some level of policy control on the prompt, or at the very least, the safety metaprompt is still sufficient in such cases (FYI @darjoo )

msft-sam (approved):

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
qutreson: As discussed, the changes done previously by Dishant aimed at exposing the AOAI policies for managed resources mainly. We also added the ability to specify custom policies for self managed resources, but that's dependent on the partner's deployment and the configured policies. This change just avoids setting the managed resource default policy onto a self-manage resource. If the partner sets a policy with the SDK, we include it in the request, if they don't set one we don't include it (which is the same as before).  No changes in that regard.

src/System Application/App/AI/src/Azure OpenAI/AzureOpenAIImpl.Codeunit.al
```diff
@@ -283,7 +284,10 @@ codeunit 7772 "Azure OpenAI Impl" implements "AI Service Name"
 
         AOAIChatCompletionParams.AddChatCompletionsParametersToPayload(Payload);
 
-        AzureOpenAIPolicy := AOAIChatCompletionParams.GetAOAIPolicyParams().GetAOAIPolicy();
+        AOAIPolicyParams := AOAIChatCompletionParams.GetAOAIPolicyParams();
+        AzureOpenAIPolicy := (ChatCompletionsAOAIAuthorization.GetResourceUtilization() = Enum::"AOAI Resource Utilization"::"Self-Managed") and AOAIPolicyParams.IsDefaultPolicy()
```
darjoo, line 288: I'd refactor this. I understand we added inline if/else but it also makes it harder to read and isn't a pattern for us.

 policy = AOAIPolicyParams.GetAOAIPolicy()
if default + self-managed then
   policy = ''

qutreson: Let's move toward the new syntax :)

darjoo (approved):

PR #6378, opened 2026-01-30
Only same unit of measure for e-docs PO matching

mynjj:
#### Summary 
If the received invoice had a unit of measure specified, then it will only be possible to match it to other order lines with the same uom. However, if no unit of measure has been assigned to the "draft" invoice, then it will be possible to match against any line, when matching, the uom will be set in the invoice line.

There's also a fix that was preventing some changes to be done when opening the matched PO from the action.

#### Work Item(s) 
Fixes [AB#619582](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/619582) [AB#619591](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/619591)


Groenbech96 (approved):
Very nice

PR #6383, opened 2026-01-30
Switches to notifications when tests are created automatically, a page should be shown now only when a test is created manually.

PredragMaricic:
#### Summary 
Switches to notifications when tests are created automatically, a page should be shown now only when a test is created manually.

#### Work Item(s) 
Fixes [AB#610799](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/610799)


attilatoury:
Why not show a list page with all the created inspections? We have this pattern elsewhere.

JakovljevicDusan (approved):

PR #6460, opened 2026-02-04
[EDocuments] Move to GA in all countries

Jose-agg:
#### Summary 
We want to progress the Payables Agent to GA in all countries. A requirement to achieve this is that the EDocument Setup Table is also enabled for all countries, and therefore in this PR I'm revoming all country filtering.

Fixes [AB#620504](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/620504)

### DIFF
--- a/src/Apps/W1/EDocument/App/src/Document/EDocumentsSetup.Table.al
+++ b/src/Apps/W1/EDocument/App/src/Document/EDocumentsSetup.Table.al
@@ -1,4 +1,5 @@
 #if not CLEANSCHEMA31
+#pragma warning disable AS0105
 // ------------------------------------------------------------------------------------------------
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
@@ -110,4 +111,5 @@ table 6107 "E-Documents Setup"
     end;
 
 }
+#pragma warning restore AS0105
 #endif


src/Apps/W1/EDocument/App/src/Document/EDocumentsSetup.Table.al
Groenbech96, line 1: Should we not remove the field 2 ?

mynjj: I agree, let's obsolete this table, it was only introduced for this

Jose-agg: Since we are removing the country-based flag, I think we should also remove the table itself, since it won't be needed anymore when the experience is enabled globally

src/Apps/W1/EDocument/App/src/Document/EDocumentsSetup.Table.al
```diff
@@ -1,18 +1,26 @@
+
```
darjoo:
```suggestion
```

src/Apps/W1/EDocument/App/src/Document/EDocumentsSetup.Table.al
```diff
@@ -1,18 +1,26 @@
...
     InherentPermissions = RX;
     ReplicateData = false;
+#if not CLEAN28
+    ObsoleteReason = 'Obsolete table, only used for storing the new experience flag.';
+    ObsoleteState = Pending;
+    ObsoleteTag = '28.0';
+#else
+    ObsoleteReason = 'This table is obsolete and should not be used.';
+    ObsoleteState = Removed;
+    ObsoleteTag = '30.0';
```
darjoo: This should be after 3 versions, so `31.0`

Jose-agg: Done

src/Apps/W1/EDocument/App/src/Document/EDocumentsSetup.Table.al
```diff
@@ -1,18 +1,26 @@
...
 // ------------------------------------------------------------------------------------------------
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See License.txt in the project root for license information.
 // ------------------------------------------------------------------------------------------------
 namespace Microsoft.eServices.EDocument;
 
-using System.Azure.Identity;
-using System.Environment;
-
 table 6107 "E-Documents Setup"
```
Groenbech96, line 12: Since this table is internal. You can just remove all references. Instead of using CLEAN28, just remove it. 
Check with Nikola, if we need normal obsoletion for internal tables. The pending -> removed, or if we can just do removed directly. So removed in 28.0

Jose-agg: We can delete the table directly, we need to obsolete it

ventselartur (approved):

PR #6501, opened 2026-02-05
Logos

darjoo:
#### Summary 
Update logos

#### Work Item(s) 
Fixes [AB#620762](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/620762)


Aleyenda:
Hey Darrick, can we also update the older NAV logos to the newer BC one, as discussed in the thread with Christian L?

> However, unless there is a relevant reason to use the old NAV logo for some extensions I would recommend they either got the generic extension icon or the new extension version of the new BC icon (both appear at the bottom row in your screenshot).

image

PR #6533, opened 2026-02-09
[AI] Filter out uninstalled apps on Copilot & capabilities page

darjoo:
#### Summary 
When copilot capability apps are uninstalled, the records remain on that page if you have not deregistered them.
This change hides those away as you may reinstall them at a later date.

#### Work Item(s) 
Fixes [AB#620260](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/620260)

### DIFF
--- a/src/System Application/Test Library/AI/src/CopilotSettingsTestLibrary.Codeunit.al
+++ b/src/System Application/Test Library/AI/src/CopilotSettingsTestLibrary.Codeunit.al
@@ -15,6 +15,7 @@ codeunit 132934 "Copilot Settings Test Library"
 
     procedure FindFirst(): Boolean
     begin
+        CopilotSettings.SetAutoCalcFields("App Publisher");
         exit(CopilotSettings.FindFirst());
     end;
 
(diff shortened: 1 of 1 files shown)


qutreson (approved):

Groenbech96 (approved):

src/System Application/App/AI/src/Copilot/CopilotSettings.Table.al
```diff
@@ -53,6 +54,11 @@ table 7775 "Copilot Settings"
             DataClassification = SystemMetadata;
             ValuesAllowed = "Not Billed", "Microsoft Billed", "Custom Billed";
         }
+        field(9; "App Installed"; Boolean)
+        {
+            FieldClass = FlowField;
+            CalcFormula = exist("NAV App Installed App" where("App ID" = field("App Id"), Publisher = field(Publisher)));
```
msft-sam: Could we end up in any weird state if the publisher name changes, but the app remains installed? Since this is an internal table, maybe we can make our publisher a flowfield too?

qutreson: Agree, I've also raised that one a couple of months ago.

PR #6546, opened 2026-02-09
[Shopify] Trial stores with different admin URL cannot connect

onbuyuka:
### Problem
When users create Shopify trial stores, Shopify provisions the store with a random internal URL (e.g., `fwlifjsl-rr.myshopify.com`) but displays a different "friendly" URL to the user (e.g., `shoping.myshopify.com`). 

During the OAuth flow:
1. User enters the friendly URL they see in Shopify admin
2. OAuth completes successfully, but Shopify returns the **real internal `myshopify_domain`** in the callback
3. The access token gets saved under the internal URL
4. Business Central tries to use the token with the original user-entered URL → **fails with "No access token" error**

### Solution
When the store URL returned from Shopify during OAuth differs from what the user entered, prompt the user to correct the store URL. If confirmed, the Shop record is updated to use the actual `myshopify_domain` that Shopify uses internally.

### Changes

**ShpfyAuthenticationMgt.Codeunit.al**
- Modified `InstallShopifyApp` to accept a `var Shop` parameter
- Added store URL mismatch detection after OAuth callback
- When mismatch detected, prompts user to correct the URL with confirmation dialog
- Fixed typo: `InstalllToStore` → `InstallToStore`

**ShpfyShop.Table.al**
- Added `SetStoreName(Store: Text)` procedure to update the Shopify URL with validation
- Updated `RequestAccessToken()` to pass the Shop record to `InstallShopifyApp`

Fixes [AB#621115](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/621115)


src/Apps/W1/Shopify/App/src/Integration/Codeunits/ShpfyAuthenticationMgt.Codeunit.al
```diff
@@ -79,15 +79,23 @@ codeunit 30199 "Shpfy Authentication Mgt."
         InstallURLTxt: Label 'https://%1/admin/oauth/authorize?scope=%2&redirect_uri=%3&state=%4&grant_options[]=%5', Comment = '%1 = Store, %2 = Scope, %3 = RedirectUrl, %4 = State, %5 = GrantOptions', Locked = true;
         InstallURLWithClientIdParamTok: Label '%1&client_id=%2', Comment = '%1 = InstallURLTxt, %2 = ClientId', Locked = true;
         NotMatchingStateErr: Label 'The state parameter value does not match.';
+        StoreMismatchLbl: Label 'The store URL returned from Shopify does not match the store. Do you want to correct the store URL?';
```
onbuyuka:
```suggestion
        StoreMismatchLbl: Label 'The store URL returned from Shopify does not match the shop. Do you want to correct the shop URL?';
```

PR #6586, opened 2026-02-10
E-Document Connector - Continia: Skip participation check when regist…

paracontinia:
#### Summary 
Added code to skip the participation check for external access points, when new Peppol participation has outbound-only profiles. In the Peppol network, participation in registry should be registered only if the client wants to receive e-documents.

#### Work Item(s) 
Fixes
[AB#621440](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/621440)


paracontinia:
@microsoft-github-policy-service agree company="Continia Software A/S"

PR #6649, opened 2026-02-12
AIAO support for content array.

Groenbech96:
#### Summary 

PR introduces a User Message Builder, that constructs an content array. Currently platform only support type text and type file. 
This is not to replace existing methods. Examples here https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_git/Platform-Core/pullrequest/239374

#### Work Item(s) 
Fixes [AB#618414](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/618414)

### DIFF
--- a/src/System Application/Test/AI/app.json
+++ b/src/System Application/Test/AI/app.json
@@ -46,6 +46,12 @@
       "name": "Environment Information Test Library",
       "publisher": "Microsoft",
       "version": "28.0.0.0"
+    },
+    {
+      "id": "0846d207-5dec-4c1b-afd8-6a25e1e14b9d",
+      "name": "Base64 Convert",
+      "publisher": "Microsoft",
+      "version": "28.0.0.0"
     }
   ],
   "screenshots": [],


src/System Application/App/AI/src/Azure OpenAI/Chat Completion/AOAIUserMessage.Codeunit.al
```diff
@@ -0,0 +1,46 @@
...
+    procedure AddTextPart(TextContent: Text)
+    begin
+        AOAIUserMessageImpl.AddTextPart(TextContent);
+    end;
+
+    /// <summary>
+    /// Adds a file content part to the user message.
+    /// </summary>
+    /// <param name="FileData">The file data to add (e.g. base64-encoded content).</param>
+    procedure AddFilePart(FileData: Text)
```
darjoo, line 33: We're not supporting `image_url` and `input_audio` am I correct?

Groenbech96: I am going to test this, since i might need to guard it for a specific model. 
But its currently only file that is supported.

Groenbech96: Waiting for platform uptake in NAV.

PR #6686, opened 2026-02-16
[MCP] Allow admin to define the default configuration

onbuyuka:
### Summary

Adds the ability for administrators to designate an MCP configuration as the **default**. When no configuration is specified by an MCP client connection, the designated default configuration is used instead of the built-in system default.

### Changes

**Public API** (MCPConfig.Codeunit.al)
- `SetAsDefaultConfiguration(ConfigId: Guid)` — designates a configuration as default (must be active)
- `ClearDefaultConfiguration()` — clears the current default and restores the system default (`Name = ''`)

**Implementation** (MCPConfigImplementation.Codeunit.al)
- `SetAsDefaultConfiguration` — clears all existing defaults via `ModifyAll`, sets the new one, emits telemetry
- `ClearDefaultConfiguration` — clears non-system defaults, re-marks the system default
- `IsDesignatedDefaultConfiguration` — reads the native `Default` Boolean field
- `MarkSystemDefaultAsDefault` — local helper to restore system default
- `CreateDefaultConfiguration` — sets `Default := true` on initial system default creation
- `CopyConfiguration` — ensures copies are created with `Default := false`
- `DeleteConfiguration` — if deleting the designated default, restores system default first
- `ActivateConfiguration` — prevents deactivation of the designated default
- `GetDimensions` — includes `IsDesignatedDefault` in telemetry

**Pages**
- MCPConfigCard.Page.al — `Default` field (read-only), `Set as Default` / `Clear Default` promoted actions, deactivation guard on `Active` field `OnValidate`
- MCPConfigList.Page.al — `Default` column, `Set as Default` / `Clear Default` actions in Advanced group

**Upgrade** (MCPUpgrade.Codeunit.al)
- `UpgradeMCPSystemDefaultAsDefault` — marks existing system default record (`Name = ''`) as `Default := true` on upgrade (tag: `MS-612454-MCPSystemDefaultAsDefault-20260216`)

**Tests** (MCPConfigTest.Codeunit.al)
- `TestSetAsDefaultConfiguration` — setting default clears system default
- `TestClearDefaultConfiguration` — clearing restores system default
- `TestOnlyOneDefaultConfiguration` — only one configuration can be default at a time
- `TestCopyConfigurationDoesNotCopyDefault` — copies don't inherit default flag
- `TestCannotSetInactiveConfigurationAsDefault` — inactive configs cannot be default
- `TestCannotDeactivateDesignatedDefault` — designated default cannot be deactivated
- `TestDeleteDesignatedDefaultRestoresSystemDefault` — deleting default restores system default

### Design decisions

- The `Default` field is a native Boolean on the platform `MCP Configuration` table — no table extension needed
- System default (`Name = ''`) starts with `Default = true`; any config can be designated as default
- At most one configuration has `Default = true` at any time
- Clearing the default is generic (parameterless) — always reverts to the system default
- Deactivation of the designated default is blocked on both the codeunit and card page levels

Fixes [AB#612454](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/612454)

### DIFF
--- a/src/System Application/App/MCP/src/Configuration/Codeunits/MCPConfigImplementation.Codeunit.al
+++ b/src/System Application/App/MCP/src/Configuration/Codeunits/MCPConfigImplementation.Codeunit.al
@@ -10,6 +10,7 @@ using System.Environment;
 #if not CLEAN28
 using System.Environment.Configuration;
 #endif
+using System.Feedback;
 using System.Reflection;
 using System.Utilities;
 
@@ -61,6 +62,10 @@ codeunit 8351 "MCP Config Implementation"
         InvalidJsonErr: Label 'The selected file is not a valid configuration file.';
         ConfigNameExistsMsg: Label 'A configuration with the name ''%1'' already exists. Please provide a different name.', Comment = '%1 = configuration name';
         ConfigurationNotFoundErr: Label 'The MCP configuration was not found.';
+        MCPServerFeedbackConfirmQst: Label 'We noticed you no longer have any active configurations. Could you share what made you decide to stop using the MCP server? Your feedback helps us improve the experience.';
+        MCPServerFeedbackQst: Label 'What could we do to improve the MCP server experience?';
+        NoActiveConfigsFeedbackTxt: Label 'No active configs feedback triggered', Locked = true;
+        GeneralFeedbackTxt: Label 'General MCP feedback triggered', Locked = true;
 
     #region Configurations
     internal procedure GetConfigurationIdByName(Name: Text[100]): Guid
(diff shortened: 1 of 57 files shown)


src/System Application/App/MCP/src/Configuration/Codeunits/MCPConfigImplementation.Codeunit.al
```diff
@@ -266,6 +277,57 @@ codeunit 8351 "MCP Config Implementation"
...
+        exit(MCPConfiguration.Default);
+    end;
+
+    internal procedure SetAsDefaultConfiguration(ConfigId: Guid)
+    var
+        MCPConfiguration: Record "MCP Configuration";
+        PreviousDefault: Record "MCP Configuration";
+    begin
+        if not MCPConfiguration.GetBySystemId(ConfigId) then
+            exit;
```
darjoo: Should it just exit if it can't find? We throw an error if it's not active, consider doing the same that it doesn't exist.

src/System Application/App/MCP/src/Configuration/Codeunits/MCPConfigImplementation.Codeunit.al
```diff
@@ -138,6 +144,9 @@ codeunit 8351 "MCP Config Implementation"
         if IsDefaultConfiguration(MCPConfiguration) then
             Error(DefaultConfigCannotBeDeletedErr);
 
+        if IsDesignatedDefaultConfiguration(MCPConfiguration) then
+            MarkSystemDefaultAsDefault();
```
darjoo, line 154: What is suppose to happen if it can't get the default and set it? We still just delete the associated MCPConfig?

src/System Application/App/MCP/src/Configuration/Codeunits/MCPConfigImplementation.Codeunit.al
```diff
@@ -266,6 +277,57 @@ codeunit 8351 "MCP Config Implementation"
...
+    internal procedure ClearDefaultConfiguration()
+    var
+        MCPConfiguration: Record "MCP Configuration";
+        SystemDefault: Record "MCP Configuration";
+    begin
+        MCPConfiguration.SetRange(Default, true);
+        MCPConfiguration.SetFilter(Name, '<>%1', '');
+        MCPConfiguration.ModifyAll(Default, false);
+
+        if SystemDefault.Get('') then begin
```
darjoo, line 320: How do these work? There's suppose to be a MCP configuration that has empty name which is the primary key?

onbuyuka: yes, that's the system default. that was how we did it in the previous releases so have to keep it that way for backwards compatibility.

src/System Application/App/MCP/src/Configuration/Codeunits/MCPConfigImplementation.Codeunit.al
onbuyuka: that's very unlikely (system default is hidden from UI, also cannot be deleted from code, config packages don't work too cause it is a system table) but yes, that's how it is supposed to work. this is a fallback in case you are deleting the designated default, not a necessity.

src/System Application/App/MCP/src/Configuration/Codeunits/MCPConfigImplementation.Codeunit.al
onbuyuka: fixed, and other occurrences too

PR #6687, opened 2026-02-16
#5879

### LINKED ISSUE #5879: [Bug]: Billing lines cannot be deleted after underlying subscription line is removed
### Describe the issue

When the subscription line is deleted, the related billing lines are marked with the update required flag. After that, the billing line can no longer be deleted because an error is thrown stating that the subscription line cannot be found.
This behavior blocks cleanup of billing data when subscription lines are removed intentionally.

### Expected behavior

Billing lines should be deletable even if the underlying subscription line no longer exists.

### Steps to reproduce

Create a new Customer Subscription contract for a customer.
Run the Billing Proposal from the Recurring Billing page for this contract.
Delete the underlying subscription line on the subscription.
Navigate back to the Recurring Billing page.
Try to delete the related billing line.

### Additional context

Screenshots shows the attempt to billing line where the underlying subscription line was deleted beforehand.

Image

dsarfraz:
#### Summary 
GetServiceCommitment during deletion of billing lines has been updated to use a safe if Get(...) pattern to prevent runtime errors if the related record does not exist.
No functional changes to existing business logic.

#### Work Item(s) 
Fixes #5879

Fixes [AB#624440](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/624440)


src/Apps/W1/Subscription Billing/App/Billing/Tables/BillingLine.Table.al
```diff
@@ -502,7 +503,7 @@ table 8061 "Billing Line"
 
     internal procedure GetServiceCommitment(var ServiceCommitment: Record "Subscription Line")
```
JesperSchulz:
```suggestion
    internal procedure GetServiceCommitment(var ServiceCommitment: Record "Subscription Line"): Boolean
```

PR #6781, opened 2026-02-23
Bug6451 email without top domain

### LINKED ISSUE #6451: [Bug]: Email Address Without Top Domain is Validated
### Describe the issue

The email address test@contoso is validated as a correct address. This makes it possible to enter bad email addresses which isn't noticed until trying to send.

Image

### Expected behavior

The email address should be stopped at validation, not when sending the email.

### Steps to reproduce

1. In Email Accounts press Send Test Email.
2. Select Other...
3. Enter test@contoso
4. The error ends up in the Outbox instead of a validation error

Alternate way to reproduce in Cronus
1. Go to customer 10000
2. Remove .com after contoso in field Email
3. Don't get a validation error

### Additional context

It is possible to fix in codeunit 8889 "Email Account Impl." in function ValidateEmailAddress().

But maybe it should be fixed in DotNet MailAddress.MailAddress() instead?

Dan-Karlsson:
#### Summary 
Validate email addresses to ensure they contain a top domain and that the top domain is at least 2 characters long

#### Work Item(s) 
Fixes #6451 

Fixes [AB#633135](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/633135)


src/System Application/App/Email/src/Account/EmailAccountImpl.Codeunit.al
```diff
@@ -294,6 +294,14 @@ codeunit 8889 "Email Account Impl."
             ValidatedEmailAddress := EmailAddress;
             if not ConvertEmailAddress(ValidatedEmailAddress) then
                 Error(InvalidEmailAddressErr, EmailAddress);
+
+            // Check that there is a top domain
+            if StrPos(CopyStr(ValidatedEmailAddress, StrPos(ValidatedEmailAddress, '@') + 2), '.') = 0 then
+                Error(InvalidEmailAddressErr, EmailAddress);
```
duichwer, line 300: Should we maybe provide more specific errors for those checks?

duichwer (changes requested):
Could you add some tests with error assertions?

Dan-Karlsson:
> Could you add some tests with error assertions?

I can't find any tests of those functions, so I'm not sure where to put them. And I'm also not sure how to write them to not risk failure with .Net-calls in different environments and similar.

src/System Application/App/Email/src/Account/EmailAccountImpl.Codeunit.al
Dan-Karlsson: I think it would be more confusing than helpful if 95% of the tests give the same error, and the last 5% is more precise. That might lead to demands of more specific error texts in the other cases also, and I don't think that is possible to solve.

In older versions where the tests were made in code instead of a .Net-call, they all returned the same error text also.

duichwer:
> > Could you add some tests with error assertions?
> 
> I can't find any tests of those functions, so I'm not sure where to put them. And I'm also not sure how to write them to not risk failure with .Net-calls in different environments and similar.

@Dan-Karlsson I think you're correct.
I couldn't find any tests for the mail validation in the system application tests either.
That doesn't mean that we shouldn't have there some tests.

There are some tests in the Base Application for the procedure `CheckValidEmailAddress` of codeunit `"Mail Management"`
https://github.com/microsoft/BusinessCentralApps/blob/main/App/Layers/W1/Tests/Integration-Internal/MailManagementTest.Codeunit.al
The procedure `CheckValidEmailAddress` internall calls the procedure of the email account codeunit 

```al
    [TryFunction]
    procedure CheckValidEmailAddress(EmailAddress: Text)
    var
        EmailAccount: Codeunit "Email Account";
        IsHandled: Boolean;
    begin
        IsHandled := false;
        OnBeforeCheckValidEmailAddr(EmailAddress, IsHandled);
        if IsHandled then
            exit;

        EmailAddress := DelChr(EmailAddress, '<>');

        // Check that only one address is validated.
        if EmailAddress.Split('@').Count() <> 2 then
            Error(InvalidEmailAddressErr, EmailAddress);

        if not EmailAccount.ValidateEmailAddress(EmailAddress) then
            Error(InvalidEmailAddressErr, EmailAddress);
    end;
```

Therefore I would suggest to some tests to https://github.com/microsoft/BCApps/blob/main/src/System%20Application/Test/Email/src/EmailAccountsTest.Codeunit.al.

src/System Application/App/Email/src/Account/EmailAccountImpl.Codeunit.al
duichwer: I will leave this up to someone of microsoft.

duichwer (changes requested):

JesperSchulz:
@Dan-Karlsson, sorry, this one fell off the radar. The DK issue seems to be a timeout. Let's re-run CI and get this one merged.

JesperSchulz (approved):

gggdttt (approved):

PR #6782, opened 2026-02-23
Table Info: Don't return to the saved position for non-AL defined indices

PhDuck:
Problem:
When an non-AL defined index is disabled (turned off) it doesn't exist anymore (neither in Database Index VT or Key VT), so it will no longer exist in the temp table, leading to an error.

Solution:
Don't attempt to return to the position for non-AL defined indices after disabling them. Instead the position will just be the first index.

This PR also adds another page which shows the index's fields and include fields.

Fixes [AB#622744](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/622744)

Example:
{E16B2371-6914-41B5-846B-15FEE42DE281}

### DIFF
--- a/src/System Application/App/Table Information/src/IndexDetails.page.al
+++ b/src/System Application/App/Table Information/src/IndexDetails.page.al
@@ -10,7 +10,7 @@ using System.Diagnostics;
 /// <summary>
 /// Displays additional details for a specific index.
 /// </summary>
-page 8705 "Index Details"
+page 8706 "Index Details"
 {
     Caption = 'Index Details';
 


src/System Application/App/Table Information/src/IndexesListPart.Page.al
```diff
@@ -144,7 +144,9 @@ page 8704 "Indexes List Part"
 
                     Rec.DeleteAll(); // Clear the temporary table to make sure the disabled index is not shown.
                     BuildInMemoryList(Rec.TableId); // Rebuild the in-memory list to get the updated index status.
-                    Rec.Get(RecordIDOfCurrentPosition); // Return to the same position in the list after refreshing the data.
+
+                    if Rec."Metadata Defined" then
+                        Rec.Get(RecordIDOfCurrentPosition); // Return to the same position, not possible for non-AL defined indexes as they are removed from the underlying table.
```
BardurKnudsen: Just for resilience, you could consider doing an "if Rec.Get( .... then"

src/System Application/App/Table Information/src/IndexesListPart.Page.al
```diff
@@ -189,6 +199,7 @@ page 8704 "Indexes List Part"
                 Image = Add;
                 ToolTip = 'Enqueues the index to be turned on in the subsequent maintenance window.';
 
+
```
ehenriksen9, line 202: Nit space

src/System Application/App/Table Information/src/IndexesListPart.Page.al
ehenriksen9, line 213: Why no space here?

src/System Application/App/Table Information/src/IndexDetails.page.al
```diff
@@ -0,0 +1,46 @@
...
+
+using System.Diagnostics;
+
+/// <summary>
+/// Displays additional details for a specific index.
+/// </summary>
+page 8706 "Index Details"
+{
+    Caption = 'Index Details';
+
```
ehenriksen9, line 16: space?

src/System Application/App/Table Information/src/IndexesListPart.Page.al
PhDuck: You mean the newline? That seems to be convention to add a newline between the properties and trigger def:

Image

src/System Application/App/Table Information/src/IndexesListPart.Page.al
PhDuck: Do you mean newline between the action definitions? It seems to be convention to not leave one:

Image

PredragMaricic (approved):

PR #6805, opened 2026-02-24
Validation  + UI update when trying to create Quality Inspection Results records with empty fields.

6CRIPT:
The Code and Description fields now display a red asterisk to indicate that they are mandatory. An error is shown if the window is closed or if a new record is inserted or an existing one modified with empty fields.

Fixes [AB#622623](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/622623)

### DIFF
--- a/src/Apps/W1/Quality Management/app/src/Configuration/Result/QltyInspectionResultList.Page.al
+++ b/src/Apps/W1/Quality Management/app/src/Configuration/Result/QltyInspectionResultList.Page.al
@@ -13,7 +13,6 @@ page 20416 "Qlty. Inspection Result List"
     SourceTable = "Qlty. Inspection Result";
     SourceTableView = sorting("Evaluation Sequence");
     PageType = List;
-    DelayedInsert = true;
     ApplicationArea = QualityManagement;
     UsageCategory = Lists;
     AboutTitle = 'About Inspection Results';


src/Apps/W1/Quality Management/app/src/Configuration/Result/QltyInspectionResult.Table.al
```diff
@@ -233,6 +233,8 @@ table 20411 "Qlty. Inspection Result"
 
         if Rec.Code = DefaultResultInProgressCodeLbl then
             Rec."Finish Allowed" := Rec."Finish Allowed"::"Do Not Allow Finish";
+
+        TestField("Description");
```
AndreiPanko: please don't 

we already have:
        field(2; "Description"; Text[100])
        {
            Caption = 'Description';
            NotBlank = true;
            ToolTip = 'Specifies a friendly description for the result.';
        }

AndreiPanko (changes requested):

src/Apps/W1/Quality Management/app/src/Configuration/Result/QltyInspectionResultList.Page.al
```diff
@@ -13,6 +13,7 @@ page 20416 "Qlty. Inspection Result List"
     SourceTable = "Qlty. Inspection Result";
     SourceTableView = sorting("Evaluation Sequence");
     PageType = List;
+    DelayedInsert = true;
```
AndreiPanko: is it needed? I would not change this setting.

6CRIPT: `DelayedInsert = true;` is used because if don't, the error coming from the `testField("Description")` will be thrown as soon as you click into the description field.

But as said, if you remove `testField("Description")`, which would make `DelayedInsert = true;` also not needed, then the application allows the user to create a rec without description.

image

Therefore `DelayedInsert = true;` is used here to avoid this error until the user actually try to save the changes; it will trigger then if the description field is empty.

6CRIPT: Finally deleted `DelayedInsert = true;` in latest commit as suggested.

PR #6849, opened 2026-02-25
Now the quality inspection card shows the status in bold + green if open

6CRIPT:
Fixes [AB#622626](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/622626)

Now if the Status is "Open" the Quality Inspection Card shows it in bold/green.
image
However, I was wondering if it makes sense also to propagate this change to the Quality Inspection List, like this:
image
The commit only includes the Card part (first image), therefore if you want, I can also push the second part to have the same format in the list.

Also Attila and I were wondering if we want the other status to be bold (but not green), not just default. Like for finished in "Sales Order"

### DIFF
--- a/src/Apps/W1/Quality Management/app/src/Document/QltyInspection.Page.al
+++ b/src/Apps/W1/Quality Management/app/src/Document/QltyInspection.Page.al
@@ -864,11 +864,7 @@ page 20406 "Qlty. Inspection"
         TempItemTrackingSetup: Record "Item Tracking Setup" temporary;
     begin
         IsOpen := Rec.Status = Rec.Status::Open;
-        
-        StatusStyleExpr :=
-            IsOpen ? 'Favorable' :
-            (Rec.Status = Rec.Status::Finished) ? 'Strong' :
-            'None';
+        StatusStyleExpr := Rec.GetStatusStyleExpression();
 
         CanReopen := not Rec.HasMoreRecentReinspection();
         CanFinish := Rec.Status <> Rec.Status::Finished;
--- a/src/Apps/W1/Quality Management/app/src/Document/QltyInspectionList.Page.al
+++ b/src/Apps/W1/Quality Management/app/src/Document/QltyInspectionList.Page.al
@@ -704,18 +704,6 @@ page 20408 "Qlty. Inspection List"
         CanReopen := RowActionsAreEnabled and not Rec.HasMoreRecentReinspection();
         CanFinish := RowActionsAreEnabled and (Rec.Status <> Rec.Status::Finished);
         CanCreateReinspection := RowActionsAreEnabled;
-        StatusStyleExpr :=
-            Rec.Status = Rec.Status::Open ? 'Favorable' :
-            (Rec.Status = Rec.Status::Finished) ? 'Strong' :
-            'None';
-    end;
-
-    trigger OnAfterGetRecord()
-    begin
-        StatusStyleExpr :=
-            Rec.Status = Rec.Status::Open ? 'Favorable' :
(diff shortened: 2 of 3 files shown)


src/Apps/W1/Quality Management/app/src/Document/QltyInspection.Page.al
```diff
@@ -102,6 +102,7 @@ page 20406 "Qlty. Inspection"
                     field(Status; Rec.Status)
                     {
                         Editable = false;
+                        StyleExpr = StatusStyleExpr;
```
BardurKnudsen, line 105: Fine.

Just FYI: There is also another pattern that can be handy:

                    Style = "Favorable";
                    StyleExpr = Rec.Status = Rec.Status::Open;

The result is the same, but you don't need to code anything

src/Apps/W1/Quality Management/app/src/Document/QltyInspection.Page.al
```diff
@@ -862,9 +864,15 @@ page 20406 "Qlty. Inspection"
         TempItemTrackingSetup: Record "Item Tracking Setup" temporary;
     begin
         IsOpen := Rec.Status = Rec.Status::Open;
+        
+        StatusStyleExpr :=
```
AndreiPanko: Maybe the Case operator is better as here you don't utilize the main benefit of ternary operator to be on one line.
Placing one ternary operator inside another one hurting readability.

JakovljevicDusan: I suggest case statement, although it would be pretty much the same. On the other hand, please move this code to procedure in table (or codeunit), since it is used on the list page as well.

src/Apps/W1/Quality Management/app/src/Document/QltyInspectionList.Page.al
```diff
@@ -695,13 +696,26 @@ page 20408 "Qlty. Inspection List"
...
         RowActionsAreEnabled: Boolean;
+        StatusStyleExpr: Text;
 
     trigger OnOpenPage()
     begin
         RowActionsAreEnabled := not IsNullGuid(Rec.SystemId);
         CanReopen := RowActionsAreEnabled and not Rec.HasMoreRecentReinspection();
         CanFinish := RowActionsAreEnabled and (Rec.Status <> Rec.Status::Finished);
         CanCreateReinspection := RowActionsAreEnabled;
+        StatusStyleExpr :=
```
JakovljevicDusan: Should be needed only in OnAfterGetCurrRecord. Remove from OnOpenPage and OnAfterGetRecord.
Hm, another question is existing code in OnOpenPage that probably does nothing (as it is reading Rec). But not issue in scope of this PR.

src/Apps/W1/Quality Management/app/src/Document/QltyInspection.Page.al
6CRIPT: Good point. But in this case I think it can not be implemented, because we are using 'Favorable' Style for Open records and 'Strong' for finished ones. So we need some code to assign the value.

src/Apps/W1/Quality Management/app/src/Document/QltyInspection.Page.al
6CRIPT: Done in last commit

src/Apps/W1/Quality Management/app/src/Document/QltyInspectionList.Page.al
6CRIPT: Done in last commit 👍

PR #7129, opened 2026-03-13
[Agent] AI Test Toolkit: addressing UX feedback and permission issues

qutreson:
#### Work Item(s) 
Fixes [AB#623338](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/623338)

### DIFF
--- a/src/Tools/AI Test Toolkit/src/Agent/AgentTestSuite.PageExt.al
+++ b/src/Tools/AI Test Toolkit/src/Agent/AgentTestSuite.PageExt.al
@@ -12,47 +12,66 @@ pageextension 149034 "Agent Test Suite" extends "AIT Test Suite"
 {
     layout
     {
+        modify(TestType)
+        {
+            trigger OnBeforeValidate()
+            begin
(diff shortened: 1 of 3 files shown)


src/Tools/AI Test Toolkit/src/Agent/AgentRunHistory.PageExt.al
```diff
@@ -65,6 +67,13 @@ pageextension 149032 "Agent Run History" extends "AIT Run History"
         }
     }
 
+    trigger OnOpenPage()
```
stkillen, line 70: Is this a security measure? Or just for display purposes?

qutreson: If you don't have permissions to see consumption data, then the method to compute the consumed credits will throw. We need to resolve whether the user has permissions so that we can skip computing it whenever the user is missing that permission.

BazookaMusic (approved):

PR #7255, opened 2026-03-19
E-Document Core Docs

Groenbech96:
## Summary

- Adds hierarchical developer documentation (CLAUDE.md + docs/) for the E-Document Core app
- 26 files across 15 directories covering architecture, data model, business logic, extensibility (24 interfaces), and patterns
- Focuses on design intent, relationships, and gotchas -- not mechanical field/procedure listings
- Every claim verified by reading actual AL source files during generation

## Documentation structure

| Level | Files |
|-------|-------|
| App root | CLAUDE.md, docs/data-model.md, docs/business-logic.md, docs/extensibility.md, docs/patterns.md |
| src/Document/ | docs/CLAUDE.md, docs/data-model.md |
| src/Service/ | docs/CLAUDE.md, docs/data-model.md |
| src/Integration/ | docs/CLAUDE.md, docs/extensibility.md |
| src/Processing/ | docs/CLAUDE.md, docs/business-logic.md |
| src/Processing/Import/ | docs/CLAUDE.md, docs/business-logic.md, docs/extensibility.md, docs/data-model.md |
| 9 other subfolders | docs/CLAUDE.md each |

[AB#626091](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/626091)

### DIFF
--- a/.github/AL-Go-Settings.json
+++ b/.github/AL-Go-Settings.json
@@ -6,7 +6,7 @@
   "runs-on": "windows-latest",
   "cacheImageName": "",
   "UsePsSession": false,
-  "artifact": "bcinsider/Sandbox/29.0.47702.0//latest",
+  "artifact": "bcinsider/Sandbox/29.0.47831.0//latest",
   "country": "base",
   "useProjectDependencies": true,
   "incrementalBuilds": {
--- a/src/Apps/W1/APIReportsFinance/App/CLAUDE.md
+++ b/src/Apps/W1/APIReportsFinance/App/CLAUDE.md
@@ -1,0 +1,62 @@
+# API Reports - Finance
+
+Read-only API layer that exposes raw financial data (GL entries, customer/vendor
+ledger entries, budgets, dimensions) through OData/REST endpoints. Unlike the
+standard BC API v2.0 financial endpoints (trialBalance, balanceSheet, etc.) which
+return pre-aggregated report data, this app exposes granular transactional records
+that consumers aggregate themselves -- a "data warehouse" approach for building
+custom financial reports.
+
+## Quick reference
+
+- **ID range**: 30300--30399
+- **Namespace**: Microsoft.API.FinancialManagement
+- **API route**: `api/microsoft/reportsFinance/beta/`
+- **API version**: beta (not GA -- endpoints may change)
+
+## How it works
+
+The app defines 9 API Pages and 6 API Queries, all read-only. Pages expose
+master/reference data (chart of accounts, customers, vendors, dimensions,
+budgets, accounting periods, business units, global settings). Queries expose
+transactional data (GL entries, GL budget entries, customer ledger entries,
+detailed customer ledger entries, vendor ledger entries, detailed vendor ledger
+entries).
+
+Every object follows the same template: `PageType = API` or `QueryType = API`,
+`DataAccessIntent = ReadOnly`, all insert/modify/delete disabled, `ODataKeyFields
+= SystemId`. The app defines no tables of its own -- it reads directly from base
+app tables (G/L Entry, Cust. Ledger Entry, etc.).
+
+The only procedural logic lives in two page triggers:
+
+- **Accounting Periods** (`APIFinanceAccPeriods.Page.al`): computes
+  `FiscalYearStartDate`, `FiscalYearEndDate`, and `EndingDate` on each record
+  fetch. The ending date is derived by peeking at the next period's start date
+  and subtracting one day.
+
+- **GL Accounts** (`APIFinanceGLAccount.Page.al`): reconstructs the chart of
+  accounts parent-child hierarchy at runtime using a `Dictionary<Integer,
+  Code[20]>` keyed by indentation level. This is order-dependent -- it relies on
+  records arriving sorted by account number with ascending indentation.
+
+## Things to know
+
+- All 15 data endpoints are completely read-only. There is zero write logic, zero
+  events, zero extensibility hooks, and zero codeunits.
+- The GL Account parent tracking via dictionary is fragile. It works because API
+  pages iterate records in sort order, but the pattern wouldn't survive
+  re-sorting or filtering that breaks indentation sequencing.
+- `globalSettings` is a virtual page -- it reads from both the Company record and
+  General Ledger Setup in `OnOpenPage`, exposing just company name, LCY code, and
+  additional reporting currency.
+- The app has several caption typos in the query objects: `entryType` is captioned
+  as `'Entry Number'` in both detailed ledger entry queries, and
+  `initialEntryGlobalDim2` is captioned as `'...Dimension 1'` instead of
+  `'...Dimension 2'`.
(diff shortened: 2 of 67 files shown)


src/Apps/W1/EDocument/App/docs/business-logic.md
```diff
@@ -0,0 +1,111 @@
...
+
+**Prepare draft** (`ImportEDocumentProcess.PrepareDraft()`): `IProcessStructuredData.PrepareDraft()` resolves BC entities from the parsed data -- finding the vendor, matching items, resolving units of measure, assigning GL accounts. This uses a chain of provider interfaces (`IVendorProvider`, `IItemProvider`, `IUnitOfMeasureProvider`, `IPurchaseLineProvider`, `IPurchaseOrderProvider`). The result is a draft with BC-specific field values filled in.
+
+**Finish draft** (`ImportEDocumentProcess.FinishDraft()`): `IEDocumentFinishDraft.ApplyDraftToBC()` creates the actual BC document (Purchase Invoice, Purchase Credit Memo, etc.) from the staging tables. It returns the RecordId of the created document, which is stored in `"Document Record ID"` on the E-Document.
+
+### Reversibility
+
+Each step can be undone via `UndoProcessingStep()`. Undoing "Finish draft" calls `IEDocumentFinishDraft.RevertDraftActions()` (which deletes the created BC document). Undoing "Prepare draft" clears the BC-resolved fields and resets the document type. Undoing "Structure received data" clears the structured data reference. This lets users go back to an earlier stage, correct data, and re-process.
```
mynjj: > (which deletes the created BC document)

no longer true, maybe we should remove tihs

src/Apps/W1/EDocument/App/docs/business-logic.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Actions framework
+
+Beyond send and receive, the framework supports arbitrary actions on documents via `EDocIntegrationManagement.InvokeAction()`. The `"Integration Action Type"` enum (extensible) dispatches to `IDocumentAction.InvokeAction()` implementations. Built-in action types are `"Sent Document Approval"` and `"Sent Document Cancellation"`, which delegate to `ISentDocumentActions.GetApprovalStatus()` and `GetCancellationStatus()` on the service integration. Connector apps can add custom action types.
+
+## Gotchas
+
+- **Commit before interface calls**: The framework commits before every interface call (format Create, service Send, receiver ReceiveDocuments, etc.) to support error trapping. This means if the interface fails, the E-Document record and logs are already persisted. The trade-off is that you cannot roll back the E-Document creation if the format export fails.
+- **Re-reading after interface calls**: After every interface call, the framework re-reads the E-Document and service records from the database (`EDocument.Get(EDocument."Entry No")`). This is because interface implementations may modify these records, and the framework needs the latest values.
```
mynjj: The original intention behind this pattern and not passing by `var` was to control the impact that extensions may have via the function signature (telling to consumers, here's the info you have to work with, but you are not allowed to modify it). But if we needed to re-get the record, because after, all code-paths can directly call `.modify` (which I guess some parts of our own code does, in particular I think it some from the status flow field iirc), then it defeats the original purpose.

What I'm trying to say is that maybe this pattern should be a thing only when having no `var` in the interface definition, so we shouldn't always do this and guide agents to always be this paranoic. So a small mention of only doing this in procs that don't have var records that can be potentially modified would be good.

src/Apps/W1/EDocument/App/docs/business-logic.md
```diff
@@ -0,0 +1,111 @@
...
+- **E-Document Status is derived**: The overall `E-Document Status` is derived from the service statuses. `EDocumentProcessing.ModifyEDocumentStatus()` computes it after every service status change using the `IEDocumentStatus` interface on the enum values.
+- **Import Processing Status is a FlowField**: On the E-Document table, `"Import Processing Status"` is a FlowField that reads from `"E-Document Service Status"`. You must call `CalcFields` before reading it.
+- **V1.0 and V2.0 coexistence**: The `"Import Process"` field on the service determines which path runs. V1.0 collapses everything into a single "Finish draft" step that calls the old `V1_ProcessEDocument` logic. The pipeline state machine still runs, but only the last step does anything for V1.0 documents.
```
mynjj, line 110: I think there's not much about v1 (from a quick skim), at least a section of the main files around that and basic flow would help

src/Apps/W1/EDocument/App/docs/patterns.md
```diff
@@ -0,0 +1,163 @@
+# Patterns
+
+This document covers the recurring code patterns used in E-Document Core and, just as importantly, the legacy patterns you should avoid. For the full processing flows, see [business-logic.md](business-logic.md). For extension contracts, see [extensibility.md](extensibility.md).
+
```
mynjj, line 4: no mention of the "context" objects via manual event subscribers, that one is not very intuitive so I think it's worth mentioning too

src/Apps/W1/EDocument/App/src/Integration/docs/extensibility.md
```diff
@@ -0,0 +1,121 @@
+# Extensibility
```
mynjj, line 1: I think this is a really nice doc to have as external doc, maybe something that patrners/e-docs extension devs can feed into their agents, worth thinking about how to package/surface them more generally (I'm imagining something like a plugin that can be installed to develop bc extensions that has the extensibility subset of our docs)

Anyway, no action here, just something I thought about

src/Apps/W1/EDocument/App/src/Processing/docs/business-logic.md
```diff
@@ -0,0 +1,77 @@
...
+
+- **Batch vs. immediate.** If the service has `"Use Batch Processing"` enabled, the document gets status `Created` and is not exported inline. A recurrent job queue entry (`"E-Doc. Recurrent Batch Send"`) picks up all pending-batch documents at the configured interval, groups them by document type, exports them as a single batch blob, and sends the batch.
+
+- **Field mapping.** Before calling the format interface's `Create()`, the framework applies field-level mappings defined in `"E-Doc. Mapping"`. Source document headers and lines are copied to temporary records with mapped field values, and a mapping log is written. This happens for both individual and batch export.
+
+- **Error isolation.** `EDocumentCreate.Codeunit.al` is a runner codeunit invoked with `Codeunit.Run()`. If the format interface throws, `GetLastErrorText()` is captured and logged against the E-Document without aborting the caller.
+
+## Order matching
+
```
mynjj, line 45: I tihnk here we need something to desambiguate between this order matching and the new one which we didn't reconcile. Something like a "common gotcha" so that po matching related things out of context don't get erroneously applied in the wrong path

mynjj: there's actually nothing about the new order matching, it seems

Groenbech96: Should be in the commit.

PR #7294, opened 2026-03-20
Add Documentation Maintenance as GitHub Action

Groenbech96:
## Summary

Adds a GitHub Action that runs Copilot CLI with the al-docs plugin on PRs to auto-generate documentation for changed AL apps.

## How it works

1. Triggers on PRs that modify `.al` files under `src/`
2. Finds changed apps by walking up to nearest `app.json`
3. Runs `copilot -p` with al-docs skill per app (read/write/glob/grep only)
4. If docs were generated, opens a draft PR targeting the feature branch

[AB#626091](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/626091)

### DIFF
--- a/tools/al-docs-plugin/.claude-plugin/plugin.json
+++ b/tools/al-docs-plugin/.claude-plugin/plugin.json
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "description": "AL codebase documentation generator - bootstrap, update, and audit hierarchical docs for Business Central AL apps",
   "author": {
-    "name": "NAV Team"
+    "name": "BC Application Team"
   },
   "keywords": ["al", "business-central", "documentation", "docs"]
 }


.github/workflows/PullRequestHandler.yaml
```diff
@@ -3,6 +3,7 @@ name: 'Pull Request Build'
 on:
   pull_request:
     branches: [ 'main', 'releases/*', 'features/*' ]
+    paths: ['THIS_PR_IS_TESTING_DOCS_WORKFLOW_ONLY']
```
aholstrup1: This needs to be removed

.github/workflows/DocumentationMaintenance.yaml
```diff
@@ -0,0 +1,121 @@
...
+          else
+            echo "has_gaps=false" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Post audit report
+        if: steps.audit.outputs.has_gaps == 'true'
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          gh pr comment ${{ github.event.pull_request.number }} --body "$(cat <<'COMMENT'
```
aholstrup1: Maybe we should add some kind of handling to make sure this isn't commented on every single iteration of your PR. That would be quite annoying I think 😄

.github/workflows/DocumentationMaintenance.yaml
```diff
@@ -0,0 +1,123 @@
...
+          while IFS= read -r APP_PATH; do
+            [ -z "$APP_PATH" ] && continue
+            APP_NAME=$(jq -r '.name // "unknown"' "$APP_PATH/app.json" 2>/dev/null || echo "unknown")
+            echo "Auditing $APP_NAME at $APP_PATH..."
+
+            OUTPUT=$(timeout 300 copilot -p \
+              "Look at the AL app at $APP_PATH. Check if it has a CLAUDE.md or docs/ directory.
+              Count the AL objects (tables, codeunits, pages). Reply in 2-3 sentences:
+              what you found, whether docs exist, and if the developer should run al-docs." \
+              --model claude-opus-4.6 \
```
aholstrup1, line 107: I wonder if the model should be defined on the skill (if that is possible). Just to ensure that the skill is invoked the same way both locally and here. I'm guessing you could get different results depending on which model you use?

.github/workflows/DocumentationMaintenance.yaml
```diff
@@ -0,0 +1,121 @@
...
+jobs:
+  AuditDocs:
+    if: github.event.pull_request.head.repo.full_name == github.event.pull_request.base.repo.full_name
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    env:
+      COPILOT_GITHUB_TOKEN: ${{ secrets.COPILOT_GH_TOKEN }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
aholstrup1: Try avoiding non-pinned dependencies. We usually prefer pinning the dependency on the commit hash like this:
`actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2`

.github/workflows/DocumentationMaintenance.yaml
Groenbech96: I think its fine. Everyone is using opus 4.6 right now, lets see what happens with newer models.

tools/al-docs-plugin/.claude-plugin/plugin.json
```diff
@@ -0,0 +1,9 @@
+{
+  "name": "al-docs",
+  "version": "0.1.0",
+  "description": "AL codebase documentation generator - bootstrap, update, and audit hierarchical docs for Business Central AL apps",
+  "author": {
+    "name": "NAV Team"
```
JesperSchulz: The what now?

tools/al-docs-plugin/skills/al-docs/references/al-scoring.md
```diff
@@ -0,0 +1,56 @@
...
+| Event subscribers present | +1 | Grep for `[EventSubscriber]` |
+| 10+ total AL objects | +2 | Count all `.al` files in the subfolder |
+| Complex codeunits (10+ procedures) | +1 | Grep for `procedure ` count per codeunit file |
+| Extension objects present | +1 | Grep for `^tableextension ` or `^pageextension ` |
+
+## Score classification
+
+| Category | Score | Documentation required |
+|----------|-------|----------------------|
+| MUST_DOCUMENT | 7+ | CLAUDE.md + relevant of data-model.md, business-logic.md, extensibility.md, or patterns.md |
```
JesperSchulz, line 38: Are we certain this is the best name?

Groenbech96: I will resolve the comment. I am open to changing them, but i dont see myself that they need to. If we find different names, lets update the plugin, and migrate the files.

PR #7296, opened 2026-03-20
Fix: Improve UX on failing validation of additional field values during E-Doc finalization

ventselartur:
## Summary

Implement the new codeunit that binds to the certain events and save the context, so that, when finalize is failed, it shows more details rather than just an error message. Example:

image

Fixes [AB#623716](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/623716) — [Payables Agent] Draft could not be finalised due to field length issues.


Groenbech96 (dismissed):

src/Apps/W1/EDocument/Test/src/Processing/EDocProcessTest.Codeunit.al
```diff
@@ -580,6 +581,66 @@ codeunit 139883 "E-Doc Process Test"
...
+        EDocumentService."Read into Draft Impl." := "E-Doc. Read into Draft"::PEPPOL;
+        EDocumentService.Modify();
+
+        // [GIVEN] An additional field is configured for Location Code (Code[10])
+        EDocPurchLineFieldSetup."Field No." := PurchaseInvoiceLine.FieldNo("Location Code");
+        EDocPurchLineFieldSetup.Insert();
+
+        // [GIVEN] An inbound e-document is received and a draft created
+        EDocImportParams."Step to Run" := "Import E-Document Steps"::"Prepare draft";
+        WorkDate(DMY2Date(1, 1, 2027));
```
adrogin: Does the test have to set a specific fixed workdate? How will it affect subsequent tests, especially a few years down the line? Workdate is not automatically reverted when the test transaction rolls back. It's probably worth resetting it if the change it necessary (although I'd prefer not to touch it unless the change can't be avoided).

AleksanderGladkov (dismissed):

src/Apps/W1/EDocument/Test/src/Processing/EDocProcessTest.Codeunit.al
ventselartur: Great catch, removed!

src/Apps/W1/EDocument/App/src/Processing/Import/EDocImportContext.Codeunit.al
```diff
@@ -0,0 +1,261 @@
...
+namespace Microsoft.eServices.EDocument.Processing.Import;
+
+using Microsoft.Purchases.Document;
+
+codeunit 6199 "E-Doc. Import Context"
+{
+    Access = Internal;
+    InherentEntitlements = X;
+    InherentPermissions = X;
+    SingleInstance = true;
```
mynjj: singleinstance defeats the purpose of making it manually bound, it shold work without it.

mynjj: Also, just personal preference, but having procedures that call Bind/Unbind seem unnecessary when the caller could call the usual API themselves.

src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/History/EDocPurchaseHistMapping.Codeunit.al
```diff
@@ -223,7 +225,10 @@ codeunit 6120 "E-Doc. Purchase Hist. Mapping"
                 continue;
             EDocPurchLineField.Get(EDocumentPurchaseLine, EDocPurchLineFieldSetup);
             NewPurchLineFieldRef := NewPurchLineRecordRef.Field(EDocPurchLineFieldSetup."Field No.");
-            NewPurchLineFieldRef.Validate(EDocPurchLineField.GetValue());
+            FieldValue := EDocPurchLineField.GetValue();
+            EDocImportContext.SetAdditionalFieldContext(NewPurchLineFieldRef.Name(), EDocPurchLineFieldSetup."Field No.", EDocPurchLineField.GetValueAsText());
+            NewPurchLineFieldRef.Validate(FieldValue);
```
mynjj, line 230: did you try calling this `NewPurchLineFieldRef.Validate` under a codeunit.run patter instead? I think it would read a bit clearer and it doesn't need the indirection of the context codeunit and multiple events. But maybe this pattern doesn't work here?

ventselartur: Tried. It won't work in case when RecRef.Validate writes to the database because it requires the commit and we should not from my point of view do the commit anywhere here

mynjj (approved):

src/Apps/W1/EDocument/App/src/Processing/Import/EDocImportContext.Codeunit.al
ventselartur: Removed

Groenbech96 (approved):

PR #7298, opened 2026-03-20
AI Eval Suite - Enable selecting the suite from the card itself

nikolakukrika:
#### Summary 
Allow direct navigation to AI Evals Suite. The last selected suite will be saved.  This will save time while navigating to the page and running tests because now the page can be opened directly. 

#### Work Item(s) 
Fixes [AB#626352](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/626352)

### DIFF
--- a/src/Tools/AI Test Toolkit/src/TestSuite/AITTestSuite.Page.al
+++ b/src/Tools/AI Test Toolkit/src/TestSuite/AITTestSuite.Page.al
@@ -49,6 +49,11 @@ page 149031 "AIT Test Suite"
 
                     trigger OnValidate()
                     begin
+                        if IsNullGuid(Rec.SystemId) then begin
+                            Rec.Code := CurrentSuiteCode;
+                            Rec.Insert();
+                        end;
+
                         ChangeTestSuite();
                     end;
 


src/Tools/AI Test Toolkit/src/TestSuite/AITTestSuite.Page.al
```diff
@@ -15,8 +15,10 @@ page 149031 "AIT Test Suite"
     PageType = Document;
     SourceTable = "AIT Test Suite";
     Extensible = true;
+    SaveValues = true;
     DataCaptionExpression = PageCaptionLbl + ' - ' + Format(Rec."Test Type") + ' (' + Format(Rec."Copilot Capability") + ') - ' + Rec."Code";
```
qutreson, line 19: Is the data caption being updated correctly?

nikolakukrika: It is. There is no need to tie it to the global variable. Rec will be updated always, it will match the global variable.

PR #7308, opened 2026-03-23
Fix manually added dimensions on e-doc draft not carried to purchase invoice

ventselartur:
## Summary
- **Bug:** When the user manually adds dimensions on an e-document draft line via the Dimensions page action, the changes are not persisted to the database. When the purchase invoice is created, `CreatePurchaseInvoiceLine` reads the e-doc line from the DB and gets `[BC] Dimension Set ID = 0`, so the dimensions are lost.
- **Root cause:** `LookupDimensions()` in `EDocumentPurchaseLine.Table.al` updates `[BC] Dimension Set ID` and shortcut dimension codes in memory but never calls `Modify()` to persist them.
- **Fix:** Added `Modify()` call in `LookupDimensions()` when the dimension set ID changes, matching the standard BC pattern (e.g., `Purchase Line.ShowDimensions()`).

## Test plan
- [x] New test `ManuallyAddedDimensionsOnDraftAreCarriedToPurchaseInvoice` in codeunit 139883
  - Creates an e-document draft via PEPPOL import
  - Calls `LookupDimensions()` to add a dimension (via ModalPageHandler)
  - Re-reads the record from the database
  - Verifies the dimension set ID was persisted
- [x] Test fails without the fix (`Dimension Set ID: 0`)
- [x] Test passes with the fix

Fixes [AB#623719](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/623719)

### DIFF
--- a/src/Apps/W1/EDocument/Test/src/LibraryEDocument.Codeunit.al
+++ b/src/Apps/W1/EDocument/Test/src/LibraryEDocument.Codeunit.al
@@ -289,14 +289,14 @@ codeunit 139629 "Library - E-Document"
         exit(EDocumentPurchaseLine);
     end;
 
-    procedure CreateInboundPEPPOLDocumentToState(var EDocument: Record "E-Document"; EDocumentService: Record "E-Document Service"; FileName: Text; EDocImportParams: Record "E-Doc. Import Parameters"): Boolean
+    procedure CreateInboundPEPPOLDocumentToState(var EDocument: Record "E-Document"; EDocumentService: Record "E-Document Service"; FileName: Text; TempEDocImportParams: Record "E-Doc. Import Parameters"): Boolean
     var
         EDocImport: Codeunit "E-Doc. Import";
         InStream: InStream;
     begin
         NavApp.GetResource(FileName, InStream, TextEncoding::UTF8);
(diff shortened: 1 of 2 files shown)


src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/EDocumentPurchaseLine.Table.al
```diff
@@ -334,6 +334,8 @@ table 6101 "E-Document Purchase Line"
             Rec, "[BC] Dimension Set ID", StrSubstNo('%1 %2', "E-Document Entry No.", "Line No."),
             "[BC] Shortcut Dimension 1 Code", "[BC] Shortcut Dimension 2 Code");
         DimMgt.UpdateGlobalDimFromDimSetID("[BC] Dimension Set ID", "[BC] Shortcut Dimension 1 Code", "[BC] Shortcut Dimension 2 Code");
+        if OldDimSetID <> "[BC] Dimension Set ID" then
```
mynjj: why aren't we using `Rec.` in this procedure?

AleksanderGladkov (approved):

PR #7326, opened 2026-03-23
[Shopify] Fix duplicate Sales Line error in refund Credit Memo creation

onbuyuka:
## Summary
- Aligned `GetLastLineNo` in `ShpfyCreateSalesDocRefund` with the `GetNextLineNo` pattern from `ShpfyProcessOrder`: returns `10000` for empty documents, `LastLineNo + 10000` for non-empty
- Replaced all `LineNo += 10000` manual increments with `GetLastLineNo(...)` calls so every insert gets the actual next available line number from the database
- Fixes duplicate Sales Line error when "Shopify Order No. on Doc. Line" is enabled and the Credit Memo has pre-existing lines (e.g. from Standard Customer Sales Codes or event subscribers)

Fixes [AB#626533](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/626533)


src/Apps/W1/Shopify/App/src/Order Return Refund Processing/Codeunits/ShpfyCreateSalesDocRefund.Codeunit.al
Groenbech96, line 140: Is there not a GetLastLineNO in BaseApp? 
And what about concurrency here?

onbuyuka: added setloadfields, shouldn't be any issue since we are only retrieve lines from a particular header - we have the same pattern in sales order creation

PR #7350, opened 2026-03-24
Show notification for uninstalled but synchronized extensions

jhejlsberg:
#### Summary 

When opening the Extension Management page, a notification is now displayed if there are extensions that have been uninstalled but still have synchronized data. The notification includes a "Show Extensions" action that opens the "Delete Orphaned Extension Data" page, allowing users to review and clean up leftover data from uninstalled extensions.

#### Work Item(s) 
Fixes [AB#616423](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/616423)


src/System Application/App/Extension Management/src/ExtensionManagement.Page.al
```diff
@@ -403,6 +417,7 @@ page 2500 "Extension Management"
         InfoStyle: Boolean;
         HelpActionVisible: Boolean;
         IsSourceSpecificationAvailable: Boolean;
+        UninstalledExtensionsMsg: Label 'There are uninstalled but still synchronized extensions.';
```
SBalslev: Maybe this should reference "Orphaned Extension Data" as this is the term used elsewhere in the page.

SusanneWindfeldPedersen: "There's orphaned data from uninstalled extensions. Use this action to review the data." --> @jhejlsberg not sure if this can be used in the scenario, otherwise rewrite to fit and send to me again, thanks

jhejlsberg: The action opens the "Delete Orphaned Extension Data" page. What about: "There's orphaned data from uninstalled extensions. Use this action to review the uninstalled extensions" @SusanneWindfeldPedersen

SusanneWindfeldPedersen (dismissed):

SBalslev (dismissed):

src/System Application/App/Extension Management/src/ExtensionManagement.Page.al
```diff
@@ -381,6 +381,20 @@ page 2500 "Extension Management"
         ActionsEnabled := false;
 
         HelpActionVisible := false;
+        ShowUninstalledExtensionsNotification();
```
thloke, line 385: Are we sure we want to be showing this every time the page is opened? There are legitimate reasons to have orphaned data since it is restored if the original extension is re-installed. Ideally there would be a way to mark orphaned data as "expected" and only show this if there are any new items, otherwise I can see this notification being ignored.

Also double check that the orphaned data page has some warning that if the data is deleted, it won't be restored when the extension is re-installed.

jhejlsberg: I have double checked the latter, and there is a pop-up opened when you select delete for a given uninstalled extension with a warning that this action can not be undone.

﻿﻿The purpose of the banner is simply to draw attention to orphaned data. It's quite minimal, so showing it on page open shouldn't be disruptive. 
﻿﻿
﻿﻿I agree that if extension data was intentionally kept for a later restore, it would be nicer to suppress the banner but I don't see this as a significant issue. But if you think it is worth implementing let me know and I will add it.

thloke: I think it's worth it, otherwise the moment you have a single uninstalled extension whose data you want to keep, this notification becomes useless since there's no way to distinguish from the data you want to keep from what you might want to clean up.

jehelles (dismissed):

SusanneWindfeldPedersen (approved):

jehelles (approved):

src/System Application/App/Extension Management/src/ExtensionManagement.Page.al
```diff
@@ -381,6 +382,30 @@ page 2500 "Extension Management"
...
+    end;
+
+    local procedure ShowUninstalledExtensionsNotification()
+    var
+        Uninstalled: Record "Extension Database Snapshot";
+        Notif: Notification;
```
haoranpb, line 391: I'd give it another name like

```suggestion
        OrphanedDataNotification: Notification;
```

src/System Application/App/Extension Management/src/ExtensionManagement.Page.al
```diff
@@ -381,6 +382,30 @@ page 2500 "Extension Management"
         ActionsEnabled := false;
 
         HelpActionVisible := false;
+        ShowUninstalledExtensionsNotification();
+    end;
+
+    local procedure ShowUninstalledExtensionsNotification()
+    var
+        Uninstalled: Record "Extension Database Snapshot";
```
haoranpb, line 390: The convention is use the full record name

```suggestion
        ExtensionDatabaseSnapshot: Record "Extension Database Snapshot";
```

src/System Application/App/Extension Management/src/DeleteOrphanedExtensionData.Page.al
```diff
@@ -113,10 +122,28 @@ page 2514 "Delete Orphaned Extension Data"
...
+                    Enabled = HasUnreviewedExtensions;
+                    Image = Approve;
+                    ToolTip = 'Mark all orphaned extension data as reviewed. This acknowledges the data has been seen and is kept intentionally.';
+
+                    trigger OnAction()
+                    var
+                        ExtensionDatabaseManagement: Codeunit "Extension Database Management";
+                    begin
+                        ExtensionDatabaseManagement.MarkAllOrphanedExtensionDataAsReviewed();
+                        UpdateHasUnreviewedExtensions();
```
haoranpb, line 143: Is `UpdateHasUnreviewedExtensions` redundant here?

src/System Application/App/Extension Management/src/ExtensionManagement.Page.al
```diff
@@ -381,6 +382,30 @@ page 2500 "Extension Management"
...
+        Notif: Notification;
+    begin
+        if not Uninstalled.ReadPermission() then
+            exit;
```
haoranpb, line 394: I'm not sure if ReadPermission will return true if they only have indirect read

PR #7365, opened 2026-03-25
[Agent] Uptake new table for Agent Task Consumption and Agent Eval Credit limit fixes

qutreson:
#### Summary 
Changes in System Application:
- Remove copilot credit related utilities from Agent Task codeunit in favor of new Agent Task Consumption codeunit.
- Change the source table of the Agent Consumption Overview table to use a table using per-row filtering. This ensures that users can only see data related to the agent they have access to, allowing to show consumption data to more users.
- The Agent Consumption Overview's opening logic has been to update to error if the page is opened for a specific agent or task that the user does not have access to.
- The Agent Consumption Overview page includes a notification for users which are not agent admins to let them know they are seeing only the consumption of agents they have access to.

Changes in AI Test Toolkit:
- Expose copilot credits consumed to all users, including the ones that are not agent admins.
- Add ability to see the Agent Consumption Overview from AI Test Suites, Lines, History, etc.
- Introduce monthly Company and Environment credits
- Make Company Eval Monthly Copilot Credits overview editable only by Agent Admins.
- Make Company Eval Monthly Copilot Credits overview editable only by Agent Admins in all companies.
- Add event for returning the platform when the credit limits have been reached.
- Adding audit logs when company eval limits are changed
- Adding verification that limits are only updated from UI sessions.

#### Work Item(s) 
Fixes [AB#625710](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/625710)

### DIFF
--- a/src/System Application/App/Agent/Interaction/AgentTask.Codeunit.al
+++ b/src/System Application/App/Agent/Interaction/AgentTask.Codeunit.al
@@ -156,7 +156,8 @@ codeunit 4303 "Agent Task"
     /// </summary>
     /// <param name="AgentTaskLogEntry">The agent task log entry to get details for.</param>
     /// <returns>The details of the agent task log entry.</returns>
-    procedure GetDetailsForAgentTaskLogEntry(var AgentTaskLogEntry: Record "Agent Task Log Entry"): Text
+    [Scope('OnPrem')]
+    procedure GetLogEntryDetails(var AgentTaskLogEntry: Record "Agent Task Log Entry"): Text
     var
         AgentTaskImpl: Codeunit "Agent Task Impl.";
     begin


src/System Application/App/Agent/Interaction/AgentTask.Codeunit.al
```diff
@@ -148,6 +148,18 @@ codeunit 4303 "Agent Task"
         exit(AgentTaskImpl.GetCopilotCreditsConsumed(AgentTaskID));
     end;
 
+    /// <summary>
+    /// Gets the details for the specified agent task log entry.
+    /// </summary>
+    /// <param name="AgentTaskLogEntry">The agent task log entry to get details for.</param>
+    /// <returns>The details of the agent task log entry.</returns>
+    procedure GetDetailsForAgentTaskLogEntry(var AgentTaskLogEntry: Record "Agent Task Log Entry"): Text
```
qutreson: Unrelated, but trying to remove references from Agent Design Experience to the internals of the System Application.

src/System Application/App/Agent/Monetization/AgentConsumptionOverview.Page.al
```diff
@@ -30,6 +27,17 @@ page 4333 "Agent Consumption Overview"
...
         area(Content)
         {
+            group(Info)
+            {
+                ShowCaption = false;
+                Visible = not CurrentUserIsAgentAdmin;
+
+                label(NonAdminDisclaimer)
+                {
+                    Caption = 'The consumption data shown only covers the agents you have access to and may be incomplete. Users with Agent-Admin have access to the complete consumption data.';
```
nikolakukrika: I do not like the text like this. It gets in the way once people learn about the system. It is an UX violation. Once the users learn how it works it is in the way, it is also taking UI real-estate. 

We should replace this label with error once you open the page - e.g.

This should be sufficient for existing scenarios:
You do not have the access to the agent to see the consumption. For more information see - help link

If we want to show the consumption for all agents:
You do not have the permissions to see the consumption for all agents. For more information see - help link

src/System Application/App/Agent/Monetization/AgentConsumptionOverview.Page.al
```diff
@@ -322,58 +329,55 @@ page 4333 "Agent Consumption Overview"
...
+        AgentTaskConsumption.Copy(Rec);
+        AgentTaskConsumption.SetRange("Task Id", Rec."Task ID");
+        TotalTaskConsumedCredits := GetCopilotCreditsConsumed(AgentTaskConsumption);
+    end;
+
+    local procedure GetCopilotCreditsConsumed(AgentTaskConsumption: Record "Agent Task Consumption") ConsumedCredits: Decimal
+    begin
+        if AgentTaskConsumption.FindSet() then
+            repeat
+                ConsumedCredits += AgentTaskConsumption."Copilot Credits";
```
nikolakukrika: Why not CalcSums?

src/System Application/App/Agent/Interaction/Internal/AgentTaskList.Page.al
```diff
@@ -226,16 +219,15 @@ page 4300 "Agent Task List"
...
-        end;
+        ConsumedCredits := 0;
 
-        UserAIConsumptionData.SetRange("Agent Task Id", Rec.ID);
-        UserAIConsumptionData.CalcSums("Copilot Credits");
-        ConsumedCredits := UserAIConsumptionData."Copilot Credits";
+        AgentTaskConsumption.SetRange("Task Id", Rec.ID);
+        if AgentTaskConsumption.FindSet() then
+            repeat
+                ConsumedCredits += AgentTaskConsumption."Copilot Credits";
```
nikolakukrika: Why not CalcSums?

src/System Application/App/Agent/Monetization/AgentConsumptionOverview.Page.al
qutreson: Not supported on the VT

src/System Application/App/Agent/Interaction/Internal/AgentTaskList.Page.al
qutreson: Not supported on the VT

src/System Application/App/Agent/Monetization/AgentConsumptionOverview.Page.al
qutreson: I've made changes to distinguish the three scenarios:
- page run with agent filter: if you don't have access to that agent, you get an error
- page run with agent task filter, if you don't have access to the agent that owns the task, you get an error
- page run without filter (from URL) you get a notification if you're not an admin since you'd see partial data

src/System Application/App/Agent/Monetization/AgentConsumptionOverview.Page.al
qutreson: Added support for it

src/System Application/App/Agent/Interaction/Internal/AgentTaskList.Page.al
qutreson: Added support for it

src/System Application/App/Agent/Monetization/AgentConsumptionOverview.Page.al
```diff
@@ -430,23 +401,82 @@ page 4333 "Agent Consumption Overview"
...
+        TaskId: BigInteger;
+    begin
+        RecallNonAdminDisclaimerNotification();
+
+        if AgentSystemPermissions.CurrentUserHasCanManageAllAgentsPermission() then
+            exit;
+
+        // There is a filter on the agent user security, check that the user has access to the agent.
+        if Rec.GetFilter("Agent User Security Id") <> '' then begin
+            Evaluate(AgentUserSecurityId, Rec.GetFilter("Agent User Security Id"));
```
nikolakukrika: Maybe we could use set filter from get filter to avoid the evaluate? It would help with future cases if the filter is more advanced.

src/System Application/App/Agent/Monetization/AgentConsumptionOverview.Page.al
```diff
@@ -430,23 +401,82 @@ page 4333 "Agent Consumption Overview"
...
+            Evaluate(TaskId, Rec.GetFilter("Task Id"));
+            if not AgentTask.Get(TaskId) then
+                Error(YouDoNotHaveAccessToTheAgentErr)
+            else
+                if not AgentSystemPermissions.CurrentUserCanUseAgent(AgentTask."Agent User Security ID") then
+                    Error(YouDoNotHaveAccessToTheAgentErr);
+            exit;
+        end;
+
+        SendNonAdminDisclaimerNotification();
```
nikolakukrika, line 460: Should we error here? We throw the error in other cases, we were not able to identify the agent.

src/Tools/AI Test Toolkit/src/Agent/AgentTestConsumptionLog.Codeunit.al
```diff
@@ -0,0 +1,74 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace System.TestTools.AITestToolkit;
+
+codeunit 149047 "Agent Test Consumption Log"
```
nikolakukrika: Nit - do we need log?
Can it be Agent Test Consumption?

nikolakukrika: It would help to avoid the clash with the table name

src/Tools/AI Test Toolkit/src/Limits/CopilotCredits/AITEvalMonthlyCopilotCred.Codeunit.al
```diff
@@ -65,127 +108,200 @@ codeunit 149039 "AIT Eval Monthly Copilot Cred." implements "AIT Eval Limit Prov
...
 
     procedure OpenConfigurationPage(Notification: Notification)
     begin
         OpenConfigurationPage()
     end;
 
+    [EventSubscriber(ObjectType::Table, Database::"AIT Eval Monthly Copilot Cred.", OnBeforeInsertEvent, '', false, false)]
+    local procedure OnBeforeInsertMonthlyCopilotCreditLimits(var Rec: Record "AIT Eval Monthly Copilot Cred.")
+    begin
+        Rec.VerifyWriteOperationAllowed();
```
nikolakukrika, line 270: These are not safe. I can use the data transfer or other means to skip the check.

nikolakukrika: We do not need to tackle it now, but we may need a platform table for vNext.

nikolakukrika (approved):

src/Tools/AI Test Toolkit/src/Limits/CopilotCredits/AITEvalMonthlyCopilotCred.Codeunit.al
qutreson: This is what was discussed. It's a best effort and documentation will reflect this.

src/System Application/App/Agent/Monetization/AgentConsumptionOverview.Page.al
qutreson: Which case are you mentioning? I already throw an error, when:
- the filter is a single task which cannot be found (likely because user has no access to it)
- the filter is a single agent which cannot be found
- the filter is a single agent the user does not have access too

These were mainly usability improvement to cover the main cases. There is no risk regarding the data exposed if the filters are more complex in the future. The source table is filtered based on the user's agent access controls. I believe it is fine as it is unless I missed something.

This said, I think the return value of Evaluate needs to be handled better here.

src/System Application/App/Agent/Monetization/AgentConsumptionOverview.Page.al
qutreson: How would this work? I am not filtering the source record, I am performing a method call which requires a Guid as parameter

src/Tools/AI Test Toolkit/src/Agent/AgentTestConsumptionLog.Codeunit.al
qutreson: We didn't really have cases where both are defined in the same scope, but I applied the rename

src/System Application/App/Agent/Monetization/AgentConsumptionOverview.Page.al
```diff
@@ -430,23 +401,89 @@ page 4333 "Agent Consumption Overview"
...
+        // There is a filter on a specific agent user security, check that the user has access to this agent.
+        if Rec.GetFilter("Agent User Security Id") <> '' then begin
+            if Evaluate(AgentUserSecurityId, Rec.GetFilter("Agent User Security Id")) then
+                if not AgentSystemPermissions.CurrentUserCanUseAgent(AgentUserSecurityId) then
+                    Error(YouDoNotHaveAccessToTheAgentErr);
+            exit;
+        end;
+
+        // There is a filter on a specific task ID, check that the user has access to this task.
+        if Rec.GetFilter("Task Id") <> '' then begin
```
BazookaMusic, line 449: I think there's a small bug here, if there is neither a filter on user security ID or task ID, you should error. Otherwise I think it's possible to bypass this check by coming in the page via the ID in the URL.

src/System Application/App/Agent/Monetization/AgentConsumptionOverview.Page.al
```diff
@@ -430,23 +401,89 @@ page 4333 "Agent Consumption Overview"
...
+        NonAdminNotification.Message := NonAdminDisclaimerMsg;
+        NonAdminNotification.Scope := NotificationScope::LocalScope;
+        NonAdminNotification.Send();
+    end;
+
+    local procedure GetNonAdminDisclaimerNotificationId(): Text
     begin
-        UserAIConsumptionData.SetRange("Agent Task Id", Rec."Agent Task ID");
-        Page.Run(Page::"Agent Consumption Overview", UserAIConsumptionData);
+        exit('b9234f5f-3e6b-437a-9f34-acdb9bf9fb8f');
```
BazookaMusic, line 486: what is this?

src/System Application/App/Agent/Monetization/AgentConsumptionOverview.Page.al
qutreson: It's ok, the virtual table is filtered anyway. This check is simply for a better UX, not for security reasons. I updated the function header to hint toward it already

src/System Application/App/Agent/Monetization/AgentConsumptionOverview.Page.al
qutreson: Notifications in AL need to have a unique ID to send them and recall them, so that's just a random Guid.

src/System Application/App/Agent/Monetization/AgentConsumptionOverview.Page.al
```diff
@@ -473,7 +507,8 @@ page 4333 "Agent Consumption Overview"
         TotalsVisible: Boolean;
         EverythingTok: Label 'Everything';
         AgentTaskNameTxt: Label 'Task #%1 - %2', Comment = '%1 - ID of the agent task, %2 - Title of the agent task';
-        YourNotAuthorizedToViewMonetizationDataErr: Label 'You are missing the required permissions to view monetization data.';
         TheEndDateIsTodayMsg: Label 'The end date is already set to today. You cannot move the date range filter further.';
+        NonAdminDisclaimerMsg: Label 'Consumption data is limited to agents you have access to. To view data for all agents, contact a user with the Agent Admin role in all companies.';
```
BazookaMusic, line 511: Should we use the precise name 'Agent - Admin'?

qutreson: I can take this in the next PR, the name is Agent - Admin but the caption is agent administrator, so it should be understandable in all cases

PR #7391, opened 2026-03-25
[E-doc] [Clearance model] Fix bug that clearance model generate replicated log

gggdttt:
#### Work Item(s) 
Fixes [AB#605044](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/605044)


src/Apps/W1/EDocument/App/src/Workflow/EDocumentWorkFlowProcessing.Codeunit.al
```diff
@@ -481,6 +482,13 @@ codeunit 6135 "E-Document WorkFlow Processing"
 
         EDocumentService.Get(WorkflowStepArgument."E-Document Service");
 
+        // If the E-Document has already been exported, skip re-exporting it.
+        if EDocServiceStatus.Get(EDocument."Entry No", EDocumentService.Code) then
```
Groenbech96, line 486: What if you want to reexport? 
Is that possible.

Try post document: Then it gets exported, then you change some editable field on posted document, and want that change to reflect in the e-document. What do you do in this case?

gggdttt: Our fix is in `EDocumentWorkFlowProcessing.ExportEDocument()` — the workflow response handler. It only runs when the workflow engine automatically fires the Export response step. The skip logic prevents a duplicate within the same workflow execution, because CreateEDocument already pre-exports the document before the workflow fires its Export response.

Manual re-export `("Recreate Document" action on the E-Document page)` calls `EDocExport.Recreate()` → `EDocExport.ExportEDocument()` directly. This path never goes through the workflow processing layer, so the status check we added is never hit.


```text
Manual Re-export (unaffected):
  Page Action "Recreate" -> EDocExport.Recreate() -> EDocExport.ExportEDocument() -> always executes

Workflow auto-trigger (what we fixed):
  Workflow Engine -> EDocumentWorkFlowProcessing.ExportEDocument() ->  [status check -> skip if already Exported]
```

gggdttt: (Correct me if I misunderstand this function)

Groenbech96 (approved):

haoranpb (approved):

PR #7444, opened 2026-03-30
[Agent] Better handle delay inserted record in Agent Configuration Rights

qutreson:
#### Summary 
The page  was updated with DelayedInsert which means the calls to Modify do not work on new records. Remove call to Modify to let the platform update records when the page is closed.

#### Work Item(s) 
Fixes [AB#623048](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/623048)

### DIFF
--- a/src/System Application/App/Agent/Setup/AgentCreationControls/AgentCreationControlPart.Page.al
+++ b/src/System Application/App/Agent/Setup/AgentCreationControls/AgentCreationControlPart.Page.al
@@ -101,7 +101,6 @@ page 4332 "Agent Creation Control Part"
         if CreationControlLookup.RunModal() = Action::LookupOK then begin
             CreationControlLookup.GetRecord(TempAgentCreationControlLookup);
             Rec."Company Name" := CopyStr(TempAgentCreationControlLookup."Key", 1, MaxStrLen(Rec."Company Name"));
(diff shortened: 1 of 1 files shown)


src/System Application/App/Agent/Setup/AgentCreationControls/AgentCreationControlPart.Page.al
```diff
@@ -101,7 +101,7 @@ page 4332 "Agent Creation Control Part"
         if CreationControlLookup.RunModal() = Action::LookupOK then begin
             CreationControlLookup.GetRecord(TempAgentCreationControlLookup);
             Rec."Company Name" := CopyStr(TempAgentCreationControlLookup."Key", 1, MaxStrLen(Rec."Company Name"));
-            Rec.Modify();
+            if Rec.Modify() then;
```
nikolakukrika: This is not that nice. It is better to do a check if the record was created or not. 
Alternative is that the Company Name update can happen on the page itself, so there is no need for Modify.

qutreson: Updated

PR #7447, opened 2026-03-30
[Shopify] Fix Total Amount not reduced by refunded tax after order edit

onbuyuka:
## Summary
- Fix `ConsiderRefundsInQuantityAndAmounts` to subtract both `Subtotal Amount` and `Total Tax Amount` from `Total Amount` (and presentment equivalents) when processing refunds from order edits
- Previously only `Subtotal Amount` was subtracted, leaving `Total Amount` overstated by the refunded tax amount

Fixes [AB#629296](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/629296)

### DIFF
--- a/src/Apps/W1/Shopify/Test/Order Refunds/ShpfyOrderRefundsHelper.Codeunit.al
+++ b/src/Apps/W1/Shopify/Test/Order Refunds/ShpfyOrderRefundsHelper.Codeunit.al
@@ -345,6 +345,24 @@ codeunit 139564 "Shpfy Order Refunds Helper"
         RefundLine.Modify(false);
     end;
 
+    internal procedure CreateRefundLineWithTaxAmount(RefundId: BigInteger; OrderLineId: BigInteger; SubtotalAmount: Decimal; TaxAmount: Decimal)
+    var
+        RefundLine: Record "Shpfy Refund Line";
+    begin
+        RefundLine."Refund Line Id" := Any.IntegerInRange(100000, 999999);
+        RefundLine."Refund Id" := RefundId;
+        RefundLine."Order Line Id" := OrderLineId;
+        RefundLine.Quantity := 1;
+        RefundLine.Amount := SubtotalAmount + TaxAmount;
+        RefundLine."Subtotal Amount" := SubtotalAmount;
+        RefundLine."Total Tax Amount" := TaxAmount;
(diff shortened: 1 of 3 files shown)


src/Apps/W1/Shopify/Test/Order Refunds/ShpfyOrderRefundsHelper.Codeunit.al
```diff
@@ -345,6 +345,24 @@ codeunit 139564 "Shpfy Order Refunds Helper"
         RefundLine.Modify(false);
     end;
 
+    internal procedure CreateRefundLineWithTaxAmount(RefundId: BigInteger; OrderLineId: BigInteger; SubtotalAmount: Decimal; TaxAmount: Decimal)
```
Groenbech96, line 348: Where is this used?

src/Apps/W1/Shopify/Test/Order Refunds/ShpfyOrderRefundsHelper.Codeunit.al
```diff
@@ -345,6 +345,24 @@ codeunit 139564 "Shpfy Order Refunds Helper"
...
+    var
+        RefundLine: Record "Shpfy Refund Line";
+    begin
+        RefundLine."Refund Line Id" := Any.IntegerInRange(100000, 999999);
```
darjoo, line 352: Why is the line Id random?

src/Apps/W1/Shopify/Test/Order Refunds/ShpfyOrderRefundsHelper.Codeunit.al
onbuyuka: in the test

src/Apps/W1/Shopify/Test/Order Refunds/ShpfyOrderRefundsHelper.Codeunit.al
onbuyuka: normally it would come from Shopify, but we do random if we are not mocking the request. this is basically testing the logic after import, for tests testing also the import we mock Shopify response which will have a set ID

darjoo (approved):

PR #7450, opened 2026-03-30
fix: remove Bash dependency from al-docs audit for CI

Groenbech96:
## Summary

- Removed `Bash(*)` from al-docs audit skill's allowed-tools, constraining it to Read/Glob/Grep only
- Added explicit "Do NOT use Bash or shell commands" instructions to all three subagent sections
- Simplified the CI workflow's `--allow-tool` list (removed find/wc/sort/uniq/jq/git-diff)
- Added XML doc comment to `ShpfyLoggingMode.Enum.al` to trigger the docs audit workflow

## Root cause

The audit skill declared `Bash(*)` but CI runs with `--no-ask-user` and restricted shell permissions. Subagents generated compound shell commands (`head`, `xargs`, `sed`, `for` loops) that got blocked with "Permission denied and could not request permission from user."

## Test plan

- [ ] Verify the Documentation Maintenance workflow runs on this PR without permission denied errors
- [ ] Verify the audit report is generated correctly using only Glob/Grep/Read

### DIFF
--- a/.github/workflows/DocumentationMaintenance.yaml
+++ b/.github/workflows/DocumentationMaintenance.yaml
@@ -103,7 +103,7 @@ jobs:
             echo "Auditing $APP_NAME at $APP_PATH..."
 
             timeout 600 copilot -p \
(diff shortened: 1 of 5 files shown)


Groenbech96:
Closing — AL-Docs CI being disabled, no longer relevant.

PR #7488, opened 2026-04-01
Fix attachment upload from E-Document factbox (Bug 619590)

Groenbech96:
## Summary
- Fix "The record is not open" error when uploading attachments from the E-Document page factbox when no attachments exist yet
- Add `OnAfterGetRecRefFail` subscriber to initialize RecRef from E-Document Entry No. (SubPageLink FilterGroup 4) when `GetRefTable` fails for Table ID = 0
- Add `OnBeforeInsertAttachment` subscriber to set `E-Document Attachment` and `E-Document Entry No.` fields so the attachment appears in the factbox

## Root cause
The E-Document page factbox links via `SubPageLink = "E-Document Entry No." = field("Entry No"), "E-Document Attachment" = const(true)` — it does not set `"Table ID"`. When there are no existing attachments (empty factbox), `Rec."Table ID" = 0`. The upload flow calls `GetRefTable(RecRef, Rec)`, which immediately returns `false` for Table ID = 0 **before** firing `OnAfterGetRefTable` (where the E-Document subscriber lives). The uninitialized `RecRef` is then passed to `SaveAttachment` → `InsertAttachment` → `RecRef.Find()` → platform error.

## Test plan
- [x] New test `UploadAttachmentToEDocSetsEDocFields` fails without the fix (Assert.IsTrue on E-Document Attachment field)
- [x] New test `GetRefTableReturnsFalseForTableIdZero` confirms the precondition
- [x] Both tests pass with the fix
- [x] Manual: open an E-Document with no attachments, upload a file from the factbox, verify it appears

[AB#619590](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/619590)

### DIFF
--- a/src/Apps/W1/EDocument/Test/src/Processing/EDocPDFMock.Codeunit.al
+++ b/src/Apps/W1/EDocument/Test/src/Processing/EDocPDFMock.Codeunit.al
@@ -23,7 +23,6 @@ codeunit 139782 "E-Doc PDF Mock" implements IStructureReceivedEDocument, IStruct
             EDocumentPurchaseHeader.Insert();
         end;
         exit(Enum::"E-Doc. Process Draft"::"Purchase Document");
-#pragma warning restore AL0432
     end;
 
     procedure View(EDocument: Record "E-Document"; TempBlob: Codeunit "Temp Blob")
--- a/src/Apps/W1/EDocument/Test/src/Processing/EDocPDFMock.Codeunit.al
+++ b/src/Apps/W1/EDocument/Test/src/Processing/EDocPDFMock.Codeunit.al
@@ -22,7 +22,6 @@ codeunit 139782 "E-Doc PDF Mock" implements IStructureReceivedEDocument, IStruct
             EDocumentPurchaseHeader."Vendor VAT Id" := '1111111111234';
             EDocumentPurchaseHeader.Insert();
         end;
-#pragma warning disable AL0432
         exit(Enum::"E-Doc. Process Draft"::"Purchase Document");
 #pragma warning restore AL0432
     end;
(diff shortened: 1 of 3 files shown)


src/Apps/W1/EDocument/Test/src/Processing/EDocPDFMock.Codeunit.al
```diff
@@ -22,7 +22,9 @@ codeunit 139782 "E-Doc PDF Mock" implements IStructureReceivedEDocument, IStruct
             EDocumentPurchaseHeader."Vendor VAT Id" := '1111111111234';
             EDocumentPurchaseHeader.Insert();
         end;
+#pragma warning disable AL0432
```
onbuyuka: this enum is not marked obsolete, why AL0432?

darjoo (dismissed):

src/Apps/W1/EDocument/Test/src/Processing/EDocPDFMock.Codeunit.al
Groenbech96:
```suggestion

```

src/Apps/W1/EDocument/Test/src/Processing/EDocPDFMock.Codeunit.al
```diff
@@ -22,7 +22,9 @@ codeunit 139782 "E-Doc PDF Mock" implements IStructureReceivedEDocument, IStruct
...
         exit(Enum::"E-Doc. Process Draft"::"Purchase Document");
+#pragma warning restore AL0432
```
Groenbech96:
```suggestion

```

src/Apps/W1/EDocument/Test/src/Processing/EDocPDFMock.Codeunit.al
Groenbech96: Right. Spilover from other PR. Fixed.

darjoo (approved):

gggdttt (approved):

PR #7618, opened 2026-04-09
Agent Message and Task - Fix modifying functions to only modify the target values (part 1)

BazookaMusic:
#### Summary 

The "Agent Message" and "Agent Task" codeunits had a major issue in the way their API worked. Methods which modified field values were using a var record parameter. What this means is that if the passed in record had modified fields and somebody calls for example `SetStatusToSent`, the code will attempt to modify all fields including the status.

**Fix**
Obsolete these methods and replace them with methods that take the PK and the new value. These will refetch the latest value, modify the field and then modify. They return the modified record as a return value.

#### Work Item(s)
Fixes [AB#629278](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/629278)


src/System Application/App/Agent/Interaction/Internal/AgentMessageImpl.Codeunit.al
```diff
@@ -30,15 +38,34 @@ codeunit 4308 "Agent Message Impl."
...
+    procedure UpdateText(TaskID: BigInteger; MessageID: Guid; NewMessageText: Text)
+    var
+        AgentTaskMessage: Record "Agent Task Message";
+    begin
+        AgentTaskMessage."Task ID" := TaskID;
+        AgentTaskMessage.ID := MessageID;
+        UpdateText(AgentTaskMessage, NewMessageText);
+    end;
+
     procedure UpdateText(var AgentTaskMessage: Record "Agent Task Message"; NewMessageText: Text)
```
qutreson, line 50: Since this method changed behavior, should we have a unit test for it to ensure we are not modifying all fields, but only the content?

BazookaMusic: Done

src/System Application/App/Agent/Interaction/Internal/AgentMessageImpl.Codeunit.al
```diff
@@ -18,6 +18,14 @@ codeunit 4308 "Agent Message Impl."
...
 
+    procedure GetText(TaskID: BigInteger; MessageID: Guid): Text
+    var
+        AgentTaskMessage: Record "Agent Task Message";
+    begin
+        AgentTaskMessage.Get(TaskID, MessageID);
+        exit(GetText(AgentTaskMessage));
+    end;
+
     procedure GetText(var AgentTaskMessage: Record "Agent Task Message"): Text
```
thpeder, line 29: The description said that we will be deprecating the "var record" based procedures. Isn't this one of those?

src/System Application/App/Agent/Interaction/Internal/AgentMessageImpl.Codeunit.al
```diff
@@ -30,15 +38,35 @@ codeunit 4308 "Agent Message Impl."
...
+        AgentTaskMessageToModify.Get(AgentTaskMessage."Task ID", AgentTaskMessage.ID);
+        AgentTaskMessageToModify.Content.CreateOutStream(ContentOutStream, AgentTaskImpl.GetDefaultEncoding());
         ContentOutStream.Write(NewMessageText);
-        AgentTaskMessage.Modify(true);
+        AgentTaskMessageToModify.Modify(true);
+
+        AgentTaskMessage.Content := AgentTaskMessageToModify.Content;
+    end;
+
+    procedure IsEditable(TaskID: BigInteger; MessageID: Guid): Boolean
```
thpeder, line 64: Might have missed it, but I didn't see any calls into this function from "UpdateText" is that because you can always update the text?

src/System Application/App/Agent/Interaction/Internal/AgentTaskImpl.Codeunit.al
```diff
@@ -82,29 +72,68 @@ codeunit 4300 "Agent Task Impl."
         exit(AgentTaskMessage);
     end;
 
+    local procedure SetMessageText(var AgentTaskMessage: Record "Agent Task Message"; MessageText: Text)
```
thpeder: Why do we have both a UpdateText and SetText?

src/System Application/App/Agent/Interaction/AgentMessage.Codeunit.al
```diff
@@ -18,6 +18,21 @@ codeunit 4307 "Agent Message"
...
     /// Get the message text for the given agent task message.
     /// </summary>
+    /// <param name="TaskID">The task ID of the message.</param>
+    /// <param name="MessageID">The unique identifier of the message.</param>
+    /// <returns>The body of the agent task message.</returns>
+    procedure GetText(TaskID: BigInteger; MessageID: Guid): Text
+    var
+        AgentMessageImpl: Codeunit "Agent Message Impl.";
+    begin
+        FeatureAccessManagement.AgentManagementAllowed(true);
```
thpeder, line 28: It is hard to know from the caller what the argument True means here. Should we have a "wrapper" function where it is clear from the name and the argument is fixed to true?

src/System Application/App/Agent/Interaction/Internal/AgentMessageImpl.Codeunit.al
qutreson: This is fine because it's not a write operation so you don't have the risk of having the Modify call throw an exception because "unexpected" fields have changed. The public API method does have a disclaimer to say you'd get the text from the record and not the latest one from the task in the DB.

src/System Application/App/Agent/Interaction/Internal/AgentTaskImpl.Codeunit.al
qutreson: I guess we could re-use the AgentMessageImpl.UpdateText here.

src/System Application/App/Agent/Interaction/Internal/AgentMessageImpl.Codeunit.al
qutreson: We're exposing this publicly. It's the caller's responsibility to ensure the message is editable first. This is not being called within the UpdateText message because the validation logic is done in the virtual table, so we're not duplicating it in app.

src/System Application/App/Agent/Interaction/AgentMessage.Codeunit.al
qutreson: We could, but I'll leave it as it is for now. This PR will be backported so reducing the number of changes.

src/System Application/App/Agent/Interaction/Internal/AgentTaskImpl.Codeunit.al
qutreson: Fixed

thpeder (approved):

nikolakukrika (approved):

PR #7619, opened 2026-04-09
Switch E-Document Core to new PEPPOL App objects

Groenbech96:
## Summary
- Replace obsoleted BaseApp PEPPOL codeunits/xmlports with new PEPPOL App equivalents in E-Document Core
- Add PEPPOL app dependency to E-Document Core app.json
- Update namespace from `Microsoft.Sales.Peppol` to `Microsoft.Peppol`

### Object replacements:
| Old (BaseApp - obsoleted) | New (PEPPOL App) |
|---|---|
| Codeunit "PEPPOL Validation" | Codeunit "PEPPOL30 Sales Validation" |
| Codeunit "PEPPOL Service Validation" | Codeunit "PEPPOL30 Service Validation" |
| Codeunit "PEPPOL Management" | Codeunit "PEPPOL30" |
| Codeunit "Serv. PEPPOL Management" | Codeunit "PEPPOL30" |
| XmlPort "Sales Invoice - PEPPOL BIS 3.0" | XmlPort "Sales Invoice - PEPPOL30" |
| XmlPort "Sales Cr.Memo - PEPPOL BIS 3.0" | XmlPort "Sales Cr.Memo - PEPPOL30" |

### Method renames:
- `CheckSalesInvoice` / `CheckSalesCreditMemo` / `CheckServiceInvoice` / `CheckServiceCreditMemo` → `ValidatePostedDocument`
- `CheckServiceHeader` → `Run` (uses OnRun trigger)
- `GetTotals` → `GetTaxTotals`
- `GetLineItemClassfiedTaxCategoryBIS` → `GetLineItemClassifiedTaxCategoryBIS` (typo fix)
- `ServPEPPOLMgt.MapServiceLineTypeToSalesLineType` → `PEPPOLMgt.MapServiceLineTypeToSalesLineType` (merged into PEPPOL30)


duichwer:
This change will probably break E Documents for Germany 
See: https://github.com/microsoft/ALAppExtensions/pull/29753
cc: @miljance

Groenbech96:
> This change will probably break E Documents for Germany 
> 
> See: https://github.com/microsoft/ALAppExtensions/pull/29753
> 
> cc: @miljance

Yep. It should be caught when bcapps is uptaken in internal pr. Then we correct those places where test no longer work.

miljance:
@Groenbech96 
Aleksandar told me that the VATEX Code should also work for PEPPOL. Could you take a look? I would not like to put it in E-Documents Core just to find out it belongs lower in the dependency tree too late.
https://github.com/microsoft/BCApps/pull/7580

miljance:
@Groenbech96 @altotovi Guys, I am wondering why Microsoft introduced the new PEPPOL app? As it seems E-Document Code will depend on it directly I do not see any reason that this App should remain in existence as it can blend into E-Document Core App. Thoughts? I mean the goal was to cleanup stuff from the Base App and this approach does not seem to be most optimal in my opinion.

Groenbech96:
> @Groenbech96 @altotovi Guys, I am wondering why Microsoft introduced the new PEPPOL app? As it seems E-Document Code will depend on it directly I do not see any reason that this App should remain in existence as it can blend into E-Document Core App. Thoughts? I mean the goal was to cleanup stuff from the Base App and this approach does not seem to be most optimal in my opinion.

Separation of concerns. Format should live outside core App. Some objects in E-Doc Core could potentially be moved to PEPPOL App.

miljance:
> > @Groenbech96 @altotovi Guys, I am wondering why Microsoft introduced the new PEPPOL app? As it seems E-Document Code will depend on it directly I do not see any reason that this App should remain in existence as it can blend into E-Document Core App. Thoughts? I mean the goal was to cleanup stuff from the Base App and this approach does not seem to be most optimal in my opinion.
> 
> Separation of concerns. Format should live outside core App. Some objects in E-Doc Core could potentially be moved to PEPPOL App.

I understand the concept of separation of concerns. My concern is that splitting functionality into many small apps with Subscribers increasingly impacts performance. E-Documents Core App already affects posting times in sales, and adding more layers will make this worse. Since E-Document Code will depend on PEPPOL directly, I would advocate for keeping it within E-Document Core App rather than introducing unnecessary complexity and performance overhead.

But of course, it is not up to me.

Groenbech96:
> I understand the concept of separation of concerns. My concern is that splitting functionality into many small apps with Subscribers increasingly impacts performance. E-Documents Core App already affects posting times in sales, and adding more layers will make this worse. Since E-Document Code will depend on PEPPOL directly, I would advocate for keeping it within E-Document Core App rather than introducing unnecessary complexity and performance overhead.
> 
> But of course, it is not up to me.

Would be great to know more about the solved posting times! That is definitely important, and the ability to split code, should ideally not slow things down once running on the platform.

PR #7623, opened 2026-04-09
[Subcontracting] Performance issues fix

AleksandricMarko:
#### Work Item(s) 
Fixes [AB#623643](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/623643)


ChethanT (dismissed):

AleksandricMarko:
Hello @SPinkow @stevengrossGOB
Could you please review this PR
Seems like there was a bug with if ItemLedgerEntry.IsEmpty() then (should be if ItemLedgerEntry**2**.IsEmpty() then)
Also, in 2 procedures DeleteSubcontractingDependentPurchaseLines was done, does it make sense to have same Purchase Line filters?

Both in file src/Apps/W1/Subcontracting/App/src/Process/Codeunits/SubcSynchronizeManagement.Codeunit.al

src/Apps/W1/Subcontracting/App/src/Process/Codeunits/Extensions/Manufacturing/SubcProdOrderCompExt.Codeunit.al
```diff
@@ -106,8 +106,10 @@ codeunit 99001524 "Subc. Prod. Order Comp. Ext."
         if ProdOrderComponent."Subcontracting Type" <> ProdOrderComponent."Subcontracting Type"::Purchase then
             exit;
 
+        PurchaseLine.SetCurrentKey("Prod. Order No.", "Prod. Order Line No.", "Routing No.", "Operation No.");
         PurchaseLine.SetRange("Prod. Order No.", ProdOrderComponent."Prod. Order No.");
         PurchaseLine.SetRange("Prod. Order Line No.", ProdOrderComponent."Prod. Order Line No.");
+        PurchaseLine.SetLoadFields("Document Type", "Document No.", "Line No.");
```
SPinkow, line 116: I don't know, if this is quite optimal. In the following loop we use a transferfields to copy all fields to the temp purchase line. I don't know the exact logic, but could it be the case that transferfields would raise a JIT load?

Maybe we can avoid that by not using the transferfields method in the loop. If I see it correctly the only fields we would need are the primary key fields of purchase line. So we can easily replace the transferfields line by three lines of field assignments of the needed fields. Then we can use the SetLoadfields and avoid any kind of JIT-Load.

src/Apps/W1/Subcontracting/App/src/Process/Codeunits/SubcSynchronizeManagement.Codeunit.al
```diff
@@ -136,23 +129,10 @@ codeunit 99001511 "Subc. Synchronize Management"
...
-
-                                TransferHeader.SetCurrentKey("Source ID", "Source Type", "Source Subtype");
-                                TransferHeader.SetRange("Source ID", PurchaseHeader."Buy-from Vendor No.");
-                                TransferHeader.SetRange("Source Type", "Transfer Source Type"::Subcontracting);
-                                TransferHeader.SetRange("Source Subtype", TransferHeader."Source Subtype"::"2");
-                                TransferHeader.SetRange("Subcontr. Purch. Order No.", PurchaseHeader."No.");
-                                if not TransferHeader.IsEmpty() then begin
-                                    TransferHeader.FindFirst();
-                                    ItemLedgerEntry2.SetRange("Order Type", "Inventory Order Type"::Production);
-                                    ItemLedgerEntry2.SetRange("Order No.", ProductionOrder."No.");
```
SPinkow, line 158: If I read this then I am not sure what we want to ensure here. Mabe the originally intention was to check if the transfer order has any item ledger entries yet? I think that we only want to delete the transfer order if nothing has been posted so far.
Using the same filtering as for ItemLedgerEntry Variable makes no sense because the ItemLedgerEntry2.IsEmpty will always be true because it would equal one of the if statements above in this procedure. If we filter on transfer item ledger entries the TransferHeader will only be deleted if there are no transfer postings. We are in the case of switching the vendor no so ifthere are some postings on the wrong vendor the user must have the possibility to finish posting the transfer and it should not be deleted.

src/Apps/W1/Subcontracting/App/src/Process/Codeunits/Extensions/Manufacturing/SubcProdOrderCompExt.Codeunit.al
AleksandricMarko: transferfields replaced

PR #7636, opened 2026-04-10
Fix Bug 626576: AI Development Toolkit silently fails to install with…

martinsrui-msft:
#### Summary
Fixes [Bug 626576]: AI Development Toolkit silently fails to install with no error message provided

Root Cause
The install of the pre-published Microsoft Apps "AI Development Toolkit" fails silently, and leaves no error entry in the installation status page because the original installation status entry that would be added during the publish flow is not there for pre-publish scenarios.

Fix
Add the initial entry for installation status when installing a  pre-published first party app via app source marketplace by calling the newly added `DotNetALNavAppOperationInvoker.CreatePreDeployedTargetReference()`.

Fixes [AB#626576](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/626576)

### DIFF
--- a/src/System Application/App/Extension Management/src/ExtensionInstallationImpl.Codeunit.al
+++ b/src/System Application/App/Extension Management/src/ExtensionInstallationImpl.Codeunit.al
@@ -18,6 +18,8 @@ codeunit 2500 "Extension Installation Impl"
     SingleInstance = false;
 
     var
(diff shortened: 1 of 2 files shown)


martinsrui-msft:
> @microsoft-github-policy-service agree company="Microsoft"

PR #7638, opened 2026-04-10
Problem purchase header security filters bug

FrankNeeckx:
If at least one security filter is applied to the “Purchase Header” table, the user will have a problem when push ‘Add file from source document’ action on the “Email Attachments” subpage. Message: "Did not find any attachments related to this email"

#### Work Item(s) 
Fixes #


darjoo:
Does this have an issue that was opened?

darjoo:
Stale

PR #7649, opened 2026-04-13
Verify Copilot PAT is read-only

Groenbech96:
## Summary
- Adds a temporary verification step to the Documentation Maintenance workflow
- Instructs Copilot CLI to attempt adding a PR comment using `COPILOT_GH_TOKEN`
- If the comment appears, the PAT has write access and needs to be restricted further
- If `gh` returns a permission error, the token is confirmed read-only

## Test plan
- [ ] Trigger the workflow on this PR and check if a "copilot here" comment appears
- [ ] If no comment appears, the PAT is confirmed read-only — remove this step


Groenbech96:
Closing — no longer needed.

PR #7689, opened 2026-04-14
Fix #7622 Shopify product export child-item variant handling

### LINKED ISSUE #7622: [Bug]: Shpfy Product Export creates unwanted variants for items added via "Add Item as Shopify Variant"
### Describe the issue

In codeunit "Shpfy Product Export", procedure UpdateProductData, the local variable Item is loaded from the Shopify product's main item at the start of the procedure but then gets overwritten inside the first loop (which iterates existing Shopify variants) when the loop encounters a variant that was added via "Add Item as Shopify Variant" (i.e. the variant's Item SystemId differs from the product's Item SystemId).

After the first loop completes, the leaked Item value is used by the second loop (which checks for "missing" BC Item Variants and creates them as Shopify variants). The second loop calls ItemVariant.SetRange("Item No.", Item."No.") using the leaked value, so it ends up iterating the child item's variants instead of the parent product's item's variants. Any of those child item BC variants that are not already mapped to a Shopify variant will be auto-created during sync, which is not the user's intent.

### Expected behavior

When an item is added as a variant of a Shopify product via "Add Item as Shopify Variant" and then mapped to a specific BC Item Variant via "Map Variant", subsequent product syncs should not auto-create additional Shopify variants for the child item's other BC Item Variants. Only the variants the user explicitly created or mapped should be present in Shopify.

### Steps to reproduce

Reproduced in a clean Cronus Demo Database with no other extensions installed, only the standard Shopify Connector.

1. Create a parent item (no BC variants required)
2. Create a child item with two BC Item Variants (e.g. CHILDVAR1 and CHILDVAR2)
3. Sync the parent item to Shopify (this creates a Shopify product)
4. Run the standard "Add Item as Shopify Variant" action on the parent product, selecting the child item
5. On the Shopify Variants page, use the standard "Map Variant" action on the new variant to map it to CHILDVAR1 only
6. Run "Sync Products" again
7. Observe that a third Shopify variant has been auto-created for CHILDVAR2, even though it was never explicitly added or mapped by the user

Image

Image

### Additional context

The issue is in the UpdateProductData procedure in src/Apps/W1/Shopify/App/src/Products/Codeunits/ShpfyProductExport.Codeunit.al.

At the start of the procedure, Item.GetBySystemId(ShopifyProduct."Item SystemId") correctly loads the parent product's item. The first loop then iterates Shopify variants and reloads Item for each one via Item.GetBySystemId(ShopifyVariant."Item SystemId"). After the loop, Item is left pointing at whichever item was loaded last, which for products with add-as-variant items will typically be the child item (those records tend to have higher Ids and so are processed last).

The second loop then uses Item."No." to set the filter on ItemVariant, which means it iterates the wrong item's variants and treats any unmapped BC Item Variants as "missing", calling CreateProductVariant for each one.

Suggested fix: reset the Item variable to the product's main item between the two loops, before the ItemVariant.SetRange call:

    Item.GetBySystemId(ShopifyProduct."Item SystemId");
    ItemVariant.SetRange("Item No.", Item."No.");

This restores the correct context for the second loop and prevents auto-creation of variants for unrelated items.

Screenshots from the Cronus reproduction are attached.

Image

Image

tVisionDeb:
Fixes #7622

## Summary
Fixes an issue where product export could continue processing variants on a child item after handling an existing Shopify variant created through "Add Item as Shopify Variant".

## Changes
- Re-fetch the parent item before the item variant loop in `Shpfy Product Export`.
- Add a regression test for the child-item variant scenario.
- Add a test subscriber that mocks Shopify GraphQL responses and verifies that no unexpected variant create call is made.

## Validation
- Validated in a clean local W1 container.
- Passed `Codeunit 139616 Shpfy Product Export 7622 Test`.

Fixes [AB#632757](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/632757)


src/Apps/W1/Shopify/Test/Products/ShpfyProductExportIssue7622Sub.Codeunit.al
```diff
@@ -0,0 +1,115 @@
...
+    local procedure OnGetContent(HttpResponseMessage: HttpResponseMessage; var Response: Text)
+    begin
+        HttpResponseMessage.Content.ReadAs(Response);
+    end;
+
+    [EventSubscriber(ObjectType::Codeunit, Codeunit::"Shpfy Product Events", 'OnBeforeSendAddShopifyProductVariant', '', true, false)]
+    local procedure OnBeforeSendAddShopifyProductVariant(ShopifyShop: Record "Shpfy Shop"; var ShopifyVariant: Record "Shpfy Variant")
+    begin
+        ProductVariantCreateCount += 1;
+    end;
```
onbuyuka: These events are removed. We don't rely on events to mock API calls anymore. Instead we use HttpClientHandler

src/Apps/W1/Shopify/Test/Products/ShpfyProductExportIssue7622Sub.Codeunit.al
```diff
@@ -0,0 +1,115 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Integration.Shopify.Test;
+
+using Microsoft.Integration.Shopify;
+
+codeunit 139614 "Shpfy Product Export 7622 Sub"
```
onbuyuka: Naming is incorrect. Please do not mention issue numbers in object names.

src/Apps/W1/Shopify/Test/Products/ShpfyProductExportIssue7622Sub.Codeunit.al
```diff
@@ -0,0 +1,115 @@
...
+    local procedure GetProductVariantCreateResponse(): HttpResponseMessage
+    var
+        HttpResponseMessage: HttpResponseMessage;
+        Body: Text;
+    begin
+        LastCreatedVariantId += 1;
+        if LastCreatedVariantId = 1 then
+            LastCreatedVariantId := 900001;
+
+        Body := StrSubstNo('{"data":{"productVariantsBulkCreate":{"productVariants":[{"legacyResourceId":%1,"createdAt":"2026-01-01T00:00:00Z","updatedAt":"2026-01-01T00:00:00Z"}],"userErrors":[]}}}', LastCreatedVariantId);
```
onbuyuka: These mock responses are already available under resources as txt files.

src/Apps/W1/Shopify/Test/Products/ShpfyProductExportIssue7622Test.Codeunit.al
```diff
@@ -0,0 +1,127 @@
...
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+
+namespace Microsoft.Integration.Shopify.Test;
+
+using Microsoft.Integration.Shopify;
+using Microsoft.Inventory.Item;
+using System.TestLibraries.Utilities;
+
+codeunit 139616 "Shpfy Product Export 7622 Test"
```
onbuyuka: Naming is incorrect. Please do not mention issue numbers in object names.

src/Apps/W1/Shopify/Test/Products/ShpfyProductExportIssue7622Test.Codeunit.al
```diff
@@ -0,0 +1,127 @@
...
+        // [WHEN] Product export runs for the shop.
+        ProductExportIssue7622Sub.ResetCounters();
+        BindSubscription(ProductExportIssue7622Sub);
+        ProductExport.SetShop(Shop);
+        Shop.SetRange(Code, Shop.Code);
+        ProductExport.Run(Shop);
+        UnbindSubscription(ProductExportIssue7622Sub);
+
+        // [THEN] No new Shopify variant is created for the unmapped child item variant.
+        LibraryAssert.AreEqual(0, ProductExportIssue7622Sub.GetProductVariantCreateCount(), 'Unexpected Shopify variant create call occurred for an unmapped child item variant.');
```
onbuyuka: I don't think it is the best idea to assert on mock calls.

src/Apps/W1/Shopify/Test/Products/ShpfyProductExportIssue7622Test.Codeunit.al
```diff
@@ -0,0 +1,127 @@
...
+        ShopifyVariant."Shop Code" := Shop.Code;
+        ShopifyVariant."Product Id" := ProductId;
+        ShopifyVariant."Item SystemId" := ItemSystemId;
+        ShopifyVariant."Item Variant SystemId" := ItemVariantSystemId;
+        ShopifyVariant."Option 1 Name" := 'Variant';
+        ShopifyVariant."Option 1 Value" := CopyStr(Any.AlphabeticText(10), 1, MaxStrLen(ShopifyVariant."Option 1 Value"));
+        ShopifyVariant.Title := CopyStr(Any.AlphabeticText(20), 1, MaxStrLen(ShopifyVariant.Title));
+        ShopifyVariant.Insert();
+        exit(ShopifyVariant);
+    end;
```
onbuyuka: Would be nice see if these are already implemented in other codeunits and maybe reused.

src/Apps/W1/Shopify/Test/Products/ShpfyProductExportIssue7622Test.Codeunit.al
onbuyuka: Can this codeunit included in ShpfyCreateProductTest? It does not add much value as a standalone codeunit with only one test case.

onbuyuka (changes requested):

src/Apps/W1/Shopify/Test/Products/ShpfyProductExportIssue7622Sub.Codeunit.al
tVisionDeb: Done. Replaced the event subscriber approach with [HttpClientHandler] and TestHttpRequestPolicy = BlockOutboundRequests. The handler is now a local procedure ProductExportChildItemVariantHttpHandler in codeunit 139601.

src/Apps/W1/Shopify/Test/Products/ShpfyProductExportIssue7622Sub.Codeunit.al
tVisionDeb: Fixed. The standalone codeunit has been deleted entirely; the test is now part of codeunit 139601 Shpfy Create Product Test with no issue numbers in any names.

src/Apps/W1/Shopify/Test/Products/ShpfyProductExportIssue7622Sub.Codeunit.al
tVisionDeb: Fixed. The inline JSON strings have been removed. The handler now loads responses from .resources/Products/ProductUpdateResponse.txt and .resources/Products/ProductVariantsBulkUpdateResponse.txt via NavApp.GetResourceAsText.

src/Apps/W1/Shopify/Test/Products/ShpfyProductExportIssue7622Test.Codeunit.al
tVisionDeb: Fixed. The codeunit has been deleted and the test merged into codeunit 139601 Shpfy Create Product Test. No issue numbers appear in any identifiers.

src/Apps/W1/Shopify/Test/Products/ShpfyProductExportIssue7622Test.Codeunit.al
tVisionDeb: Agreed. Removed the mock call count assertion. The test now only asserts on actual database state: it verifies that the Shopify variant record still has the correct ItemSystemId after export, which is the observable outcome that the bug would have broken.

src/Apps/W1/Shopify/Test/Products/ShpfyProductExportIssue7622Test.Codeunit.al
tVisionDeb: Reviewed the existing helpers in ShpfyCreateProductTest and other product test codeunits. The new local procedures (CreateItemVariantForExport, CreateShopifyProductForExport, CreateMappedShopifyVariantForExport) follow the same pattern as the existing CreateItem, CreateShopifyProduct, and CreateShopifyVariant helpers in the same codeunit but set up data specifically for the export flow, so I kept them as separate locals rather than conflating the two setups.

src/Apps/W1/Shopify/App/src/Products/Codeunits/ShpfyProductExport.Codeunit.al
```diff
@@ -651,6 +651,9 @@ codeunit 30178 "Shpfy Product Export"
                                 UpdateProductVariant(ShopifyVariant, Item, ItemVariant, TempCurrVariant);
                         end;
                 until ShopifyVariant.Next() = 0;
+            // Re-fetch the parent product's item, as the loop above may have overwritten
+            // the Item variable with a child item from "Add Item as Shopify Variant".
+            Item.GetBySystemId(ShopifyProduct."Item SystemId");
```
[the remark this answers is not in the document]
tVisionDeb: Fixed. Changed to 'if not Item.GetBySystemId(ShopifyProduct."Item SystemId") then exit;' so the function exits cleanly if the item can no longer be found.

src/Apps/W1/Shopify/Test/Products/ShpfyCreateProductTest.Codeunit.al
```diff
@@ -2958,4 +2965,145 @@ codeunit 139601 "Shpfy Create Product Test"
...
+
+        // [GIVEN] A Shopify product mapped to the parent item and one existing Shopify variant
+        // [GIVEN] mapped to only one child item variant.
+        ShopifyProduct := CreateShopifyProductForExport(ParentItem.SystemId);
+        ShopifyVariant := CreateMappedShopifyVariantForExport(ShopifyProduct.Id, ChildItem.SystemId, ChildItemVariantMapped.SystemId);
+
+        // [WHEN] Product export runs for the shop.
+        ProductExport.SetShop(ExportShop);
+        ExportShop.SetRange(Code, ExportShop.Code);
+        ProductExport.Run(ExportShop);
```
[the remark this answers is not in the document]
tVisionDeb: Fixed. Added OutboundHttpRequests.AssertEmpty() immediately after ProductExport.Run(ExportShop) to verify all expected HTTP calls were consumed.

onbuyuka (dismissed):

onbuyuka:
There are compilation failures on the test app:

  Compilation started for project 'Shopify Connector Test' containing '76' files at '12:27:07.683'.
  
  Error: AL0132 'Codeunit Microsoft.Integration.Shopify."Shpfy Communication Mgt."' does not contain a definition for 'SetTestInProgress'

tVisionDeb:
Fixed and rebased onto main. The compilation error was caused by \SetTestInProgress\ being removed from \Shpfy Communication Mgt.\ in a recent upstream commit. Removed the call and the associated local variable — the \[HttpClientHandler]\ attribute with \TestHttpRequestPolicy = BlockOutboundRequests\ handles HTTP interception without it. Branch rebased onto latest main.

onbuyuka (dismissed):

tVisionDeb:
Fixed. The AA0198 build errors were caused by \ProductInitTest\ being mistakenly included in the codeunit-level \ar\ block (added alongside our new export test variables). Every pre-existing procedure in the file already declares its own local \ProductInitTest\, which triggered the 'identical to a global variable' warning across ~25 procedures. Removed the global declaration — our new test code does not use it.

onbuyuka (approved):

JesperSchulz (approved):

PR #7697, opened 2026-04-15
[Agent] Add SetBillingType to Agent Task Builder

Groenbech96:
## Summary
- Add `SetBillingType(BillingType: Enum "AgentTaskBillingType")` to the public `Agent Task Builder` API
- Wire through `AgentTaskBuilderImpl` and `AgentTaskImpl.CreateTask` following the existing `SetExternalId` pattern
- Enables callers to set billing type (Default/Exclude) at task creation time, unblocking PayablesAgent trial billing support

[AB#634133](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/634133)


src/System Application/App/Agent/Interaction/Internal/AgentTaskBuilderImpl.Codeunit.al
```diff
@@ -20,6 +20,7 @@ codeunit 4310 "Agent Task Builder Impl."
         GlobalAgentUserSecurityId: Guid;
         GlobalTaskTitle: Text[150];
         GlobalExternalID: Text[2048];
+        GlobalBillingType: Enum "Agent Task Billing Type";
```
qutreson: Minor: Should we set the default in Initialize to be explicit about it?

qutreson (dismissed):

src/System Application/App/Agent/Interaction/AgentTaskBuilder.Codeunit.al
```diff
@@ -93,6 +93,19 @@ codeunit 4315 "Agent Task Builder"
         exit(this);
     end;
 
+    /// <summary>
+    /// Set the billing context for the task.
+    /// </summary>
+    /// <param name="BillingContext">The billing context to set on the task.</param>
+    /// <returns>This instance of the Agent Task Builder.</returns>
+    [Scope('OnPrem')]
+    procedure SetBillingContext(BillingContext: Enum "Agent Task Billing Context"): codeunit "Agent Task Builder"
```
nikolakukrika, line 102: Instead of setting the billing context, is it better to mark the task as free trial and etc?

nikolakukrika: It will be easier to understand and find if we are using concepts instead of the enum

nikolakukrika (approved):

ventselartur (approved):

PR #7735, opened 2026-04-16
[Subcontracting] Bug 631232: Change position of "purchase with service" lines with "suctontracting…

SPinkow:
* 
#### Summary 
This pull request introduces a new automated test for subcontracting purchase orders with backward flushing and includes related utility improvements and minor refactoring in the manufacturing and subcontracting modules.

### New test scenario for backward flushing in subcontracting

* Added the `PostSubcontPurchOrder_PurchWithService_BackwardFlush` test to `SubcPurchSubcontTest.Codeunit.al`, which verifies that when posting a subcontracting purchase order for a BOM component with Subcontracting Type "Purchase with Service" and Flushing Method "Backward," the component is both received and consumed correctly, resulting in zero net inventory.

* Introduced the `EnsureGeneralPostingSetupIsValid` local procedure to ensure that the required general posting setup exists and is unblocked before posting purchase documents in tests.

### Refactoring and event handling

* Refactored the handling of subcontracting purchase order line insertion in `SubcReqWkshMakeOrd.Codeunit.al` by removing the `NextLineNo` parameter from `HandleSubcontractingAfterPurchOrderLineInsert` and adding a new event subscriber for `OnInsertPurchOrderLineOnAfterCheckInsertFinalizePurchaseOrderHeader` to ensure proper transfer of subcontracting production order components.
* 
#### Work Item(s) 
Fixes [AB#631232](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/631232)

### DIFF
--- a/src/Apps/W1/Subcontracting/App/src/Process/Codeunits/Extensions/Manufacturing/SubcReqWkshMakeOrd.Codeunit.al
+++ b/src/Apps/W1/Subcontracting/App/src/Process/Codeunits/Extensions/Manufacturing/SubcReqWkshMakeOrd.Codeunit.al
@@ -12,14 +12,24 @@ codeunit 99001516 "Subc. Req. Wksh. Make Ord."
     [EventSubscriber(ObjectType::Codeunit, Codeunit::"Req. Wksh.-Make Order", OnAfterInsertPurchOrderLine, '', false, false)]
     local procedure OnAfterInsertPurchOrderLine(var PurchOrderLine: Record "Purchase Line"; var NextLineNo: Integer; var RequisitionLine: Record "Requisition Line")
     begin
-        HandleSubcontractingAfterPurchOrderLineInsert(PurchOrderLine, NextLineNo, RequisitionLine);
+        HandleSubcontractingAfterPurchOrderLineInsert(PurchOrderLine, RequisitionLine);
     end;
 
-    local procedure HandleSubcontractingAfterPurchOrderLineInsert(var PurchaseLine: Record "Purchase Line"; var NextLineNo: Integer; var RequisitionLine: Record "Requisition Line")
+    [EventSubscriber(ObjectType::Codeunit, Codeunit::"Req. Wksh.-Make Order", OnInsertPurchOrderLineOnAfterCheckInsertFinalizePurchaseOrderHeader, '', false, false)]
+    local procedure OnInsertPurchOrderLineOnAfterCheckInsertFinalizePurchaseOrderHeader(var RequisitionLine: Record "Requisition Line"; var PurchaseHeader: Record "Purchase Header"; var NextLineNo: Integer)
+    var
+        AdditionallyPurchaseLine: Record "Purchase Line";
+        SubcPurchaseOrderCreator: Codeunit "Subc. Purchase Order Creator";
+    begin
+        AdditionallyPurchaseLine."Document Type" := PurchaseHeader."Document Type";
+        AdditionallyPurchaseLine."Document No." := PurchaseHeader."No.";
+        SubcPurchaseOrderCreator.TransferSubcontractingProdOrderComp(AdditionallyPurchaseLine, RequisitionLine, NextLineNo);
+    end;
+
+    local procedure HandleSubcontractingAfterPurchOrderLineInsert(var PurchaseLine: Record "Purchase Line"; var RequisitionLine: Record "Requisition Line")
     var
         SubcPurchaseOrderCreator: Codeunit "Subc. Purchase Order Creator";
     begin
         SubcPurchaseOrderCreator.InsertProdDescriptionOnAfterInsertPurchOrderLine(PurchaseLine, RequisitionLine);
-        SubcPurchaseOrderCreator.TransferSubcontractingProdOrderComp(PurchaseLine, RequisitionLine, NextLineNo);
     end;
 }
--- a/src/Apps/W1/Subcontracting/App/src/Process/Codeunits/Extensions/Manufacturing/SubcReqWkshMakeOrd.Codeunit.al
+++ b/src/Apps/W1/Subcontracting/App/src/Process/Codeunits/Extensions/Manufacturing/SubcReqWkshMakeOrd.Codeunit.al
@@ -18,12 +18,14 @@ codeunit 99001516 "Subc. Req. Wksh. Make Ord."
     [EventSubscriber(ObjectType::Codeunit, Codeunit::"Req. Wksh.-Make Order", OnInsertPurchOrderLineOnAfterCheckInsertFinalizePurchaseOrderHeader, '', false, false)]
     local procedure OnInsertPurchOrderLineOnAfterCheckInsertFinalizePurchaseOrderHeader(var RequisitionLine: Record "Requisition Line"; var PurchaseHeader: Record "Purchase Header"; var NextLineNo: Integer)
     var
-        AdditionallyPurchaseLine: Record "Purchase Line";
+        PurchaseLineWithService: Record "Purchase Line";
         SubcPurchaseOrderCreator: Codeunit "Subc. Purchase Order Creator";
     begin
-        AdditionallyPurchaseLine."Document Type" := PurchaseHeader."Document Type";
-        AdditionallyPurchaseLine."Document No." := PurchaseHeader."No.";
-        SubcPurchaseOrderCreator.TransferSubcontractingProdOrderComp(AdditionallyPurchaseLine, RequisitionLine, NextLineNo);
+        if RequisitionLine."Prod. Order No." = '' then
+            exit;
+        PurchaseLineWithService."Document Type" := PurchaseHeader."Document Type";
+        PurchaseLineWithService."Document No." := PurchaseHeader."No.";
+        SubcPurchaseOrderCreator.TransferSubcontractingProdOrderComp(PurchaseLineWithService, RequisitionLine, NextLineNo);
     end;
 
     local procedure HandleSubcontractingAfterPurchOrderLineInsert(var PurchaseLine: Record "Purchase Line"; var RequisitionLine: Record "Requisition Line")
--- a/src/Apps/W1/Subcontracting/Test/src/Codeunits/Tests/SubcPurchSubcontTest.Codeunit.al
+++ b/src/Apps/W1/Subcontracting/Test/src/Codeunits/Tests/SubcPurchSubcontTest.Codeunit.al
@@ -378,7 +378,7 @@ codeunit 139991 "Subc. Purch. Subcont. Test"
         LibraryManufacturing.CreateProductionBOMLine(
             ProductionBOMHeader, ProductionBOMLine, '', ProductionBOMLine.Type::Item, ComponentItem."No.", 1);
         ProductionBOMLine.Validate("Routing Link Code", RoutingLink.Code);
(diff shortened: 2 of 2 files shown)


ChethanT:
## Analysis

**Correctness:**

The `AdditionallyPurchaseLine` local variable is initialized with only `Document Type` and `Document No.` from `PurchaseHeader` — exactly what `TransferSubcontractingProdOrderComp` needs to call `PurchaseHeader.Get(...)` at line 110. Inside the loop, `InitPurchOrderLine` calls `PurchaseLine.Init()` first, clearing all fields and starting fresh for each component line, so the partially-initialized carrier record causes no issue.

`NextLineNo` is passed `by var` through both the event signature and into `TransferSubcontractingProdOrderComp`, so line number increments from component lines correctly propagate back to `InsertPurchOrderLine`, where `InitPurchOrderLine` then advances by another 10000 for the finished-good line.

For non-subcontracting requisition lines, `TransferSubcontractingProdOrderComp` exits early after `ProdOrderRoutingLine.Get(...)` fails (since `RequisitionLine."Prod. Order No."` is empty). No functional change for non-subcontracting flows.

**Side Effects:**

`InsertProdDescriptionOnAfterInsertPurchOrderLine` (the description line insertion) is still called from `OnAfterInsertPurchOrderLine` — unchanged. `GetLineNoBeforeInsertedLineNo` computes a slot between the line preceding the finished-good line and the finished-good line itself. With the new ordering, the slot falls between the last component line and the finished-good line. This is correct and expected.

No `Commit()` / `Error()` anti-patterns introduced. No dead code created.

**Risk Assessment: Low**

The change is narrowly scoped to the `Req. Wksh.-Make Order` flow. Only subcontracting requisition lines (with non-empty `Prod. Order No.`) produce any behavioral change. The `NextLineNo` mechanism is the same `var` propagation pattern already used throughout `InsertPurchOrderLine`. Non-subcontracting purchase planning is unaffected.

*What if we don't make this change?* Users creating subcontracting purchase orders where BOM components have `Subcontracting Type = Purchase` and `Flushing Method = Backward` will see no consumption ILE after posting receipt, leaving component inventory non-zero. There is no simple workaround other than manually creating a consumption journal entry.

**Test Coverage:**

The test `PostSubcontPurchOrder_PurchWithService_BackwardFlush` covers the bug scenario end-to-end and is well-structured. However, two assertions are weaker than they should be:

1. **Purchase ILE** — only checks `RecordIsNotEmpty`, not quantity. Should assert `ItemLedgerEntry.Quantity = Qty`.
2. **Consumption ILE** — only checks `RecordIsNotEmpty`. Should assert `ItemLedgerEntry.Quantity = -Qty` (negative, as consumption is an outgoing entry).

Additionally, `ProductionBOMLine."Subcontracting Type" := "Subcontracting Type"::Purchase;` is a direct field assignment without `Validate`. While common in test setup code, `Validate` would ensure any dependent business logic fires.

---

## Recommendation: Accept with Suggestions

1. **Strengthen consumption and purchase ILE quantity assertions** — replace `RecordIsNotEmpty` checks with exact quantity assertions. After `ItemLedgerEntry.FindFirst()` for the purchase entry, add `ItemLedgerEntry.TestField(Quantity, Qty)`. For the consumption entry: `ItemLedgerEntry.FindFirst()` + `ItemLedgerEntry.TestField(Quantity, -Qty)`. This ensures the test would catch partial or zero-quantity entries.

2. **Use `Validate` for `Subcontracting Type`** — change `ProductionBOMLine."Subcontracting Type" := "Subcontracting Type"::Purchase;` to `ProductionBOMLine.Validate("Subcontracting Type", "Subcontracting Type"::Purchase);` to avoid bypassing field validation logic.

3. **Rename `AdditionallyPurchaseLine`** — the variable name is non-standard. Use `SubcPurchaseLine` or `PurchLineForSubcComp` for clarity.

4. **Consider early exit in new subscriber** — add `if RequisitionLine."Prod. Order No." = '' then exit;` at the top of `OnInsertPurchOrderLineOnAfterCheckInsertFinalizePurchaseOrderHeader` to avoid calling `TransferSubcontractingProdOrderComp` for non-subcontracting lines.

🤖 Review by [Claude Code](https://claude.ai/code)

ChethanT:
## Quality Findings

### 🔵 Low — Style

**`SubcReqWkshMakeOrd.Codeunit.al` line 21**

`AdditionallyPurchaseLine` does not follow BC naming conventions — variable names should reflect the AL object's purpose. "Additionally" is not a BC-style prefix and does not describe the variable's role (a Purchase Line used to carry the PO header key into `TransferSubcontractingProdOrderComp`).

**Fix**: Rename to `SubcPurchaseLine` or `PurchLineForSubcComp`.

```al
        SubcPurchaseLine: Record "Purchase Line";
```

---

### 🔵 Low — Performance

**`SubcReqWkshMakeOrd.Codeunit.al` line 18**

`OnInsertPurchOrderLineOnAfterCheckInsertFinalizePurchaseOrderHeader` fires for **every** requisition line processed through `Req. Wksh.-Make Order`, not just subcontracting lines. For non-subcontracting lines, `TransferSubcontractingProdOrderComp` performs an unnecessary `ProdOrderRoutingLine.Get()` before exiting. Impact is negligible in practice (this is an admin/planning operation), but an early-exit guard in the subscriber would avoid the redundant DB lookup.

**Fix**: Add an early exit at the top of the subscriber:

```al
    local procedure OnInsertPurchOrderLineOnAfterCheckInsertFinalizePurchaseOrderHeader(...)
    begin
        if RequisitionLine."Prod. Order No." = '' then
            exit;
        ...
    end;
```

🤖 Review by [Claude Code](https://claude.ai/code)

ChethanT (dismissed):
Please review the comments above from our agent.

AleksandricMarko (dismissed):

alexei-dobriansky:
### Analysis

**Correctness:**

- **Timing is verified against BaseApp master** (`app/layers/w1/baseapp/Inventory/Requisition/ReqWkshMakeOrder.Codeunit.al`): the new event `OnInsertPurchOrderLineOnAfterCheckInsertFinalizePurchaseOrderHeader` already exists (declared at line 1718, fired at 669), and fires precisely between `NextLineNo := 0` (line 662, on header finalize) and BaseApp's `InitPurchOrderLine` (line 678, which does `NextLineNo := NextLineNo + 10000` at line 574). The fix exploits the right hook.
- **Carrier safety:** `TransferSubcontractingProdOrderComp` reads only `Document Type` / `Document No.` from the carrier `AdditionallyPurchaseLine` (`SubcPurchaseOrderCreator.Codeunit.al:110`), and `InitPurchOrderLine` calls `PurchaseLine.Init()` (line 357) before each component insert. No stale-field leakage.
- **`NextLineNo` propagation** is by-`var` end-to-end and verified in BaseApp (`InitPurchOrderLine` increments `NextLineNo += 10000` at line 574 *after* the new event has fired).
- **Multi-line PO behavior:** if a single PO carries two subcontracting requisition lines, the second invocation of `InsertPurchOrderLine` re-enters without finalizing a new header (so `NextLineNo` is not re-zeroed). The new event still fires; components get `NextLineNo + 10000…`, finished-good gets the next slot. Per-requisition-line ordering is preserved.
- **Non-subcontracting requisition lines:** the new subscriber fires unconditionally, but `TransferSubcontractingProdOrderComp` exits early at line 104 (`ProdOrderRoutingLine.Get` returns false for empty `Prod. Order No.`). `NextLineNo` is unchanged. Functionally inert.
- **Description line:** `InsertProdDescriptionOnAfterInsertPurchOrderLine` still runs from `OnAfterInsertPurchOrderLine` and uses `GetLineNoBeforeInsertedLineNo` to find a slot just below the finished-good line. With components inserted *before* the finished-good line, the description still ends up immediately above the finished-good line — same user-visible behavior.

**Side Effects:**

- No public API change. `HandleSubcontractingAfterPurchOrderLineInsert` is `local`.
- One extra DB lookup per non-subcontracting requisition line via the new subscriber. Subcontracting MO planning is not throughput-critical — impact negligible.
- No `Commit()` / `Error()` anti-patterns. No code becomes dead.
- Event subscriber compatibility: existing partner subscribers on `OnAfterInsertPurchOrderLine` still receive the same signature; that event is unchanged.

**Risk Assessment: Low.**

- Narrow scope: one subscriber relocated, one new subscriber added, one helper signature simplified.
- No costing/posting logic touched — only line *ordering* changes. End ILE state for finished-good output and component consumption is identical; their order changes so backward flushing can consume already-received components.
- *What if we don't ship?* Customers on subcontracting + Purchase-with-Service + Backward flushing must post in two manual steps. The single-step path is broken. Workaround exists but is workflow-disruptive.

**Test Coverage:**

`PostSubcontPurchOrder_PurchWithService_BackwardFlush` exercises the ADO repro end-to-end via `ReleasedProdOrderRtng.CreateSubcontracting.Invoke()` and `LibraryPurchase.PostPurchaseDocument`, then asserts: positive output ILE for finished good with `Quantity = Qty`, Purchase ILE present for component, Consumption ILE present for component, `ComponentItem.Inventory = 0`. The test would fail without the fix.

Gaps worth addressing:

1. **No direct assertion on PO line ordering.** The test verifies a downstream effect (consumption ILE present, inventory zero). The literal invariant the fix establishes is `ComponentLine."Line No." < FinishedGoodLine."Line No."`. Asserting this directly — between `CreateSubcontracting.Invoke()` and `PostPurchaseDocument` — would tie the test to the fix point, not to BC's backward-flushing semantics. If a future BC change tweaks how backward flushing decides availability, this test could pass for the wrong reason.
2. **Weak ILE quantity assertions** (already raised by prior automated reviews) — Purchase / Consumption ILE checks should be `Quantity = Qty` / `Quantity = -Qty`, not just `RecordIsNotEmpty`.
3. **Direct field assignment** for `ProductionBOMLine."Subcontracting Type"` — should be `Validate` (already raised).
4. **`Vendor.Modify()`** without `(true)` — bypasses validation in test setup. Minor.
5. **`EnsureGeneralPostingSetupIsValid`** persists changes (un-blocks posting setups; inserts new ones via `SuggestSetupAccounts`) without rollback. Fine for one test; risk of test pollution if more tests adopt it. Existing `LibraryERM` helpers may be preferable.

### Recommendation: **Accept with Suggestions**

The fix is correct, timing-verified against BaseApp, narrowly targeted, and the new test pins the bug end-to-end. New findings on top of what prior reviews already noted:

1. **Assert PO line ordering directly** — after `CreateSubcontracting.Invoke()` and before `PostPurchaseDocument`, look up the finished-good and component `Purchase Line` records and assert `ComponentLine."Line No." < FinishedGoodLine."Line No."`. This tests the exact invariant the fix establishes, independent of backward-flushing semantics.
2. **(Concur with prior reviews)** Strengthen ILE quantity assertions; use `Validate("Subcontracting Type", ...)`; rename `AdditionallyPurchaseLine` (e.g., `SubcPurchaseLine`); consider an early-exit guard `if RequisitionLine."Prod. Order No." = '' then exit;` at the top of the new subscriber so non-subcontracting requisition lines short-circuit cleanly.
3. **Test hygiene (minor)** — `Vendor.Modify(true)`; reconsider `EnsureGeneralPostingSetupIsValid` vs existing `LibraryERM` helpers.

SPinkow:
@ChethanT

AleksandricMarko (approved):

ChethanT (approved):

PR #7755, opened 2026-04-20
[Retention Policy] Handle indirect permissions for table truncation

Jiawen-CS:
#### Summary 
When a user has only indirect delete permissions on a table (e.g., "Retention Policy Log Entry"), the `RecRef.Truncate(true)` call in TruncateTableRecords fails because it requires direct permissions. This PR adds an integration event pattern to delegate truncation to a subscriber that holds the required indirect permissions.

#### Details
- Added OnTruncateRecordsIndirectPermissionRequired integration event to Apply Retention Policy codeunit
- Added new codeunit 3921 Retention Policy Logs Truncate that subscribes to the event and truncates the "Retention Policy Log Entry" table with indirect permissions
- Modified TruncateTableRecords in RetentionPolicySetupImpl to check effective delete permissions and raise the event when permissions are indirect instead of truncating directly
- Updated object ID range in [app.json] from 3920 to 3921

### System Application / Retention Policy
**Files:** `ApplyRetentionPolicy.Codeunit.al`, RetentionPolicyLogsTruncate.Codeunit.al, `RetentionPolicySetupImpl.Codeunit.al`, app.json
**Guidance:** Verify the integration event pattern is correct, the indirect permission check logic is sound, and the new codeunit properly handles the `Handled` pattern.

#### Work Item(s) 
Fixes #[AB#631806](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/631806/)


src/System Application/App/Retention Policy/src/Apply Retention Policy/ApplyRetentionPolicy.Codeunit.al
```diff
@@ -138,4 +138,17 @@ codeunit 3910 "Apply Retention Policy"
...
     end;
 #pragma warning restore
+
+    /// <summary>
+    /// This event is raised when the user truncating the table has indirect permissions to delete records in the table.
+    /// A subscriber to this event with indirect permissions can truncate the records on behalf of the user.
+    /// </summary>
+    /// <param name="RecRef">The record reference for the table to be truncated.</param>
+    /// <param name="Handled">Indicates whether the event has been handled.</param>
+#pragma warning disable AA0072
```
haoranpb, line 159: Is the pragma here really needed?

src/System Application/App/Retention Policy/src/Retention Policy Setup/RetentionPolicySetupImpl.Codeunit.al
```diff
@@ -477,10 +480,14 @@ codeunit 3903 "Retention Policy Setup Impl."
 
     procedure TruncateTableRecords(RetentionPolicySetup: Record "Retention Policy Setup")
```
haoranpb: Let's move the logic as discussed, don't think this belongs in Retention Policy Setup.

haoranpb (changes requested):

src/System Application/App/Retention Policy/src/Apply Retention Policy/ApplyRetentionPolicy.Codeunit.al
Jiawen-CS: Right here follow the `OnApplyRetentionPolicyIndirectPermissionRequired` which uses pragma.

darjoo (approved):

Jiawen-CS:
@aholstrup1 Can you review this PR? Thanks in advance!

src/System Application/App/Retention Policy/src/Apply Retention Policy/ApplyRetentionPolicy.Codeunit.al
```diff
@@ -138,4 +149,17 @@ codeunit 3910 "Apply Retention Policy"
...
+    [IntegrationEvent(false, false)]
```
pri-kise, line 160: I understand that this is the fast way, but wasn't the Record Reference Module introduced to handle such cases?

I know that the current interface version does not support Truncate.
https://github.com/microsoft/BCApps/blob/main/src/System%20Application/App/Record%20Reference/src/RecordReference.Codeunit.al
https://github.com/microsoft/BCApps/blob/main/src/System%20Application/App/Record%20Reference/src/RecordReference.Interface.al

But maybe this can be improved.

e.g. Record Reference interface is currently used here for normal deletion:
https://github.com/microsoft/BCApps/blob/main/src/System%20Application/App/Retention%20Policy/src/Apply%20Retention%20Policy/RetenPolDeleteImpl.Codeunit.al

Jiawen-CS: Thanks for advising. Right now, Truncate function in BC is not totally the same as SQL's Truncate, so here we don't consider about this. But for the further developing, we would discuss with Platform teams about it.

aholstrup1 (approved):

haoranpb (approved):

PR #7776, opened 2026-04-21
Subscription Billing: improve performance for large datasets

### LINKED ISSUE #7690: [Bug][SubscriptionBilling]: Performance: Billing document creation does not scale to large datasets (4M+ subscription lines)
### Describe the issue

Subscription Billing was tested at a large customer with ~4,000,000 Subscription Lines. The billing run (proposal + document creation) took an unacceptable amount of time. The following issues were identified:

### 1. Missing database indexes

Key tables lack covering indexes for queries executed during billing: Billing Line, Cust. Sub. Contract Line, Usage Data Billing, Subscription Line, Usage Data Generic Import.

### 2. No progress monitoring

There is no progress dialog during any long-running step. At scale each step can run for hours with no indication of progress. Affected steps:

- **Billing Proposal creation**
- **Billing Document creation**
- **Usage Data Import processing**

### 3. Unnecessary validation chains during line creation

`Validate("No.")`, `Validate("Unit of Measure Code")`, and `Validate("Unit Price")` trigger the full pricing engine, cost engine, availability checks, and aggregator recalculations on every line - even though all values are already known from the Billing Line.

### 4. Redundant database round-trips

- Line numbering via `FindLast()` instead of in-memory counters
- Per-line `Get()` for master data (contracts, service objects) instead of caching
- O(n²) `CheckServiceCommitmentDataConsistency` pre-check

### 5. No per-document Commit()

The entire run is a single transaction. A failure after hours of processing loses all work.

### 6. No Job Queue support

The billing path does not guard dialogs with `GuiAllowed`, preventing Job Queue execution.

### Expected behavior

- Covering indexes on all billing-related tables
- Progress dialogs (elapsed time, throughput, ETA) on all long-running steps
- Direct field assignment instead of full validation chains when values are pre-determined
- Master data caching and in-memory line numbering to eliminate redundant DB reads
- Per-document `Commit()` for recoverability
- Job Queue support with `GuiAllowed` guards

### Steps to reproduce

1. Set up an environment with ~4,000,000 Subscription Lines across multiple Customer Contracts.
2. Open the **Recurring Billing** page and run **Create Billing Proposal**.
3. Observe that no progress dialog is shown. The process runs for hours with no feedback and may time out.
4. After the billing proposal finishes, run **Create Billing Documents** from the same page.
5. Observe that again no progress dialog is shown. If the process times out or fails, a complete rollback occurs and no documents are preserved.

### Additional context

_No response_

jeffreybulanadi:
Fixes #7690

## Summary

Subscription Billing document creation is unscalable at 4M+ subscription lines due to missing indexes, crashes in Job Queue, an O(n2) query pattern, redundant database reads, and no per-document transaction boundary. This PR addresses all root causes identified in the issue.

## Root Causes Fixed

**1. Missing database indexes**
- `BillingLine`: Added key `SK7` on `(Partner, Subscription Contract No., Subscription Contract Line No., Document Type, Document No.)` covering `GetBillingLineNo()` on every line insertion
- `SubscriptionLine`: Added key `Key4` on `(Partner, Subscription Contract No., Next Billing Date)` covering `ProcessContractServiceCommitments()`
- `UsageDataBilling`: Added key `key2` on `(Partner, Subscription Contract No., Subscription Contract Line No., Document Type, Document No.)` mirroring the billing line lookup pattern
- `UsageDataGenericImport`: Added key `key2` on `(Usage Data Import Entry No., Processing Status)` covering status-filtered scans during import

**2. Job Queue crash (GuiAllowed = false)**
- `BillingProposal`: `ProposalWindow.Open/Update/Close` now guarded with `if GuiAllowed`. The `Count()` call that feeds the progress dialog is also skipped when `GuiAllowed = false`.
- `CreateBillingDocuments`: `Window.Open/Update/Close` already had an `AutomatedBilling` guard but was missing `GuiAllowed` checks. Both guards are now applied. `ProcessingFinishedMessage` exits early when `not GuiAllowed`.

**3. Redundant GetBillingLineNo() calls inside inner loop**
- In `InsertSalesLineFromTempBillingLine` and `InsertPurchaseLineFromTempBillingLine`, `GetBillingLineNo()` was called inside a `UsageDataBilling.FindSet()` loop with constant parameters. Hoisted outside the loop.

**4. Redundant contract Get() on every billing line**
- `CreateTempBillingLines` called `CustomerContract.Get()` and `VendorContract.Get()` for every billing line. Added cache variables `CachedCustomerContractNo` and `CachedVendorContractNo` to skip the read when the contract has not changed.

**5. Single transaction for entire batch**
- `CreateSalesDocumentsPerContract`, `CreatePurchaseDocumentsPerContract`, `CreateSalesDocumentsPerCustomer`, and `CreatePurchaseDocumentsPerVendor` each call `Commit()` after completing each document. This limits the transaction scope to one document at a time, preventing lock escalation and enabling partial-batch recovery.

**6. Progress visibility in BillingProposal**
- Added a `Dialog` progress window in `CreateBillingProposal` so users see per-contract progress during long proposal creation runs. Includes contract number, partner number, and a processed/total counter.

## Files Changed

- `src/Apps/W1/Subscription Billing/App/Billing/Tables/BillingLine.Table.al` -- added SK7
- `src/Apps/W1/Subscription Billing/App/Service Commitments/Tables/SubscriptionLine.Table.al` -- added Key4
- `src/Apps/W1/Subscription Billing/App/Usage Based Billing/Tables/UsageDataBilling.Table.al` -- added key2
- `src/Apps/W1/Subscription Billing/App/Usage Based Billing/Tables/UsageDataGenericImport.Table.al` -- added key2
- `src/Apps/W1/Subscription Billing/App/Billing/Codeunits/BillingProposal.Codeunit.al` -- progress dialog + GuiAllowed guards
- `src/Apps/W1/Subscription Billing/App/Billing/Codeunits/CreateBillingDocuments.Codeunit.al` -- GuiAllowed guards, Commit() per document, hoisted GetBillingLineNo(), cached contract Get()

## How to Test

1. Set up a BC environment with a large number of Customer and Vendor Subscription Contracts (10,000+ lines recommended).
2. Create a Billing Template and run recurring billing to generate a billing proposal -- verify the progress dialog shows contract progress.
3. Run "Create Billing Documents" -- verify documents are created without error and the progress dialog updates correctly.
4. Schedule the billing template via Job Queue (AutomatedBilling = true) -- verify the run completes without a "Dialog not allowed" error.
5. With 4M+ subscription lines, verify that document creation completes without lock timeout or transaction size errors.
6. After an interrupted run, verify that already-committed documents are present and no partial data loss occurred.

Fixes [AB#634052](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/634052)

### DIFF
--- a/src/Apps/W1/Subscription Billing/App/Billing/Tables/BillingLine.Table.al
+++ b/src/Apps/W1/Subscription Billing/App/Billing/Tables/BillingLine.Table.al
@@ -243,6 +243,9 @@ table 8061 "Billing Line"
         key(SK6; "Billing Template Code", Partner)
         {
         }
+        key(SK7; Partner, "Subscription Contract No.", "Subscription Contract Line No.", "Document Type", "Document No.")
+        {
+        }
     }
 
     trigger OnDelete()
--- a/src/Apps/W1/Subscription Billing/App/Billing/Tables/BillingLine.Table.al
+++ b/src/Apps/W1/Subscription Billing/App/Billing/Tables/BillingLine.Table.al
@@ -243,7 +243,7 @@ table 8061 "Billing Line"
         key(SK6; "Billing Template Code", Partner)
         {
         }
-        key(SK7; Partner, "Subscription Contract No.", "Subscription Contract Line No.", "Document Type", "Document No.")
+        key(SK7; "Subscription Contract No.", "Subscription Contract Line No.", "Document No.", "Document Type", Partner)
         {
         }
     }
--- a/src/Apps/W1/Subscription Billing/App/Service Commitments/Tables/SubscriptionLine.Table.al
+++ b/src/Apps/W1/Subscription Billing/App/Service Commitments/Tables/SubscriptionLine.Table.al
@@ -611,7 +611,7 @@ table 8059 "Subscription Line"
         {
 
         }
-        key(Key4; Partner, "Subscription Contract No.", "Next Billing Date")
+        key(Key4; "Subscription Contract No.", "Next Billing Date", Partner)
         {
         }
     }
--- a/src/Apps/W1/Subscription Billing/App/Service Commitments/Tables/SubscriptionLine.Table.al
+++ b/src/Apps/W1/Subscription Billing/App/Service Commitments/Tables/SubscriptionLine.Table.al
@@ -611,6 +611,9 @@ table 8059 "Subscription Line"
         {
 
         }
+        key(Key4; Partner, "Subscription Contract No.", "Next Billing Date")
+        {
+        }
     }
 
     trigger OnInsert()
--- a/src/Apps/W1/Subscription Billing/App/Usage Based Billing/Tables/UsageDataBilling.Table.al
+++ b/src/Apps/W1/Subscription Billing/App/Usage Based Billing/Tables/UsageDataBilling.Table.al
@@ -300,6 +300,9 @@ table 8006 "Usage Data Billing"
             SumIndexFields = Quantity, Amount;
             MaintainSiftIndex = true;
         }
+        key(key2; Partner, "Subscription Contract No.", "Subscription Contract Line No.", "Document Type", "Document No.")
+        {
+        }
 
     }
     trigger OnInsert()
--- a/src/Apps/W1/Subscription Billing/App/Usage Based Billing/Tables/UsageDataBilling.Table.al
+++ b/src/Apps/W1/Subscription Billing/App/Usage Based Billing/Tables/UsageDataBilling.Table.al
@@ -300,7 +300,10 @@ table 8006 "Usage Data Billing"
             SumIndexFields = Quantity, Amount;
             MaintainSiftIndex = true;
         }
-        key(key2; Partner, "Subscription Contract No.", "Subscription Contract Line No.", "Document Type", "Document No.")
+        key(key2; "Subscription Contract No.", "Subscription Contract Line No.", "Document No.", "Document Type", Partner)
+        {
+        }
+        key(key3; "Usage Data Import Entry No.", "Document No.")
         {
         }
 
--- a/src/Apps/W1/Subscription Billing/App/Billing/Codeunits/BillingProposal.Codeunit.al
+++ b/src/Apps/W1/Subscription Billing/App/Billing/Codeunits/BillingProposal.Codeunit.al
@@ -26,6 +26,7 @@ codeunit 8062 "Billing Proposal"
         PurchaseCreditMemoExistsForBillingLineQst: Label 'There is a purchase credit memo that needs to be posted before an invoice can be created. Do you want to open the credit memo?';
         BillingLinesForAllContractLinesExistsErr: Label 'There are billing lines for all contract lines. For contract lines with billing lines, the invoice must be created in recurring billing.';
         NotAuthorizedToClearOrDeleteDocumentErr: Label 'You are not authorized to clear billing templates or delete billing documents. To perform these actions, you must be set up as an Auto Contract Billing user in the User Setup.';
+        BillingProposalProgressTxt: Label 'Creating Billing Proposal...\Contract No. #1#################################\Partner No. #2#################################\Processed: #3########## / #4##########', Comment = '%1=Contract No., %2=Partner No., %3=Contracts processed, %4=Total contracts';
 
     procedure InitTempTable(var TempBillingLine: Record "Billing Line" temporary; GroupBy: Enum "Contract Billing Grouping")
     var
@@ -154,8 +155,11 @@ codeunit 8062 "Billing Proposal"
         BillingTemplate: Record "Billing Template";
         CustomerContract: Record "Customer Subscription Contract";
         VendorContract: Record "Vendor Subscription Contract";
+        ProposalWindow: Dialog;
         FilterText: Text;
         BillingRhythmFilterText: Text;
+        TotalContractsCount: Integer;
+        ContractCounter: Integer;
     begin
         SalesHeaderGlobal.Reset();
         BillingTemplate.Get(BillingTemplateCode);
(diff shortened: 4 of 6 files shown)


src/Apps/W1/Subscription Billing/App/Billing/Tables/BillingLine.Table.al
```diff
@@ -243,6 +243,9 @@ table 8061 "Billing Line"
         key(SK6; "Billing Template Code", Partner)
         {
         }
+        key(SK7; Partner, "Subscription Contract No.", "Subscription Contract Line No.", "Document Type", "Document No.")
+        {
+        }
```
miljance: Where does these Key kicks in? Partner is an Enum field and thus has low cardinality. If filtered on all fields from this key, I would consider either moving this field after other fields with high cardinality or removing the low cardinality fields from the key completely.

src/Apps/W1/Subscription Billing/App/Service Commitments/Tables/SubscriptionLine.Table.al
```diff
@@ -611,6 +611,9 @@ table 8059 "Subscription Line"
         {
 
         }
+        key(Key4; Partner, "Subscription Contract No.", "Next Billing Date")
+        {
+        }
```
miljance: Suggestion to consider skipping low cardinality fields in first place of the key.

Image

src/Apps/W1/Subscription Billing/App/Usage Based Billing/Tables/UsageDataBilling.Table.al
```diff
@@ -300,6 +300,9 @@ table 8006 "Usage Data Billing"
             SumIndexFields = Quantity, Amount;
             MaintainSiftIndex = true;
         }
+        key(key2; Partner, "Subscription Contract No.", "Subscription Contract Line No.", "Document Type", "Document No.")
+        {
+        }
```
miljance, line 308: Suggestion to consider skipping low cardinality fields in the first place of the key.

Some more key suggesions:
"Usage Data Import Entry No.", "Processing Status" - low cardinality for "Processing Status.
"Usage Data Import Entry No.", "Document No."

miljance (changes requested):
This looks like a great work. I have some more suggestions for improvements and few more questions:

1. Usage Data Import processing was not improved and is further lacking the progress windows. This is possible to be done at some later point if the speedup is desired.
2. Unnecessary validation chains during line creation - was not addressed. Could be done in follow-up.
3. The performance bottleneck problem in CheckServiceCommitmentDataConsistency was not addressed - Could be done in follow-up.
4. Line numbering via FindLast() instead of in-memory counters - Could be done in follow-up.
5. I know it is out of the current scope, but I would at least in some of the follow ups display some more information in the Dialog Window. The praxis has shown that a user would like to know what the average time is spent on a contract or document during document creation and billing proposal and perhaps the estimated remaining time based on the contract or document counter.

jeffreybulanadi:
Good catch on the key ordering. Partner is an enum with two values (Customer/Vendor), and leading a composite index with it means 50% of the index range is traversed before the more selective fields narrow the scan. Same issue applies to Document Type.

The keys have been reordered to lead with Subscription Contract No. (Code[20], high cardinality) across all three tables:

- BillingLine SK7: Subscription Contract No., Subscription Contract Line No., Document No., Document Type, Partner
- SubscriptionLine Key4: Subscription Contract No., Next Billing Date, Partner
- UsageDataBilling key2: Subscription Contract No., Subscription Contract Line No., Document No., Document Type, Partner

Also added UsageDataBilling key3 on (Usage Data Import Entry No., Document No.) per the suggestion in the last comment. Document No. carries higher cardinality than Processing Status, so it makes a more selective second column for entry-level lookups.

src/Apps/W1/Subscription Billing/App/Billing/Tables/BillingLine.Table.al
```diff
@@ -243,6 +243,9 @@ table 8061 "Billing Line"
         key(SK6; "Billing Template Code", Partner)
         {
         }
+        key(SK7; "Subscription Contract No.", "Subscription Contract Line No.", "Document No.", "Document Type", Partner)
```
djukicmilica, line 246: @jeffreybulanadi

Adding new keys can be costly and may negatively impact performance. Based on the analysis, this key degrades performance for some operations, so I recommend removing it.


Image

djukicmilica (changes requested):

src/Apps/W1/Subscription Billing/App/Billing/Tables/BillingLine.Table.al
miljance: Good point. The function GetBillingLineNo might be a bottleneck indeed which I already observed on the real project. Instead of adding the key I suggest removing the function GetBillingLineNo completely and passing the "Billing Line No." as a denormalized field. That would bring more benefit to the performance then the key.

miljance:
@JesperSchulz @djukicmilica Can I please create another PR that contain remaining 50% percent of optimizations I was planning to do? This original issue was created by me after all, and I would like to avoid additional effort for the second run.

djukicmilica:
> @JesperSchulz @djukicmilica Can I please create another PR that contain remaining 50% percent of optimizations I was planning to do? This original issue was created by me after all, and I would like to avoid additional effort for the second run.

Yes, please. :)

djukicmilica:
Hi @jeffreybulanadi , can you please check the comments? :)

alexei-dobriansky:

<details>

This PR improves Subscription Billing for large datasets by adding table keys, adding progress dialogs, guarding dialog use with `GuiAllowed`, caching contract reads, hoisting `GetBillingLineNo()`, and committing after each created document. The customer and vendor document paths are kept mostly parallel, and the changed code does not depend on a BaseApp publisher. The main functional change is the new per-document transaction boundary, so an error after one document can now leave earlier documents and billing-line updates committed. The diff helps several hot paths, but it does not cover all root causes listed in issue #7690, and it overlaps heavily with PR #8553.
</details>

<details>

**S1** - *Make the W1 build pass*
The current status for this head has failed W1 app build and test jobs. Please rerun or fix the failures before merge, because this PR changes Subscription Billing app objects and the build is the first validation of those changes.

**S2** - *Add a transaction regression test*
The new `Commit()` calls change rollback behavior for billing document creation. Add a test with at least two documents where a later document fails, and verify that the earlier document and its billing-line updates are intentionally preserved.

**S3** - *Add automated no-GUI coverage*
The PR changes several dialog paths for Job Queue and non-GUI execution. Add or extend an automated billing test so the path runs without UI messages and still creates documents or logs errors correctly.

**S4** - *Resolve the overlap with PR 8553*
This PR and PR #8553 change the same Subscription Billing codeunits and tables for the same large-dataset issue. Please make it clear which PR is the intended fix, or keep this PR scoped as a partial fix so issue #7690 is not closed too early.
</details>

<details>

**Risk:** The highest risk is the new commit boundary in `CreateBillingDocuments.Codeunit.al`, because failed runs can now leave partial documents and updated billing lines by design. The new keys can improve reads, but they also add write cost to billing and usage tables. No public API or event signature changes were found, and no BaseApp publisher check was needed.

**Necessity:** The large-dataset scenario is valid and important; issue #7690 describes runs with about 4 million subscription lines and missing Job Queue support. The scope is still incomplete for that issue: usage import progress, validation-chain reduction, in-memory line numbering, and the `CheckServiceCommitmentDataConsistency` bottleneck are not addressed here. Because PR #8553 covers the same area more broadly, the merge path should be clarified before this PR is accepted.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=7776 round=1 by=alexei-dobriansky at=2026-07-02 lastSha=70ef7ff5fbb75d56e51eb69be1b91061ef983464 suggestions=S1,S2,S3,S4

djukicmilica:
Hi @jeffreybulanadi , can you please check the comments? :)

miljance:
> Hi @jeffreybulanadi , can you please check the comments? :)

Hi @djukicmilica, I have created another PR. I asked for this change since I have had (a bad) experience from a large scare project. Can we please close this PR in favor of my PR? https://github.com/microsoft/BCApps/pull/8553

AndersLarsenMicrosoft:
fixed with https://github.com/microsoft/BCApps/pull/8553

PR #7777, opened 2026-04-21
Rest Client: use ContainsSecret when removing default request headers

jeffreybulanadi:
Fixes #7581

## Summary

Calling `SetAuthorizationHeader` (or `SetDefaultRequestHeader`) twice on the same `Rest Client` instance throws `Invalid HTTP header. Please, make sure the format of the header is correct.` The root cause is that the `SecretText` overload of `SetDefaultRequestHeader` in codeunit 2351 "Rest Client Impl." uses `Contains()` to check for an existing header before removing it. `Contains()` only finds plain-text headers. When a header was previously stored as a `SecretText` value, `Contains()` returns false, the `Remove()` is skipped, and the subsequent `Add()` fails with a collision.

## Root Cause

`HttpHeaders.Contains(Name)` and `HttpHeaders.ContainsSecret(Name)` are separate checks. A header added with a `SecretText` value is invisible to `Contains()` and must be queried with `ContainsSecret()`.

The sibling codeunit `HttpRequestMessageImpl` (2349) in the same module already uses the correct combined check:

```al
if RequestHttpHeaders.Contains(HeaderName) or RequestHttpHeaders.ContainsSecret(HeaderName) then
    RequestHttpHeaders.Remove(HeaderName);
```

`SetDefaultRequestHeader` in `RestClientImpl` used only `Contains()`, missing the secret-header case.

## Fix

Apply the same `Contains(Name) or ContainsSecret(Name)` pattern to both overloads of `SetDefaultRequestHeader` in `RestClientImpl.Codeunit.al`.

The `Text` overload is also updated so that a secret header can be cleanly replaced by a plain-text header (symmetric fix).

## Files Changed

- `src/System Application/App/Rest Client/src/RestClientImpl.Codeunit.al` -- add `ContainsSecret` check to both overloads of `SetDefaultRequestHeader`
- `src/System Application/Test/Rest Client/src/RestClientTests.Codeunit.al` -- add three regression tests

## How to Test

**Reproduce the bug (before fix):**
```al
trigger OnAction()
var
    RestClient: Codeunit "Rest Client";
begin
    RestClient.Initialize();
    RestClient.SetAuthorizationHeader(SecretStrSubstNo('Bearer test'));
    RestClient.SetAuthorizationHeader(SecretStrSubstNo('Bearer test')); // throws before fix
end;
```

**After fix:** the code above completes without error.

The three new test procedures in `RestClientTests.Codeunit.al` cover:
1. `TestSetAuthorizationHeaderTwice` -- exact reproduction from the issue
2. `TestSetDefaultRequestHeaderSecretTwice` -- general `SecretText` overload called twice
3. `TestSetDefaultRequestHeaderReplaceSecretWithText` -- secret header replaced by plain-text header

### DIFF
--- a/src/System Application/App/Rest Client/src/RestClientImpl.Codeunit.al
+++ b/src/System Application/App/Rest Client/src/RestClientImpl.Codeunit.al
@@ -77,7 +77,7 @@ codeunit 2351 "Rest Client Impl."
     procedure SetDefaultRequestHeader(Name: Text; Value: Text)
     begin
         CheckInitialized();
-        if CurrHttpClientInstance.DefaultRequestHeaders.Contains(Name) then
+        if CurrHttpClientInstance.DefaultRequestHeaders.Contains(Name) or CurrHttpClientInstance.DefaultRequestHeaders.ContainsSecret(Name) then
             CurrHttpClientInstance.DefaultRequestHeaders.Remove(Name);
         CurrHttpClientInstance.DefaultRequestHeaders.Add(Name, Value);
     end;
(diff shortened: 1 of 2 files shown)


JesperSchulz (approved):

JesperSchulz:
Let's re-run the build pipeline.

JesperSchulz:
There seems to be a compilation error: AL0133 Argument 2: cannot convert from 'Text' to 'SecretText'

darjoo:
@jeffreybulanadi FYI on the build failures.

Aleyenda:
Closing this as the build failures have not been addressed. @jeffreybulanadi, if you have time to look into it, feel free to reopen.

PR #7782, opened 2026-04-21
Bug 631974: [EDocument] [28.0] Copilot capability registration + historical matching fixes

dayland:
- Add self-healing RegisterCapabilityIfNeeded in AI Tool Processor Setup
- Expand SetLoadFields to include all fields needed by historical matching
- Enrich temp table with product codes from E-Doc Purchase Line History
- Fix ProductCode search to use Vendor Item No instead of G/L Account No

Fixes [AB#631974](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/631974)

### DIFF
--- a/src/Apps/W1/EDocument/App/src/Processing/AI/Tools/EDocHistoricalMatching.Codeunit.al
+++ b/src/Apps/W1/EDocument/App/src/Processing/AI/Tools/EDocHistoricalMatching.Codeunit.al
@@ -219,6 +219,7 @@ codeunit 6177 "E-Doc. Historical Matching" implements "AOAI Function", IEDocAISy
     local procedure LoadHistoricalDataIntoTempTable(var TempPurchInvLine: Record "Purch. Inv. Line" temporary; VendorNo: Code[20]; HistoricalMatchingConfig: Text)
     var
         PurchInvLine: Record "Purch. Inv. Line";
+        EDocPurchaseLineHistory: Record "E-Doc. Purchase Line History";
         AllocationAccount: Record "Allocation Account";
         FeatureTelemetry: Codeunit "Feature Telemetry";
         OneYearAgoDate: Date;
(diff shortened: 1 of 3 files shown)


src/Apps/W1/EDocument/App/src/Processing/AI/Tools/EDocHistoricalMatching.Codeunit.al
```diff
@@ -253,6 +254,18 @@ codeunit 6177 "E-Doc. Historical Matching" implements "AOAI Function", IEDocAISy
             until (PurchInvLine.Next() = 0) or (RecordCount >= MaxHistoricalRecords);
         end;
 
+        // Enrich temp records with product codes from e-document history
+        EDocPurchaseLineHistory.SetCurrentKey("Vendor No.", "Product Code");
```
Groenbech96, line 258: I could not read about this in the bug. 

What are we gaining by adding Product code?
Would a produce code match not be directly matched already? 

Just for me to understand. 
Maybe consider only backporting what is on the rootcause analysis of the bug? If this is a improvement, then maybe just 28.x? and HF the CP registration issues.

Groenbech96 (approved):

src/Apps/W1/EDocument/App/src/Processing/AI/Tools/EDocHistoricalMatching.Codeunit.al
dayland: Valid point, this was a latent bug that already existed where the product code matching never worked because it was comparing an extracted product code value from the invoice with the "No." field in the history which is wrong as that is the internal number that the line is mapped to, not the customer product number. So, the agent fixed 3 original root issues 1) CP capability registration 2) missing selected fields in the load history 3) invalid product code direct match. 

Should I simplify the HF code here or update the bug with the full root cause analysis? Seems the agent only put in the CP capability registration details in the bug ticket.

AndersLarsenMicrosoft (approved):

PR #7786, opened 2026-04-21
[SOA] Enhance attachment handling in agent task messages with ignored reason support

tomasevicst:
#### Summary 
This PR adds support for tracking and populating the ignored reason when handling attachments in agent task messages.

**Changes:**
- Modified `AgentTaskMessageBuilder.Codeunit.al `to expose ignored reason functionality
- Updated `AgentMessageImpl.Codeunit.al` to handle ignored reason in attachment processing
- Enhanced `AgentTaskMsgBuilderImpl.Codeunit.al` to populate ignored reason for attachments

#### Work Item(s) 
Fixes [AB#629496](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/629496)


src/System Application/App/Agent/Interaction/Internal/AgentTaskMsgBuilderImpl.Codeunit.al
```diff
@@ -146,6 +158,7 @@ codeunit 4311 "Agent Task Msg. Builder Impl."
         TempAgentTaskFileToAttach.Modify();
 
         GlobalIgnoreAttachmentsList.Add(TempAgentTaskFileToAttach.ID, Ignored);
+        GlobalIgnoredReasonByFileId.Add(TempAgentTaskFileToAttach.ID, IgnoredReason);
```
attilatoury: Should we add the IgnoredReason only when Ignored = true? Right now I can call this with Ignored = false and a non-empty IgnoredReason

src/System Application/App/Agent/Interaction/Internal/AgentMessageImpl.Codeunit.al
```diff
@@ -69,10 +69,28 @@ codeunit 4308 "Agent Message Impl."
             exit;
 
         TempAgentTaskFile.Content.CreateInStream(FileInstream, AgentTaskImpl.GetDefaultEncoding());
-        AddAttachment(AgentTaskMessage, TempAgentTaskFile."File Name", TempAgentTaskFile."File MIME Type", FileInstream);
+        AddAttachment(AgentTaskMessage, TempAgentTaskFile."File Name", TempAgentTaskFile."File MIME Type", FileInstream, '');
+    end;
+
+    procedure AddAttachment(var AgentTaskMessage: Record "Agent Task Message"; var TempAgentTaskFile: Record "Agent Task File" temporary; IgnoredReason: Text[250])
```
attilatoury, line 62: Looks like this code is duplicated between this procedure and the one above, can we have one call the other instead?

src/System Application/App/Agent/Interaction/Internal/AgentTaskMsgBuilderImpl.Codeunit.al
attilatoury: Also the Add will fail with a runtime error if the value is already there so maybe we can check before adding it

attilatoury (dismissed):

src/System Application/App/Agent/Interaction/Internal/AgentTaskMsgBuilderImpl.Codeunit.al
tomasevicst: Good catch

tomasevicst: I agree. Set instead of Add will handle that. Good catch.

src/System Application/App/Agent/Interaction/Internal/AgentMessageImpl.Codeunit.al
tomasevicst: Good catch.

attilatoury (approved):

JakovljevicDusan (approved):

PR #7792, opened 2026-04-22
Surface e-document processing errors in page actions

mynjj:
## Summary
- Page actions (Analyze PDF, Prepare Draft, Open Draft, Analyze Document, Reset Draft) silently swallowed pipeline processing errors, returning without surfacing failures to the user or agent
- Centralize the check-and-throw pattern into `EDocumentErrorHelper.ThrowIfHasErrors` and apply it to all non-Finalize entry points
- Refactor existing `FinalizeEDocument` to use the same helper, removing duplicate code

image

Fixes [AB#631965](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/631965)

[AB#620089](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/620089)


Groenbech96 (approved):
The error message fact box should contain the errors if I am not mistaken?
Can the agent not use that? 

Anyway, not sure what the general pattern here is.

AndersLarsenMicrosoft (approved):

src/Apps/W1/EDocument/App/src/Helpers/EDocumentErrorHelper.Codeunit.al
```diff
@@ -105,6 +105,24 @@ codeunit 6115 "E-Document Error Helper"
...
+    procedure ThrowIfHasErrors(EDocument: Record "E-Document")
+    var
+        TempErrorMessage: Record "Error Message" temporary;
+        ErrorMessage: Record "Error Message";
+    begin
+        if not HasErrors(EDocument) then
+            exit;
+        ErrorMessage.SetRange("Context Record ID", EDocument.RecordId);
+        ErrorMessage.CopyToTemp(TempErrorMessage);
+        Commit();
```
ventselartur, line 122: Won't we commit anything unnecessary in this case? I think you can utilize this one to make sure that we have not started the write transaction and won't actually commit anything to the database:
https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/methods-auto/database/database-isinwritetransaction-method

ventselartur (approved):

AleksanderGladkov (approved):

src/Apps/W1/EDocument/App/src/Helpers/EDocumentErrorHelper.Codeunit.al
mynjj: This commit is to persist the error in the ErrorMessage table before throwing, in this scenario it is called at the end of the process, where we have already rolled back if there was such an error (all the process runs under `if cu.run` pattern), so it makes no harm here.

PR #7798, opened 2026-04-22
[E-Documents Core] [Peppol] - Enabling EDI capabilities with E-Documents. Receive path (Sales Order ← E‑Document).

GMatuleviciute:

Implementation

Enables electronic document import for sales orders in PEPPOL BIS 3.0 format, allowing suppliers to send structured order documents to customers.

Key additions:

- Creates an E-Document when a sales order is received.
- Insures correct mapping and document creation.
- Prevents deletion of sales orders linked to active (non-canceled) E-Documents to protect data integrity.
- Adds test coverage for E-Document creation on import, deletion guard, and successful deletion after cancellation.

Fixes #


Groenbech96:
Nvm, it is a draft.

PR #7826, opened 2026-04-23
Shopify: preserve Customer filters during customer mapping

### LINKED ISSUE #4436: [Bug]: Incorrectly removing filters from a record during customer mapping with Shopify
### Describe the issue

The "Shpfy Customer Mapping" Codeunit uses a procedure called FindMapping.
This Procedure has the ability to apply filters on the Customer before the Function DoFindMapping() try to map.

Image

But it uses the FindCustomer Variable without passing the filters from the Customer Parameter.

Image

### Expected behavior
Filters of the Customer Record are considered.

### Steps to reproduce

Try to apply additional filters for Customer Mapping.

### Additional context

Our Customer want´s to be sure, that he only finds and updates the customers he want´s to.
Therefore additional filters are needed because Business-Email and Private-Email is sometimes the same, but one hand the order is B2B and on the other hand B2C.

jeffreybulanadi:
Fixes #4436

## Problem

In ShpfyCustomerMapping.Codeunit.al, the DoFindMapping local procedure declares its own FindCustomer: Record Customer variable for the ShopifyToBC direction. This variable starts with no filters, so any filters that subscribers set on the Customer parameter via OnBeforeFindMapping are silently discarded. In scenarios such as B2B/B2C where multiple customers share the same email or phone number, the caller has no reliable way to restrict which customer is matched.

## Root Cause

FindCustomer is a fresh local variable so caller filters applied via OnBeforeFindMapping are never inherited before the email or phone lookup.

## Fix

Call FindCustomer.CopyFilters(Customer) before each lookup, and again after Clear(FindCustomer) (which resets all filters) in the phone-number path.

## Tests

Two new integration tests in ShpfyCustomerMappingTest (codeunit 139569):

- TestFindMappingRespectsCustomerFilters: two BC customers share the same email; a manual event subscriber restricts the Customer view to the second customer via SetRange; asserts FindMapping maps the Shopify customer to the filtered customer only.
- TestFindMappingRespectsPhoneFilter: same pattern exercising the phone-number lookup path.

A new subscriber codeunit Shpfy Cust. Mapping Filter Sub (ID 139700) follows the EventSubscriberInstance = Manual pattern with BindSubscription/UnbindSubscription established across the Shopify test suite.

### DIFF
--- a/src/Apps/W1/Shopify/App/src/Customers/Codeunits/ShpfyCustomerMapping.Codeunit.al
+++ b/src/Apps/W1/Shopify/App/src/Customers/Codeunits/ShpfyCustomerMapping.Codeunit.al
@@ -108,6 +108,7 @@ codeunit 30118 "Shpfy Customer Mapping"
         case Direction of
             Direction::ShopifyToBC:
                 begin
+                    FindCustomer.CopyFilters(Customer);
                     if ShopifyCustomer.Email <> '' then begin
                         FindCustomer.SetFilter("E-Mail", '@' + ShopifyCustomer.Email);
                         if FindCustomer.FindFirst() then begin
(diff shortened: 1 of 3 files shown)


onbuyuka:
@jeffreybulanadi do you need this change? asking since this is not an issue created by you and a pretty old one too. we rather avoid changes if they are not needed.

onbuyuka:
not planned

PR #7849, opened 2026-04-25
Subscription Billing: carry Customer Price Group to Subscription Line on Extend Contract

### LINKED ISSUE #5105: [Bug]:  The customer price group is not transferred to the subscription line when the “Extend Contract” function is used. [Subscription Billing]
### Describe the issue

When the “Extend contract” function is used in the customer contract, the customer price group is only transferred to the subscription, but not to the subscription line.

### Expected behavior

If a line is created in the customer contract using the “Get Subscription Lines” function, the customer price group is transferred to both the subscription and the subscription line.
This behavior is also expected for the “Extend Contract” function.

### Steps to reproduce

Create a new Customer Subscription contract for a Customer to whom a costumer price group is assigned.
Use function "Get Subscription Lines" to get the first line.
Use function "Extend Contract" and add the same item as in the first line.

Each lines is assigned to a subscription. 
The first subscription shows customer price group in the subscription and the subscription line.
The second line shows customer price group only in the subscription.

### Additional context

Screenshots from the line which is created by "Extend contract":

Image

Image

jeffreybulanadi:
## Summary

Fixes #5105

When the **Extend Contract** function is used on a Customer Subscription Contract, the Customer Price Group was transferred to the Subscription Header (Subscription) but **not** to the Subscription Lines.

The Get Subscription Lines function correctly transfers the Customer Price Group to both, so this fix aligns the Extend Contract path with that expected behavior.

## Root Cause

In `SubscriptionHeader.Table.al`, `InsertServiceCommitmentsFromServCommPackage` set the Subscription Line's Customer Price Group to `ServiceCommitmentPackage."Price Group"` instead of `Rec."Customer Price Group"`:

The Subscription Package's Price Group field is a selection filter (which customer price groups a package applies to) and can be blank even when the customer has a price group assigned. The Subscription Line's Customer Price Group should always reflect the customer's actual price group, not the package's filter value.

## Changes

**SubscriptionHeader.Table.al**
- Fixed `InsertServiceCommitmentsFromServCommPackage` to assign `Rec."Customer Price Group"` to new Subscription Lines instead of `ServiceCommitmentPackage."Price Group"`.

**ExtendContractTest.Codeunit.al**
- Corrected the assertion in `CheckAssignedSalesServiceCommitmentValues` that was comparing the Subscription Line's Customer Price Group against the package's Price Group (the wrong value).
- Added new regression test `ExtendContractTransfersCustomerPriceGroupToSubscriptionLine` that creates a customer with a Customer Price Group, extends a contract, and verifies the Customer Price Group is correctly transferred to the Subscription Lines.

## How to Test

1. Create a Customer Price Group in Business Central.
2. Create a Customer and assign the Customer Price Group to the customer.
3. Create a Customer Subscription Contract for that customer.
4. Set up a subscription item with a standard Subscription Package (Price Group left blank on the package).
5. On the Customer Subscription Contract card, use the Extend Contract action to add the item.
6. Open the created Subscription and inspect its Subscription Lines.

Before fix: Subscription Lines show a blank Customer Price Group even though the customer has one assigned.
After fix: Subscription Lines show the same Customer Price Group as the Subscription Header and the customer.

Fixes [AB#634137](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/634137)

### DIFF
--- a/src/Apps/W1/Subscription Billing/Test/Customer Contracts/ExtendContractTest.Codeunit.al
+++ b/src/Apps/W1/Subscription Billing/Test/Customer Contracts/ExtendContractTest.Codeunit.al
@@ -362,8 +362,8 @@ codeunit 148152 "Extend Contract Test"
     [Test]
     procedure ExtendContractTransfersCustomerPriceGroupToSubscriptionLine()
     var
-        ExtendContractMgt: Codeunit "Extend Sub. Contract Mgt.";
         TempExtendServiceCommPackage: Record "Subscription Package" temporary;
+        ExtendContractMgt: Codeunit "Extend Sub. Contract Mgt.";
     begin
         // [SCENARIO] When Extend Contract is used for a customer with a Customer Price Group,
         // the Customer Price Group is transferred to both the Subscription Header and the Subscription Line.
--- a/src/Apps/W1/Subscription Billing/Test/Customer Contracts/ExtendContractTest.Codeunit.al
+++ b/src/Apps/W1/Subscription Billing/Test/Customer Contracts/ExtendContractTest.Codeunit.al
@@ -359,6 +359,52 @@ codeunit 148152 "Extend Contract Test"
         ExtendContract.ItemDescription.AssertEquals(ItemTranslation.Description);
     end;
 
+    [Test]
+    procedure ExtendContractTransfersCustomerPriceGroupToSubscriptionLine()
+    var
+        ExtendContractMgt: Codeunit "Extend Sub. Contract Mgt.";
+        TempExtendServiceCommPackage: Record "Subscription Package" temporary;
+    begin
+        // [SCENARIO] When Extend Contract is used for a customer with a Customer Price Group,
+        // the Customer Price Group is transferred to both the Subscription Header and the Subscription Line.
+
+        // [GIVEN] Subscription Item with a standard package
+        Initialize();
+        ContractTestLibrary.InitContractsApp();
+        ContractTestLibrary.CreateItemWithServiceCommitmentOption(Item, Enum::"Item Service Commitment Type"::"Service Commitment Item");
+        SetupItemWithMultipleServiceCommitmentPackages();
+
+        // [GIVEN] Customer with a Customer Price Group assigned
+        LibrarySales.CreateCustomerPriceGroup(CustomerPriceGroup);
+        ContractTestLibrary.CreateCustomer(Customer);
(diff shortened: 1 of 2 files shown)


AndersLarsenMicrosoft (dismissed):

src/Apps/W1/Subscription Billing/Test/Customer Contracts/ExtendContractTest.Codeunit.al
```diff
@@ -359,6 +359,52 @@ codeunit 148152 "Extend Contract Test"
         ExtendContract.ItemDescription.AssertEquals(ItemTranslation.Description);
     end;
 
+    [Test]
+    procedure ExtendContractTransfersCustomerPriceGroupToSubscriptionLine()
+    var
+        ExtendContractMgt: Codeunit "Extend Sub. Contract Mgt.";
```
djukicmilica: Error: AA0021 Variable declarations should be ordered by type. Variables should be sorted like this: Record, Report, Codeunit, XmlPort, Page, Query, Notification, BigText, DateFormula, RecordId, RecordRef, FieldRef, and FilterPageBuilder. ...

djukicmilica (changes requested):

djukicmilica (changes requested):
.

src/Apps/W1/Subscription Billing/Test/Customer Contracts/ExtendContractTest.Codeunit.al
djukicmilica: @jeffreybulanadi

src/Apps/W1/Subscription Billing/Test/Customer Contracts/ExtendContractTest.Codeunit.al
```diff
@@ -359,6 +359,52 @@ codeunit 148152 "Extend Contract Test"
...
+        TempExtendServiceCommPackage: Record "Subscription Package" temporary;
+    begin
```
JesperSchulz, line 367:
```suggestion
    var
        TempExtendServiceCommPackage: Record "Subscription Package" temporary;
        ExtendContractMgt: Codeunit "Extend Sub. Contract Mgt.";
    begin
```

JesperSchulz (approved):

djukicmilica (approved):

PR #7850, opened 2026-04-25
RecordSelection: exclude Media fields from column list

### LINKED ISSUE #1198: [Bug]: RecordSelection should not display Media Fields
### Describe the issue

Currently the RecordSelection Codeunit uses the page summary to select the fields for the record selection.

This results in non-helping media fields being displayed in a list view.

### Expected behavior

Media should be ignored.

### Steps to reproduce

```al
                field("Record SystemId"; RecordSystemId)
                {
                    Caption = 'Record SystemId';
                    Editable = false;
                    ToolTip = 'Specifies the SystemId of the record.';
                    trigger OnAssistEdit()
                    var
                        TempRecordSelectionBuffer: Record "Record Selection Buffer" temporary;
                        RecordSelection: Codeunit "Record Selection";
                        RecordSystemId : Guid;
                    begin
                        CheckForTableId(true);
                        if RecordSelection.Open(Database::167, 2147483647, TempRecordSelectionBuffer) then
                            RecordSystemId := TempRecordSelectionBuffer."Record System Id";
                    end;
                }
```
image

### Additional context

Furthermore it would be usefult to replace fields for custom apps. 

jeffreybulanadi:
Fixes #1198

## Summary

The RecordSelection codeunit uses the page summary to determine which fields to show in the lookup list. This included Media and MediaSet fields, which are unhelpful in a list view as they render as empty strings or raw GUIDs.

## Changes

**RecordSelectionImpl.Codeunit.al**
- Added RemoveMediaFields procedure that filters Media and MediaSet fields from the page summary field list using the Field system table
- Updated GetPageSummaryFields to accept a TableId parameter and call RemoveMediaFields after building the list
- Updated callers ToText and GetFieldsFromPage to pass TableId

**RecordSelectionTestTable.Table.al**
- Added a SomeMedia Media field
- Updated the Brick fieldgroup to include SomeMedia so the regression scenario is covered

**RecordSelectionTest.Codeunit.al**
- Added RecordSelectionMediaFieldsNotIncludedTest to verify Media fields are excluded from the field count
- Added RecordSelectionToTextSkipsMediaFieldsTest to verify ToText output is not affected by Media fields

### DIFF
--- a/src/System Application/App/Record Selection/src/RecordSelectionImpl.Codeunit.al
+++ b/src/System Application/App/Record Selection/src/RecordSelectionImpl.Codeunit.al
@@ -172,6 +172,8 @@ codeunit 9556 "Record Selection Impl."
             exit(false);
 
         GenericList := NavPageSummaryALFunctions.GetSummaryFields(PageId);
+        if IsNull(GenericList) then
+            exit(false);
 
         foreach PageSummaryField in GenericList do
             PageSummaryFieldList.Add(PageSummaryField);
--- a/src/System Application/App/Record Selection/src/RecordSelectionImpl.Codeunit.al
+++ b/src/System Application/App/Record Selection/src/RecordSelectionImpl.Codeunit.al
@@ -57,7 +57,7 @@ codeunit 9556 "Record Selection Impl."
             exit('');
 
         PageId := GetRelatedPageId(TableId);
-        RelatedPageExist := GetPageSummaryFields(PageId, FieldCount, PageSummaryFieldList);
+        RelatedPageExist := GetPageSummaryFields(TableId, PageId, FieldCount, PageSummaryFieldList);
 
         if not RelatedPageExist then
             GetPrimaryKeyFields(FromRecordRef, FromKeyRef, FieldCount);
@@ -148,7 +148,7 @@ codeunit 9556 "Record Selection Impl."
         FromRecordRef: RecordRef;
         PageSummaryFieldList: List of [Integer];
     begin
-        if not GetPageSummaryFields(PageId, FieldCount, PageSummaryFieldList) then
+        if not GetPageSummaryFields(TableId, PageId, FieldCount, PageSummaryFieldList) then
             exit(false);
 
         FromRecordRef.Open(TableId);
@@ -162,7 +162,7 @@ codeunit 9556 "Record Selection Impl."
         exit(true);
     end;
 
-    local procedure GetPageSummaryFields(PageId: Integer; var FieldCount: Integer; var PageSummaryFieldList: List of [Integer]): Boolean
+    local procedure GetPageSummaryFields(TableId: Integer; PageId: Integer; var FieldCount: Integer; var PageSummaryFieldList: List of [Integer]): Boolean
     var
         GenericList: DotNet GenericList1;
         NavPageSummaryALFunctions: DotNet NavPageSummaryALFunctions;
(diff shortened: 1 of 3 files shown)


src/System Application/Test/Record Selection/src/RecordSelectionTest.Codeunit.al
```diff
@@ -141,6 +141,44 @@ codeunit 135136 "Record Selection Test"
...
+        // [GIVEN] Data in the test table which has a Media field in the Brick fieldgroup
+        Initialize();
+        PermissionsMock.Set('Rec. Selection Read');
+
+        // [WHEN] GetRecordsFromTableId is called for the table
+        FieldCount := RecordSelectionImpl.GetRecordsFromTableId(Database::"Record Selection Test Table", PrimaryKeyCaptions, TempRecordSelectionBuffer);
+
+        // [THEN] The media field is not included in the field count
+        Assert.AreEqual(3, FieldCount, 'Media fields should be excluded from the field count.');
+        Assert.AreEqual('', PrimaryKeyCaptions[4], 'Media field caption should not be present in the captions.');
```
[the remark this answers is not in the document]
jeffreybulanadi: Addressed in the latest commit. Rewrote RecordSelectionMediaFieldsNotIncludedTest to use the public Record Selection codeunit (RecordSelection.ToText) instead of the internal RecordSelectionImpl. Also removed the duplicate RecordSelectionToTextSkipsMediaFieldsTest since the rewritten test covers the same scenario through the public API.

src/System Application/App/Record Selection/src/RecordSelectionImpl.Codeunit.al
```diff
@@ -176,10 +176,34 @@ codeunit 9556 "Record Selection Impl."
         foreach PageSummaryField in GenericList do
             PageSummaryFieldList.Add(PageSummaryField);
```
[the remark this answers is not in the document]
jeffreybulanadi: Addressed in the latest commit. Added an IsNull(GenericList) guard in GetPageSummaryFields immediately after NavPageSummaryALFunctions.GetSummaryFields(PageId), returning false when the list is null. This matches the same pattern used in PageSummaryProviderImpl.Codeunit.al.

darjoo (approved):

src/System Application/App/Record Selection/src/RecordSelectionImpl.Codeunit.al
```diff
@@ -162,7 +162,7 @@ codeunit 9556 "Record Selection Impl."
         exit(true);
     end;
 
-    local procedure GetPageSummaryFields(PageId: Integer; var FieldCount: Integer; var PageSummaryFieldList: List of [Integer]): Boolean
+    local procedure GetPageSummaryFields(TableId: Integer; PageId: Integer; var FieldCount: Integer; var PageSummaryFieldList: List of [Integer]): Boolean
```
nikolakukrika, line 165: This is very risky. Unless we update all references we will be calling the code with wrong IDs.

nikolakukrika: Better approach is to use the source table from the page and filter the list out.

nikolakukrika (changes requested):

src/System Application/App/Record Selection/src/RecordSelectionImpl.Codeunit.al
```diff
@@ -148,7 +148,7 @@ codeunit 9556 "Record Selection Impl."
         FromRecordRef: RecordRef;
         PageSummaryFieldList: List of [Integer];
     begin
-        if not GetPageSummaryFields(PageId, FieldCount, PageSummaryFieldList) then
+        if not GetPageSummaryFields(TableId, PageId, FieldCount, PageSummaryFieldList) then
```
nikolakukrika, line 151: same comment here, we cannot simply switch from pageID to table ID. 
It will break the logic in different ways.
TableID could be zero right now, PageID could be non-zero. See the comment below on the fix.

Groenbech96:
Closing as stale.

PR #7860, opened 2026-04-27
[MCP] Pages UX adjustments

onbuyuka:
## Summary

Apply PM-reviewed UX changes to the BC MCP (Model Context Protocol) configuration pages from the deck attached to bug 632734.

**MCP Config Card** (`MCPConfigCard.Page.al`)
- General fasttab restructured into explicit `Column1` (Name + Description) and `Column2` (all toggles: Active, Default, Dynamic Tool Mode, Discover Read Only Objects, Allow Prod Changes).
- Tool Mode descriptive text moved inside the General group so collapsing General also hides the explanation, freeing vertical space for the Available Tools subform.

**Available Tools subform** (`MCPConfigToolList.Page.al`)
- Column order: Object Id (kept first as the drilldown/select entry point), Object Name, Allow Read, Allow Create, Allow Modify, Allow Delete, Allow Bound Actions, Object Type, API Version.

**MCP API Tools lookup** (`MCPAPIConfigToolLookup.Page.al`)
- Column order: Name, Entity Name, ID, API Publisher, API Group, API Version.

**API Publishers page** (`MCPAPIPublisherLookup.Page.al`)
- Renamed to "API Groups". Column order: API Group, API Publisher.

**Add Tools by API Group dialog** (`MCPToolsByAPIGroup.Page.al` + new `ResolvePublisherForGroup` in `MCPConfigImplementation.Codeunit.al`)
- API Group is now the leading field. On selection (lookup or manual entry), the publisher auto-fills when only one publisher exposes the chosen group. When multiple publishers share the group name (which is allowed), the existing publisher lookup pops up filtered to those rows so the user disambiguates explicitly. The previous "select publisher first" precondition is gone.

Fixes [AB#632734](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/632734)


src/System Application/App/MCP/src/Configuration/Pages/MCPConfigCard.Page.al
```diff
@@ -24,81 +24,92 @@ page 8351 "MCP Config Card"
...
+                        end;
+                    }
+                    field(DiscoverReadOnlyObjects; Rec.DiscoverReadOnlyObjects)
+                    {
+                        ToolTip = 'Specifies whether to allow discovery of read-only objects not defined in the configuration. Only supported with dynamic tool mode.';
+                        Editable = not IsDefault and Rec.EnableDynamicToolMode and not Rec.Active;
+                    }
+                    field(AllowProdChanges; Rec.AllowProdChanges)
+                    {
+                        ToolTip = 'Allows create, update and delete tools for the specified MCP configuration. Disallowing this will make the tools read-only.';
```
gggdttt, line 89: Minor: I guess we should start from 'Specifies'?

gggdttt (dismissed):

src/System Application/App/MCP/src/Configuration/Pages/MCPConfigCard.Page.al
onbuyuka: allows also works

nikolakukrika (dismissed):

onbuyuka:
conflict - wait until https://github.com/microsoft/BCApps/pull/6670 is in

JesperSchulz (dismissed):

src/System Application/App/MCP/src/Configuration/Codeunits/MCPConfigImplementation.Codeunit.al
```diff
@@ -668,6 +668,30 @@ codeunit 8351 "MCP Config Implementation"
...
 
+    internal procedure ResolvePublisherForGroup(var MCPAPIPublisherGroup: Record "MCP API Publisher Group"; var APIPublisher: Text; APIGroupValue: Text)
+    begin
+        if APIGroupValue = '' then begin
+            APIPublisher := '';
+            exit;
+        end;
+
+        MCPAPIPublisherGroup.Reset();
+        MCPAPIPublisherGroup.SetRange("API Group", APIGroupValue);
```
haoranpb, line 679: Hmm, then why not define it as a local variable in the function? Or am I missing something

src/System Application/App/MCP/src/Configuration/Codeunits/MCPConfigImplementation.Codeunit.al
```diff
@@ -668,6 +668,30 @@ codeunit 8351 "MCP Config Implementation"
...
+
+        if not MCPAPIPublisherGroup.FindSet() then begin
```
haoranpb, line 681:
```suggestion
        if MCPAPIPublisherGroup.IsEmpty() then begin
```

src/System Application/App/MCP/src/Configuration/Codeunits/MCPConfigImplementation.Codeunit.al
onbuyuka: table is temporary, we need the records inserted to that variable

src/System Application/App/MCP/src/Configuration/Codeunits/MCPConfigImplementation.Codeunit.al
darjoo: This causes a problem. The subsequent calls to MCPAPIPublisherGroup will not have any records.
I recommend to leave as is.

onbuyuka: we need the records in the page.run

Groenbech96 (approved):

darjoo (approved):

PR #3697, opened 2025-05-17
Support for SFTP

joandrsn:
#### Summary
I have added a module for handling SFTP connections directly from Business Central.
image

This requires a SFTP service. A raspberry pi on the local network is sufficent for testing, but [Azure Blob Storage can also provides a testing scenario](https://learn.microsoft.com/en-us/azure/storage/blobs/secure-file-transfer-protocol-support).

#### Work Item(s)
Fixes #3686

#### DLL install process
1. Download these two nuget packages:
https://www.nuget.org/packages/ssh.net/
https://www.nuget.org/packages/BouncyCastle.Cryptography/
2. Extract the DLL's:
`lib\net8.0\Renci.SshNet.dll`
`lib\net6.0\BouncyCastle.Cryptography.dll`
3. Place in service tier's `Add-Ins`-folder and in one of the `al.assemblyProbingPaths`.

My internal checklist for this PR:
- [x] Get approval of SSH.NET + BouncyCastle.Cryptography for Business Central
- [x] Get a different number range (Not 50000..100000)
- [x] Update `app.json` to reflect the module
- [x] Namespaces for all files
- [x] Handle different types of exceptions (Connection errors, etc)
- [x] ~~Remove SFTP Client debugging page~~
- [x] Expand the "SFTP Operation Response" to contain Errors (Like ABS Operation Response)
- [x] Consider a stateless API
- [x] Add tests
- [x] Handle upload/download better if possible
  - Currently I'm downloading all the bytes and passing it to a MemoryStream, but I would like to simply get a stream. However I cannot seem to invoke the functions (The last parameter `Action<ulong>? downloadCallback = null` is tricky to make in Business Central dotnet integration).
  - Upload could use [UploadFile(Stream, string, Action<ulong>?)](https://sshnet.github.io/SSH.NET/api/Renci.SshNet.SftpClient.html#Renci_SshNet_SftpClient_UploadFile_System_IO_Stream_System_String_System_Action_System_UInt64__)
  - Download could use [DownloadFile(string, Stream, Action<ulong>?)](https://sshnet.github.io/SSH.NET/api/Renci.SshNet.SftpClient.html#Renci_SshNet_SftpClient_DownloadFile_System_String_System_IO_Stream_System_Action_System_UInt64__)

Fixes [AB#351397](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/351397)

PR #7892, opened 2026-04-28
[28.x] Deprecate SFTP connector

JesperSchulz:
#### Summary 
From version 29.0, we won't be able to support SFTP connections as introduced with #3697. The module and the connector are hence deprecated again.

#### Work Item(s) 
Fixes [AB#632775](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/632775)


JesperSchulz:
Deprecation decision postponed. Security hardening of module replaces deprecation plans for now (#9360).

PR #7913, opened 2026-04-30
EDocument: scope PEPPOL BIS 3.0 .xml file extension to correct document format

### LINKED ISSUE #7200: [Bug]: E-Document - E-Document Log Data Storage content is always exported as an .xml file
### Describe the issue

When exporting the file associated with an **E-Document Log** entry, it is always exported with an `.xml` extension because of the following piece of code:

```
namespace Microsoft.eServices.EDocument.IO.Peppol;
codeunit 6166 "EDoc Import PEPPOL BIS 3.0"
{
    ...
    [EventSubscriber(ObjectType::Table, Database::"E-Document Log", 'OnBeforeExportDataStorage', '', false, false)]
    local procedure SetFileExt(EDocumentLog: Record "E-Document Log"; var FileName: Text)
    begin
        FileName += '.xml';
    end;
    ...
}
```

This behavior breaks the file export function for third-party integrations where the imported data is structured in formats other than `xml`.

### Expected behavior

The specified event subscriber should only append the default file extension (`.xml`) when
- **Document Format** of the **E-Document Log** entry is _PEPPOL BIS 3.0_
- An extension has not yet been appended to the `FileName`.

### Steps to reproduce

1. Set up a new **E-Document Service** with the following parameters:
   - **Document Format**: any other than _PEPPOL BIS 3.0_
   - **Service Integration**: _none_
2. On the **E-Documents** page use the **New from file** action to upload any _valid_ `json` file.
3. On the card of the new **E-Document** navigate to **Related** / **Logs**.
4. On the **E-Document Logs** list select the first log entry, which should have the _Imported_ **E-Document Status**.
5. Use the **Export File** action to download the associated content.
6. The exported file content is `json` but the file extension is `.xml`.

### Additional context

This approach in general is not ideal as the file extension should be provided by the `IEDocFileFormat` implementation of an `E-Doc. Data Storage` entry that is being exported, not by event subsribers in E-Document apps. But unfortunatelly, **File Format** is not always specified on the `Data Storage` entry (actually, it is _always_ `Unspecified` for the _"Batch Imported"_ `E-Document Log Entries`).

So the suggested change would at least make it possible for specific `Document Format`/`Service Integration` implementations to decide the file extension on their own `E-Document Log` entries.

jeffreybulanadi:
## Summary

Closes #7200

The SetFileExt event subscriber in EDocImportPEPPOLBIS30.Codeunit.al (codeunit 6166) was unconditionally appending '.xml' to every E-Document Log export, regardless of the Document Format on the E-Document Service. This caused third-party integrations using non-PEPPOL formats (e.g., json, csv) to receive exported files with an incorrect .xml extension.

## Root cause

The subscriber appended '.xml' for all Document Formats, not only PEPPOL BIS 3.0.

## Change

The subscriber now guards with two conditions before appending '.xml':

1. The E-Document Log entry's Document Format is PEPPOL BIS 3.0 (the only format this codeunit is responsible for)
2. The FileName does not already have a file extension, allowing other subscribers or implementations to set their own extension first

## No breaking changes

- PEPPOL BIS 3.0 log exports continue to receive '.xml' when no extension is already set
- Other format log exports (Data Exchange, custom) are no longer overridden with '.xml'
- The 'using System.IO;' namespace for 'File Management' is already declared in this codeunit

Fixes [AB#634372](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/634372)

### DIFF
--- a/src/Apps/W1/EDocument/App/src/Format/EDocImportPEPPOLBIS30.Codeunit.al
+++ b/src/Apps/W1/EDocument/App/src/Format/EDocImportPEPPOLBIS30.Codeunit.al
@@ -611,7 +611,13 @@ codeunit 6166 "EDoc Import PEPPOL BIS 3.0"
 
     [EventSubscriber(ObjectType::Table, Database::"E-Document Log", 'OnBeforeExportDataStorage', '', false, false)]
     local procedure SetFileExt(EDocumentLog: Record "E-Document Log"; var FileName: Text)
+    var
+        FileManagement: Codeunit "File Management";
     begin
+        if EDocumentLog."Document Format" <> Enum::"E-Document Format"::"PEPPOL BIS 3.0" then
+            exit;
+        if FileManagement.GetExtension(FileName) <> '' then
+            exit;
         FileName += '.xml';
     end;
 


Groenbech96 (approved):
You have build errors. My guess is import statements.

JesperSchulz (approved):

JesperSchulz:
Build fails with:

`MCPConfigImplementation.Codeunit.al(526,85): error AL0132: '"Object Type"' does not contain a definition for 'Query' `

JesperSchulz:
Re-running build.

Groenbech96:
Pull latest main into branch.

Groenbech96:
@copilot merge main into the branch

JesperSchulz:
Retrying CI.

PR #7915, opened 2026-04-30
Helper method to display agent model name with Auto option

mbaranovych:
#### Summary 
Add helper method to display agent model name with Auto option that can be reused across pages
Mark new model methods as on prem for now.
#### Work Item(s) 
[AB#620625](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/620625)


src/System Application/App/Agent/Setup/AgentCard.Page.al
```diff
@@ -214,6 +222,11 @@ page 4315 "Agent Card"
             ProfileDisplayName := UserSettings.GetProfileName(TempUserSettingsRecord);
         end;
 
+        if Rec."Model Name" = '' then
```
qutreson: There will likely be other places where we need to show this Auto string, for instance in the agent design experience.

What about creating a codeunit "Agent Model" and have a method named GetModelName(Guid), or maybe adding this to the Agent.Codeunit which already has GetModelId?

qutreson (dismissed):

src/System Application/App/Agent/Setup/AgentCard.Page.al
nikolakukrika: +1

nikolakukrika (dismissed):

src/System Application/App/Agent/Setup/AgentCard.Page.al
mbaranovych: Added method, however we will not change agent card at the moment

qutreson (approved):

nikolakukrika (approved):

PR #7932, opened 2026-04-30
Remove redundant Prod. Order No. fields from Subc. Item Ledger Entry (AB#620746)

ChethanT:
## Summary

- The Subcontracting ILE table extension had two custom fields (`"Prod. Order No."` field 99001510 and `"Prod. Order Line No."` field 99001511) that were redundant with the base `Item Ledger Entry` fields `"Order No."` (field 91) and `"Order Line No."` (field 92) when `Order Type = Production`.
- These fields were written during posting but never read back, since all existing Subcontracting queries against ILE already use the base fields.
- Removed the two redundant fields, their `Key99001500` segment, the two hidden columns from the ILE page extension, and the two assignment lines from `UpdateNewItemLedgerEntry`.
- Added regression test `PostingSubcontractingPurchReceiptCreatesILEWithSubcFields` (codeunit 139989) verifying posting still works and the remaining custom ILE fields are accessible.

## Root cause

The original Subcontracting app duplicated production-order traceability that BC's base ILE already provides via `"Order Type"/"Order No."/"Order Line No."`. The custom fields added unnecessary schema weight and created two sources of truth.

## Test plan

- [x] `PostingSubcontractingPurchReceiptCreatesILEWithSubcFields` (new, codeunit `Subc. Subcontracting Test` 139989) — verifies posting creates ILE without error and remaining Subcontracting ILE fields are accessible
- [x] Verified no remaining references to the removed fields on `Item Ledger Entry` in App or Test source

Fixes [AB#620746](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/620746)


ChethanT:
Closing as Won't Fix. Investigation confirmed the custom ILE fields Prod. Order No. (99001510) and Prod. Order Line No. (99001511) are not redundant — they serve a distinct semantic role that cannot be replaced by the base Order No. / Order Line No. fields without breaking subcontracting transfer tracking. Full reasoning documented on ADO bug AB#620746.

PR #8031, opened 2026-05-07
(Bug 633224): [Subcontracting] View purchase orders prompt opens card pages instead of list when orders already exist

alexei-dobriansky:
### Problem

When **Create Subcontracting Order** is invoked from a Prod. Order Routing for which purchase orders already exist, the system prompts *"Purchase order(s) have already been created. Do you want to view them?"* and — on **Yes** — opens an individual **Purchase Order** card per existing PO. With multiple selected routing lines or multiple existing POs, the user gets a stack of card windows instead of a single consolidated list. The first-time path (no existing POs) correctly opens a list view via `ShowCreatedPurchaseOrder`, so the two paths were inconsistent.

Root cause: `Subc. Purchase Order Creator.CheckProdOrderRtngLine` runs once **per routing line** and each invocation independently prompted and opened its own page, with the per-line branch always falling through to `PageManagement.PageRun(PurchaseHeader)` (the card).

### Fix

- Stripped the existing-PO prompt and per-line page-opening out of `CheckProdOrderRtngLine` — it now only does validation (status, prod-order-line existence, work-center fields).
- Added `ShowExistingPurchaseOrdersForRoutingLines(var ProdOrderRoutingLine)` which iterates the selection, prompts **once** if any matching `Purchase Line` exists, and on Yes calls `PageManagement.PageRun(PurchaseLine)` filtered by `Document Type=Order`, `Type=Item`, `Prod. Order No.` — the same call/filter as `ShowCreatedPurchaseOrder`'s multi-PO branch, so both Scenario A (newly created) and Scenario B (already exists) now route through identical page-dispatch logic.
- The new method is invoked from `Subc. Prod. Order Rtng.` page extension's `Create Subcontracting Order` action, after `SetSelectionFilter` and before the create loop.

Fixes [AB#633224](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/633224)

### DIFF
--- a/src/Apps/W1/Subcontracting/App/src/Process/Codeunits/SubcPurchaseOrderCreator.Codeunit.al
+++ b/src/Apps/W1/Subcontracting/App/src/Process/Codeunits/SubcPurchaseOrderCreator.Codeunit.al
@@ -209,10 +209,7 @@ codeunit 99001557 "Subc. Purchase Order Creator"
 
     local procedure CheckProdOrderRtngLine(ProdOrderRoutingLine: Record "Prod. Order Routing Line"; var ProdOrderLine: Record "Prod. Order Line")
     var
-        PurchaseHeader: Record "Purchase Header";
-        PurchaseLine: Record "Purchase Line";
         WorkCenter: Record "Work Center";
-        ConfirmManagement: Codeunit "Confirm Management";
     begin
         if ProdOrderRoutingLine.Status <> "Production Order Status"::Released then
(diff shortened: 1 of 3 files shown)


ChethanT:
@spinkow

AleksandricMarko (approved):

PredragMaricic (approved):

PR #8048, opened 2026-05-07
[Shopify] Fix #7724: Resolve Option 1 Name from existing Shopify Variant records

### LINKED ISSUE #7724: [Bug]: Shopify variant create uses wrong option name after initial import from Shopify
## Describe the issue

When a product is imported from Shopify (`Sync Item = From Shopify`) and the shop is later switched to `To Shopify`, product sync may create `productVariantsBulkCreate` mutations using `optionName: "Variant"` instead of the product's actual Shopify option name (e.g. `Size`). Shopify rejects the mutation, the new BC item variants are never created in Shopify, and BC logs:

> The response from Shopify contains no JSON

This is reproducible in Cronus and not customer-data-specific.

## Expected behavior

For existing Shopify products, new variant creation from BC should reuse the product's existing Shopify option name (e.g. `Size`). Fallback to `Variant` should only occur when no Shopify option metadata exists (i.e. the product was originally created from BC).

## Steps to reproduce

1. In Shopify, create a product with option name `Size` and value `Test`.
2. In BC, set `Sync Item = From Shopify` and run product sync.
3. Confirm the imported Shopify variant shows `Option 1 Name = Size`, `Option 1 Value = Test`.
4. Switch the shop to `Sync Item = To Shopify`.
5. Add BC item variants (e.g. `VARIANT 1`, `VARIANT 2`) on the mapped item.
6. Run product sync (Report 30108).
7. Check the job queue log — the outbound payload will show `optionName: "Variant"` instead of `"Size"`, and the new variants will not appear in Shopify.

## Additional context

- Repro product ID: `10755490906449`
- Failing payload: `optionValues: [{ optionName: "Variant", name: "VARIANT 2" }]`
- Error in BC: `The response from Shopify contains no JSON`
- Root cause: `FillInProductVariantData` in `ShpfyProductExport.Codeunit.al` hardcodes `'Variant'` as option name (lines ~423 and ~476) instead of deriving it from existing Shopify variant metadata for the product.
- Suggested fix: Resolve option name from existing `Shpfy Variant` rows for the product before falling back to `'Variant'`. BC-origin products (where no Shopify option metadata exists) are unaffected.

tVisionDeb:
## Summary

Fixes #7724: Variant option names should be resolved from existing Shopify Variant records instead of using hardcoded values.

## Changes

- Modified `ShpfyProductExport.Codeunit.al` to call `ResolveOption1Name()` procedure when syncing product variant option names
  - Added new local procedure `ResolveOption1Name()` that queries existing variants for the correct option name
  - Ensures consistency with Shopify's actual variant configuration

## Validation

- Compiled successfully on BC-28 (Runtime 17.0): 0 errors
- Compiled successfully on BC-29 (Runtime 18.0): 0 errors
- Test app compiled successfully on both runtimes: 0 errors

## Test Execution Results

Full test suite executed on BC-28 (container BC-20260506, version 28.0.46665.48932):

- **190 test methods passed, 20 failed** across 26 codeunits (24 passed, 2 failed)
- **All product/variant-related tests passed** (Codeunit 139601 Shpfy Create Product Test - all methods pass)

### Failing tests (pre-existing, unrelated to this change):

1. **Codeunit 139606 - Shpfy Shipping Test** (1 failure):
   - `UnitTestExportFulfillmentServiceShipment` - Fails with API version check error ("The Shopify Admin API used by your current Shopify connector is no longer supported"). This is a platform-level version mismatch in the test environment.

2. **Codeunit 139608 - Shpfy Orders API Test** (19 failures):
   - All fail with: "The field Customer Disc. Group of table Sales Line contains a value (GCMURLJYCQ) that cannot be found in the related table (Customer Discount Group)"
   - This is a pre-existing test data setup issue where randomized customer data references a non-existent Customer Discount Group record. Unrelated to product export logic.

### Relevant passing tests (product export with variants):
- UnitTestCreateTempProductFromItemWithVariants
- UnitTestCreateTempProductFromItemWithVariantsAndExtendedText
- UnitTestCreateTempProductFromItemWithVariantsAndItemAttributes
- UnitTestCreateTempProductFromItemWithVariantsAndSKUIsVariantCode
- (and 40+ additional product/variant test methods)

### DIFF
--- a/src/Apps/W1/Shopify/App/src/Products/Codeunits/ShpfyProductExport.Codeunit.al
+++ b/src/Apps/W1/Shopify/App/src/Products/Codeunits/ShpfyProductExport.Codeunit.al
@@ -434,12 +434,11 @@ codeunit 30178 "Shpfy Product Export"
                 ShopifyVariant."Country/Region of Origin Code" := GetCountryISOCode(Item."Country/Region of Origin Code");
             end;
             if ShopifyVariant."Option 1 Name" = '' then
-                ShopifyVariant."Option 1 Name" := 'Variant';
-            if ShopifyVariant."Option 1 Name" = 'Variant' then
-                if ItemAsVariant then
-                    ShopifyVariant."Option 1 Value" := Item."No."
-                else
-                    ShopifyVariant."Option 1 Value" := ItemVariant.Code;
+                ShopifyVariant."Option 1 Name" := ResolveOption1Name(ShopifyVariant."Product Id");
+            if ItemAsVariant then
+                ShopifyVariant."Option 1 Value" := Item."No."
+            else
+                ShopifyVariant."Option 1 Value" := ItemVariant.Code;
             ShopifyVariant."Shop Code" := Shop.Code;
             ShopifyVariant."Item SystemId" := Item.SystemId;
             ShopifyVariant."Item Variant SystemId" := ItemVariant.SystemId;
@@ -879,6 +878,31 @@ codeunit 30178 "Shpfy Product Export"
                 TempCurrVariant := ShopifyVariant;
                 TempCurrVariant.Insert();
             end;
+    end;
+
+    local procedure ResolveOption1Name(ProductId: BigInteger): Text[50]
+    var
+        ShpfyVariant: Record "Shpfy Variant";
+        VariantOption1Name: Text[50];
+        FirstVariantName: Boolean;
+    begin
+        ShpfyVariant.SetRange("Product Id", ProductId);
+        if not ShpfyVariant.FindSet() then
+            exit('Variant');
+        FirstVariantName := true;
+        repeat
+            if ShpfyVariant."Option 1 Name" <> '' then
+                if FirstVariantName then begin
+                    VariantOption1Name := ShpfyVariant."Option 1 Name";
+                    FirstVariantName := false;
+                end
+                else
+                    if VariantOption1Name <> ShpfyVariant."Option 1 Name" then
+                        exit('Variant');
+        until ShpfyVariant.Next() = 0;
+        if not FirstVariantName then
+            exit(VariantOption1Name);
+        exit('Variant');
     end;
 
     local procedure RevertVariantChanges(VariantId: BigInteger)
--- a/src/Apps/W1/Shopify/App/src/Products/Codeunits/ShpfyProductExport.Codeunit.al
+++ b/src/Apps/W1/Shopify/App/src/Products/Codeunits/ShpfyProductExport.Codeunit.al
@@ -391,6 +391,7 @@ codeunit 30178 "Shpfy Product Export"
     var
         Product: Record "Shpfy Product";
         ItemAsVariant: Boolean;
+        JustResolvedOption1Name: Boolean;
     begin
         if Shop."Sync Prices" or OnlyUpdatePrice then
             if Item.Blocked or Item."Sales Blocked" then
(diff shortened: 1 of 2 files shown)


JesperSchulz (approved):

tVisionDeb:
@onbuyuka Hi! The build checks were auto-deleted after 72 hours. This PR has been waiting for review for 3 weeks and has approval from @JesperSchulz.

Could you:
1. Manually trigger a fresh build, OR
2. Let me know if I should close/reopen to trigger new checks?

Happy to address any feedback - the fix is ready and tested.

JesperSchulz:
Let's re-run the build pipeline and get a 2nd pair of 👀 on this one.

onbuyuka (changes requested):
Thanks for picking this up, @tVisionDeb! The `ResolveOption1Name` helper is exactly the shape we discussed in the issue, and I appreciate that the scope stayed tight to the single-axis overload. There is one issue I'd like to get sorted before this merges, plus a couple of smaller asks.

## 🚨 Regression: `Option 1 Value` is now overwritten on every variant update

The PR removed the outer `if ShopifyVariant."Option 1 Name" = 'Variant' then` guard around the value assignment, so the new block runs `Option 1 Value := Item."No."` / `ItemVariant.Code` **unconditionally**:

```al
if ShopifyVariant."Option 1 Name" = '' then
    ShopifyVariant."Option 1 Name" := ResolveOption1Name(ShopifyVariant."Product Id");
if ItemAsVariant then
    ShopifyVariant."Option 1 Value" := Item."No."
else
    ShopifyVariant."Option 1 Value" := ItemVariant.Code;
```

That was safe in the create path (the caller does `Clear(TempShopifyVariant)` first — see `CreateProductVariant` around L246/L258), but it breaks the update path. `UpdateProductVariant` at L839 passes the already-populated `ShopifyVariant` straight into `FillInProductVariantData` (L845), and that record is reached from L670 and L728 with the existing `Option 1 Name` / `Option 1 Value` already loaded from the DB.

Concrete repro on a Shopify-imported product (the exact scenario the issue was filed for):

| Existing variant in BC | Before PR | After PR |
|---|---|---|
| `Option 1 Name = 'Color'`, `Option 1 Value = 'Red'` | Guard skipped, `Red` preserved → `productVariantsBulkUpdate` ships `Color = 'Red'` ✅ | Value overwritten to `ItemVariant.Code` (e.g. `VAR001`) → `Color = 'VAR001'` is pushed to Shopify on the next sync ❌ |

So every re-sync of an existing Shopify-origin variant would silently corrupt its option value on the storefront. That's a worse regression than the original bug.

**Suggested minimal fix** — keep the original guard, but extend it to also cover the path where we just resolved a placeholder name:

```al
var
    JustResolvedOption1Name: Boolean;
begin
    ...
    if ShopifyVariant."Option 1 Name" = '' then begin
        ShopifyVariant."Option 1 Name" := ResolveOption1Name(ShopifyVariant."Product Id");
        JustResolvedOption1Name := true;
    end;
    if (ShopifyVariant."Option 1 Name" = 'Variant') or JustResolvedOption1Name then
        if ItemAsVariant then
            ShopifyVariant."Option 1 Value" := Item."No."
        else
            ShopifyVariant."Option 1 Value" := ItemVariant.Code;
```

That way:
- New BC variants (placeholder `'Variant'`) keep behaving as before.
- Shopify-imported variants with a real option name keep their existing value on update.
- The newly-resolved name path (your fix) still gets a sensible value seeded the first time.

## Tests

I don't see any test additions in this PR. Could you add coverage for the 5 scenarios from the issue comment ([ref](https://github.com/microsoft/BCApps/issues/7724#issuecomment-4372731886))? The update-path regression above would have been caught by the "manual override is preserved" scenario, and the new `ResolveOption1Name` helper deserves its own test for the ambiguous / no-variants / all-blank fallbacks.

`ShpfySkippedRecordLogTest.Codeunit.al` (around L307–L431) is a good model — it already exercises `FillInProductVariantData` directly, and you can pre-populate a `Shpfy Variant` record per scenario.

If you go down this path, please use the `[HttpClientHandler]` pattern (see `Test/docs/testing.md`) — the legacy `IsTestInProgress` / `ShpfyCommunicationEvents` approach is being phased out.

## Minor PR hygiene

- Title convention for Shopify PRs is `[Shopify] …` — could you update to `[Shopify] Fix #7724: Resolve Option 1 Name from existing Shopify Variant records`?

Happy to re-review once the value-overwrite guard is back in place and the tests are added. Thanks again!

<sub>internal ref: copilot-cli session 06b4db55-89b0-4eae-90ca-e3fba11a0186</sub>

tVisionDeb:
Closing temporarily to finish local test validation before re-submitting for review. Will reopen once tests pass.

PR #8049, opened 2026-05-07
(Bug 633226): [Subcontracting] Standard Task not propagated from Routing to Prod. Order Routing or Subcontracting Worksheet; prices not picked up

alexei-dobriansky:
## Summary

The Subcontracting Worksheet flow was dropping `Standard Task Code` between the Prod. Order Routing Line and the Requisition Line, the field on the worksheet was read-only, and Subcontractor Prices keyed on a Standard Task were therefore never applied. This change:

- Propagates `Standard Task Code` from `Prod. Order Routing Line` to `Requisition Line` in both worksheet-population paths (Calculate Subcontracts… on the worksheet, and direct PO creation from a Prod. Order Routing line).
- Makes the `Standard Task Code` field editable on the Subcontracting Worksheet so users can override it.
- Causes the standard-task-bound subcontractor price to flow through to the Subcontracting Purchase Order's `Direct Unit Cost`.

The bug enumerated four symptoms (Routing → Prod. Order Routing, Prod. Order Routing → Worksheet, field not editable, prices not picked up). Routing → Prod. Order Routing was already handled by `Prod. Order Routing Line.CopyFromRoutingLine`; no change there. The remaining three are addressed below.

### `SubcCalcSubcontractsExt.Codeunit.al`
Extended the existing `OnAfterTransferProdOrderRoutingLine` subscriber on report `Subc. Calculate Subcontracts` to also `Validate("Standard Task Code", ProdOrderRoutingLine."Standard Task Code")` on the new requisition line. Using `Validate` (not direct assignment) fires the field's OnValidate trigger in `Subc. RequisitionLine`, which calls `UpdateSubcontractorPrice` → `Subc. Price Management.GetSubcPriceForReqLine`. The price lookup already filters `Subcontractor Price` by `RequisitionLine."Standard Task Code"` exactly, so the standard-task-bound price is applied immediately.

### `SubcPurchaseOrderCreator.Codeunit.al`
Same fix in `InsertReqWkshLine`, which is the alternate worksheet-line builder used when a Subcontracting PO is created directly from a Prod. Order Routing Line (bypassing the worksheet UI).

### `SubcSubcontractingWorksheet.Page.al`
Removed `Editable = false` from the `Standard Task Code` field and added a tooltip clarifying that editing the value re-applies the matching subcontractor price. The field's OnValidate already triggers price re-lookup, so manual edits/clears on the worksheet immediately update `Direct Unit Cost`.

Fixes [AB#633226](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/633226)


ChethanT:
@spinkow

AleksandricMarko (approved):

AleksandricMarko (dismissed):

ChethanT (approved):

ChethanT (approved):

VolodySyn (dismissed):

SPinkow:
@ChethanT:
Please finish this after WIP Item Transfer PR and Renaming of Subcontracting Fields for IT

ChethanT (dismissed):
We will wait for WIP Item work to be done before merging this

alexei-dobriansky:
### Analysis

**Scenario Validity:**
Valid, clear, and important. AB#633226 (with the `[AI-REPRO]` comment, clarity score 6/10) enumerates four concrete symptoms in the Subcontracting routing → worksheet → PO chain and prescribes the expected end-to-end propagation. Subcontracting pricing is financially sensitive, and standard-task-bound prices are a documented configuration surface — dropping the key silently makes them unreachable. The Routing → Prod. Order Routing leg was already handled by `Prod. Order Routing Line.CopyFromRoutingLine`, so the PR correctly scopes itself to the three remaining symptoms.

**Correctness:**
The fix has three change points; each lands correctly:

1. `SubcCalcSubcontractsExt.Codeunit.al` — the `OnAfterTransferProdOrderRoutingLine` subscriber on report `Subc. Calculate Subcontracts` now `Validate`s `Standard Task Code` from `ProdOrderRoutingLine`. Using `Validate` (not direct assignment) is deliberate: the OnValidate trigger in `SubcRequisitionLine.TableExt.al:25-33` is guarded by `(Type = Type::Item) and ("No." <> '') and ("Prod. Order No." <> '') and (xRec."Standard Task Code" <> "Standard Task Code")`. At the moment the subscriber fires, the report has already populated Type, No., Prod. Order No., and Vendor No., so the guard is satisfied when the routing carries a non-blank task code, and the call falls through to `UpdateSubcontractorPrice` → `Subc. Price Management.GetSubcPriceForReqLine` (`SubcPriceManagement.Codeunit.al:403`). That procedure strictly filters `SetRange("Standard Task Code", RequisitionLine."Standard Task Code")` (line 419), so the std-task-bound price wins when one exists.

2. `SubcPurchaseOrderCreator.Codeunit.al:541` — the new `Validate` sits *after* `Prod. Order No.` (line 518), `Work Center No.` (line 534), and `Vendor No.` (line 537) have been set, so the OnValidate guard's preconditions hold and the price re-lookup runs against fully-initialized line state. The xRec/Rec comparison works on the in-memory record because xRec mirrors Rec prior to the first Validate of a new field.

3. `SubcSubcontractingWorksheet.Page.al:244` — dropping `Editable = false` allows manual override; the field's existing OnValidate handles re-pricing.

**No-match safety:** When the routing has a Standard Task Code but no subcontractor price is configured for it, `GetSubcPriceForReqLine` does `FindLast()`; if no row matches, the entire `Direct Unit Cost := DirectCost` assignment (line 442) is skipped because it's inside the `if FindLast() then begin … end` block. The Direct Unit Cost that was computed earlier in the flow is preserved — no silent zero-out, no regression to the un-tagged-price scenario.

**Side Effects:**
- The first leg of the chain (`CopyFromRoutingLine`) was already correct, so propagation now goes through end-to-end. No code becomes unreachable.
- Routings with a blank Standard Task Code: xRec = Rec = `''`, OnValidate guard short-circuits, no price re-lookup — behavior preserved.
- Routings with std task but no std-task-bound price: gracefully falls back to whatever was computed earlier (per the no-match safety above).
- The page becoming editable is a UX expansion, not a contraction; the OnValidate already shipped, so user edits were already supported at the data layer.
- Event subscribers / partner extensions on the requisition line's OnValidate of `Standard Task Code` will now fire in two new code paths. The field is in `SubcRequisitionLine.TableExt.al` (BCApps-owned), so the surface that could subscribe is limited and the contract being added is consistent with what already fires on manual edit.

**Risk Assessment: Low.**
- Narrow surface: two single-line `Validate` insertions and one removed `Editable = false`.
- Financial area, but the lookup function is conservative (no overwrite on no-match) and the strict `SetRange` on Standard Task Code is the documented price-table contract.
- **If we don't make this change:** users who configure subcontractor prices keyed on Standard Task can never have those prices applied through the Subcontracting Worksheet → PO flow; the routing → prod-order leg silently drops the key and the worksheet field can't be set manually. There is no workaround short of editing the resulting Purchase Line's Direct Unit Cost by hand on every PO, which defeats the price-table feature.

**Test Coverage:**
`StandardTaskCodePropagatedAndDrivesSubcPriceLookup` (220 lines added in `SubcSubcontractingTest.Codeunit.al`) is thorough:
- Sets up two subcontracting routing lines on the same work center — one with a Standard Task, one without — and two subcontractor prices (PriceWithStdTask = 2 × PriceWithoutStdTask) so the assertions discriminate between the two pricing paths.
- Asserts Standard Task Code propagation Routing → Prod. Order Routing Line.
- Runs `Subc. Calculate Subcontracts`, asserts both worksheet lines get the *correct* Direct Unit Cost (tagged vs un-tagged), proving the new Validate in the subscriber actually drives the price lookup — not just sets the field.
- Exercises the editability change by `Validate`-clearing and re-setting Standard Task Code on the worksheet line and asserting Direct Unit Cost flips correctly each way.
- Runs `Carry Out Action Msg. - Req.` and asserts the resulting Subcontracting Purchase Lines carry the right Direct Unit Cost per operation.

Gaps worth noting (non-blocking):
1. **Direct-PO-from-routing path (`SubcPurchaseOrderCreator.InsertReqWkshLine`) is not exercised by the new test.** All test coverage hits the Calculate Subcontracts (subscriber) path. The InsertReqWkshLine fix is symmetric and analytically correct, but unit-tested only through code review, not regression suite.
2. The page-level `Editable = true` is exercised at the record/Validate layer, not at the TestPage layer (no `TestPage "Subc. Subcontracting Worksheet"` interaction). The data behavior is what matters and is covered, so this is informational only.

### Recommendation: Accept with Suggestions

1. **Add a focused test for `SubcPurchaseOrderCreator.InsertReqWkshLine`** — create a Subcontracting Purchase Order directly from a Prod. Order Routing Line that carries a Standard Task Code, with a std-task-bound subcontractor price, and assert the resulting Purchase Line's `Direct Unit Cost` matches the tagged price. This is the only changed code path not exercised by the existing test, and it shares structure with the worksheet path so the regression cost is real if it ever drifts.
2. **(Optional) Consider a no-matching-price regression assertion** — i.e., routing has a Standard Task Code but no price is configured for it; assert the worksheet line keeps the un-tagged-or-default Direct Unit Cost rather than being zeroed. The current code is correct here because of the `if FindLast() then` guard, but the assertion would lock that behavior in.

ChethanT (approved):

AleksandricMarko (approved):

PR #8051, opened 2026-05-07
[Edit in Excel] Expose ExternalizeODataObjectName and add Page Control Field publisher

### LINKED ISSUE #5390: [BC Idea]: Edit in Excel export personalized fields
### BC Idea Link

https://experience.dynamics.com/ideas/idea/?ideaid=b220967d-37d3-e911-b083-0003ff68ac6b

### Description

> Currently, edit in excel when using the export function, it exports every field in the table, regardless of personalization done in Dynamics 365 Business Central. I would like the system, when exporting, to be able to take into account the personalization currently in place. Additionally, with the impact of dimensions, I'd like the system to add the 8 dimension fields available on the journals and also export those in edit in excel.

As a first option - expose `ExternalizeODataObjectName` from `codeunit 1482 "Edit in Excel Impl."` in `codeunit 1481 "Edit in Excel"` - then at least other app can utilize naming without reinventing the bicycle and copy implementer with their logic.

As a second option I propose at least add a publisher where other apps can define what fields to allow in the payload in `codeunit 1482 "Edit in Excel Impl."`:
```al
...
    local procedure SetupFieldColumnBindings(var EditinExcelWorkbook: Codeunit "Edit in Excel Workbook"; PageNo: Integer)
...
        PageControlField.SetRange(PageNo, PageNo);
        PageControlField.SetCurrentKey(Sequence);
        PageControlField.SetAscending(Sequence, true);
        EditInExcel.OnPageControlFieldFilterBeforeFindSetInSetupFieldColumnBindings(PageControlField); // new line
        if PageControlField.FindSet() then
...
```
and event itself in `codeunit 1481 "Edit in Excel"`
```al
    /// <summary>
    /// This event is called before applying filters to the Page Control Field in SetupFieldColumnBindings procedure.
    /// </summary>
    /// <param name="PageControlField"></param>
    /// <param name="PageId"></param>
    internal procedure OnPageControlFieldFilterBeforeFindSetInSetupFieldColumnBindings(var PageControlField: Record "Page Control Field"; PageId: Integer)
    begin
    end;
```

Internal work item: [AB#325948](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/325948)

Franco111000:
#### Summary

Adds two extension points to the public Edit in Excel codeunit so other apps can integrate with Edit in Excel without duplicating internal conversion logic and without forking the Page Control Field selection.

- Exposes `ExternalizeODataObjectName` as a public procedure on codeunit 1481 "Edit in Excel". The conversion already exists on the internal codeunit 1482 "Edit in Excel Impl." but was not reachable from other apps. Callers can now derive the same OData externalized field and service names that Edit in Excel uses when generating the workbook payload.

- Adds `OnPageControlFieldFilterBeforeFindSetInSetupFieldColumnBindings` on codeunit 1481, raised inside `SetupFieldColumnBindings` after PageNo and Sequence have been applied but before `FindSet`. Subscribers receive the `Page Control Field` record by reference and can add filters or change the sort order to control which fields end up in the workbook (for example, to honor user personalization).

Tests added for the public wrapper round-trip and for event invocation with the correct PageId.

#### Work Item(s)
Fixes #5390

Fixes [AB#325948](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/325948)

### DIFF
--- a/src/System Application/App/Edit in Excel/src/EditInExcel.Codeunit.al
+++ b/src/System Application/App/Edit in Excel/src/EditInExcel.Codeunit.al
@@ -6,6 +6,7 @@
 namespace System.Integration.Excel;
 
 using System.Integration;
+using System.Reflection;
 
 /// <summary>
 /// This codeunit provides an interface to running Edit in Excel for a specific page.
@@ -81,6 +82,18 @@ codeunit 1481 "Edit in Excel"
         EditInExcelImpl.GetEndPointAndCreateWorkbookWStructuredFilter(TenantWebService."Service Name", EditinExcelFilters, '');
     end;
 
+    /// <summary>
+    /// Converts a name (e.g. a field name or service name) to its OData externalized form. Mimics the conversion the Edit in Excel runtime applies to field and service names when generating the workbook payload, so that other apps can derive the same external names without duplicating the conversion logic.
+    /// </summary>
+    /// <param name="Name">The name to convert.</param>
+    /// <returns>The OData-externalized name.</returns>
+    procedure ExternalizeODataObjectName(Name: Text): Text
+    var
+        EditInExcelImpl: Codeunit "Edit in Excel Impl.";
+    begin
+        exit(EditInExcelImpl.ExternalizeODataObjectName(Name));
+    end;
+
 
     /// <summary>
     /// This event is called when Edit in Excel is invoked, handling JSON structured filters. It also allows overriding the Edit in Excel functionality.
@@ -106,6 +119,16 @@ codeunit 1481 "Edit in Excel"
     //  <param name="Handled">Specifies whether the event has been handled and no further execution should occur.</param>
     [IntegrationEvent(false, false)]
     internal procedure OnEditInExcelWithFilters(ServiceName: Text[240]; var EditinExcelFilters: Codeunit "Edit in Excel Filters"; SearchFilter: Text; var Handled: Boolean)
+    begin
+    end;
+
+    /// <summary>
+    /// This event is raised inside SetupFieldColumnBindings before iterating the Page Control Field record set, after the procedure has applied the PageNo filter and Sequence sort order. Subscribers can add further filters or change the sort order to control which page fields are exported to Excel (for example, to honor user personalization).
+    /// </summary>
+    /// <param name="PageControlField">The Page Control Field record, already filtered by PageNo and sorted by Sequence. Subscribers may add further filters.</param>
+    /// <param name="PageId">The ID of the page being exported.</param>
+    [IntegrationEvent(false, false)]
+    internal procedure OnPageControlFieldFilterBeforeFindSetInSetupFieldColumnBindings(var PageControlField: Record "Page Control Field"; PageId: Integer)
     begin
     end;
 }
--- a/src/System Application/App/Edit in Excel/src/EditInExcel.Codeunit.al
+++ b/src/System Application/App/Edit in Excel/src/EditInExcel.Codeunit.al
@@ -123,12 +123,12 @@ codeunit 1481 "Edit in Excel"
     end;
 
     /// <summary>
-    /// This event is raised inside SetupFieldColumnBindings before iterating the Page Control Field record set, after the procedure has applied the PageNo filter and Sequence sort order. Subscribers can add further filters or change the sort order to control which page fields are exported to Excel (for example, to honor user personalization).
+    /// This event is raised before iterating the Page Control Field record set used to build the Edit in Excel workbook column bindings. Subscribers can add further filters or change the sort order to control which page fields are exported to Excel (for example, to honor user personalization).
     /// </summary>
-    /// <param name="PageControlField">The Page Control Field record, already filtered by PageNo and sorted by Sequence. Subscribers may add further filters.</param>
+    /// <param name="PageControlField">The Page Control Field record with relevant filters and sorting already applied. Subscribers can add further filters.</param>
     /// <param name="PageId">The ID of the page being exported.</param>
     [IntegrationEvent(false, false)]
-    internal procedure OnPageControlFieldFilterBeforeFindSetInSetupFieldColumnBindings(var PageControlField: Record "Page Control Field"; PageId: Integer)
+    internal procedure OnBeforePageControlFieldFindSet(var PageControlField: Record "Page Control Field"; PageId: Integer)
     begin
     end;
 }
--- a/src/System Application/Test/Edit in Excel/src/EditInExcelTest.Codeunit.al
+++ b/src/System Application/Test/Edit in Excel/src/EditInExcelTest.Codeunit.al
@@ -9,6 +9,7 @@ using System;
 using System.Integration;
 using System.Integration.Excel;
 using System.TestLibraries.Integration.Excel;
+using System.Reflection;
 using System.TestLibraries.Utilities;
 codeunit 132525 "Edit in Excel Test"
 {
@@ -21,6 +22,8 @@ codeunit 132525 "Edit in Excel Test"
         EditInExcel: Codeunit "Edit in Excel";
         IsInitialized: Boolean;
         EventServiceName: Text[240];
+        EventPageControlFieldFilterPageId: Integer;
+        EventPageControlFieldFilterFired: Boolean;
 
     [Test]
     procedure TestEditInExcelCreatesWebService()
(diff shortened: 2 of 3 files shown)


PaulFurlet (approved):
I like that ExternalizeODataObjectName wrapper is added - thing that I did not think through back that time.
And Test unit looks good to me.

Franco111000:
Thanks @PaulFurlet, appreciate the review and the approval!

JesperSchulz (dismissed):

src/System Application/Test/Edit in Excel/src/EditInExcelTest.Codeunit.al
```diff
@@ -9,6 +9,7 @@ using System;
 using System.Integration;
 using System.Integration.Excel;
 using System.TestLibraries.Integration.Excel;
+using System.Reflection;
 using System.TestLibraries.Utilities;
 codeunit 132525 "Edit in Excel Test"
```
JesperSchulz, line 15:
```suggestion
namespace System.Test.Integration.Excel;

using System;
using System.Integration;
using System.Integration.Excel;
using System.Reflection;
using System.TestLibraries.Integration.Excel;
using System.TestLibraries.Utilities;

codeunit 132525 "Edit in Excel Test"
```

src/System Application/Test/Edit in Excel/src/EditInExcelTest.Codeunit.al
```diff
@@ -399,4 +403,73 @@ codeunit 132525 "Edit in Excel Test"
...
+    begin
+        // [SCENARIO] The public ExternalizeODataObjectName procedure on codeunit "Edit in Excel" exposes the same conversion as the internal implementation, so other apps can match Edit in Excel OData externalized names without duplicating the conversion logic.
+        Init();
+
+        // [GIVEN] A name that triggers OData escaping (digit prefix, en dash, space)
+        Input := '3 lager – reklassfication field';
+
+        // [WHEN] The public wrapper and the internal implementation are called with the same input
+        PublicResult := EditInExcel.ExternalizeODataObjectName(Input);
+        ImplResult := EditinExcelTestLibrary.ExternalizeODataObjectName(Input);
```
encimita: Here we are testing that the two functions return the same output, not that they return the _expected_ output.
And in fact, if you look at what `ExternalizeOdataObjectName` in the Library does, it just calls down to the same code as the other function.

Instead, hardcode the expected output, so something like this (pseudocode):

```
        Input := '3 lager – reklassfication field';
        ExpectedOutput := '_x4567_ lager _x1234_ reklassfication field';

        PublicResult := ...;

        LibraryAssert.AreEqual(PublicResult, ExpectedOutput, '...');
```

src/System Application/App/Edit in Excel/src/EditInExcel.Codeunit.al
```diff
@@ -108,4 +121,14 @@ codeunit 1481 "Edit in Excel"
     internal procedure OnEditInExcelWithFilters(ServiceName: Text[240]; var EditinExcelFilters: Codeunit "Edit in Excel Filters"; SearchFilter: Text; var Handled: Boolean)
     begin
     end;
+
+    /// <summary>
+    /// This event is raised inside SetupFieldColumnBindings before iterating the Page Control Field record set, after the procedure has applied the PageNo filter and Sequence sort order. Subscribers can add further filters or change the sort order to control which page fields are exported to Excel (for example, to honor user personalization).
+    /// </summary>
+    /// <param name="PageControlField">The Page Control Field record, already filtered by PageNo and sorted by Sequence. Subscribers may add further filters.</param>
```
encimita: Don't be too specific in the param definition. Tomorrow the filters that are set might be different. Go for something a bit more generic like (please adjust for casing and typos) "The page control field record, with relevant filters and sorting already applied. Subscribers can add further filters".

src/System Application/Test/Edit in Excel/src/EditInExcelTest.Codeunit.al
encimita: I'd also suggest to add a couple more examples to test against. Since this is now exposed to all partners out there, we want it to be extra solid. Add an example where no modification happens (e.g. "CustomerNumber"), one with arabic or japanese characters, and maybe a couple more of your discretion :)

src/System Application/Test/Edit in Excel/src/EditInExcelTest.Codeunit.al
```diff
@@ -399,4 +403,73 @@ codeunit 132525 "Edit in Excel Test"
...
+    begin
+        EventPageControlFieldFilterPageId := 0;
+        EventPageControlFieldFilterFired := false;
+    end;
+
+    [EventSubscriber(ObjectType::Codeunit, Codeunit::"Edit in Excel", 'OnPageControlFieldFilterBeforeFindSetInSetupFieldColumnBindings', '', false, false)]
+    local procedure CapturePageControlFieldFilterBeforeFindSet(var PageControlField: Record "Page Control Field"; PageId: Integer)
+    begin
+        EventPageControlFieldFilterPageId := PageId;
+        EventPageControlFieldFilterFired := true;
```
encimita: Do we expect this event to be raised several times throughout one test run? (this is a genuine question, as I don't remember the details of the underlying code)

If not, you can make this even more strict by adding something like

```
    if EventPageControlFieldFilterFired then
        Error('Event OnPageControlFieldFilterBeforeFindSetInSetupFieldColumnBindings is expected to be raised only once');
```

src/System Application/App/Edit in Excel/src/EditInExcel.Codeunit.al
```diff
@@ -108,4 +121,14 @@ codeunit 1481 "Edit in Excel"
...
+    /// <param name="PageId">The ID of the page being exported.</param>
+    [IntegrationEvent(false, false)]
+    internal procedure OnPageControlFieldFilterBeforeFindSetInSetupFieldColumnBindings(var PageControlField: Record "Page Control Field"; PageId: Integer)
```
encimita: This is a very C/AL way of naming events, IMHO. I'd prefer a shorter name like `OnBeforeFindPageControlFields` or `OnBeforePageControlFieldFindSet`. The fact that this event is raised in SetupFieldColumnBindings is an implementation detail that does not need to be specified in the event name. In 10 years from now it might be raised from somewhere else (within the Edit in Excel flow that the codeunit scopes to of course).

encimita:
Left a few comments just because Edit in Excel has been historically a painful integration to maintain, so I am extra picky when it comes to new code tied to it. But the change overall makes sense! Thanks for taking the time :)

src/System Application/Test/Edit in Excel/src/EditInExcelTest.Codeunit.al
PaulFurlet: I would not agree on specifying exactly what we expect here.
It should be within test of the original procedure.
The purpose of the exposed function is to allow external app to call implementer.
And we need to check whether the "hidden" has the same result as "exposed", so custom fields follow the same naming rule as standard. If there is an issue - it must be caught and fixed in the original implementer, not in the "custom" code.

Franco111000:
Hi @encimita, thanks for the framing! Pushed `5b32fd938` covering all four asks:

1. Event renamed to `OnBeforePageControlFieldFindSet`. I went with this over `OnBeforeFindPageControlFields` because it puts the record subject first (matches the var name); correct me if I'm wrong, happy to switch.
2. Param doc made generic, no more PageNo / Sequence baked in.
3. Wrapper test now asserts hardcoded outputs and covers Arabic, Japanese, and accented Latin BMP cases. I thought this was better than re-covering the ASCII edge cases, since `TestEditInExcelNumberInFieldNameReplacement` already does that with hardcoded outputs.
4. Switched the fires-once tracker to a counter; the literal `Error if fired` would have broken sibling tests that also call `EditPageInExcel` (subscriber binding stays global once `Init()` runs).

src/System Application/Test/Edit in Excel/src/EditInExcelTest.Codeunit.al
Franco111000: Hi @PaulFurlet, fair point. I went with hardcoded outputs because @encimita explicitly asked for them, but you're right that this duplicates what `TestEditInExcelNumberInFieldNameReplacement` already does (it asserts hardcoded outputs against the impl), and the delegation-equivalence shape we had originally is arguably the cleaner thing to test from the wrapper.

@encimita, given Paul's argument, would you like me to revert to the delegation-equivalence pattern? I can keep the extra Arabic / Japanese / accented inputs you asked for, just comparing wrapper output against impl output instead of hardcoded strings, so conversion-rule correctness stays in `TestEditInExcelNumberInFieldNameReplacement` where it already lives. Correct me if I'm wrong on either side.

encimita: I did not spot that these cases were already covered by `TestEditInExcelNumberInFieldNameReplacement`! Thanks @Franco111000 for pointing that out.

I am still confused about what the test was previously covering (at the time of my original comment). It was testing that our test library function returns the same as what the facade exposes, but these literally both call down to the same function. What are we trying to secure against? That the test library function flips to something else? Or that the facade/exposed function flips to something else? @PaulFurlet

Regardless, if this is what we want to test, we should then secure against all corner cases that we have already listed in `TestEditInExcelNumberInFieldNameReplacement` rather than just one. If we deemed these strings risky and worth testing against when evaluating the implementation behaviour, then they are also risky when it comes to the two functions (implementation and exposed) diverging?

Franco111000: Good question. The way I see it, the test isn't really about conversion correctness (that lives in `TestEditInExcelNumberInFieldNameReplacement`), it's about refactor drift between the two wrappers. Concrete scenarios it would catch but the impl test would not:

- Impl gets a new param; public wrapper passes a default, test library wrapper still passes the old form, outputs diverge on some inputs.
- One wrapper gets a "fast path" optimization that drifts from the other.
- Public wrapper gains pre-processing (trim, normalize) that the test library wrapper doesn't.
- Impl gets obsoleted, one wrapper inlines a copy that goes stale.

If you both agree that's the right framing, I'll rewrite the wrapper test to delegation-equivalence (Paul's shape) and mirror all the inputs from `TestEditInExcelNumberInFieldNameReplacement` plus the Arabic / Japanese / accented Latin extras (encimita's "extra solid for partners" intent), no hardcoded outputs. Correct me if I'm wrong on the framing.

PaulFurlet: @encimita I would say, it is completely non-obligatory, just in case existing Edit in Excel Test Library starts to call something else, it might have different output than we have.
So it can look like we are cross-testing :)

BTW, I noticed there are two different implementers of `ExternalizeODataObjectName`:
* [codeunit 1482 "Edit in Excel Impl."](https://github.com/microsoft/BCApps/blob/664265cbb8e78f0405791fd6e2a585c7be0d03c3/src/System%20Application/App/Edit%20in%20Excel/src/EditinExcelImpl.Codeunit.al#L201)
* [codeunit 9751 "Web Service Management Impl."](https://github.com/microsoft/BCApps/blob/664265cbb8e78f0405791fd6e2a585c7be0d03c3/src/System%20Application/App/Web%20Service%20Management/src/WebServiceManagementImpl.Codeunit.al#L678) - looks relatively similar, probably would be worth using one approach

Franco111000: Good catch on spotting both, and you're right they're both there. Looking at the diff though, I think they're more divergent than they appear at first:

- CU 9751 (WSM) doesn't byte-encode digit prefixes, apostrophes, plus signs, or en-dash. `'`, `+` are treated as plain special chars (replaced with `_`); en-dash isn't in the special set at all.
- CU 1482 (Edit in Excel) does all four encodings (`_x0033_`, `_x0027_`, `_x002B_`, `_x2013_`) and has an `IsByteValueUnderscore` dictionary so those byte-encoded underscores don't get deduped away.

So they only agree on plain ASCII like `'Customer No.'`. They diverge on `'3field'`, `"name's"`, `'c+c'`, anything with en-dash.

Git blame shows both came in the same 2023 initial code drop. The EditInExcel one then got progressively stricter (`IsByteValueUnderscore` dedup exception added in 2024, byte-value uppercase in 2025) to match what the OData payload generator produces on the server. The WSM one hasn't been touched since 2023, probably because service names with leading digits get prefixed with `WS` elsewhere and don't need byte-encoding.

I think they intentionally serve different jobs: WSM normalizes service NAMES used in URL paths (coarser), EditInExcel converts FIELD NAMES inside the workbook payload (stricter, has to round-trip). Unifying them would be breaking for one side or the other depending on which rule set wins. Probably worth a separate issue with the System App + Web Service teams if it's worth pursuing, but I'd lean toward leaving it out of this PR. Correct me if I'm wrong.

encimita: @Franco111000 I see. Thanks for the explanation! Fair, that sounds good :) Go ahead!
If you want to make me extra happy you can add the new accented/nonlatin strings in `TestEditInExcelNumberInFieldNameReplacement` as well! 😄 

@PaulFurlet Got it! Makes sense as an extra level of testing (as I mentioned in another comment, Edit in Excel has been historically a bit hard to maintain, so any additional testing layer is very welcome)!
For the two different functions, they do look similar indeed! With a quick code search I can see the WebService one is only called in some DemoTool functions (in Microsoft code; plus any partner extension out there of course). I agree they should probably be merged at some point, but I don't have an overview if that can cause side effects (since it would be a breaking change if we start "encoding" something differently in one of the two code paths).

PaulFurlet: @Franco111000 you did much more research than me, I just spotted similarity without intent to do anything about it :)
I agree we should concentrate on current implementation changes, nothing more.
And my vote is pro-cross-testing output, not copying existing checks.

encimita: Agree to leave Web Service Management out of the picture for now! :) And to have the test in its original form (before my comment), but with a bunch more test case strings.

Franco111000: Pushed `4d666ffc5` with the consensus design:

- Wrapper test back to delegation-equivalence (@PaulFurlet's preferred shape), with five inputs: clean ASCII (`CustomerNumber`), ASCII with a space (`Posting Date`), Arabic (`مرحبا`), Japanese hiragana (`こんにちは`), accented Latin (`café`). No hardcoded strings, no copying of existing checks.
- Also added the same three non-ASCII inputs to `TestEditInExcelNumberInFieldNameReplacement` with hardcoded pass-through expected outputs, as the extra-happy bonus @encimita asked for. Locks down the impl's behavior on non-ASCII BMP inputs so a future change that breaks them would be caught even without going through the wrapper.

Correct me if I'm off on either.

encimita: Great! One last thing @Franco111000 : could you also re-add the original `'3 lager – reklassfication field'` input that you had in the original iteration to the test? I think it was good to have one test string starting with a number as well.

Franco111000: Done, added in `1455bd6e1` as the sixth input. Now covers digit prefix, en-dash, and space rules in one string.

encimita (approved):

JesperSchulz (approved):

JesperSchulz:
Let's run the build once more. We've seen many instabilities lately.

JesperSchulz:
Let's re-run CI.

PR #8072, opened 2026-05-09
[Shopify] Use ISO country code when filtering Shpfy Tax Area on customer/company export

### LINKED ISSUE #6129: [Bug]: Shopify Connector: EU country code "EL" (Greece) incompatible with Shopify's ISO 3166-1 requirement
### Describe the issue

The Shopify Connector fails when exporting customers/companies with the EU country code "EL" for Greece. Business Central correctly uses "EL" for VIES (VAT Information Exchange System) validation and EU regulatory compliance, but Shopify's API requires the ISO 3166-1 alpha-2 standard code "GR". The connector does not map between these two legitimate standards.

### Expected behavior

The connector should map EL to GR when going from BC to Shopify and vice-versa

### Steps to reproduce

- Set up Business Central in an EU country (e.g., Denmark)
- Configure Country/Region for Greece with Code "EL" (correct for EU/VIES purposes)
- Create a customer with country code "EL"
- Attempt to export the customer to Shopify
- Export fails with GraphQL validation error

### Additional context

_No response_

Franco111000:
## What & why

`Shpfy Tax Area` is populated from Shopify's `shipsToCountries` GraphQL enum (and the bundled `provinces.yml`) and stores ISO 3166-1 alpha-2 codes (e.g. `GR`). The customer and company export procedures filtered it using BC's local `Country/Region Code` (e.g. `EL` for Greece per EU/VIES). For any country whose BC code differs from its ISO code, the Tax Area filter never matched and Province Code / Province Name silently failed to populate on the Shopify-side address.

This PR resolves the ISO code once via `Country/Region.Get` + `TestField("ISO Code")` and reuses it for both Tax Area filtering and the Shopify-side address assignment. The local ISO variable is declared as `Code[2]`, matching `Country/Region."ISO Code"` and the destination `Shpfy Customer Address."Country/Region Code"` / `Shpfy Company Location."Country/Region Code"` fields, which avoids the AA0139 overflow that blocked the previous attempt in PR #6130.

I also harmonized the order of operations in `Shpfy Company Export.FillInShopifyCompanyLocation` so the Company Information fallback runs before the county lookup, matching `Shpfy Customer Export.FillInShopifyCustomerData`. Today the company-side flow runs the county lookup with a possibly-empty `Customer."Country/Region Code"`, so a customer with no country falls through with blank province even when the company has a country configured.

## Linked work

Fixes #6129

## How I validated this

- Added two regression tests, `UnitTestFillInShopifyCustomerDataISOCountryCodeMapping` in `Shpfy Customer Export Test` and `UnitTestFillInShopifyCompanyLocationISOCountryCodeMapping` in `Shpfy Company Export Test`. Each creates a `Country/Region` fixture with `Code = 'EL'` and `"ISO Code" = 'GR'`, sets up a `Shpfy Tax Area` keyed by `'GR'`, and asserts the Shopify-side address `"Country/Region Code"` resolves to `'GR'` and Province Code / Name populate from the Tax Area.
- Existing `UnitTestFillInShopifyCustomerDataCounty` and `UnitTestFillInShopifyCustomerData` (and the company equivalents) are unaffected because the BC demo data they rely on (`'US'`) has matching BC code and ISO code.
- I did not run a clean local compile: the Shopify app's `.alpackages` is empty (AL-Go for PTE pattern) so it would require fetching symbols out of band. I rely on CI for the build. PR #6130's CI failure on AA0139 was the specific signal I designed the `Code[2]` local around to avoid.

## Risk & compatibility

- Behavioral change in `Shpfy Company Export`: when `Customer."Country/Region Code"` is blank and `Company Information` has a country, the county lookup now resolves against that fallback country. Today the company-side county block runs with an empty `Customer.Country` and finds nothing. This brings the company flow in line with the existing order in `Shpfy Customer Export`.
- The procedures preserve the existing "skip the address country assignment when `Country/Region` is missing" behavior, so a customer whose BC country code does not exist in `Country/Region` continues to leave the Shopify-side country blank rather than being assigned the raw BC code. This is intentionally different from PR #6130, which silently fell back to the BC code.
- No breaking changes, no permissions / upgrade / telemetry impact.
- Builds on the approach validated by @onbuyuka and @JesperSchulz before PR #6130 was closed for inactivity.

Fixes [AB#636783](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/636783)

### DIFF
--- a/src/Apps/W1/Shopify/App/src/Companies/Codeunits/ShpfyCompanyExport.Codeunit.al
+++ b/src/Apps/W1/Shopify/App/src/Companies/Codeunits/ShpfyCompanyExport.Codeunit.al
@@ -123,6 +123,7 @@ codeunit 30284 "Shpfy Company Export"
         TaxRegistrationIdMapping: Interface "Shpfy Tax Registration Id Mapping";
         CountyCodeTooLongErr: Text;
         PaymentTermsId: BigInteger;
+        ISOCountryCode: Code[2];
     begin
         TempCompanyLocation := CompanyLocation;
 
@@ -133,8 +134,20 @@ codeunit 30284 "Shpfy Company Export"
         CompanyLocation.City := Customer.City;
         CompanyLocation.Recipient := Customer.Name;
 
+        if (Customer."Country/Region Code" = '') and CompanyInformation.Get() then
+            Customer."Country/Region Code" := CompanyInformation."Country/Region Code";
+
+        // Shpfy Tax Area is keyed by Shopify's ISO 3166-1 alpha-2 codes (e.g. "GR")
+        // which can differ from BC's Country/Region Code (e.g. "EL" used for EU/VIES).
+        // Resolve once and reuse for both Tax Area filtering and the Shopify-side location.
+        if CountryRegion.Get(Customer."Country/Region Code") then begin
+            CountryRegion.TestField("ISO Code");
+            ISOCountryCode := CountryRegion."ISO Code";
+            CompanyLocation."Country/Region Code" := ISOCountryCode;
+        end;
+
         if Customer.County <> '' then begin
-            TaxArea.SetRange("Country/Region Code", Customer."Country/Region Code");
+            TaxArea.SetRange("Country/Region Code", ISOCountryCode);
             if not TaxArea.IsEmpty() then
                 case Shop."County Source" of
                     Shop."County Source"::Code:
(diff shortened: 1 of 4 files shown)


onbuyuka (approved):

Groenbech96:
@copilot pull main into branch

Groenbech96 (approved):

PR #8076, opened 2026-05-11
[28.0][Agent] Expose setting/getting model on Agent Task level

qutreson:
## What & why

Enabling developers using the agent SDK to set the model to use on agent tasks.

## Linked work

Fixes [AB#620625](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/620625)

## How I validated this

## Risk & compatibility

None

### DIFF
--- a/src/System Application/App/Agent/Interaction/Internal/AgentTaskBuilderImpl.Codeunit.al
+++ b/src/System Application/App/Agent/Interaction/Internal/AgentTaskBuilderImpl.Codeunit.al
@@ -20,6 +20,7 @@ codeunit 4310 "Agent Task Builder Impl."
         GlobalAgentUserSecurityId: Guid;
         GlobalTaskTitle: Text[150];
         GlobalExternalID: Text[2048];
+        GlobalModelId: Code[30];
 
     [Scope('OnPrem')]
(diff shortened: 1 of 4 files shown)


nikolakukrika (dismissed):

PredragMaricic (dismissed):

src/System Application/App/Agent/Interaction/Internal/AgentTaskBuilderImpl.Codeunit.al
```diff
@@ -65,6 +66,13 @@ codeunit 4310 "Agent Task Builder Impl."
         exit(this);
     end;
 
+    [Scope('OnPrem')]
```
thpeder, line 69: I haven't look through all of the application objects, so this might make sense. But Why is this here but not on any of the other functions?

thpeder (approved):

src/System Application/App/Agent/Interaction/Internal/AgentTaskBuilderImpl.Codeunit.al
qutreson: Discussed, app has a pattern of using Impl codeunits which are internal for the implementation of the public codeunits. This one just marks everything as OnPrem to ensure it's not used since Internal scope is not a 'security' feature.

nikolakukrika (approved):

PR #8085, opened 2026-05-11
[MCP] Server Features in MCP configuration: API Tools, Dynamic Tool Mode, Data Query Tools (Preview)

onbuyuka:
## Summary

Adds a **Server Features** surface to the MCP configuration card so admins can opt configurations into the **Data Query Tools (Preview)** MCP runtime, alongside the existing **Dynamic Tool Mode** and **API Tools** controls. Each feature is a self-contained handler behind an interface, so the card is data-driven and extensible. Activation persists on the real `MCP Configuration` fields shipped in [BC-Platform #44811](https://microsoft.ghe.com/bic/BC-Platform/pull/44811); only the Data Query Tools server runtime lands separately (see _Platform dependency_).

**Feature handlers (interface-based)**
- New `interface "MCP Server Features"` — `SetActive` / `IsActive` / `HasSettings` / `OpenSettings` / `Description` / `LoadSystemTools` / `TryGetParentFeature`. `enum "MCP Server Feature"` implements it, binding each value to a handler codeunit; resolving `ServerFeature := Rec.Feature` replaces the old per-feature `case` blocks.
- Both the `MCP Server Feature List` and `MCP System Tool List` pages are fully generic: each walks `ServerFeature.Ordinals()` + `FromInteger(...)` (the same pattern as `MCP Config Warning Type`) and dispatches every action — activation, settings, descriptions, **and system-tool loading** — through the handler. **Adding a feature is a new enum value + handler codeunit — no page edits.**
- Each feature owns its description, activation, settings, system tools, and rules; `SetActive` delegates to `codeunit "MCP Config Implementation"`, which reads/writes the real `MCP Configuration` fields (`EnableDynamicToolMode`, `EnableApiTools`, `EnableAlQueryTools`); `IsActive` reads them back.

**Card UX**
- `Server Features` list-part with per-row Activate / Deactivate / Configure, mirroring `page 7775 "Copilot AI Capabilities"` → `Copilot Capabilities GA`. **Configure** appears only for features that have settings — gated on a per-row `Configurable` flag set from the handler's `HasSettings()`. **Dynamic Tool Mode renders indented as a sub-feature of API Tools** (`TryGetParentFeature` + an `Indentation` column on the temp table).
  - **API Tools** — gates the **Available APIs** list-part (`Visible = not IsDefault and APIToolsActive`), where the admin curates which API pages / queries the client can reach. No settings.
  - **Dynamic Tool Mode** — its Configure dialog hosts `Discover Read-Only Objects`; activation flows to `EnableDynamicToolMode`, deactivation cascade-clears `DiscoverReadOnlyObjects`. **Enabling it requires API Tools to be enabled first** — `impl.EnableDynamicToolMode` enforces the gate (`APIToolsRequiredForDynamicErr`).
  - **Data Query Tools (Preview)** — contributes the Data Query Tools system tools when active. No settings.
- **Available APIs** (`page 8352`) — the curated API list. Its **`Select APIs`** action opens a **single lookup listing both API pages and API queries** (new temp `table "MCP API Object Buffer"` + `page "MCP API Object Lookup"`, populated by `impl.LookupAPIObjects`); the per-row `Object Id` lookup shares the same path. The part was renamed `'Available Tools'` → `'API Tools'` → **`'Available APIs'`** to disambiguate it from the API Tools *feature* row.
- The single settings `StandardDialog` (`page 8369 "MCP Server Feature Settings"`) is used only by Dynamic Tool Mode; its handler runs it via `OpenSettings`, writing on OK.
- The **System Tools** list-part (right-rail FactBox) is rebuilt by iterating the active features and calling each handler's `LoadSystemTools`.

**Configuration list (`page 8350`)**
- The **`Default`** field is read-only on the card; **Set as Default** / **Clear Default** actions (promoted) live on the configuration list instead.
- Read-only **API Tools** / **Data Query Tools** status columns (bound to `Rec.EnableApiTools` / `Rec.EnableAlQueryTools`) replaced the previous Dynamic Tool Mode / Discover Read-Only Objects columns.

**Public API**
- `codeunit 8350 "MCP Config"` exposes `EnableAPITools` and `EnableDataQueryTools` alongside the existing `EnableDynamicToolMode` / `EnableDiscoverReadOnlyObjects`. The two feature procedures read/write the `MCP Configuration.EnableApiTools` / `.EnableAlQueryTools` fields.

**New objects**
- `interface "MCP Server Features"`
- `enum 8351 "MCP Server Feature"` (`implements "MCP Server Features"`; `API Tools`, `Dynamic Tool Mode`, `Data Query Tools`)
- `table 8355 "MCP Server Feature"` (temporary; carries `Configurable` + `Indentation`)
- `table 8357 "MCP API Object Buffer"` (temporary; unifies API pages + queries for the Select APIs lookup)
- `page 8368 "MCP Server Feature List"`, `page 8369 "MCP Server Feature Settings"`, `page 8376 "MCP API Object Lookup"`
- handler codeunits `8369 "MCP API Tools Feature"`, `8368 "MCP Data Query Tools Feature"`, `8370 "MCP Dyn. Tool Mode Feature"`

**Deprecated**
- Pages `8353 "MCP API Config Tool Lookup"` / `8367 "MCP Query Config Tool Lookup"` are **obsoleted** (`ObsoleteState = Pending`, `ObsoleteTag = '29.0'`, wrapped in `#if not CLEAN29`) — superseded by the unified `MCP API Object Lookup` (`8376`). They're public System Application objects, so they're deprecated rather than deleted (removing them breaks dependent extensions — `AS0029`). Their internal `MCP Config Implementation` lookup procedures (`LookupAPIPageTools` / `LookupAPIQueryTools`) and the test-library wrappers were removed (neither is public surface).

**Permissions**
- `MCP - Objects` lists the new app tables (`MCP Server Feature`, `MCP API Object Buffer`).

## Platform dependency

**[BC-Platform PR #44811](https://microsoft.ghe.com/bic/BC-Platform/pull/44811) has merged, and this PR is productionized against it** — no mocks remain (`grep -rn "MOCK:|PLATFORM-PENDING" "src/System Application/App/MCP/"` is empty).

- **Activation** persists on `MCP Configuration.EnableApiTools` (field 8, `InitValue = true`) / `.EnableAlQueryTools` (field 9), read/written through four `MCP Config Implementation` procedures. An upgrade (`MCP Upgrade.EnableApiToolsOnExistingConfigurations`) turns API Tools on for **existing** configurations (`InitValue` only covers new ones).
- **Data Query Tools system tools** come from `"MCP Utilities".GetSystemToolsInDataQuery()`.

The platform uses the **"AL Query Tools"** field name; our UX stays **"Data Query Tools"** (it maps onto `EnableAlQueryTools`).

> The Data Query Tools **server runtime** (compile + execute of client-submitted AL queries) is a separate, larger platform dependency — this PR only *configures and advertises* the tools. See **Appendix E** in `docs/features/al-query-mcp-config-ui/design.md`.

## Test plan
- [x] `MCP Config Test` (`codeunit 130130`): activation tests asserting the real `EnableApiTools` / `EnableAlQueryTools` field state; a `Server Features` TestPage set on `page 8351 "MCP Config Card"` (the list shows all three features in order, `Configure` is enabled only on Dynamic Tool Mode, activating Dynamic Tool Mode flips its row Status and sets `EnableDynamicToolMode`, and the part is hidden on the default configuration; API Tools is enabled before Dynamic Tool Mode so the gate is satisfied); `TestLookupAPIObjects` exercising the unified API lookup; and Export/Import round-trip tests asserting the two new fields survive.
- [x] Manual walkthrough of the Server Features list / Configure / Activate-Deactivate cycle and the Select APIs lookup against a local BC instance.

Fixes [AB#631012](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/631012)


src/System Application/App/MCP/src/Configuration/Pages/MCPServerFeatureList.Page.al
```diff
@@ -0,0 +1,178 @@
...
+    var
+        ServerFeature: Interface "MCP Server Features";
+    begin
+        ServerFeature := Rec.Feature;
+        ServerFeature.SetActive(ParentSystemId, NewActive);
+
+        // Reload so a cascaded change to another feature (disabling API Tools turns off
+        // Dynamic Tool Mode) is reflected on every row, not just the one toggled.
+        Reload(ParentSystemId, ActionsEnabled);
+        CurrPage.Update();
```
onbuyuka:
```suggestion
        CurrPage.Update(false);
```

src/System Application/App/MCP/src/Configuration/Pages/MCPConfigToolList.Page.al
```diff
@@ -9,7 +9,7 @@ using System.Reflection;
 
 page 8352 "MCP Config Tool List"
 {
-    Caption = 'Available Tools';
+    Caption = 'Available APIs';
```
darjoo, line 12: Should the rest of the tooltips etc also say API instead of tool?

darjoo (dismissed):

ankosiorek (dismissed):

src/System Application/App/MCP/src/Configuration/Pages/MCPConfigToolList.Page.al
onbuyuka: Good catch — updated the tooltips on this page to say "API" instead of "tool" to match the renamed 'Available APIs' caption (Allow Read/Create/Modify/Delete/Bound Actions and API Version). The Object Type/Id/Name tooltips still say "object" since each row is backed by a page or query object. Done in 70be5a9.

nikolakukrika (dismissed):

gggdttt:
> 🟡 Medium Severity — Upgrade Iteration 1
> 
> ### Missing negative test for Dynamic Tool Mode + API Tools gate
> The test `TestEnableDynamicToolMode` was updated to first enable API Tools before enabling Dynamic Tool Mode, ensuring the positive path succeeds. However, there is no test that verifies calling `EnableDynamicToolMode(ConfigId, true)` when API Tools is disabled raises `APIToolsRequiredForDynamicErr`. Without a negative test, a future regression removing or relocating the guard would go undetected.
> 
> **Recommendation:**
> 
> * Add a test procedure (e.g., `TestEnableDynamicToolModeRequiresAPITools`) that calls `MCPConfig.EnableDynamicToolMode(ConfigId, true)` on a config where API Tools is disabled and asserts that the expected error is thrown.
> 
> ```
> [Test]
> procedure TestEnableDynamicToolModeRequiresAPITools()
> var
>     ConfigId: Guid;
> begin
>     ConfigId := CreateMCPConfig(false, false, true, false);
>     // API Tools is off by default on new config
>     asserterror MCPConfig.EnableDynamicToolMode(ConfigId, true);
>     Assert.ExpectedError('API Tools must be enabled');
> end;
> ```
> 
> _Line mapping was unavailable, so this was posted as an issue comment._
> 
> 👍 useful · ❤️ especially valuable · 👎 wrong - [reply with why](https://github.com/microsoft/BCAppsCampAIRHack/tree/main/docs)

This one looks make sense. Should we consider it?

gggdttt (dismissed):

darjoo (dismissed):

gggdttt (approved):

darjoo (approved):

PR #8116, opened 2026-05-12
Bug 634237: [Subcontracting] Create separate transfer orders per subcontracting purchase order

ChethanT:
## Summary
Create a separate subcontracting transfer order when the purchase order is different instead of appending lines to an existing transfer order with a mismatched purchase-order reference.

## Root Cause
The subcontracting transfer-order creation flow reused an open transfer order based on vendor and locations without scoping the lookup to the originating subcontracting purchase order. That allowed lines from different purchase orders to be mixed into the same transfer order.

## Fix
- scope transfer-order reuse to the same subcontracting purchase order
- keep each transfer order linked to its originating purchase order
- update the regression test to verify different purchase orders produce different transfer orders and preserve purchase-order linkage

## Test Coverage
- Subc. Subcontracting Test.CreateTransferOrderFromSecondSubcontractingOrderOpensReusedTransferOrder

[AB#634237](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/634237)


ChethanT:
@spinkow

AndreiPanko (approved):

AleksandricMarko (approved):

PR #8130, opened 2026-05-13
Fixes #8156

### LINKED ISSUE #8156: [Bug]: If at least one security filter is applied to the “Purchase Header” table, the user will have a problem when push ‘Add file from source document’ action on the “Email Attachments” subpage
### Describe the issue

Hi,
If at least one security filter is applied to the “Purchase Header” table, the user will have a problem when push ‘Add file from source document’ action on the “Email Attachments” subpage. Message: "Did not find any attachments related to this email"

With this event, we will be able to add the following condition:
SourceRecordRef.SecurityFiltering(SecurityFilter::Ignored);
before reading permissions.

Thx in advance,
Frank Neeckx

### Expected behavior

When reproducing, We expect to be able to select document attachments.

### Steps to reproduce

1. add a security filter to the puchase header table
2. Add attachments to a blanked purchase order
3. Push: send by mail on the blanked purchase order
4. Push action : ‘Add file from source document’ action on the “Email Attachments” subpage. 
5. Result : Error : "Did not find any attachments related to this email"

### Additional context

A new onbefore event to be added to the function : procedure FilterRemovedSourceRecords(var EmailRelatedRecord: Record "Email Related Record") in the [Codeunit][8900][Email Impl]

FrankNeeckx:
Hi,
If at least one security filter is applied to the “Purchase Header” table, the user will have a problem when push ‘Add file from source document’ action on the “Email Attachments” subpage. Message: "Did not find any attachments related to this email"

With this event, we will be able to add the following condition: SourceRecordRef.SecurityFiltering(SecurityFilter::Ignored); before reading permissions.

Thx in advance,
Frank Neeckx

## Linked work

Fixes #8156 

## How I validated this

### DIFF
--- a/src/System Application/App/Email/src/Email/EmailImpl.Codeunit.al
+++ b/src/System Application/App/Email/src/Email/EmailImpl.Codeunit.al
@@ -748,7 +748,7 @@ codeunit 8900 "Email Impl"
         SourceRecordRef: RecordRef;
         Handled: Boolean;
     begin
-        OnBeforeFilterRemovedSourceRecords(EmailRelatedRecord,Handled);
+OnBeforeFilterRemovedSourceRecords(EmailRelatedRecord, Handled);
         if Handled then
             exit;
         repeat
--- a/src/System Application/App/Email/src/Email/EmailImpl.Codeunit.al
+++ b/src/System Application/App/Email/src/Email/EmailImpl.Codeunit.al
@@ -746,7 +746,11 @@ codeunit 8900 "Email Impl"
     var
         AllObj: Record AllObj;
         SourceRecordRef: RecordRef;
+        Handled: Boolean;
     begin
+        OnBeforeFilterRemovedSourceRecords(EmailRelatedRecord,Handled);
+        if Handled then
+            exit;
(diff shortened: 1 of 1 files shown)


FrankNeeckx:
Hi,
Do we need to create a separate issue, for the BE localization, even though its a system app modification?

Thx
Frank Neeckx

Groenbech96:
Remove the ishandled and get it to build please.

src/System Application/App/Email/src/Email/EmailImpl.Codeunit.al
```diff
@@ -746,7 +746,11 @@ codeunit 8900 "Email Impl"
     var
         AllObj: Record AllObj;
         SourceRecordRef: RecordRef;
+        Handled: Boolean;
     begin
+OnBeforeFilterRemovedSourceRecords(EmailRelatedRecord, Handled);
```
nikolakukrika, line 751: We cannot add IsHandled events in this place.
Please use the different implementation.

nikolakukrika (changes requested):

PR #8170, opened 2026-05-15
Bug 634267: [Subcontracting] Transfer Order Reopen not persisting from factbox drill-down (Bug 634267)

ChethanT:
## Problem

When a Transfer Order is opened from the **Subcontracting Details Factbox** (e.g. clicking *No. of Transfer Orders* on a Purchase Order), actions like **Reopen** that modify `Rec` directly appear to succeed on screen but the change is silently lost when the page closes.

### Root Cause

`ShowTransferOrdersAndReturnOrder` in `SubcPurchFactboxMgmt` collected Transfer Headers using `Mark`/`MarkedOnly` and passed that marked record directly to `PageManagement.PageRun`/`PageRunList`. When the Transfer Order page opened bound to a marked record, operations that modify `Rec` (e.g. Reopen → `Rec.Validate(Status, Open); Rec.Modify()`) wrote to the marked record set in memory rather than the real database table. The change was discarded when the page closed.

Actions like **Post** were unaffected because they call `TransferHeader.Get()` internally on a separate variable, bypassing `Rec`.

## Fix

Introduced a separate `TransferHeaderToOpen` record variable that is populated from the real database:

- **Single record**: `TransferHeaderToOpen.Get(TransferHeader.\"No.\")` fetches the real DB record, then `PageManagement.PageRun(TransferHeaderToOpen)`.
- **Multiple records**: Uses `SelectionFilterManagement.GetSelectionFilterForTransferHeader` to build a \"No.\" filter from the marked set, applies it to `TransferHeaderToOpen`, then `PageManagement.PageRunList(TransferHeaderToOpen)`.

## Test

Added `FactboxDrilldownTransferOrderReopenPersists`:
1. Sets up subcontracting with transfer components
2. Creates and releases a Transfer Order
3. Opens it via `ShowTransferOrdersAndReturnOrder` (the factbox drill-down path)
4. Page handler invokes Reopen
5. Asserts the Transfer Order status persists as **Open** after the page closes

[AB#634267](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/634267)

### DIFF
--- a/src/Apps/W1/Subcontracting/App/src/Process/Codeunits/SubcPurchFactboxMgmt.Codeunit.al
+++ b/src/Apps/W1/Subcontracting/App/src/Process/Codeunits/SubcPurchFactboxMgmt.Codeunit.al
@@ -314,6 +314,8 @@ codeunit 99001560 "Subc. Purch. Factbox Mgmt."
                     end;
                 NoOfTransferHeaders > 1:
                     begin
+                        // As we do not expect more than a handful tranfer orders linked to the purchase order, there is no need to 
+                        // add extra processing if the number of records linked are more than allowed.
                         TransferHeaderToOpen.SetFilter("No.", SelectionFilterMgt.GetSelectionFilterForTransferHeader(TransferHeader));
                         PageManagement.PageRunList(TransferHeaderToOpen);
                     end;
--- a/src/Apps/W1/Subcontracting/App/src/Process/Codeunits/SubcPurchFactboxMgmt.Codeunit.al
+++ b/src/Apps/W1/Subcontracting/App/src/Process/Codeunits/SubcPurchFactboxMgmt.Codeunit.al
@@ -12,6 +12,7 @@ using Microsoft.Purchases.Document;
 using Microsoft.Purchases.History;
 using Microsoft.Utilities;
 using System.Reflection;
+using System.Text;
 
 codeunit 99001560 "Subc. Purch. Factbox Mgmt."
 {
@@ -229,9 +230,11 @@ codeunit 99001560 "Subc. Purch. Factbox Mgmt."
         ProductionOrder: Record "Production Order";
         PurchaseLine: Record "Purchase Line";
         TransferHeader: Record "Transfer Header";
+        TransferHeaderToOpen: Record "Transfer Header";
         TransferLine: Record "Transfer Line";
         DataTypeManagement: Codeunit "Data Type Management";
         PageManagement: Codeunit "Page Management";
+        SelectionFilterMgt: Codeunit SelectionFilterManagement;
         RecRef: RecordRef;
         NoOfTransferHeaders: Integer;
     begin
@@ -305,10 +308,15 @@ codeunit 99001560 "Subc. Purch. Factbox Mgmt."
                 NoOfTransferHeaders = 0:
                     Message(NoTransferExistsMsg);
                 NoOfTransferHeaders = 1:
-                    if TransferHeader.FindFirst() then
-                        PageManagement.PageRun(TransferHeader);
+                    if TransferHeader.FindFirst() then begin
+                        TransferHeaderToOpen.Get(TransferHeader."No.");
+                        PageManagement.PageRun(TransferHeaderToOpen);
+                    end;
                 NoOfTransferHeaders > 1:
-                    PageManagement.PageRunList(TransferHeader);
+                    begin
+                        TransferHeaderToOpen.SetFilter("No.", SelectionFilterMgt.GetSelectionFilterForTransferHeader(TransferHeader));
+                        PageManagement.PageRunList(TransferHeaderToOpen);
+                    end;
             end;
         end;
     end;
--- a/src/Apps/W1/Subcontracting/Test/src/Codeunits/Tests/SubcSubcontractingTest.Codeunit.al
+++ b/src/Apps/W1/Subcontracting/Test/src/Codeunits/Tests/SubcSubcontractingTest.Codeunit.al
@@ -2098,7 +2098,6 @@ codeunit 139989 "Subc. Subcontracting Test"
         // [GIVEN] Subcontracting setup with transfer components and a released transfer order
         Initialize();
         SubcontractingMgmtLibrary.UpdateManufacturingSetupWithSubcontractingLocation();
-        SubcontractingMgmtLibrary.SetupInventorySetup();
         Subcontracting := true;
         UnitCostCalculation := UnitCostCalculation::Units;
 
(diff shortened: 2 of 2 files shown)


alexei-dobriansky:
### Analysis

**Scenario Validity:**
- Valid and very clear. The linked ADO bug and its `[AI-REPRO]` comment spell out an exact, supported workflow: create a subcontracting purchase order, create and release its transfer order, open that transfer order again by drilling down from **No. of Transfer Orders**, choose **Reopen**, close the page, and drill down again; the status incorrectly comes back as `Released` instead of staying `Open`.
- The strongest repro statement is: **"Observe: Status = `Released` — the Reopen did not persist. The change appeared to succeed in step 5 but was silently lost when the page closed."** That is a real defect, not a preference issue, because the user is navigating through first-party UI on a normal subcontracting workflow documented in Microsoft Learn's subcontract manufacturing flow.

**Correctness:**
- The root cause is real in code. `ShowTransferOrdersAndReturnOrder` builds a marked `Transfer Header` set from matching transfer lines (`SubcPurchFactboxMgmt.Codeunit.al:297-303`). Before this PR, that marked record was passed straight into `PageManagement.PageRun` / `PageManagement.PageRunList` (PR diff on the same method). That means the page was opened on the marked in-memory record set rather than a fresh database-backed header.
- That matters because the Transfer Order page's **Reopen** action calls `ReleaseTransferDoc.Reopen(Rec)` (`C:\depot\master\app\layers\w1\baseapp\Inventory\Transfer\TransferOrder.Page.al:600-605`), and `ReleaseTransferDocument.Reopen` mutates the passed record directly with `Validate(Status, Open); Modify();` (`C:\depot\master\app\layers\w1\baseapp\Inventory\Transfer\ReleaseTransferDocument.Codeunit.al:87-97`). If `Rec` is not the real table row, the UI can show `Open` transiently while nothing persists.
- The single-record fix is correct and directly addresses that root cause. When exactly one transfer header is found, the PR now does `TransferHeaderToOpen.Get(TransferHeader."No."); PageManagement.PageRun(TransferHeaderToOpen);` (`SubcPurchFactboxMgmt.Codeunit.al:310-313`). That rebinds the page to the real `Transfer Header` row before any page action runs, so the BaseApp `Modify()` call now writes to the database.
- The multi-record branch is also reasonable. Instead of opening a marked record set, it converts the marked headers into a `No.` filter via `SelectionFilterMgt.GetSelectionFilterForTransferHeader(TransferHeader)` and runs the list on `TransferHeaderToOpen` (`SubcPurchFactboxMgmt.Codeunit.al:315-320`). That matches the standard BaseApp pattern used on transfer pages themselves when converting a selection into a durable `No.` filter (`C:\depot\master\app\layers\w1\baseapp\Inventory\Transfer\TransferOrder.Page.al:749-750`).
- The new regression test does exercise the fixed path. `FactboxDrilldownTransferOrderReopenPersists` creates the subcontracting scenario, calls `ShowTransferOrdersAndReturnOrder(PurchaseLine, true, false)`, invokes **Reopen** through the page handler, then re-gets the header from the table and asserts `Status::Open` (`SubcSubcontractingTest.Codeunit.al:2076-2149`, `2657-2662`). That would have failed before this change and now proves the exact reported bug is fixed.
- Edge cases: the PR also affects the return-order drilldown and routing factbox entry points because they all funnel through the same helper (`SubcPurchaseLineFactbox.Page.al:30-55`, `SubcRoutingInfoFactbox.Page.al:63-82`). I do not see a correctness issue there because the only behavioral change is replacing marked-page records with real DB-backed records.

**Side Effects:**
- Previously working behavior should remain intact for the core path. `PageManagement.PageRun` / `PageRunList` still open the same transfer documents; the only change is which record instance is passed in.
- This should actually reduce inconsistency between actions. The ADO bug already notes that actions like **Post** persisted because they fetched their own real record, while **Reopen** did not. Opening the page on a real header makes actions behave consistently instead of mixing durable and non-durable behavior on the same page.
- I do not see new event-ordering or extension-surface risk here. The PR does not add/remove events or alter public signatures; it only changes the record binding before calling existing page-management APIs.
- One thing still unproven by tests is the `NoOfTransferHeaders > 1` list-page branch. The implementation looks sound and is patterned after BaseApp, but that branch was changed and is not directly covered by the added test.

**Risk Assessment:**
- **Low to Medium.** The change is narrow (one helper method plus one regression test), but it touches transfer-order state changes, which do affect warehouse/manufacturing flow. The narrowness of the code change keeps the blast radius contained.
- If the fix were wrong, users could still open the wrong subset of transfer orders from the multi-record path. I do not see evidence of that in the code, but that is the main residual risk.
- **What if we don't make this change?** Users can reopen a transfer order from the subcontracting factbox, see `Open` on screen, and still have the status silently revert to `Released` when they leave the page. That is worse than a hard error because it misleads the user into thinking the document changed. There is a workaround—open the transfer order through a path that fetches the real header—but the bug affects a first-party drilldown path that users are expected to trust.

**Test Coverage:**
- Added test: `FactboxDrilldownTransferOrderReopenPersists` (`SubcSubcontractingTest.Codeunit.al:2076-2149`). It covers the exact reported scenario and asserts at the right level: after the page closes, it re-reads the `Transfer Header` from the table and checks that the status is really `Open`.
- The handler `HandleTransferOrderReopen` genuinely drives the changed code path by pressing the page action that ultimately mutates `Rec` (`SubcSubcontractingTest.Codeunit.al:2657-2662`). This is not a token test.
- Missing coverage is non-blocking but worth adding: there is no regression for the multi-record/list-page branch introduced at `SubcPurchFactboxMgmt.Codeunit.al:315-320`, and no explicit smoke test for the routing-factbox/return-order call sites that also use this helper.

### Recommendation: Accept with Suggestions

1. **Add one multi-record regression test** — The PR changes not only the single-record branch but also the `NoOfTransferHeaders > 1` path (`SubcPurchFactboxMgmt.Codeunit.al:315-320`). Please add a test that creates two transfer orders for the same drilldown context, opens the list through `ShowTransferOrdersAndReturnOrder`, and verifies the page shows the intended real headers rather than a marked in-memory set.
2. **Consider one shared-caller smoke test** — Because the same helper backs the routing-info and return-transfer drilldowns (`SubcRoutingInfoFactbox.Page.al:63-82`, `SubcPurchaseLineFactbox.Page.al:48-55`), a small additional test on one of those entry points would prove the fix is truly helper-wide rather than only correct for the purchase-order factbox scenario.

alexei-dobriansky (approved):

attilatoury (approved):

PR #8222, opened 2026-05-20
Bug 634269: [Subcontracting] Fix second return from subcontractor failing after first posted

ChethanT:
## Problem

Cannot create a second return from subcontractor after the first return transfer order is posted and deleted. The error is:

> Transfer-from Code and Transfer-to Code cannot be the same

## Root Cause

The `OnAfterDeleteEvent` subscriber in `SubcTransferLineExt.Codeunit.al` calls `UpdateLocationCodeInProdOrderCompAfterDeleteTransferLine` for ALL transfer line deletions, including **return** transfer lines. This procedure was designed only for **outbound** transfer cleanup (restoring the component location from SUBC back to MAIN when an outbound transfer is deleted/cancelled).

When it fires for a return line deletion (after posting), it:
1. Restores `ProdOrderComponent."Location Code"` to the Original Location Code
2. Clears `"Subc. Original Location Code"` to `''`

On the next return attempt, the system sees `Orig Location Code = ''`, sets it to the current (corrupted) location, and computes Transfer-from = Transfer-to = SUBC → error.

## Fix

Added a guard in `OnAfterDeleteEvent`:
```al
if Rec."Subc. Return Order" then
    exit;
```

This ensures location restoration only runs for outbound transfer line deletions, not return lines.

## Testing

New test codeunit 149917 that:
1. Creates outbound transfer, posts it
2. Creates first return, posts it (return transfer deleted)
3. Creates second return — verifies it succeeds without error

Fixes [AB#634269](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/634269)


src/Apps/W1/Subcontracting/Test/src/Codeunits/Tests/SubcReturnTransfTest.Codeunit.al
```diff
@@ -0,0 +1,202 @@
...
+using Microsoft.Manufacturing.WorkCenter;
+using Microsoft.Purchases.Document;
+using Microsoft.Warehouse.Structure;
+using System.TestLibraries.Utilities;
+
+codeunit 149917 "Subc. Return Transf. Test"
+{
+    // [FEATURE] Subcontracting Return Transfer Order - second return after first is posted
+    Subtype = Test;
+    TestPermissions = Disabled;
```
ChethanT, line 24:
```suggestion
    TestPermissions = Disabled;
    TestType = IntegrationTest;
```

PR #8289, opened 2026-05-23
Add the ability in the Email Module to store Message Headers

mynjj:
Addresses [AB#625413](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/625413)

## What

Adding the ability in the Email Module to pass filters to be used at email retrieval for populating `Message Headers`, as well as primitives in the `Email Message` to manage them.

## Why

Part of the "skip review for trusted senders" slice. The headers will be used for determining authentication properties of the received email.

## Summary of the changes

- Adds a new BLOB field "Message Headers" that won't be populated by default.
- Adds a new field to the temporary "Filters" table to use as parameter in the email retrieval
- Exposes a surface for connectors to store the headers. It's up to connector to honor the contract on the parameter. Such surface is:
    - AddHeader: Sets (appending) the header specified
    - SetHeader: Sets (replacing) the header specified
    - TryGetHeader: From the stored JSON, retrieves the header value requested.
    - FlushHeaders: Instead of full round-trips persisting for each header on every call to `AddHeader`, the updates are done on the instance of the EmailMessage. Persistance happens when flushing. 

### Other technical considerations

- Maintaining a cache of the headers JSON
- Only persisted when required (filters that default to false). Preliminar "smoke-analysis" reveals an increase of around 50% on DB size on emails if we were to store all the headers for all the emails.
image

## PTE for testing

[OutlookHeadersTester.zip](https://github.com/user-attachments/files/28313134/OutlookHeadersTester.zip)

Requires the changes in the [M365 API app](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_git/NAV/pullrequest/247802)

Scenarios behave as expected:
- Retrieving email without the "Load headers" filter don't store anything in the headers blob:
image

- Retrieving them with the "Load headers" filter, stores them, and the proposed API allows retrieving them

image

### DIFF
--- a/src/System Application/App/Email/src/Email/Inbox/EmailRetrievalFilters.Table.al
+++ b/src/System Application/App/Email/src/Email/Inbox/EmailRetrievalFilters.Table.al
@@ -67,6 +67,10 @@ table 8885 "Email Retrieval Filters"
             InitValue = "Include";
             Caption = 'Category Filter Type';
         }
+        field(11; "Load Headers"; Boolean)
+        {
+            Caption = 'Load Headers';
+        }
     }
 
     keys
--- a/src/System Application/App/Email/src/Message/EmailMessage.Codeunit.al
+++ b/src/System Application/App/Email/src/Message/EmailMessage.Codeunit.al
@@ -236,9 +236,9 @@ codeunit 8904 "Email Message"
     /// <returns><c>true</c> when a value was found.</returns>
     /// <remarks>Reads include any unflushed mutations issued via <see cref="AddHeader"/> or
     /// <see cref="SetHeader"/> on this codeunit instance.</remarks>
-    procedure TryGetHeader(HeaderName: Text; var Value: Text): Boolean
+    procedure GetHeader(HeaderName: Text; var Value: Text): Boolean
     begin
-        exit(EmailMessageImpl.TryGetHeader(HeaderName, Value));
+        exit(EmailMessageImpl.GetHeader(HeaderName, Value));
     end;
 
     /// <summary>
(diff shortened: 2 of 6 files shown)


src/System Application/App/Email/src/Email/Inbox/EmailRetrievalFilters.Table.al
```diff
@@ -67,6 +67,10 @@ table 8885 "Email Retrieval Filters"
             InitValue = "Include";
             Caption = 'Category Filter Type';
```
mynjj, line 68: leaving a thread open to wait for Darrick

src/System Application/App/Email/src/Message/EmailMessage.Codeunit.al
```diff
@@ -197,6 +197,61 @@ codeunit 8904 "Email Message"
...
+    /// Reads a single header value from the currently loaded message. Lookups are case-insensitive
+    /// on the header name. Repeated headers are returned as one string with values joined by a
+    /// line feed character in the order they were added.
+    /// </summary>
+    /// <param name="HeaderName">Header name to look up (case-insensitive, whitespace-trimmed).</param>
+    /// <param name="Value">Returned value when the header is present; empty otherwise.</param>
+    /// <returns><c>true</c> when a value was found.</returns>
+    /// <remarks>Reads include any unflushed mutations issued via <see cref="AddHeader"/> or
+    /// <see cref="SetHeader"/> on this codeunit instance.</remarks>
+    procedure TryGetHeader(HeaderName: Text; var Value: Text): Boolean
```
darjoo: Just call this `GetHeader`

darjoo (dismissed):

darjoo (approved):

Groenbech96 (approved):

PR #8327, opened 2026-05-27
(Bug 636059): [Subcontracting] Subcontractor pricing ignores UoM filter in routing and requisition worksheet — picks wrong price line based on sort order

alexei-dobriansky:
## What & why

`Subc. Price Management` (codeunit 99001508) picked the wrong **Subcontractor Price** row when prices existed in multiple Units of Measure. Two callers were broken:

- **`SetRoutingPriceListCost`** (Prod. Order Routing, and Standard Cost rollup via `CalcRtngCostPerUnit`) had no UoM filter — `FindLast()` returned whichever row sorted last by primary key. With prices in both PCS and SET, SET sorted after PCS and was always picked, then converted via the alt-UoM factor (e.g. `1004/10 = 100.40` instead of `1001`).
- **`GetSubcPriceForReqLine`** applied its UoM filter only when the caller passed `FixedUOM <> ''`. On the Subcontracting Worksheet path that passes `''`, the same `FindLast()` defect surfaced.

Applies the exact-or-blank `'%1|%2'` filter pattern for UoM that `GetSubcPriceForPurchLine` (the working purchase-order path) and BaseApp purchase pricing already use. The blank value remains a valid fallback row.

## Linked work

Fixes [AB#636059](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/636059)


src/Apps/W1/Subcontracting/Test/src/Codeunits/Tests/SubcSubcontractingTest.Codeunit.al
```diff
@@ -3412,6 +3416,156 @@ codeunit 139989 "Subc. Subcontracting Test"
             'Direct Unit Cost must equal the price-list cost when the worksheet UoM matches the price-list UoM.');
     end;
 
+    [Test]
+    procedure RoutingPriceUsesOrderUoMWhenMultipleUoMPricesExist()
```
ChethanT, line 3420: This codeunit is growing huge, pleae add this to a new test codeunit, so that we can move all pricing related tests to that.

ChethanT (approved):

AleksandricMarko (approved):

PR #8438, opened 2026-06-03
Fix data-driven AI tests failing when TestPermissions is not Disabled

Groenbech96:
#### Summary

Data-driven tests run by the AI Test Suite Runner fail when the test method declares a `TestPermissions` value other than `Disabled`.

The data-driven test framework reads the `Test Input` table from a subscriber to `OnBeforeTestMethodRun` (codeunit 130460 `Test Input`). The platform raises that event **inside** the test method's `TestPermissions` scope, so the effective permissions are overridden and the `IndirectRead` on table `Test Input` fails:

```
Sorry, the current permissions prevented the action.
(TableData 130452 Test Input IndirectRead: Test Runner)
The user permissions have been overridden by the TestPermissions property.
```

Because the permission execution context is overridden by `TestPermissions`, the codeunit's own `Permissions` property is not honored in that window, so the read cannot succeed there.

#### Work Item(s)

Fixes [AB#636024](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/636024)

#### Fix

Load the test input earlier, while still running under full permissions. The AI Test Toolkit input handler already has a pre-permission hook (`OnBeforeRunTestMethodLine`, raised before the test runner codeunit is invoked). It now pre-loads the input into the single-instance `Test Input` globals via a new `PreloadTestInput` method, so the later read inside the restricted `TestPermissions` scope becomes a no-op (the existing early-exit guard returns before the `Get`).

- `Test Input` (130460): extracted the load logic into a shared `LoadTestInput` helper and exposed `PreloadTestInput(GroupCode, InputCode)`.
- `AIT Test Run Input Handler` (149045): calls `PreloadTestInput` from its existing `OnBeforeRunTestMethodLine` subscriber.

### DIFF
--- a/src/Tools/Test Framework/Test Runner/src/DataDrivenTest/DataInputs/TestInput.Codeunit.al
+++ b/src/Tools/Test Framework/Test Runner/src/DataDrivenTest/DataInputs/TestInput.Codeunit.al
@@ -24,13 +24,29 @@ codeunit 130460 "Test Input"
 
     internal procedure InitializeTestInputsBeforeTestMethodRun(CodeunitID: Integer; CodeunitName: Text[30]; FunctionName: Text[128]; FunctionTestPermissions: TestPermissions; var CurrentTestMethodLine: Record "Test Method Line")
     begin
-        if CurrentTestMethodLine."Data Input" = '' then
+        LoadTestInput(CurrentTestMethodLine."Data Input Group Code", CurrentTestMethodLine."Data Input");
+    end;
+
+    // Pre-loads the test input into the (single instance) globals before the test method runs. The data-driven
+    // test framework reads the "Test Input" table from a subscriber to OnBeforeTestMethodRun, which the platform
+    // raises inside the test method's TestPermissions scope. When a test declares a TestPermissions value other
+    // than Disabled, the platform overrides the effective permissions and the IndirectRead on the "Test Input"
+    // table fails. Calling this from a context that still runs under full permissions (before the test runner is
+    // invoked) caches the input so the later read inside the restricted scope is a no-op.
+    procedure PreloadTestInput(TestInputGroupCode: Code[100]; TestInputCode: Code[100])
+    begin
+        LoadTestInput(TestInputGroupCode, TestInputCode);
+    end;
+
+    local procedure LoadTestInput(TestInputGroupCode: Code[100]; TestInputCode: Code[100])
+    begin
+        if TestInputCode = '' then
             exit;
 
-        if (CurrentTestMethodLine."Data Input" = DataPerTest.Code) and (CurrentTestMethodLine."Data Input Group Code" = DataPerTest."Test Input Group Code") then
+        if (TestInputCode = DataPerTest.Code) and (TestInputGroupCode = DataPerTest."Test Input Group Code") then
             exit;
 
-        DataPerTest.Get(CurrentTestMethodLine."Data Input Group Code", CurrentTestMethodLine."Data Input");
+        DataPerTest.Get(TestInputGroupCode, TestInputCode);
 
         DataPerTestTestInput.Initialize(DataPerTest.GetInput(DataPerTest));
     end;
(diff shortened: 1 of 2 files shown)


src/Tools/Test Framework/Test Runner/src/DataDrivenTest/DataInputs/TestInput.Codeunit.al
```diff
@@ -24,13 +24,29 @@ codeunit 130460 "Test Input"
...
+
+    // Pre-loads the test input into the (single instance) globals before the test method runs. The data-driven
+    // test framework reads the "Test Input" table from a subscriber to OnBeforeTestMethodRun, which the platform
+    // raises inside the test method's TestPermissions scope. When a test declares a TestPermissions value other
+    // than Disabled, the platform overrides the effective permissions and the IndirectRead on the "Test Input"
+    // table fails. Calling this from a context that still runs under full permissions (before the test runner is
+    // invoked) caches the input so the later read inside the restricted scope is a no-op.
+    procedure PreloadTestInput(TestInputGroupCode: Code[100]; TestInputCode: Code[100])
+    begin
+        LoadTestInput(TestInputGroupCode, TestInputCode);
```
nikolakukrika, line 45: I'm wondering if we should update the permission sets.
To include the Test Input in the Basic or the permission set that stays when the permissions are lowered.
We are accessing the other tables and they are not affected.

nikolakukrika (dismissed):

src/Tools/Test Framework/Test Runner/src/DataDrivenTest/DataInputs/TestInput.Codeunit.al
Groenbech96: Lets do this, incase the current fix does not work.

src/Tools/Test Framework/Test Runner/src/DataDrivenTest/DataInputs/TestInput.Codeunit.al
```diff
@@ -24,13 +24,29 @@ codeunit 130460 "Test Input"
 
     internal procedure InitializeTestInputsBeforeTestMethodRun(CodeunitID: Integer; CodeunitName: Text[30]; FunctionName: Text[128]; FunctionTestPermissions: TestPermissions; var CurrentTestMethodLine: Record "Test Method Line")
     begin
-        if CurrentTestMethodLine."Data Input" = '' then
+        LoadTestInput(CurrentTestMethodLine."Data Input Group Code", CurrentTestMethodLine."Data Input");
+    end;
+
+    // Pre-loads the test input into the (single instance) globals before the test method runs. The data-driven
```
darjoo: NIT: If it's comments on the function, to follow our standard documentation. 
```<summary>...</summary> <remarks>...</remarks>``` etc.

**Review by darjoo — DISMISSED** [review]

**Comment by Groenbech96** [comment]
@copilot Can you look at darjoos comment to follow standard?

**Review by darjoo — APPROVED** [review]

File: src/Tools/Test Framework/Test Runner/src/DataDrivenTest/DataInputs/TestInput.Codeunit.al [hunk 3]
```diff
@@ -24,13 +24,36 @@ codeunit 130460 "Test Input"
...
+    /// The data-driven test framework reads the "Test Input" table from a subscriber to OnBeforeTestMethodRun,
+    /// which the platform raises inside the test method's TestPermissions scope. When a test declares a
+    /// TestPermissions value other than Disabled, the platform overrides the effective permissions and the
+    /// IndirectRead on the "Test Input" table fails. Calling this from a context that still runs under full
+    /// permissions (before the test runner is invoked) caches the input so the later read inside the restricted
+    /// scope is a no-op.
+    /// </remarks>
+    procedure PreloadTestInput(TestInputGroupCode: Code[100]; TestInputCode: Code[100])
+    begin
+        LoadTestInput(TestInputGroupCode, TestInputCode);
```
gggdttt, line 45: Is there any special reason that we need a local procedure "LoadTestInput"?  If not maybe we can put code under PreLoadTestInput directlly

gggdttt (approved):

PR #8440, opened 2026-06-03
[AI Test Toolkit] Refresh Run History when changing View By

onbuyuka:
## Summary
On the **AI Eval Suite Run History** page (`page 149032 "AIT Run History"`), changing the **View By** control (Version/Tag) — or toggling **Apply Filter** — did not update the displayed data.

## Root cause
Two compounding defects in the refresh path:

1. **Repaint.** `UpdateRunHistory()` rebuilds the temporary source table (`DeleteAll` + re-`Insert` inside `codeunit 149036 "AIT Run History".GetHistory`) and then calls the **parameterless** `CurrPage.Update()`, which defaults to `Update(true)`. That tries to *save* the (just-deleted) current record rather than redraw. The canonical pattern after repopulating a temporary page source is `CurrPage.Update(false)`.

2. **Stale FlowFilter.** `GetHistory` set the `"Line No. Filter"` FlowFilter only when `LineNo <> 0` and never cleared it. Since `DeleteAll` does not clear filters, unchecking **Apply Filter** (`LineNo -> 0`) left the previous line filter on the FlowFields, so the columns recomputed against the wrong line and showed zeros/old data.

## Fix
- `AITRunHistory.Page.al` — `CurrPage.Update(false)` after rebuilding the temp records.
- `AITRunHistory.Codeunit.al` — clear the `"Line No. Filter"` FlowFilter when `LineNo = 0` (add an `else SetRange("Line No. Filter")`).

## Impact / risk
Low — internal AI Test Toolkit tooling page, read-only temporary source. Both changes are the standard idioms for this scenario.

## Verification
- Statically traced: `View By` OnValidate → `UpdateRunHistory` → `GetHistory` (sole data builder) → `CurrPage.Update`. FlowField columns key on `"Line No. Filter"`.
- ⚠️ Not runtime-verified on an NST (AL does not run locally here) — relying on PR CI for compile + tests; manual repro on the Run History page recommended before merge.

Linked work item: [AB#622420](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/622420)

### DIFF
--- a/src/Tools/AI Test Toolkit/src/Logs/AITRunHistory.Page.al
+++ b/src/Tools/AI Test Toolkit/src/Logs/AITRunHistory.Page.al
@@ -245,6 +245,6 @@ page 149032 "AIT Run History"
             LineNo := 0;
 
         AITRunHistory.GetHistory(TestSuiteCode, LineNo, ViewBy, Rec);
-        CurrPage.Update();
+        CurrPage.Update(false);
     end;
 }
--- a/src/Tools/AI Test Toolkit/src/Logs/AITRunHistory.Codeunit.al
+++ b/src/Tools/AI Test Toolkit/src/Logs/AITRunHistory.Codeunit.al
@@ -41,5 +41,7 @@ codeunit 149036 "AIT Run History"
 
         if (LineNo <> 0) then
             TempAITRunHistory.SetRange("Line No. Filter", LineNo)
+        else
+            TempAITRunHistory.SetRange("Line No. Filter");
     end;
 }


onbuyuka:
## CI note: the Clean build failure is a known BCApps incremental-build infra issue — not this change

The `... (Clean)` build failure is **not** caused by this PR. The incremental Clean build compiles the modified **AI Test Toolkit** app against the **System Application bundled in the BC platform artifact** (`sandbox/29.0.51094.0`) instead of the freshly-built System Application. That artifact predates microsoft/BCApps#8381, which added `Agent Task.OpenAgentTaskLogEntries` / `GetStepsDoneCount` (called by `AITAgentLogEntryPart.Page.al`), so the compiler reports `AL0132 ... does not contain a definition for ...`.

Evidence it is infrastructure, not this change:
- The **Default** build of this same app **passes** (compiles against fresh symbols).
- main`s **full** build passes (System Application compiled from source).
- The two methods exist, unwrapped (present under CLEAN29), in current `main` source and in the last-green baseline `a97db24da`.
- PRs that modify System Application (e.g. #8441, #8469) pass Clean because they rebuild it from source; only AI-Test-Toolkit-only PRs hit this.
- Build log shows the Clean compile resolving `Microsoft_System Application.app` from the BcContainerHelper compiler-folder cache (the platform artifact), not the freshly-built `29.0.74476.0`.

No code change is appropriate (a `#if not CLEAN29` guard on non-obsolete methods would be wrong). This goes green once the platform-artifact pin advances past #8381. Holding for that.

nikolakukrika (approved):

src/Tools/AI Test Toolkit/src/Logs/AITRunHistory.Page.al
```diff
@@ -245,6 +245,6 @@ page 149032 "AIT Run History"
             LineNo := 0;
 
         AITRunHistory.GetHistory(TestSuiteCode, LineNo, ViewBy, Rec);
-        CurrPage.Update();
+        CurrPage.Update(false);
```
nikolakukrika, line 248: Why is this needed? We did not introduce a transaction above?

nikolakukrika (approved):

src/Tools/AI Test Toolkit/src/Logs/AITRunHistory.Page.al
onbuyuka: GetHistory does changes to Rec: https://github.com/microsoft/BCApps/blob/7b75c8aa5d40b0c0860c9333390bfc707f1eaad4/src/Tools/AI%20Test%20Toolkit/src/Logs/AITRunHistory.Codeunit.al#L12-L43

darjoo (approved):

PR #8441, opened 2026-06-03
[Azure AD Plan] Update access for GDAP users who were previously guests

onbuyuka:
## Summary
A delegated-admin (**GDAP**) user who had previously logged in to the environment **as a guest** got the GDAP plan row, but their user access (user groups / permissions) was never updated — so they did not effectively receive the GDAP license. IcM 1888383.

## Root cause
In `codeunit 9018 "Azure AD Plan Impl.".AssignPlanToUserWithDelegatedRole`, the caller (`AzureADUserMgmtImpl.Run`) passes `SkipUpdateUserAccess = UserLoggedInEnvironment`, which is **true for any returning user**. After deleting the old guest plan and inserting the GDAP plan, the procedure hit two early-exits before the access update:

```al
if SkipUpdateUserAccess then exit;   // true for returning users
...
if HasPlans then exit;               // true for former guests (had a guest plan)
```

so `OnUpdateUserAccessForSaaS` (the user-group / permission assignment) and the SUPER handling never ran for a former-guest GDAP user.

The earlier guard `if IsPlanAssignedToUser(PlanId, UserSID) then exit;` already prevents re-processing when the GDAP plan is unchanged, so the two removed checks are redundant for correctness — and crucially, that guard means the access update only re-runs when the plan **actually changes** (no per-login perf regression for normal returning users).

## Fix
Remove the two redundant early-exits and the now-unused `HasPlans` local, so access is updated on the guest → GDAP transition. Adds regression test `GDAPUserWithPreviousGuestPlanGetsAccessUpdated`.

## Note on provenance
This **revives the previously abandoned PR #7490** (closed without merging on 2026-04-01). I reproduced the same code change and test against current `main`. Posting as **draft** so the original owners can confirm there was no hidden blocking concern behind the earlier abandonment before this lands.

## Verification
- Traced the full caller→callee path; current `main` matches the pre-fix state exactly. Verified all test-library helpers referenced by the new test exist on current `main`.
- ⚠️ Not runtime-verified on an NST here (AL does not run locally) — relying on PR CI. The abandoned PR reported the baseline test failed pre-fix (bug reproduced) and passed post-fix; re-running the Azure AD Plan suite on CI is the gate.

Linked work item: [AB#612420](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/612420)

### DIFF
--- a/src/System Application/App/Azure AD Plan/src/AzureADPlanImpl.Codeunit.al
+++ b/src/System Application/App/Azure AD Plan/src/AzureADPlanImpl.Codeunit.al
@@ -662,7 +662,7 @@ codeunit 9018 "Azure AD Plan Impl."
         PlanIds: Codeunit "Plan Ids";
         PlanConfigurationImpl: Codeunit "Plan Configuration Impl.";
         UserPermissions: Codeunit "User Permissions";
-        UserGroupsAdded, ShouldRemoveSuper, HasPlans : Boolean;
+        UserGroupsAdded, ShouldRemoveSuper : Boolean;
         PlanId: Guid;
     begin
         case true of
@@ -682,9 +682,6 @@ codeunit 9018 "Azure AD Plan Impl."
 
         Session.LogMessage('0000IC4', StrSubstNo(AssigningPlanForDelegatedRoleTxt, PlanId, not SkipUpdateUserAccess), Verbosity::Normal, DataClassification::SystemMetadata, TelemetryScope::ExtensionPublisher, 'Category', UserSetupCategoryTxt);
 
-        // Check if the user have any plans assigned before removing
-        HasPlans := DoesUserHavePlans(UserSID);
-
         // Delete any existing plans for the user
         UserPlan.SetRange("User Security ID", UserSID);
         UserPlan.DeleteAll();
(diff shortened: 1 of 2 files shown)


src/System Application/App/Azure AD Plan/src/AzureADPlanImpl.Codeunit.al
```diff
@@ -695,16 +692,12 @@ codeunit 9018 "Azure AD Plan Impl."
         UserPlan."User Security ID" := UserSID;
         UserPlan.Insert();
 
-        // Exit if user access should not be updated
-        if SkipUpdateUserAccess then
-            exit;
-
+        // Fix for ADO #612420: Removed early-exit checks on SkipUpdateUserAccess and HasPlans
```
nikolakukrika, line 695: Lets not check in the bug numbers in the code

src/System Application/App/Azure AD Plan/src/AzureADPlanImpl.Codeunit.al
```diff
@@ -695,16 +692,12 @@ codeunit 9018 "Azure AD Plan Impl."
         UserPlan."User Security ID" := UserSID;
         UserPlan.Insert();
 
-        // Exit if user access should not be updated
-        if SkipUpdateUserAccess then
```
nikolakukrika, line 699: I believe this line should stay as it is coming in as a parameter. If we remove it, we have a dead parameter that is used only in the telemetry.

nikolakukrika: I believe that this PR requires more changes, probably to the caller functions too.

nikolakukrika (changes requested):

PR #8442, opened 2026-06-03
[AI Test Toolkit] Export full eval run results (cost, time, dataset)

onbuyuka:
## Summary
The **Export Results** / **Download Test Summary** action on the AI Eval log pages (report `149030 "AIT Test Summary"`) did not export the *full* run results. It only emitted: codeunit/test name, status, accuracy, turns, input, output, and error. The reporter needs the cost and timing data to analyse a run.

## Root cause
The export passes the complete filtered `AIT Log Entry` view to the report (no row/marked-row limiting) — so the limitation is purely the report's **column set**. The dataset omitted fields that already exist on the table and are shown on screen:
- **Cost/time:** `Tokens Consumed`, `Duration (ms)`, `Start Time`, `End Time`
- **Identity:** `Version`, `Tag`, `Test Input Group Code`, `Test Input Code`, `Test Input Description`, `Run ID`, `Entry No.`

## Fix
Add those columns to the `dataitem(Results; "AIT Log Entry")` dataset. A modern Excel-layout report writes **every** dataset column into the workbook's *Data* sheet, so the new fields are exported in the raw data without changing the formatted `.xlsx` layout.

## Follow-up (optional, not in this PR)
If the formatted/pretty sheet of `AITestSummary.xlsx` should also display the new columns, the layout file can be regenerated from the updated report (Report → Run → *Edit layout*). The raw Data sheet already carries them.

## Impact / risk
Additive only — no codeunit/page changes; the export already passed the full filtered view. Field names verified against `AITLogEntry.Table.al`.

## Verification
- Verified each added field exists on `AIT Log Entry` (fields 1, 4, 5, 9, 13, 15, 18, 24, 25, 26, 50).
- ⚠️ Not runtime-verified on an NST (AL does not run locally here); relying on PR CI for compile. A manual export check is recommended to confirm the Data sheet content.

Linked work item: [AB#633442](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/633442)

### DIFF
--- a/src/Tools/AI Test Toolkit/src/Logs/AITTestSummary.Report.al
+++ b/src/Tools/AI Test Toolkit/src/Logs/AITTestSummary.Report.al
@@ -54,6 +54,39 @@ report 149030 "AIT Test Summary"
             column(Error; ErrorCallstack)
             {
             }
+            column(EntryNo; Results."Entry No.")
+            {
+            }
+            column(Version; Results.Version)
+            {
+            }
+            column(Tag; Results.Tag)
+            {
+            }
+            column(RunID; Results."Run ID")
+            {
+            }
+            column(TestInputGroupCode; Results."Test Input Group Code")
+            {
+            }
+            column(TestInputCode; Results."Test Input Code")
+            {
+            }
+            column(TestInputDescription; Results."Test Input Description")
+            {
+            }
+            column(TokensConsumed; Results."Tokens Consumed")
+            {
+            }
+            column(DurationMs; Results."Duration (ms)")
+            {
+            }
+            column(StartTime; Results."Start Time")
+            {
+            }
+            column(EndTime; Results."End Time")
+            {
+            }
 
             trigger OnAfterGetRecord()
             begin
--- a/src/Tools/AI Test Toolkit/src/Logs/AITTestSummary.Report.al
+++ b/src/Tools/AI Test Toolkit/src/Logs/AITTestSummary.Report.al
@@ -75,9 +75,6 @@ report 149030 "AIT Test Summary"
             column(TestInputDescription; Results."Test Input Description")
             {
             }
-            column(TokensConsumed; Results."Tokens Consumed")
-            {
-            }
             column(DurationMs; Results."Duration (ms)")
             {
             }


onbuyuka:
## CI note: the Clean build failure is a known BCApps incremental-build infra issue — not this change

The `... (Clean)` build failure is **not** caused by this PR. The incremental Clean build compiles the modified **AI Test Toolkit** app against the **System Application bundled in the BC platform artifact** (`sandbox/29.0.51094.0`) instead of the freshly-built System Application. That artifact predates microsoft/BCApps#8381, which added `Agent Task.OpenAgentTaskLogEntries` / `GetStepsDoneCount` (called by `AITAgentLogEntryPart.Page.al`), so the compiler reports `AL0132 ... does not contain a definition for ...`.

Evidence it is infrastructure, not this change:
- The **Default** build of this same app **passes** (compiles against fresh symbols).
- main`s **full** build passes (System Application compiled from source).
- The two methods exist, unwrapped (present under CLEAN29), in current `main` source and in the last-green baseline `a97db24da`.
- PRs that modify System Application (e.g. #8441, #8469) pass Clean because they rebuild it from source; only AI-Test-Toolkit-only PRs hit this.
- Build log shows the Clean compile resolving `Microsoft_System Application.app` from the BcContainerHelper compiler-folder cache (the platform artifact), not the freshly-built `29.0.74476.0`.

No code change is appropriate (a `#if not CLEAN29` guard on non-obsolete methods would be wrong). This goes green once the platform-artifact pin advances past #8381. Holding for that.

darjoo (dismissed):

src/Tools/AI Test Toolkit/src/Logs/AITTestSummary.Report.al
```diff
@@ -54,6 +54,39 @@ report 149030 "AIT Test Summary"
...
+            column(TestInputGroupCode; Results."Test Input Group Code")
+            {
+            }
+            column(TestInputCode; Results."Test Input Code")
+            {
+            }
+            column(TestInputDescription; Results."Test Input Description")
+            {
+            }
+            column(TokensConsumed; Results."Tokens Consumed")
```
nikolakukrika: We do not want to export the tokens consumed. The reason for this is that Agent tests do not log the tokens. The second reason is that if we implement it, it is internal information.

nikolakukrika (dismissed):

src/Tools/AI Test Toolkit/src/Logs/AITTestSummary.Report.al
Groenbech96: Done in d03b6fc89 — removed the TokensConsumed column from the report. Agent tests do not log tokens and the value is internal, so it should not be exported.

darjoo (approved):

gggdttt (approved):

PR #8483, opened 2026-06-04
Fix Azure AD user-sync test failing in NAV uptake gate (bug 637527)

Groenbech96:
## Summary
Fixes `TestFailedUpdateRecordsErrorMessageOnBuffer` (codeunit 132928 `Azure AD User Sync Test`) failing in the NAV uptake gate (`RunALTestSystemModules_Group1`), tracked as ADO bug 637527.

## Root cause
The test exercises the error-on-buffer path added in #8393 ("Surface errors from failed M365 user-sync updates in the wizard"): a failed per-user update should have its error recorded on the buffer via the isolated `OnApplyUpdateFromAzureGraph` event, not propagate.

`Initialize()` calls `User.DeleteAll()`. In the System Application test environment the database is built from the **System DACPAC**, which seeds the `MSOLSYNC` daemon user (`Eng/Core/Scripts/ExportSystemDacpac.ps1` → `CreateMSOnlineSyncDaemonUser.ps1`, kept via `Remove-NavTenantDatabaseUserData -KeepUsers`). So `User.DeleteAll()` deletes a **real row** and opens a write transaction. `ApplyUpdatesFromAzureGraph` then raises the isolated event **inside that write transaction**, where the subscriber error cannot be isolated — so `User.Get` throws "The User does not exist" and it propagates, failing the test.

This stayed green wherever the `User` table was empty (`DeleteAll` is a no-op → no write transaction), which is why it passed before the NAV uptake.

## Fix
`Commit()` after `Initialize()` so the event is raised outside a write transaction, matching the requirement already documented in the test's own comment.

> Note: this is an experiment to confirm the mechanism (test-side commit). If the product path can also hit a write-transaction caller, a `Commit()` before the raise in `ProcessAllUpdatesForUser` would be the more robust fix.

[AB#637527](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/637527)


Groenbech96:
@copilot Merge main into branch

PR #8522, opened 2026-06-08
[Subcontracting] Check negative quantities and arbitrary values on WIP Adjustment

### LINKED ISSUE #8325: [Subcontracting] WIP Adjustment allows negative quantities and arbitrary values without validation
## Bug

**Area:** AI Business Solutions / SCM / Manufacturing

## Repro Steps

1. Open a Released Production Order with subcontracting
2. Open the WIP Adjustment action
3. Enter a negative quantity (e.g., -50)
4. Observe: the system accepts it without validation
5. Enter an arbitrary large value exceeding the production order quantity
6. Observe: also accepted without validation

## Expected Behavior

WIP Adjustment should validate entries - either reject negative values or require justification/confirmation. Arbitrary values exceeding the production order quantity should be flagged.

---

Internal workitem: [AB#636826](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/636826)

stevengrossGOB:
## What & why

- Implement validation to prevent negative new quantities.
- Ensure new quantity does not exceed production order line quantity.
- Add corresponding test cases to verify error handling.

## Linked work

Fixes [AB#636826](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/636826)

## How I validated this

- Ran the new Test Functions

### DIFF
--- a/src/Apps/W1/Subcontracting/App/src/Process/WIP Item/SubcWIPAdjustment.Page.al
+++ b/src/Apps/W1/Subcontracting/App/src/Process/WIP Item/SubcWIPAdjustment.Page.al
@@ -385,8 +385,8 @@ page 99001561 "Subc. WIP Adjustment"
         if NewQty < 0 then
             Error(NewQuantityMustNotBeNegativeErr);
         ProdOrderLine.SetLoadFields("Quantity (Base)");
-        if ProdOrderLine.Get(Rec."Prod. Order Status", Rec."Prod. Order No.", Rec."Prod. Order Line No.") then
-            if NewQty > ProdOrderLine."Quantity (Base)" then
-                Error(NewQuantityExceedsProdOrderQtyErr, ProdOrderLine."Quantity (Base)");
+        ProdOrderLine.Get(Rec."Prod. Order Status", Rec."Prod. Order No.", Rec."Prod. Order Line No.");
+        if NewQty > ProdOrderLine."Quantity (Base)" then
+            Error(NewQuantityExceedsProdOrderQtyErr, ProdOrderLine."Quantity (Base)");
     end;
 }
--- a/src/Apps/W1/Subcontracting/App/src/Process/WIP Item/SubcWIPAdjustment.Page.al
+++ b/src/Apps/W1/Subcontracting/App/src/Process/WIP Item/SubcWIPAdjustment.Page.al
@@ -5,6 +5,7 @@
 namespace Microsoft.Manufacturing.Subcontracting;
 
 using Microsoft.Inventory.Item;
+using Microsoft.Manufacturing.Document;
 
 page 99001561 "Subc. WIP Adjustment"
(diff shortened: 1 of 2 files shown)


AleksandricMarko (dismissed):

alexei-dobriansky:

### What this PR does
This adds validation on both `New Quantity (Base)` page fields so the target WIP quantity cannot be negative and cannot exceed the related production order line quantity. The guard is placed in `OnValidate`, before `NewQuantities.Set(...)`, so normal page entry is blocked before the value can be stored and later posted by `CreateAdjustmentEntries()`. The `< 0` and `>` boundaries look intentional: zero and exactly the production quantity remain allowed. Tests were added for the two invalid inputs, and the existing positive/negative adjustment tests still cover normal posting flows.

### Suggestions

**S1** - *Fail closed when the order line is missing*
`ValidateNewQuantity` silently skips the upper-bound check if `ProdOrderLine.Get(...)` fails, which leaves the "arbitrary value" protection dependent on perfect source data. Since the WIP ledger entry has a production order line relation, consider using a required `Get`/clear error instead of treating a missing line as permission to post any value.

**S2** - *Cover valid boundary values explicitly*
The new tests assert the two error paths, but they only check error substrings and do not prove that valid boundary values such as `0`, exactly the production order quantity, and a normal in-range value with an existing `Prod. Order Line` still succeed. Adding those cases would make the intended `< 0` and `> Quantity (Base)` boundaries unambiguous.

**Risk:** The changed page writes adjustment ledger entries from the `NewQuantities` dictionary in `SubcWIPAdjustment.Page.al`, so a wrong guard can either block legitimate subcontracting WIP reductions or create incorrect positive/negative WIP adjustment entries. The scope is otherwise narrow: no public API/event signatures changed, and the validation runs before the dictionary update.

**Necessity:** The change is justified because the dialog previously accepted negative and unbounded target WIP quantities. The added tests in `SubcWIPAdjustmentTest.Codeunit.al` make the negative and over-production error paths mechanical to validate, but the missing-line fallback and boundary success cases should be tightened.

---
> [AI-PR-REVIEW] version=1 system=github pr=8522 round=1 by=alexei-dobriansky at=2026-06-10T15:19:54Z lastSha=d2e8683e1b5957c63bf4af699a1616b8df48d5a6 suggestions=S1,S2

alexei-dobriansky:
@stevengrossGOB please review the suggestions

alexei-dobriansky (commented):
.

stevengrossGOB:
@alexei-dobriansky @AleksandricMarko I have incorporated your suggestions.

alexei-dobriansky (approved):

ChethanT (approved):

PR #8570, opened 2026-06-10
[Subcontracting] Fix Subcontracting item charge posting for last operation

stevengrossGOB:
## What & why

Discovered during Subcontracting Tests:
- Add logic to handle item charge assignment for last operation in subcontracting receipts.
- Introduce new test codeunit for validating item charge posting scenarios.
- Ensure correct linking of value entries to item ledger entries during posting.

## Linked work

Fixes  [AB#632496](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/632496)

## How I validated this

- New automatic test that covers the failed scenario
- manual test for last/ not last operation

### DIFF
--- a/src/Apps/W1/Subcontracting/App/src/Process/Codeunits/Extensions/Purchase/SubcPurchPostExt.Codeunit.al
+++ b/src/Apps/W1/Subcontracting/App/src/Process/Codeunits/Extensions/Purchase/SubcPurchPostExt.Codeunit.al
@@ -12,6 +12,7 @@ using Microsoft.Inventory.Ledger;
 using Microsoft.Inventory.Posting;
 using Microsoft.Inventory.Tracking;
 using Microsoft.Manufacturing.Capacity;
+using Microsoft.Manufacturing.Document;
 using Microsoft.Purchases.Document;
 using Microsoft.Purchases.History;
 using Microsoft.Purchases.Posting;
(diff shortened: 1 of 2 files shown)


alexei-dobriansky:

### What this PR does
This PR changes subcontracting item-charge posting so charges assigned to a receipt for the last routing operation stay on the item-ledger-entry posting path, while intermediate operations keep the existing capacity/output posting path.

The change looks targeted: `SubcPurchPostExt.Codeunit.al` now copies the production order and inventory posting fields before branching, then exits before rewriting the journal line to capacity/output when the posted receipt operation has no next operation. That addresses the reported root cause instead of just masking the capacity-ledger lookup error, and the new test posts an item charge against a last-operation subcontracting receipt and asserts the resulting value entry links back to the output item ledger entry. I verified the relevant BaseApp publishers exist and are raised at the expected points in purchase posting.

### Suggestions

**S1** - *Add a non-last regression test*
The production code now has separate last-operation and non-last-operation branches. Please add an automated item-charge posting test for an intermediate subcontracting operation so the existing capacity-path behavior is protected from future regressions.

**Risk:** The regression surface is limited to subcontracting item-charge invoice posting for posted purchase receipt lines in `src/Apps/W1/Subcontracting/App/src/Process/Codeunits/Extensions/Purchase/SubcPurchPostExt.Codeunit.al`. The last-operation decision depends on the current Released/Finished production routing line, so changed or missing routing state before the item-charge invoice could still fall back to the non-last path. No public API or event signature changes were introduced. BaseApp event publishers were verified in `app\layers\w1\baseapp\Purchases\Posting\PurchPost.Codeunit.al` and `app\layers\w1\baseapp\Manufacturing\Posting\MfgPurchPost.Codeunit.al`; linked work item not independently verified.

**Necessity:** The change is necessary for a valid costing scenario: posting item charges after receiving the last subcontracting operation should apply cost to the produced item ledger entry, not fail through capacity-ledger posting. The scope is appropriately narrow and includes a test for the failing path.

---
> [AI-PR-REVIEW] version=1 system=github pr=8570 round=1 by=alexei-dobriansky at=2026-06-10T15:20:42Z lastSha=accf675c8e7e7a7a15a0d3e8339fae2a8f6e3bf8 suggestions=S1

alexei-dobriansky (approved):

ChethanT (approved):

PR #8615, opened 2026-06-15
[Backport releases/28.x] Fix #7622 Shopify product export child-item variant handling

### LINKED ISSUE #7622: [Bug]: Shpfy Product Export creates unwanted variants for items added via "Add Item as Shopify Variant"

tVisionDeb:
#### Summary
Backport of PR #7689 (merged to `microsoft:main` on 29 April 2026).

Cherry-picks the fix for issue #7622 onto `releases/28.x`. The fix prevents Shopify product export from creating unwanted variants for items that were added as a Shopify variant via the child-item mapping.

HS Code / tariff tests (part of a separate feature not present in `releases/28.x`) were excluded from the cherry-pick as they are unrelated to this fix.

#### Work Item(s)

Fixes [AB#647317](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/647317)

### DIFF
--- a/src/Apps/W1/Shopify/App/src/Products/Codeunits/ShpfyProductExport.Codeunit.al
+++ b/src/Apps/W1/Shopify/App/src/Products/Codeunits/ShpfyProductExport.Codeunit.al
@@ -657,6 +657,10 @@ codeunit 30178 "Shpfy Product Export"
                                 UpdateProductVariant(ShopifyVariant, Item, ItemVariant, TempCurrVariant);
                         end;
                 until ShopifyVariant.Next() = 0;
+            // Re-fetch the parent product's item, as the loop above may have overwritten
+            // the Item variable with a child item from "Add Item as Shopify Variant".
+            if not Item.GetBySystemId(ShopifyProduct."Item SystemId") then
+                exit;
             ItemVariant.SetRange("Item No.", Item."No.");
             ItemUnitofMeasure.SetRange("Item No.", Item."No.");
             if ItemVariant.FindSet(false) then
--- a/src/Apps/W1/Shopify/Test/.resources/Products/ProductUpdateResponse.txt
+++ b/src/Apps/W1/Shopify/Test/.resources/Products/ProductUpdateResponse.txt
@@ -1,0 +1,1 @@
+{"data":{"productUpdate":{"product":{"onlineStoreUrl":"","onlineStorePreviewUrl":"","updatedAt":"2026-01-01T00:00:00Z"},"userErrors":[]}}}
--- a/src/Apps/W1/Shopify/Test/.resources/Products/ProductVariantsBulkUpdateResponse.txt
+++ b/src/Apps/W1/Shopify/Test/.resources/Products/ProductVariantsBulkUpdateResponse.txt
@@ -1,0 +1,1 @@
+{"data":{"productVariantsBulkUpdate":{"productVariants":[],"userErrors":[]}}}
--- a/src/Apps/W1/Shopify/Test/Products/ShpfyCreateProductTest.Codeunit.al
+++ b/src/Apps/W1/Shopify/Test/Products/ShpfyCreateProductTest.Codeunit.al
@@ -15,11 +15,17 @@ codeunit 139601 "Shpfy Create Product Test"
     Subtype = Test;
     TestType = IntegrationTest;
     TestPermissions = Disabled;
+    TestHttpRequestPolicy = BlockOutboundRequests;
 
 
     var
+        ExportShop: Record "Shpfy Shop";
         Any: Codeunit Any;
         LibraryAssert: Codeunit "Library Assert";
+        OutboundHttpRequests: Codeunit "Library - Variable Storage";
+        LibraryRandom: Codeunit "Library - Random";
+        ShpfyInitializeTest: Codeunit "Shpfy Initialize Test";
+        ExportIsInitialized: Boolean;
 
     [Test]
     procedure UnitTestCreateTempProductFromItem()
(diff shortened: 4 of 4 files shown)


onbuyuka (dismissed):

tVisionDeb:
This PR is mostly a straight backport of #7689, but one additional fix was needed for the test to pass in 28.x CI.

**Additional fix (commit e878f5e):** The test uses \[HttpClientHandler]\ for HTTP mocking, but \Shpfy Communication Mgt.\ (codeunit 30103) is declared \SingleInstance = true\. Every call to \CreateShop()\ sets \IsTestInProgress = true\ on that singleton, which routes all HTTP calls through the old \OnClientSend\/\OnGetContent\ event system instead of \HttpClient.Send()\. Since \[HttpClientHandler]\ only intercepts at \HttpClient.Send()\, no mock response was ever set - resulting in the empty response that caused the 'no JSON for productUpdate' CI failure across all 17 locales.

The fix adds \CommunicationMgt.SetTestInProgress(false)\ to \InitializeProductExport()\, matching the existing pattern in \ShpfyCreateItemAPITest\ (which has the comment 'Disable Event Mocking' for exactly this reason). The original test was validated in an isolated codeunit where \CreateShop()\ had not previously been called; when merged into the existing codeunit 139601 the singleton state from earlier tests in the same session caused the failure.

tVisionDeb:
To add context on the testing approach: the Shopify connector originally mocked HTTP via event subscribers on \OnClientSend\/\OnGetContent\, controlled by the \IsTestInProgress\ flag. Newer test codeunits (written after BC introduced \[HttpClientHandler]\) call \SetTestInProgress(false)\ explicitly to opt out of the old event system - you can see the pattern in \ShpfyCreateItemAPITest\ and others.

In main, the test for #7689 was written in a fresh codeunit that started with a clean singleton state. In this backport it was placed into the existing codeunit 139601, which was written before the \[HttpClientHandler]\ pattern existed and whose setup code never resets the flag. The 28.x backport therefore straddles the transition between the two approaches, and the \SetTestInProgress(false)\ call is the bridge.

Groenbech96:
I removed the workitem fixes. 
I dont think external links should be there. Work items should refer to BC issues.

Groenbech96 (dismissed):
Looks ok.

Aleyenda:
@tVisionDeb there seems to be a bunch of failures on the CI. Can you check and fix them?

@onbuyuka FYI

tVisionDeb:
Thanks for the review, @Aleyenda. I've pushed a fix for the CI failures.

**Root cause:** In the new `UnitTestProductExportDoesNotCreateVariantsForChildItemVariants` test, `InitializeProductExport()` called `CommunicationMgt.SetTestInProgress(false)` **before** `ShpfyInitializeTest.CreateShop()`. `CreateShop()` re-enables `IsTestInProgress` on the singleton `Shpfy Communication Mgt.`, which routed the `productUpdate` call through the old event mock instead of the `[HttpClientHandler]` — producing the empty response ("The response from Shopify contains no JSON") that failed across all locales.

**Fix (test-only):** Reordered so `SetTestInProgress(false)` runs **after** `CreateShop()`, matching the established pattern in `ShpfyCreateItemAPITest`. Commit `fd0791ac9f`.

The build workflows are showing `action_required` — could you approve/re-run them when you have a moment so the fix can be validated? Thanks!

tVisionDeb:
> @tVisionDeb there seems to be a bunch of failures on the CI. Can you check and fix them?
> 
> @onbuyuka FYI

Is this ok now?

onbuyuka (approved):

ReneGubler:
Any news when this will be fixed? We have 5 customers waiting for a fix

JesperSchulz (approved):

PR #8677, opened 2026-06-19
[E-Documents Core] - Internal visibility for Digital Vouchers

GMatuleviciute:

Implementation

Added internal visibility for Digital Vouchers to access E-Documents. 

Important

This Pull request related with https://github.com/microsoft/ALAppExtensions/pull/29793


ventselartur:
@GMatuleviciute, could you please elaborate a bit on this change? Why this change is required and how it is related to the PR? https://github.com/microsoft/ALAppExtensions/pull/29793

I think we have been talking that E-Document Core should not take a dependency on Digital vouchers. It is other way around. Digital voucher can have a dependency from E-Document Core if needed.

GMatuleviciute:
> @GMatuleviciute, could you please elaborate a bit on this change? Why this change is required and how it is related to the PR? [microsoft/ALAppExtensions#29793](https://github.com/microsoft/ALAppExtensions/pull/29793)
> 
> I think we have been talking that E-Document Core should not take a dependency on Digital vouchers. It is other way around. Digital voucher can have a dependency from E-Document Core if needed.

@ventselartur Sure, I have added internal visibility, but not a dependency, to the E-Documents. Even though we created events to subscribe to E-Documents, we still need internal visibility to access them. This is because the codeunits that need to be subscribed to in the E-Doc solution are internal.
Code in Digital Voutchers:
image
E-Doc codeunit: 
image

Groenbech96:
Closing as we wont be giving internal visibility to.

Groenbech96:
Make the events public. Internal visible to is incorrect deisgn.

AndriusAndrulevicius:
> Make the events public. Internal visible to is incorrect deisgn.

@Groenbech96 The property internalsVisibleTo is needed not only because of the events but also because we're using the procedure `GetTempBlob` (which is internal) from table `"E-Doc. Data Storage"`  https://github.com/microsoft/ALAppExtensions/pull/29793/changes#diff-baf95b88a386b44a71c436b7206edb9dac11a79fd8e2d478f7936f7e408cfc8fR354
Should we make this procedure global as well?

GMatuleviciute:
@Groenbech96 rebuild solution to avoid internal visibility and migrated Digital Vouchers changes from ALAppExtensions repos. You can find changes here: https://github.com/microsoft/BCApps/pull/9208

As this pull request is now outdated, I'm closing it.

PR #8746, opened 2026-06-23
Added the interface method override

eikawata:
## What & why

A new interface method was added to IAgentMetadata. This has a default implementation, but this is currently enforced by consumers to explicitly override it.

## Linked work

Fixes [AB#640126](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/640126)

## How I validated this

There's nothing to test for this as it is only adding a default method which is not consumed yet.

## Risk & compatibility

The risk is low as no one uses the added method right now.

### DIFF
--- a/src/System Application/Test/Agent/src/Setup/MockAgent/Integration/MockAgentMetaProv.Codeunit.al
+++ b/src/System Application/Test/Agent/src/Setup/MockAgent/Integration/MockAgentMetaProv.Codeunit.al
@@ -66,6 +66,11 @@ codeunit 133952 "Mock Agent Meta. Prov." implements IAgentMetadata, IAgentFactor
         exit(Page::"Agent Task Message Card");
     end;
 
+    procedure GetCreateAgentTaskPageId(AgentUserId: Guid): Integer
+    begin
+        exit(0);
+    end;
+
     procedure GetDefaultProfile(var TempAllProfile: Record "All Profile" temporary)
     begin
         MockAgentSetup.GetDefaultProfile(TempAllProfile);


SBalslev (approved):

tomasmsft (approved):

src/System Application/Test/Agent/src/Setup/MockAgent/Integration/MockAgentMetaProv.Codeunit.al
```diff
@@ -66,6 +66,11 @@ codeunit 133952 "Mock Agent Meta. Prov." implements IAgentMetadata, IAgentFactor
         exit(Page::"Agent Task Message Card");
     end;
 
+    procedure GetCreateAgentTaskPageId(AgentUserId: Guid): Integer
+    begin
+        exit(0);
```
qutreson, line 71: I think we should try to remove the override of AL0925 as an error in the Compiler.Ruleset.json.

That rule is an informational diagnostic and should not be blocking.

Fixing the ruleset would also ensure future uptakes of platform with default methods will not fail.

qutreson: See #8761

eikawata: Thanks. I'll close this in favor of the new PR.

PR #8749, opened 2026-06-23
[Subcontracting] Bug 638694, 638816: Fix subcontracting return and WIP transfer issues

ChethanT:
### Summary
When performing multiple partial receipt + return cycles on a subcontracting purchase order, the second 'Create Return from Subcontractor' action fails with:
`Transfer-to Code must have a value in Transfer Header: No.=XXXX. It cannot be zero or empty.`

### Root Cause
`UpdateLocationCodeInProdOrderCompAfterDeleteTransferLine` clears `Subc. Original Location Code` on the prod order component after a return transfer is posted/deleted. On subsequent return attempts, `HandleComponentReturnForPurchLine` passes the now-empty field as the Transfer-to Code to `InsertTransferHeader`, which fails validation.

### Fix
Moved the fallback re-population logic to **before** the `InsertTransferHeader` call in `SubcCreateSubCReturnOrder.Report.al`. Previously this ran after the call (too late).

### Test
Added `SecondReturnTransferSucceedsAfterPartialReceiptAndReturn` (codeunit 139991) — [SCENARIO 638694].

---

## Bug 638816: Fix error when toggling off Direct Transfer on WIP transfer order

(See previous commit in this branch)

---

Fixes [AB#638694](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/638694)
Fixes [AB#638816](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/638816)

### DIFF
--- a/src/Apps/W1/Subcontracting/Test/Tests/SubcPurchSubcontTest.Codeunit.al
+++ b/src/Apps/W1/Subcontracting/Test/Tests/SubcPurchSubcontTest.Codeunit.al
@@ -511,6 +511,94 @@ codeunit 139991 "Subc. Purch. Subcont. Test"
     end;
 
     [Test]
+    [HandlerFunctions('DoConfirmCreateProdOrderForSubcontractingProcess,HandleTransferOrder,MessageHandler')]
+    procedure SecondReturnTransferSucceedsAfterPartialReceiptAndReturn()
+    var
+        Item: Record Item;
+        HomeLocation: Record Location;
+        MachineCenter: array[2] of Record "Machine Center";
+        ProdOrderComponent: Record "Prod. Order Component";
+        ProductionOrder: Record "Production Order";
+        PurchaseHeader: Record "Purchase Header";
+        PurchaseLine: Record "Purchase Line";
+        ReturnTransferHeader: Record "Transfer Header";
+        ReturnTransferLine: Record "Transfer Line";
+        TransferHeader: Record "Transfer Header";
+        WorkCenter: array[2] of Record "Work Center";
+        PurchaseOrderPage: TestPage "Purchase Order";
+        ProductionQty: Decimal;
+        FirstReceiptQty: Decimal;
+        SecondReceiptQty: Decimal;
+    begin
+        // [SCENARIO 638694] Creating a second return transfer order after partial receipts must not fail
+        // with "Transfer-to Code must have a value" when Subc. Original Location Code was cleared
+        // by a previous return cycle.
+        Initialize();
+        ProductionQty := 10;
+        FirstReceiptQty := 4;
+        SecondReceiptQty := 3;
+
+        // [GIVEN] A subcontracting purchase order with qty=10 and transfer-to-vendor components
+        SetupSubContractingProdOrder(Item, HomeLocation, WorkCenter, MachineCenter, ProductionOrder, "Component Supply Method"::"Transfer to Vendor", ProductionQty);
+        CreateSubcontractingPurchaseOrderForProdOrder(PurchaseHeader, PurchaseLine, Item, WorkCenter, ProductionOrder);
+
+        // [GIVEN] Outbound transfer order is created and posted (components sent to subcontractor)
+        CreateTransferOrderForPurchaseOrder(PurchaseHeader);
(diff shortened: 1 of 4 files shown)


src/Apps/W1/Subcontracting/Test/Tests/SubcPurchSubcontTest.Codeunit.al
```diff
@@ -510,6 +510,94 @@ codeunit 139991 "Subc. Purch. Subcont. Test"
...
     end;
 
+    [Test]
+    [HandlerFunctions('DoConfirmCreateProdOrderForSubcontractingProcess,HandleTransferOrder,MessageHandler')]
+    procedure SecondReturnTransferSucceedsAfterPartialReceiptAndReturn()
+    var
+        Item: Record Item;
+        HomeLocation: Record Location;
+        MachineCenter: array[2] of Record "Machine Center";
+        ProdOrderComponent: Record "Prod. Order Component";
```
ChethanT:
```suggestion
```

alexei-dobriansky (approved):

alexei-dobriansky (changes requested):
wait a sec

alexei-dobriansky:

### What this PR does
This PR fixes two unrelated subcontracting bugs.

**638694** - On a second "Create Return from Subcontractor", `InsertTransferHeader` received an empty Transfer-to Code and failed validation. The cause is real: after an earlier transfer line is deleted, `UpdateLocationCodeInProdOrderCompAfterDeleteTransferLine` clears `Subc. Original Location Code` (SubcTransferManagement.Codeunit.al:311-314), so the next return passes an empty value. The fix re-populates that field from `Location Code` *before* the call. This is the right fix: it uses the same fallback the code already applied *after* the call (lines 245-246), and at second-return time `Location Code` is the original/home location, so the Transfer-to Code is correct. The added test asserts the Transfer-to Code equals the home location.

**638816** - Toggling off Direct Transfer re-validated `Transfer WIP Item` to a value it already had, which re-ran `CheckForExistingReservationsOrItemTracking()` and errored. The fix only validates when the value actually changes. I verified the BaseApp publisher (`TransferHeader.UpdateTransLines`) calls `TransferLine.Modify(true)` for every line after this event fires, so the line is still persisted and no `Modify` is missing in the subscriber.

### Suggestions

**S1** - *Post-call re-population of `Subc. Original Location Code` is now dead*
After the new line copies `Location Code` into `Subc. Original Location Code` before `InsertTransferHeader`, the same assignment further down (SubcCreateSubCReturnOrder.Report.al:245-246) can never run, because the field is no longer empty at that point. Consider removing those two lines so it's clear there is a single source of truth. Keep the bin-code block right below it (247-248) - that one is still needed.

**Risk:** Low and well contained. Both changes are inside the Subcontracting app and gated by the subcontracting feature flag; no public API, event signature, or table field changed. The 638694 fix reuses an existing fallback value, so it does not introduce a new location-selection rule. The 638816 fix only skips a no-op validate, so its only behavioral effect is not throwing on an unchanged value. The BaseApp `Modify(true)` after the event was confirmed, so the WIP description/validate updates are saved.

**Necessity:** Both are clear, reproducible defects that block normal partial-receipt return cycles and Direct Transfer toggling on WIP transfer orders, with no reasonable workaround for the user. The scope is minimal and appropriate for a bug fix, and each fix ships with a regression test that exercises the exact failing path. Bundling two unrelated bugs in one PR is a minor process note, not a blocker.

---
> [AI-PR-REVIEW] version=1 system=github pr=8749 round=1 by=alexei-dobriansky at=2026-06-24T08:46:00Z lastSha=5f15283f8f31b303bdaecc29251c4b0dff549a0c suggestions=S1

AleksandricMarko (dismissed):

alexei-dobriansky (commented):
please check the suggestions

AleksandricMarko (approved):

PredragMaricic (approved):

alexei-dobriansky (approved):

PR #8762, opened 2026-06-24
Add System Application dependency to Test Framework apps

thloke:
## What

Adds an explicit dependency on **System Application** (`63ca2fa4-4f03-4f2b-a480-172fef340d3f`) to the four Test Framework library apps:

- Any
- Library Assert
- Test Runner
- Library Variable Storage

## Why

These apps previously declared **no application-level dependency** (platform only). As a result the app dependency resolver treats them as compatible across application builds, so a tenant can end up with a newer build of these apps published than the application schema being applied during an environment upgrade — which blocks the upgrade sync (forward-only schema guard).

Declaring a dependency on System Application version-gates the apps to the application build, modelling the dependency the same way other first-party apps (e.g. Business Foundation) already do.

## Circular dependency check

None introduced:

- The umbrella `System Application` app declares `dependencies: []`.
- **0** of the 111 runtime sub-apps under `System Application/App/` depend on any Test Framework app.
- The only consumers of the Test Framework inside System Application are the separate `Test` / `Test Library` / `Partner Test` packages, which sit **above** the framework in the graph.

Graph stays acyclic: `System Application (App) -> Test Framework -> System Application Test`.

> [!NOTE]
> Draft / experimental — raised in the context of an upgrade-compatibility investigation. Architectural note: this inverts the conceptual layering (the Test Framework would sit above the entire System Application runtime surface).

### DIFF
--- a/src/Tools/Test Framework/Test Libraries/Any/app.json
+++ b/src/Tools/Test Framework/Test Libraries/Any/app.json
@@ -11,7 +11,14 @@
   "url": "https://go.microsoft.com/fwlink/?LinkId=724011",
   "contextSensitiveHelpUrl": "https://learn.microsoft.com/dynamics365/business-central/dev-itpro/developer/devenv-testing-application/",
   "logo": "ExtensionLogo.png",
-  "dependencies": [],
+  "dependencies": [
+    {
+      "id": "63ca2fa4-4f03-4f2b-a480-172fef340d3f",
+      "name": "System Application",
+      "publisher": "Microsoft",
+      "version": "29.0.0.0"
+    }
+  ],
   "screenshots": [],
   "platform": "29.0.0.0",
   "idRanges": [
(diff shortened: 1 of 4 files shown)


alexei-dobriansky:

# **Recommendation: Accept**

<details>

This change adds an explicit System Application dependency to four Test Framework app manifests. The fix targets dependency resolution during upgrades, and the dependency versions match the apps' own version and platform version.
</details>

<details>

**Fit:** Strong

The reported upgrade-compatibility problem is addressed at the manifest dependency level, which is where the resolver makes the compatibility decision. The scope is limited to the affected Cloud Test Framework apps.
</details>

<details>

No suggestions.
</details>

<details>

**Risk:** The main risk is dependency graph or install-order regression. The System Application umbrella app has no dependencies, and the runtime System Application app graph does not depend on these Test Framework apps.

**Necessity:** The change is justified because it version-gates the test framework packages to the application build and prevents mismatched package resolution during upgrades.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=8762 round=1 by=alexei-dobriansky at=2026-09-11T12:58:36Z lastSha=663e9c14faf7a7941997097368a2a11208413e65 reviewKey=b53982aa311730d4317ce92dd91bb72d2a838f68add795a88de5594b8911e285 suggestions=none

PR #8776, opened 2026-06-24
[Subcontracting] SubcPurchaseOrderCreator has guards (Gen. Prod. Posting Group, blocked items) that SubcCalculateSubcontracts lacks

VolodySyn:
Fixes [AB#635775](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/635775)


PredragMaricic (approved):

alexei-dobriansky:

### What this PR does
This PR consolidates two validation guards into shared procedures on `Subcontracting Management` (`CheckSubcontractingWorkCenter` and `CheckProdNotBlockedForOutput`) and applies them in both subcontracting entry points: the `Subc. Calculate Subcontracts` report and the `Subc. Purchase Order Creator`. The goal is to make both paths fail fast and consistently when a subcontracting work center has no Gen. Prod. Posting Group, or when the manufactured item/variant is blocked for production output.

The refactor is sound. The extracted procedures match what the Purchase Order Creator already did inline (work-center `TestField`s), and the duplicated inline code is correctly removed with a single remaining caller, so no dead references are left. The work-center check is correctly gated by `ProdOrderRoutingLine.Type = "Work Center"` in the report. One real behavior change is intentional and worth highlighting: the old report flow used a `Confirm` ("...Do you want to continue?") that let the user skip a single blocked item and keep calculating, whereas the new `CheckProdNotBlockedForOutput` raises a hard `Error`. This is defensible - the new `ProductionBlockedOutputItemErr` label is byte-for-byte identical to base manufacturing's message in `MfgItem.TableExt.al`, so subcontracting now behaves like the rest of the platform, which errors hard on blocked output. It also removes a latent issue where the old `CanCreateRequisitionLineFromProdOrderLine` returned `false` when `not GuiAllowed()`, silently skipping lines in headless runs.

### Suggestions

**S1** - *Add a test for the blocked-output variant path*
The four new tests cover the work-center and item-level guards well, but the variant branch of `CheckProdNotBlockedForOutput` (the `ProductionBlockedOutputItemVariantErr` path) is never exercised. A short test that blocks an item variant for output and asserts the error would close the only untested branch of the new code.

**S2** - *Confirm the hard-error for blocked output is the intended UX*
Because Calculate Subcontracts is a batch over many orders, a single blocked item now aborts the whole run instead of letting the user skip it and continue. This matches base manufacturing and the work item's intent, so it is likely correct - just confirming the removal of the skip-and-continue option is deliberate, as it is a user-visible change for anyone who relied on the previous confirmation.

**Risk:** Low-to-moderate and contained within the Subcontracting app. No public signatures change; the two new procedures are `internal`. The notable change is the `Confirm` -> `Error` semantics for blocked-output items in Calculate Subcontracts (whole-batch abort instead of per-item skip) and the fact that the blocked-output guard is now enforced in the Purchase Order Creator where it previously was not - both align with standard manufacturing behavior, but partners or users relying on the old confirm/skip flow will see a behavior change. CI is green, including the W1 app and test builds.

**Necessity:** Justified. Without these guards, Calculate Subcontracts could create requisition lines for a work center with a blank Gen. Prod. Posting Group or a blocked item, deferring the failure to posting time where it is harder to diagnose. Failing fast at calculation time is the correct fix, and the scope is appropriate for the linked bug. The change ships with targeted regression tests for both entry points.

---
> [AI-PR-REVIEW] version=1 system=github pr=8776 round=1 by=alexei-dobriansky at=2026-06-25T10:05:00Z lastSha=52ad089a16afbf4986d830409512cdf20605710c suggestions=S1,S2

alexei-dobriansky (approved):

alexei-dobriansky (commented):
Please confirm that S2 is intentional.

PR #8819, opened 2026-06-25
Harden Copilot PR-review JSON parser against resume-fence corruption

JesperSchulz:
## Problem

When the review agent breaks off mid-string and resumes, a single malformed in-string value can make `ConvertFrom-Json` reject the **entire** findings report — so a full set of findings is posted as **0**. Observed in a production run that found 42 issues across 11 sub-skills and posted none.

Two compounding bugs in `Parse-BCQualityReport`:

1. **Resume-fence corruption.** The agent glitched mid-string inside a `suggested-code` value, emitting a bare ` ```json ` resume fence inside an unterminated string (no "Placeholder" marker). `Repair-InterruptedAgentJson` keys off that marker so it does not fire, and `Remove-StructuralFences` deliberately preserves in-string fences — so neither existing repair clears it, and the whole document fails to parse.
2. **Wrong fragment wins.** Even after repair, the candidate loop accepted the *first* parseable fragment. The transcript echoes each sub-skill's own JSON and the top-level `findings[]` is itself a large balanced array — so a bare sub-skill report (or the bare array) parsed first and the real orchestrator report was silently dropped.

## Fix

- **`Repair-ResumeFenceJson`** — a third, conservative repair path. Key-anchored, transcript-safe splice that removes a bare resume fence sitting inside an *unterminated* string, keeping the re-emitted property and discarding the broken partial. It acts only on the unambiguous triple signature (unterminated property → bare fence → re-emission of the same key), so well-formed output and legitimate in-string code fences are never touched.
- **Tiered candidate preference** — prefer the orchestrator report shape (`sub-results`/`dispatch`) → then any `findings`-bearing object → then the first parseable candidate. Leniency preserved; the correct document always wins.
- **`Test-InterruptedJsonRepair.ps1`** — new regression suite. Extracts the parser functions from `Invoke-CopilotPRReview.ps1` via the PowerShell AST so tests cannot drift, covers synthetic Shapes 1–3 plus guards (clean report unaffected, in-string ` ```al ` code sample preserved), and pins three real-runner fixtures that the unpatched parser returned 0 for.

## Verification

All 30 assertions pass, including the real fixture recovering **42 findings / 11 sub-results** (the unpatched parser returned 0), plus the two existing real-runner fixtures (25 and 22 findings).

## Context

This was first found and fixed in microsoft/BCAppsBCQuality (which carries a port of this reviewer); this PR brings the same hardening upstream. The `$DomainMap` is intentionally left unchanged here.

Fixes: [AB#640481](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/640481)

### DIFF
--- a/tools/Code Review/scripts/Invoke-CopilotPRReview.ps1
+++ b/tools/Code Review/scripts/Invoke-CopilotPRReview.ps1
@@ -1038,6 +1038,109 @@ function Remove-StructuralFences {
     return $sb.ToString()
 }
 
+function Repair-ResumeFenceJson {
+    <#
+    Repairs the interrupted-emission shape where the agent breaks off
+    mid-string — leaving an UNTERMINATED JSON string — and resumes by
+    re-opening a fresh ```json fence WITHOUT emitting a "Placeholder ..."
+    marker. Observed in production (run 28168894037), inside a
+    `suggested-code` value:
+
+        "confidence": "high",
+        "suggested-code": "    ODataKeyFields =
+                                              <- raw newline(s) inside the string
+        ```json                               <- bare resume fence, no marker
+              "suggested-code": "    ODataKeyFields = SystemId;"
+
+    A raw newline inside a string literal is always invalid JSON, so this
+    shape is unambiguous. A single malformed value like this makes
+    ConvertFrom-Json reject the ENTIRE document, discarding every finding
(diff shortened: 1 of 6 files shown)


JesperSchulz:
JSON hardening is a general issue, which we're going to address in a coordinated effort, rather than on patch basis. Closing this PR.

PR #8820, opened 2026-06-25
Authenticate trusted base/PR-head fetches when a token is available

JesperSchulz:
## Problem

The `trusted` checkout in the Copilot PR review runner uses `persist-credentials: false` so the tool-enabled Copilot CLI (which later processes untrusted PR content) can never read a token from `.git/config`, and the base/PR-head refs are fetched anonymously. That works for the public BCApps repo, but **private repos and ports** of this reviewer fail in Discovery with:

```
fatal: could not read Username for 'https://github.com': No such device or address
```

aborting the review before any analysis runs. (Hit in the private microsoft/BCAppsBCQuality port after it synced this reviewer.)

## Fix

Wrap the two pre-CLI discovery fetches in `Invoke-WithGitHubFetchAuth`, which injects the workflow's `GH_TOKEN` as an **ephemeral** `http.extraheader` via `GIT_CONFIG_*` env vars:

- The credential is **never written to `.git/config`**, so it stays invisible to the tool-enabled CLI — the `persist-credentials: false` security intent is fully preserved.
- Passing via env (not `git -c http...extraheader=...`) keeps the token off the command line, out of the process table, and out of `Invoke-GitCommand`'s thrown error text.
- It is cleared the instant the two fetches finish; the later worktree/diff operations are local-only.

When no token is present the fetch proceeds anonymously, so **current public-repo behaviour is unchanged** unless a token happens to be in the environment.

## Note for maintainers

On the public BCApps runner `GH_TOKEN` *is* set in the generate-phase env, so this would switch its base/PR-head fetches from anonymous to authenticated. That's harmless and more robust, but if you'd prefer to keep BCApps strictly anonymous I can gate the auth behind an opt-in env var instead — happy to adjust.

### DIFF
--- a/tools/Code Review/scripts/Invoke-CopilotPRReview.ps1
+++ b/tools/Code Review/scripts/Invoke-CopilotPRReview.ps1
@@ -361,6 +361,51 @@ function Invoke-GitCommand {
     return $output
 }
 
+function Invoke-WithGitHubFetchAuth {
+    <#
+    Runs $Action with an ephemeral GitHub credential injected via GIT_CONFIG_*
+    environment variables, so the trusted checkout's base/PR-head fetches succeed
+    against a PRIVATE repository.
+
+    The trusted checkout deliberately uses persist-credentials:false so the
+    tool-enabled Copilot CLI (which later processes untrusted PR content) can
+    never read a token from .git/config. That is fine for a PUBLIC repo, where
+    the base/PR-head refs fetch anonymously — but this reviewer also runs on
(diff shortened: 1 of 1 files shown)


JesperSchulz:
Not needed.

JesperSchulz:
For the record: the fetch-auth fix is only needed for **private** repos. BCApps is public, so base/PR-head fetches succeed anonymously. This change belongs only in the private BCAppsBCQuality fork (microsoft/BCAppsBCQuality#54). The parser-hardening port (#8819) still stands.

PR #8825, opened 2026-06-25
Decouple Copilot PR-review orchestrator from BCQuality's domain taxonomy

JesperSchulz:
## What & why

The Copilot PR-review orchestrator (`tools/Code Review/scripts/Invoke-CopilotPRReview.ps1`) currently maps each finding's `from-sub-skill` id to a display label via a static `$DomainMap`. That map has to be edited every time BCQuality adds a new domain, which couples this repo to BCQuality's internal taxonomy.

BCQuality is being updated (separate, companion PR) to emit a human-readable `domain` label on every finding. This change makes the orchestrator **prefer** that emitted label and keep `$DomainMap` only as a **legacy fallback**.

## Changes

- `Resolve-FindingDomain` now reads an explicit `domain` (or `Domain`) property off each finding and renders it verbatim when present. Only when it's absent does it fall back to the static `$DomainMap` keyed on `from-sub-skill`, and finally to `Other`.
- The comment above `$DomainMap` is re-documented as a legacy fallback that no longer needs an entry per new domain.
- **Slug-hardening of dedup metadata:** added a `ConvertTo-DomainSlug` helper and routed the `agent_domain` metadata + heading-pattern matching through it. The heading patterns now accept multi-word domain labels, and both the emitted metadata and the parsed comparison are slug-encoded, so multi-word labels (e.g. "Breaking Changes") round-trip correctly through the cross-iteration dedup logic instead of being truncated/mismatched.
- `tools/Code Review/README.md` "Domain labels" section updated to describe the emitted-label behavior with the legacy-map fallback.

## Backward compatibility

Fully backward-compatible. Findings that do **not** carry a `domain` field still resolve exactly as before via the legacy `$DomainMap` → `Other`. The new behavior only activates once findings start arriving with a `domain` label.

The BCQuality pin (`tools/BCQuality/bcquality.config.yaml`) is intentionally **not** changed here. Until the pin is later bumped to a BCQuality commit that emits `domain`, the legacy fallback preserves current behavior exactly.

## Recommended rollout order

1. Merge this PR (safe no-op until BCQuality emits `domain`).
2. Merge the companion BCQuality PR that emits a `domain` label on every finding.
3. Bump the BCApps BCQuality pin to that BCQuality commit to activate the decoupled path.

## Validation

- `Invoke-CopilotPRReview.ps1` parses cleanly (`[System.Management.Automation.Language.Parser]::ParseFile` → `parse OK`).

Fixes [AB#640493](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/640493)


JesperSchulz:
Closing as superseded by BCApps #9210, which removed the in-repository review engine and delegated it to microsoft/BC-ALAgents. After syncing with current main, this PR has an empty diff and must not be merged. The consumer fix for AB#640493 is moving to BC-ALAgents and will be linked from BCQuality #54 when available.

PR #8830, opened 2026-06-26
[Main] [Subscription Billing] Infinite loop in CalculateBillingPeriod when HarmonizeNextBillingTo caps at stale "Next Billing To" date

sanjmaurya:
[AB#640056](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/640056)

Bug 640056: [master] [ALL-E][Subscription Billing] Infinite loop in CalculateBillingPeriod when HarmonizeNextBillingTo caps at stale "Next Billing To" date

**Issue:-** CalculateBillingPeriod in BillingProposal.Codeunit.al enters an infinite loop (hang / no response) when generating a billing proposal for a harmonized customer subscription contract whose "Next Billing To" date is stale (older than the current billing period being calculated).

**Cause :-** Root Cause
The while loop advances the billing period end date by repeatedly calling CalculateNextBillingToDateForServiceCommitment:

For a harmonized contract, CalculateNextBillingToDateForServiceCommitment calls HarmonizeNextBillingTo, which caps the computed date at CustomerContract."Next Billing To":

When the contract's "Next Billing To" is a stale date earlier than BillingDate, every iteration gets capped back to that same stale value. So BillingPeriodEnd never advances past it, the loop condition BillingPeriodEnd < BillingDate stays true forever, and the loop never terminates.
Solution
Add a progress guard that breaks the loop as soon as BillingPeriodEnd stops advancing (i.e., is capped to the same/earlier date). A PreviousBillingPeriodEnd tracker is introduced, and the loop only continues while BillingPeriodEnd > PreviousBillingPeriodEnd:

### DIFF
--- a/src/Apps/W1/Subscription Billing/App/Billing/Codeunits/BillingProposal.Codeunit.al
+++ b/src/Apps/W1/Subscription Billing/App/Billing/Codeunits/BillingProposal.Codeunit.al
@@ -483,6 +483,7 @@ codeunit 8062 "Billing Proposal"
     local procedure CalculateBillingPeriod(ServiceCommitment: Record "Subscription Line"; BillingDate: Date; BillToDate: Date; var BillingPeriodStart: Date; var BillingPeriodEnd: Date)
     var
         UsageDataBilling: Record "Usage Data Billing";
+        PreviousBillingPeriodEnd: Date;
     begin
         BillingPeriodEnd := 0D;
         BillingPeriodStart := ServiceCommitment."Next Billing Date";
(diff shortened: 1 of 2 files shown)


alexei-dobriansky:

# **Recommendation: Accept**

<details>

This PR adds a non-progress guard to `CalculateBillingPeriod` and adds a regression test for a harmonized customer subscription contract where `Next Billing To` caps the calculated period before the requested Billing Date.

The loop now keeps the previous `BillingPeriodEnd` and only repeats when the recalculated date moves forward. I traced the explicit `BillToDate`, usage-based billing, normal recurring billing, and harmonized cap paths. In the stale-cap path the next calculation can return the same date, so the new guard stops the loop instead of spinning forever. The following `UpdateBillingLine` path still creates the capped billing line and advances the subscription from that line.
</details>

<details>

*No blocking or optional suggestions.*
</details>

<details>

**Risk:** This area controls billing proposal periods and next billing dates for subscription contracts, so a wrong change could create too few or too many billing lines. The code change is narrow: it only changes the loop guard in codeunit 8062 and does not change public signatures, events, tables, or BaseApp-dependent subscribers. The added test covers the stale harmonized cap that caused the loop.

**Necessity:** The linked bug describes an active hang in billing proposal creation for harmonized contracts, with SQL timeout as the later symptom. The fix is needed because users cannot reliably create the proposal in that state, and the scope is limited to stopping the non-advancing billing-period calculation.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=8830 round=1 by=alexei-dobriansky at=2026-07-02 lastSha=1214474c71eab0c79ae1d870d58255f5a71c1935 suggestions=none

v-ankitgoyal (approved):

AndersLarsenMicrosoft (approved):

dcenic (approved):

PR #8857, opened 2026-06-29
[Master]-Planning Error Log confuses Users "Planning Parameters were taken from Item Card because Missing SKU Planning Policy is Item Card but SKU does not exist for Item xxx at Location yyy."

neeleshsinghal:


neeleshsinghal:
Base was wrong

PR #8908, opened 2026-06-30
Add composite layout theme and header/footer management to the Base A…

v-mortenr:
## What & why

Adds **composite report layouts** to the Base Application: a report's Word layout can compose a shared **Theme** part and a shared **Header/Footer** part, configured independently and resolved at render time. Parts resolve across four levels — *this layout → report default → company → global default* — most-specific wins, with theme and header/footer resolved independently. This lets administrators manage branding centrally instead of duplicating it in every layout.

Includes:
- New Reporting objects (`src/Layers/W1/BaseApp/Foundation/Reporting`): `Composite Layout Lookup Helper` (encode/decode part references + resolve the effective theme/header-footer and its source level) and the management/assignment pages (`Tenant Report Layout Cfg`/`Selection`, theme & header/footer assignment, new-part dialog, theme-and-header/footer box & list).
- `Report Layouts` page and `Report Layouts Impl.` extended with subtype-aware layout creation (Theme / Header-Footer) and the composite resolver.
- `Company Information` surfaced for the shared company metadata (W1 + localization layers).
- `D365 Bus Full Access` grants for the new configuration tables.
- `Composite Layout Tests` codeunit (resolver precedence + theme/header-footer independence).

## Linked work

[AB#636293](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/636293)

## How I validated this

- **Automated** — `Composite Layout Tests` codeunit covers: no configuration resolves to *None* (blank source); resolution and source at each level (this layout / report default / company / global default); header and theme resolving independently from different levels; the more-specific level winning over global; and composite-name encode/decode round-trip (lowercase GUID). Setup is data-agnostic (touches only data it creates).
- **Manual** — built the Base Application locally (no new analyzer warnings) and ran in Business Central: configured theme and header/footer parts at the report/company/global levels via the Report Layouts UI, rendered a Word report, and confirmed the header/footer and theme were composed correctly and that resolution falls back to the next level when a part is cleared.

## Risk & compatibility

- **Additive** — new pages/codeunit, new configuration tables, and additive permission grants; no objects removed and no breaking signature changes (a new `InsertNewLayout` overload is added; the original is kept).
- Gated behind the existing **Document Report Experience** feature key; with no composite parts configured, report rendering is unchanged.
- `Company Information` changes span W1 and the localization layers to surface the shared metadata consistently — please sanity-check the localization-layer edits during review.


v-mortenr:
@microsoft-github-policy-service agree company="Abakion A/S"

src/Layers/APAC/BaseApp/Foundation/Company/CompanyInformation.Page.al
```diff
@@ -416,6 +417,48 @@ page 1 "Company Information"
                     end;
                 }
             }
+            group(ReportLayoutDefaults)
+            {
+                Caption = 'Report Layout Defaults';
```
mynjj: Maybe the name here should be a bit more generic so that we can use it for other things? I remember (when discussing a different slice with Kennie) that he proposed having a section in Company Information called "Reporting" and he mentioned that your project would need it too. Should we maybe check with him?

src/Layers/W1/BaseApp/Foundation/Reporting/ReportLayouts.page.al
```diff
@@ -5,6 +5,7 @@
 
 namespace Microsoft.Shared.Report;
```
mynjj, line 6: minor: several of these files have `.page` instead of the usual `.Page` in their names

mynjj: ah, this one is not new, so no need to rename (but maybe we should keep the convention for the new ones)

BardurKnudsen (dismissed):

src/Layers/W1/BaseApp/Foundation/Reporting/ReportLayoutNewDialog.page.al
```diff
@@ -113,28 +123,46 @@ page 9662 "Report Layout New Dialog"
...
     trigger OnOpenPage()
+    var
+        FeatureKeyManagement: Codeunit "Feature Key Management";
     begin
-        FormatOptions := FormatOptions::Excel;
         CreateEmptyLayout := false;
         ExcelMultipleDataSheets := "Excel Sheet Configuration"::Default;
         LayoutName := '';
         AvailableInAllCompanies := true;
+        DocumentReportExperienceEnabled := FeatureKeyManagement.IsDocumentReportExperienceEnabled();
```
mynjj, line 133: When introducing features like this, since eventually they will be activated by default, we usually wrap them with `CLEAN` preprocessor symbols so that when the enforcing becomes a reality it becomes straightforward to clean-up the code. 

Whenever we branch off say for 29.0, as part of a clean-up we activate the symbol "CLEAN26" (3 releases behind) and if there's any code wrapped between `#if not CLEAN26` `#endif` will go away. In this case you would use CLEAN29 for it to be enforced on 32 (or you check when you want this by default)

See examples of usages of `FeatureManagementFacade.IsEnabled`, they usually are accompanied by these symbols.

v-mortenr: @mynjj Good point — wrapping the feature gating in `#if not CLEAN29` pattern is the right approach. To keep this PR focused and unblock the merge, I will handle the CLEAN29 wrapping in a dedicated follow-up task rather than in this PR, and link it here. AB#641000

src/Layers/APAC/BaseApp/Foundation/Company/CompanyInformation.Page.al
v-mortenr: @mynjj good point on the naming. Note that `Reporting` is already used in a localization layer, so we need the name to account for that. I will check with Kennie about aligning this with the proposed Company Information "Reporting" section.

src/Layers/W1/BaseApp/Foundation/Reporting/ReportLayouts.page.al
v-mortenr: Done in 026015a — renamed the new page files to `.Page.al`. Left the existing `ReportLayouts.page.al` as-is since it is not new, per your note.

src/Layers/W1/BaseApp/Foundation/Reporting/ReportLayoutNewDialog.page.al
mynjj: I recommend you do it already in this PR: it's a bit easier as reviewer to see that you are not missing any place to consider for this conditional cleanup logic when the feature is new and I can see all the scope in a single place.

Up to you, I also respect the other approach.

v-mortenr: @mynjj I will do a follow up PR for the Clean 29 wrapping, I have a team waiting to get started on testing a lot of reports work as expected with the new merge functionality and we are getting closer to the Gates deadline.

src/Layers/APAC/BaseApp/Foundation/Company/CompanyInformation.Page.al
v-mortenr: @mynjj Confirmed with Kennie — he suggested a single "Reporting" section in Company Information (W1), used for document reporting and, later, Power BI settings, with localizations adding their own fields into it.

Done in c93aa688: renamed the `ReportLayoutDefaults` group to `Reporting` across all localizations, so `Default Theme` and `Default Header/Footer` now live in that shared section. In RU — which already has a statutory `Reporting` group — the two fields fold into the existing group, with the document-report-experience gating moved from the group level to each field (the group must stay visible for its statutory fields). W1 BaseApp compiles clean.

src/Layers/W1/BaseApp/Permissions/d365busfullaccess.permissionset.al
```diff
@@ -221,6 +221,8 @@ permissionset 959 "D365 BUS FULL ACCESS"
                   tabledata "Tenant Profile Extension" = imd,
                   tabledata "Tenant Profile Page Metadata" = imd,
                   tabledata "Tenant Profile Setting" = imd,
+                  tabledata "Tenant Report Layout Cfg" = RIMD,
```
mynjj, line 224: other permissionsets don't need these?

mynjj (dismissed):

PredragMaricic (dismissed):

v-mortenr:
@mynjj

LegacyTestsBucket1 failures

Two groups:

1. My composite-layout tests (ReportDefault/Company/GlobalDefaultResolvesAs…) — failed with Actual='MyHF' due to a test-isolation bug: Initialize() didn't reset config rows, so a row from an earlier test leaked in. Fixed in 510086a (reset the suite's rows in Initialize).

2. Unrelated pre-existing tests (sales/service doc totals, item-tracking, checklist, GLTransferWithClosingDate, ServiceOrderRelease) — I don't believe these are caused by this PR:
- The PR edits none of these codeunits.
- The failing set differs per locale (US vs NL fail on different tests) — a real regression would fail consistently.
- Symptoms are shared-setup corruption (wrong VAT/discount amounts, off-by-one counts, missing posting setup), which report-layout changes can't produce.
- LegacyTestsBucket1 runs without per-test rollback on a shared company; adding a new test codeunit reorders the bucket and exposes latent order-dependencies among these legacy tests.

@PredragMaricic — are these known-unstable legacy tests, and is a rerun the accepted path if they fail again, or should I log them separately? (The pipeline's RerunUnstableFailures stage suggests they're known-flaky.)

mynjj (dismissed):

src/Layers/W1/BaseApp/Foundation/Reporting/ReportLayoutsImpl.codeunit.al
```diff
@@ -467,12 +504,71 @@ codeunit 9660 "Report Layouts Impl."
...
+    local procedure ValidateCompositePartReferences(SelectedReportLayoutList: Record "Report Layout List"): Text
+    var
+        TenantReportLayoutCfg: Record "Tenant Report Layout Cfg";
+        Warnings: TextBuilder;
+        PartNotFoundTxt: Label 'Header/Footer part "%1" was not found in Report Layout List.', Comment = '%1 = composite part name';
+        ThemeNotFoundTxt: Label 'Theme part "%1" was not found in Report Layout List.', Comment = '%1 = composite part name';
+        PartObsoleteTxt: Label 'Header/Footer part "%1" is marked as obsolete.', Comment = '%1 = composite part name';
+        ThemeObsoleteTxt: Label 'Theme part "%1" is marked as obsolete.', Comment = '%1 = composite part name';
+    begin
+        TenantReportLayoutCfg.SetRange("Report ID", SelectedReportLayoutList."Report ID");
```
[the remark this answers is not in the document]
v-mortenr: A display helper that strips the AppId (<guid>::<name> → <name>) — then filters Report Layout List by Name + Layout Subtype only. If two apps ship a part with the same name and same subtype, FindFirst() can match the wrong one. The AppId is present in the composite reference and simply isn't used.

But the blast radius is small: this routine only produces a warning message about whether a referenced part exists / is obsolete. It does not drive rendering or resolution (that's the platform resolver + Composite Layout Lookup Helper, which uses the full composite string). So worst case is a misleading "not found/obsolete" warning on a cross-app name collision — not wrong output. Legitimate to fix, not a blocker.

PredragMaricic (dismissed):

alexei-dobriansky:

<details>

This PR adds Base Application UI and helper code for composite Word report layouts: reusable Theme and Header/Footer parts, report/layout/company/global defaults, Company Information defaults, permission grants, and resolver tests. The resolver shape is coherent and the tests cover the main precedence rules, including independent theme and header/footer fallback. The change is not ready because the new administration UI can write tenant-wide configuration without a clear permission gate, and the assignment dialog can save changes even when the user cancels.
</details>

<details>

**S1** - *Gate composite layout administration*
The new Report Layouts actions open pages that grant `Tenant Report Layout Cfg` write permissions themselves. Add `AccessByPermission` or an equivalent reporting-admin check before users can manage tenant-wide theme/header-footer defaults. Otherwise a user who can open Report Layouts can change report branding defaults without explicit permission to that configuration table.

**S2** - *Do not save cancelled assignments*
`Header/Footer Theme Assignment` inserts or updates `Tenant Report Layout Cfg` while the dialog is open, and callers ignore the `RunModal()` result. Stage the selected parts in page variables and write the row only when the dialog closes with OK. A Cancel action must not change the effective theme or header/footer for a report.

**S3** - *Match composite references by app ID*
`ValidatePartReference` decodes `AppId::LayoutName` to only the layout name and then searches by name and subtype. Keep and use the App ID part as well. This avoids validating the wrong part when two apps provide a theme or header/footer with the same name.

**S4** - *Use stored-length composite values consistently*
`HasCompositePartReferences` filters on the full encoded composite name, while the other code paths store and filter the same value with `CopyStr(..., MaxStrLen(...))`. Apply the same stored-length value there too. Without that, long part names can skip the delete warning even though the part is assigned.
</details>

<details>

**Risk:** The feature is gated, but when enabled it changes effective Word report rendering through layout, report, company, and global defaults. Wrong writes in `Tenant Report Layout Cfg` can change customer-facing report branding across reports or companies. The public surface looks additive, and the original `InsertNewLayout` overload is kept, but the new elevated pages and dialog save behavior are high-risk UI paths.

**Necessity:** The linked work item is a Deliverable, so feature mode applies. The feature need is clear: expose platform composite Word layout support so administrators can manage shared branding parts centrally. The scope is broad but aligned with that deliverable; the blockers are in the admin write path, not in the resolver concept.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=8908 round=1 by=alexei-dobriansky at=2026-07-02T09:54:25.1557327Z lastSha=510086a1da07365c40d88b0c971abc2df48fb9f5 suggestions=S1,S2,S3,S4

alexei-dobriansky (changes requested):
Please review the suggestions

v-mortenr:
Addressed the review feedback from @alexei-dobriansky in the latest commits.

**Test isolation (the earlier `LegacyTestsBucket1` failures).** `Composite Layout Tests`.`Initialize()` now resets the suite's own configuration rows (report `50000` and the report-`0` wildcard company/global rows) before each test. These run in a non-isolated Legacy bucket, so the layout-level row left by `LayoutLevelAssignmentResolvesAsThisLayout` was leaking into the report/company/global-default tests and winning resolution (`Actual = MyHF`). All per-country `LegacyTestsBucket1` jobs are green after the fix.

**S1 – Gate composite layout administration.** The assignment page and lookup helper hold indirect RIMD on `Tenant Report Layout Cfg`, so I added `AccessByPermission = tabledata "Tenant Report Layout Cfg" = M;` to the three composite actions — only a user who actually holds Modify on that table can reach them.

**S2 – Do not save cancelled assignments.** `Header/Footer Theme Assignment` now uses a temporary source table, stages the selection, and writes `Tenant Report Layout Cfg` only when the dialog closes with OK. Cancel no longer changes the effective theme/header-footer.

**S3 – Match composite references by app ID.** Added `DecodeAppId` and used it in `ValidatePartReference` to filter `Report Layout List` by `Application ID` as well as name/subtype, so a same-named part of the same subtype in another app isn't matched.

**S4 – Stored-length composite values.** `HasCompositePartReferences` now filters at the stored field length (`CopyStr`/`MaxStrLen`), matching the assign/clear paths, so a long part name no longer skips the delete warning.

The remaining `Build Unit Tests (LegacyTestsBucket1)` (W1) failure is `GLTransferWithClosingDate` (Cost Register entry count), which is unrelated to this change — a pre-existing order-dependent legacy test surfaced by bucket ordering.

PredragMaricic (approved):

alexei-dobriansky:

# **Recommendation: Accept**

<details>

The commits since round 1 harden the composite layout administration path. The Report Layouts actions now require Modify permission on `Tenant Report Layout Cfg`, the assignment dialog stages changes in a temporary record and writes only on OK, validation keeps the encoded app ID, and deletion checks filter with the stored field length. These changes match the shape requested in round 1 and are limited to the admin/write validation paths.
</details>

<details>

| ID | Title | Status | Note |
|----|-------|--------|------|
| S1 | Gate composite layout administration | Addressed | `ReportLayouts.page.al` now gates all three composite administration actions with `AccessByPermission = tabledata "Tenant Report Layout Cfg" = M` at lines 423-467. |
| S2 | Do not save cancelled assignments | Addressed | `HeaderFooterThemeAssignment.Page.al` now uses a temporary source table and persists only when `CloseAction = Action::OK` at lines 21-23 and 113-141. |
| S3 | Match composite references by app ID | Addressed | `DecodeAppId` was added at `CompositeLayoutLookupHelper.Codeunit.al` lines 67-83, and `ValidatePartReference` filters by `Application ID` at `ReportLayoutsImpl.codeunit.al` lines 556-563. |
| S4 | Use stored-length composite values consistently | Addressed | `HasCompositePartReferences` now applies `CopyStr(..., MaxStrLen(...))` before filtering `Header Part Name` and `Theme Part Name` at `ReportLayouts.page.al` lines 794-800. |
</details>

<details>

</details>

<details>

**Risk:** The remaining risk is in tenant/company branding administration. If this code is wrong, users could be blocked from managing composite defaults or the wrong header/footer or theme could be written. The round-2 diff does not introduce new events or event subscribers, and the BaseApp objects reviewed are in-repo under `src/Layers/W1/BaseApp`; the public surface remains additive/internal rather than a breaking signature change.

**Necessity:** The linked Deliverable needs Base App administration for platform composite Word layouts. The hardening commit is necessary because it closes the unsafe write paths found in round 1 without expanding the feature scope.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=8908 round=2 by=alexei-dobriansky at=2026-07-03 lastSha=6d3b54a64ab61b612267a4eb6a0fd1700cc5f8fd suggestions=S1:addressed,S2:addressed,S3:addressed,S4:addressed parentRound=1

alexei-dobriansky (approved):

MortenRa (approved):

PR #8911, opened 2026-06-30
[Master]- "Filter by Attribute" action does not apply the filter under specific circumstances.Changes

DeepsShukla:


alexei-dobriansky:

<details>

This PR changes the Item List **Filter by Attributes** action to always use the temporary item buffer path. It also adds a regression test for combining an attribute filter with an existing Vendor No. filter.

The main Vendor No. exclusion scenario is valid and the new temp-record path should intersect the attribute result with the page filters. However, the clear path was not updated for the new design. If the user clears the rows in the filter dialog and presses OK, the page can remain on the old temporary attribute result because `RunOnTempRec` and the temp buffer are not reset. BaseApp event publisher verification was not applicable because this PR does not add or move event subscribers.
</details>

<details>

**S1** - *Empty filter dialog leaves stale results*
When `TempFilterItemAttributesBuffer.IsEmpty()` is true, `ClearAttributesFilter()` exits without clearing `TempItemFilteredFromAttributes` or setting `RunOnTempRec := false`. After a user clears the filter rows in the dialog, the page can still show the previous attribute result. Move the temp-record reset into the shared clear path and add a test for clearing through the dialog.

**S2** - *Clear Attributes removes user No. filter*
The new temp-record design no longer stores the attribute filter in `Rec."No."`, so `ClearAttributesFilter()` should not remove a user-supplied `No.` filter. Keep existing page filters when Clear Attributes is used, and only clear the attribute temp buffer.
</details>

<details>

**Risk:** The regression surface is the Item List filter flow in APAC, GB, IT, NA, RU, and W1. This is a common list page, so stale or dropped filters can make users see the wrong item set. There is no public API or event signature change.

**Necessity:** The linked Bug has clear repro steps: Attribute Filter must work together with an existing Vendor No. filter, including `<>Vendor`. The fix is needed, but the clear-filter behavior must be corrected before merge.
</details>

---
> [AI-PR-REVIEW] by=alexei-dobriansky round=1 sha=ece49e139ddf94f34edb53486d5a91e3e89c75b5 version=1 system=github pr=8911 suggestions=S1,S2

alexei-dobriansky (approved):
Please address the suggestions.

DeepsShukla:
> Keep existing page filters when Clear Attributes is used, and only clear the attribute temp buffer.

@alexei-dobriansky I checked this one currently ClearAttributeFilter- Keep existing page filters when Clear Attributes is used, and only clear the attribute filter.

v-ankitgoyal (approved):

PR #8918, opened 2026-06-30
[Master]-Chance of success % does not validate correctly when using the action type "Update" in OpportunityChanges

DeepsShukla:


mazhelez:
@DeepsShukla Please change the name of your branch to exclude the `%` sign.

alexei-dobriansky:

### What this PR does
This PR tries to make the Update Opportunity wizard reset `Chances of Success %` when the user changes `Action Type` from `Next` back to `Update`. The scenario is valid, and the chosen path is the right area: the page calls `WizardActionTypeValidate2()` when `Action Type` changes. However, the fix does not fully address the root cause because most branches copy the chance value from `TempSalesCycleStageFirst` instead of the temp record for the selected stage. For an existing opportunity, `CreateStageList()` does not populate `TempSalesCycleStageFirst`, so the `Update` branch is not guaranteed to restore the current stage chance.

### Suggestions

**S1** - *Wrong stage source for chance value*
Use the temp Sales Cycle Stage that matches the selected `Action Type` when setting `Chances of Success %`. The `Next`, `Previous`, `Skip`, `Update`, and `Jump` branches now read from `TempSalesCycleStageFirst`, so the `Update` path does not read the current stage. This means the PR can still save the wrong chance value.

**S2** - *Random test values can hide regression*
Set fixed, different chance values in the new regression test. The helper uses random values for stage 1 and stage 2, so they can be equal and then the test no longer proves that `Update` changed the value back.

**Risk:** The change is narrow to the Update Opportunity wizard and does not change public APIs or event signatures. The regression surface is CRM opportunity updates: an incorrect chance value changes probability and calculated current value for an opportunity.

**Necessity:** The change is required because users can save the wrong `Chances of Success %` after switching from `Next` to `Update`. The scope is right, but the assignment must use the selected stage record before this can merge.

---
> [AI-PR-REVIEW] version=1 system=github pr=8918 round=1 by=alexei-dobriansky at=2026-06-30T16:00:35Z lastSha=b9ba2a4b92ca36ff79191cf938971d9222426d91 suggestions=S1,S2

alexei-dobriansky (changes requested):
please address the suggestions

DeepsShukla:
> please address the suggestions

Changes done

PR #8919, opened 2026-06-30
 [Main] Issue with nonspecific reservation with item tracking - the reservation engine creates invalid reservation entries when reshuffling after a warehouse pick is registered

sanjmaurya:
[Bug 640407](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/640407): [master] [ALL-E] Issue with nonspecific reservation with item tracking - the reservation engine creates invalid reservation entries when reshuffling after a warehouse pick is registered

Fixes [AB#640407](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/640407)

**Issue:-**
When a sales order has nonspecific reservations on a lot/expiration‑tracked (FEFO) item, registering a warehouse pick makes the reservation engine reshuffle reservations and create invalid reservation entries. This surfaces as an item‑tracking mismatch error during pick registration / shipment posting:

"Qty. to Handle (Base) in the item tracking assigned to the document line for item X is currently Y. It must be Z."

This happens when other sales orders also hold nonspecific reservations on the same lots.

**Cause:-**
In WhseActivityRegister.Codeunit.al, DeleteNonSpecificReservationEntries released reservations too aggressively:

It called ReservationManagement.DeleteReservEntries(false, ReservationEntry."Quantity (Base)"), deleting the entire nonspecific reservation entry for the picked line. This freed lot quantity that other orders still had reserved on the same lot, and even freed quantity this order itself still needs to pick from that lot.
It never considered the paired (opposite‑sign) reservation entry, so the surplus/lot bookkeeping went out of sync.
The loop exit condition was inverted: until (ReservationEntry.Next() = 0) or (QtyToPick >= 0) (should stop when the quantity to pick is consumed, i.e. <= 0).
The net effect: reservation entries no longer matched the item‑tracking that the pick/shipment expected → invalid entries and the "Qty. to Handle (Base) … must be Z" error.

**Solution:-**
Rewrote DeleteNonSpecificReservationEntries to release only the surplus on the paired lot that the current order will not pick itself, keeping the reservation for what it does pick:

Fetch the paired entry via PairedReservationEntry.Get(ReservationEntry."Entry No.", not ReservationEntry.Positive); exit if none.
Compute ReleaseQtyBase = Abs(ReservationEntry."Quantity (Base)") - CalcQtyToPickOnLotBase(WhseActivLine, PairedReservationEntry."Lot No."), capped at the remaining QtyToPick.
If the release covers the whole entry → delete both the entry and its paired entry; otherwise partially reduce both entries so the remaining lot quantity stays reserved for the other orders.
Added helper CalcQtyToPickOnLotBase, which sums Qty. to Handle (Base) of the warehouse activity Take lines for the same source on a given lot.
Fixed the loop condition to until (ReservationEntry.Next() = 0) or (QtyToPick <= 0) and passed WhseActivLine (instead of re‑reading SalesLine), with ReservationEntry now passed var


v-ankitgoyal (dismissed):

alexei-dobriansky:

<details>

This PR changes warehouse pick registration so a sales line with nonspecific reservations keeps the reservation quantity for the lots that the current pick will handle, and releases surplus reservations for other lots. It also fixes the loop stop condition and adds a FEFO warehouse pick regression test for the linked Bug 640407 scenario.

The approach matches the repro when each lot has a single reservation entry. The remaining concern is that the new calculation is per lot, but the loop processes reservation entries one by one. If one lot is represented by more than one reservation entry, the same picked quantity is counted again for each entry, so the fix can still leave reservation entries that do not match the pick item tracking.
</details>

<details>

**S1** - *Over-keeps split reservations on the same lot*
`CalcQtyToPickOnLotBase()` subtracts the full picked quantity for a lot from every reservation entry on that lot. If the lot is split across two entries, the code can keep too much reservation on that lot and still leave the pick under-reserved. Track a remaining picked quantity per lot and decrease it after each kept entry.

**S2** - *Test split reservations for one lot*
The new test posts one inventory entry per lot, so it cannot catch the split-entry case above. Add a regression test where the same lot is split across two item ledger/reservation entries and the pick uses only part of that lot. This should fail before the production fix and pass after it.
</details>

<details>

**Risk:** This is sensitive reservation-engine and item-tracking code used during warehouse pick registration and shipment posting. A wrong reshuffle can create reservation entries that no longer match warehouse item tracking, which is the same class of failure as the linked bug. The PR does not change public signatures or event publishers, but it directly deletes and modifies `Reservation Entry` records.

**Necessity:** The linked bug is a concrete Severity 2 warehouse scenario with FEFO, lot tracking, and multiple sales orders reserving the same lots. The change is needed, but it must handle normal reservation splits for the same lot before it is safe to merge.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=8919 round=1 by=alexei-dobriansky at=2026-07-03T08:25:54Z lastSha=b72c1ab922e43fe31bb348a8d0e3354e2de2bfc4 suggestions=S1,S2

alexei-dobriansky (changes requested):
Please review the suggestions

alexei-dobriansky:

# **Recommendation: Accept**

<details>

Since round 1, the PR adds a per-lot dictionary while removing nonspecific reservations. The code now calculates the picked quantity to keep for a lot once, then consumes that remaining quantity as each reservation entry for the same lot is processed.

That addresses the split-reservation issue: a lot split across two reservation entries no longer keeps the full picked quantity on each entry. The new regression test creates the same lot through two item ledger entries, registers the FEFO pick, posts the shipment, and then verifies the other orders can also pick and ship. The later merge-from-main commits were not treated as PR-authored fixes.
</details>

<details>

| S1 | Over-keeps split reservations on the same lot | Addressed | Fixed in `86d9a8c` by tracking the remaining quantity to keep per lot and decreasing it after each processed reservation entry. |
| S2 | Test split reservations for one lot | Addressed | Fixed in `86d9a8c` by adding `RegisterFEFOPickWithNonspecificReservationsOnLotSplitAcrossTwoEntries()`. |
</details>

<details>

*None - the PR-authored change since round 1 addresses the previous suggestions.*
</details>

<details>

**Risk:** The touched code still mutates `Reservation Entry` records during warehouse pick registration, so the regression surface is sensitive. The round-2 change is narrow, stays inside local procedures, and does not change public events or procedure signatures.

**Necessity:** The linked Bug 640407 is still a concrete warehouse FEFO and nonspecific-reservation scenario. Without this change, pick registration can leave reservation entries that do not match item tracking and can block shipment posting.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=8919 round=2 by=alexei-dobriansky at=2026-07-06T08:00:40Z lastSha=070520e76e9a86ba1edb496cca6c69252e2f5613 suggestions=S1:addressed,S2:addressed parentRound=1

alexei-dobriansky (approved):

v-ankitgoyal (approved):

PR #8930, opened 2026-06-30
Fix misplaced single-line PR-review suggestion anchoring

JesperSchulz:
## Problem

The Copilot PR Review orchestration (`tools/Code Review/scripts/Invoke-CopilotPRReview.ps1`) can anchor a `suggestion` block to the **wrong line**, so applying the suggestion corrupts the file.

Seen on #8893: a style suggestion (prefix the call with `this.`, CodeCop AA0248) was anchored to a **comment line** (2196) instead of the statement it edits (2198). Auto-applying would overwrite the comment and leave the real call unprefixed.

## Root cause

`Resolve-SuggestionPlacement`'s single-line branch re-anchored only by **exact** whitespace-insensitive equality between the suggested line and a file line within ±8 of the model anchor. But a single-line suggestion almost always *edits* a line, so it is by definition **not equal** to the line it replaces (`this.` was inserted). No match was ever found, and the code fell through to *"trust the anchor"* — returning the model's unreliable line number, which pointed at a nearby comment.

## Fix

- Re-anchor single-line suggestions by **character-LCS similarity** over a ±8 window. Pick the clear winner using an absolute confidence floor plus an ambiguity margin over the runner-up.
- When no candidate is confident/unambiguous, return `$null` so the caller **suppresses** the suggestion (posts a manual snippet) instead of trusting a wrong anchor — i.e. never produce an auto-applicable misplaced suggestion.
- Multi-line (additive-span) behaviour is unchanged.

Fixes [AB#640948](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/640948)

### DIFF
--- a/tools/Code Review/scripts/Invoke-CopilotPRReview.ps1
+++ b/tools/Code Review/scripts/Invoke-CopilotPRReview.ps1
@@ -556,17 +556,20 @@ function Resolve-SuggestionPlacement {
     # A one-line suggestion almost always *edits* a line, so it is NOT equal to
     # the line it replaces (e.g. inserting 'this.' for CodeCop AA0248). Exact
     # equality therefore fails for the common case and the model's reported
-    # anchor is unreliable (it frequently points at a neighbouring comment or
-    # blank line). We instead score every file line in a window around the
-    # anchor by similarity to the suggested line and take the clear winner.
-    # When no candidate is similar enough, or the best is ambiguous against the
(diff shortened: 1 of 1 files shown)


mazhelez (approved):

JesperSchulz:
@WaelAbuSeada will take care of this bug.

PR #8938, opened 2026-06-30
Fix SOA mailbox uniqueness and synchronized identity suffix generation

qasimikram:
## Summary
- Fix mailbox uniqueness so the same email can be used across agents when folder differs
- Enforce duplicate prevention for same email + folder combination and include conflicting agent details in error
- Rework default identity generation so agent name and initials always use the same suffix index
- Extend identity suffix search range to 99 while preserving gap handling

## Azure DevOps Bugs
- [AB#640649](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/640649)
- [AB#640653](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/640653)

## Validation
- AL diagnostics checked for updated setup files

### DIFF
--- a/src/Apps/W1/SalesOrderAgent/app/src/Setup/SOASetup.Codeunit.al
+++ b/src/Apps/W1/SalesOrderAgent/app/src/Setup/SOASetup.Codeunit.al
@@ -547,7 +547,8 @@ codeunit 4400 "SOA Setup"
     internal procedure CheckMailboxUnique(SOASetup: Record "SOA Setup")
     var
         OtherSOASetup: Record "SOA Setup";
-        MailboxAlreadyUsedErr: Label 'This email account is already used by another Sales Order Agent instance.';
+        MailboxAlreadyUsedWithoutFolderErr: Label 'This email account is already used by Sales Order Agent "%1" (%2).', Comment = '%1 = agent name, %2 = agent initials';
+        MailboxAndFolderAlreadyUsedErr: Label 'This email account and folder combination is already used by Sales Order Agent "%1" (%2).', Comment = '%1 = agent name, %2 = agent initials';
     begin
         if IsNullGuid(SOASetup."Email Account ID") then
             exit;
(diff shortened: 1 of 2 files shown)


PredragMaricic (approved):

src/Apps/W1/SalesOrderAgent/app/src/Setup/SOASetup.Codeunit.al
```diff
@@ -547,16 +547,24 @@ codeunit 4400 "SOA Setup"
...
         if IsNullGuid(SOASetup."Email Account ID") then
             exit;
 
         OtherSOASetup.SetRange("Email Account ID", SOASetup."Email Account ID");
         OtherSOASetup.SetRange("Email Connector", SOASetup."Email Connector");
         OtherSOASetup.SetFilter(ID, '<>%1', SOASetup.ID);
-        if not OtherSOASetup.IsEmpty() then
-            Error(MailboxAlreadyUsedErr);
+
+        if SOASetup."Email Folder Id" = '' then begin
```
PredragMaricic, line 558: what if SOASetup."Email Folder Id" = '' and no other SOA Setup is empty?

PredragMaricic: Is actually "Email Folder Id" is good for making part of "unique key". NBot sure, but could it be that different "Email Folder Ids" have same folder name ?

qasimikram: Email folder ids are unique, and that is what we are trying to achieve here i.e. if the combination of Email + Folder is used by some other agent we throw an error. 

"Email Folder Id" would only be empty in version 28.2 and older, with new implementation in 28.3 we always set folder id, if user doesn't set anything then we default to Inbox. So the empty logic is for backward compatibility.

AleksandricMarko (approved):

PR #8941, opened 2026-06-30
Fix agent setup field lookup casing

martinsrui-msft:
Fixes the agent setup page validation to locate the required "User Security ID" field case-insensitively.

[AB#640664](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/640664)

### DIFF
--- a/src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
+++ b/src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
@@ -533,7 +533,7 @@ codeunit 4301 "Agent Impl."
         SetupPageRecordRef.Open(PageMetadata.SourceTable, PageMetadata.SourceTableTemporary);
 
         FieldMetadata.SetRange(TableNo, PageMetadata.SourceTable);
-        FieldMetadata.SetFilter(FieldName, '@''%1''', UserSecurityIdTok);
+        FieldMetadata.SetFilter(FieldName, StrSubstNo('@%1', UserSecurityIdTok));
         if not FieldMetadata.FindFirst() then
             Error(SetupPageSourceTableMissingFieldErr, SetupPageId, UserSecurityIdTok);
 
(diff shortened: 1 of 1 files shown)


src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
```diff
@@ -533,7 +533,7 @@ codeunit 4301 "Agent Impl."
         SetupPageRecordRef.Open(PageMetadata.SourceTable, PageMetadata.SourceTableTemporary);
 
         FieldMetadata.SetRange(TableNo, PageMetadata.SourceTable);
-        FieldMetadata.SetRange(FieldName, UserSecurityIdTok);
+        FieldMetadata.SetFilter(FieldName, '@' + UserSecurityIdTok);
```
nikolakukrika: You may need to escape the UserSecurityIdToken. Set range is escaping the values. SetFilter is not. 
There is a syntax with adding '' but you would need to test this.

nikolakukrika (dismissed):

src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
martinsrui-msft: I tested this with a demo app, and the syntax that I could get working was this one below, so I went with that:

FieldMetadata.SetFilter(FieldName, StrSubstNo('@%1', UserSecurityIdTok));

martinsrui-msft: | Filter | Result |
|--------|--------|
| `User Security ID` | not found (case-sensitive) |
| `'User Security ID'` | not found (quotes = case-sensitive) |
| `@User Security ID` | ✅ FOUND (`@` = case-insensitive) |
| `@'User Security ID'` | ❌ not found (`@` breaks quotes) |
| `@'User (Sec) Id'` | ❌ ERROR: `Did not expect '('` |

nikolakukrika (approved):

krupybalu (approved):

PR #8942, opened 2026-06-30
[master] Trail Balance by Period doesn't show the balance for some of the G/L Accounts in the PDF version

neeleshsinghal:
Workitem: [Bug 640991](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/640991): [master] [All-E]Trail Balance by Period doesn't show the balance for some of the G/L Accounts in the PDF version

Fixes [AB#640991](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/640991)

**Issue:** In report 38 "Trial Balance by Period", G/L Account amounts with ~10+ digits render as blank cells in the PDF for negative values (e.g., -1,234,567,890.00) while positive values display.

**Cause:** The 13 ColumnValuesAsText_N_ value textboxes in TrialBalancebyPeriod.rdlc were missing the explicit <CanGrow>true</CanGrow> setting, so the cell could not grow vertically when the extra minus-sign character pushed the right-aligned text past the 48.44pt column width, causing the PDF renderer to drop the value.

**Solution**: Added <CanGrow>true</CanGrow> to each of the 13 ColumnValuesAsText_N_ textboxes in TrialBalancebyPeriod.rdlc, matching the pattern already used in TrialBalance.rdlc.


v-ankitgoyal (approved):

alexei-dobriansky:

<details>

This PR updates the RDLC layout for report 38, "Trial Balance by Period", so the 13 period amount textboxes can grow when a large negative value does not fit in the narrow period column. The linked bug is a PDF layout issue, not a dataset issue: the report already calculates `ColumnValuesAsText[1..13]` from G/L net change, including the opening period and End-Total rows. The change covers every period value cell that uses those dataset fields, so it addresses the described blank negative amount without changing date filters, CalcFields, or balance logic.
</details>

<details>

**S1** - *Add focused PDF validation evidence*
Please add a short validation note, or a test if the PDF output can be checked, for a 10+ digit negative amount in this report. The existing report layout smoke test only saves the report as PDF and does not prove that the large negative value is visible.

**S2** - *Check similar period balance layouts*
Please check the same large negative amount in the similar G/L period reports, especially `FiscalYearBalance.rdlc`, which also has `ColumnValuesAsText_*` cells without `CanGrow`. If the same PDF issue appears there, it should be fixed in the same bug or tracked as a follow-up.
</details>

<details>

**Risk:** This is a financial report, so hidden negative balances in PDF output are high impact. The code risk is limited because the change is RDLC-only, uses the same pattern as other amount textboxes, and does not change report data, public APIs, events, or BaseApp publishers. No `[AI-REPRO]` comment was present, but the work item repro steps are concrete.

**Necessity:** The change is needed because the PDF can show a blank cell instead of a real negative G/L balance, and users may rely on the PDF as a financial report. The scope is targeted: all 13 period amount cells are changed, which covers opening balance, later periods, Posting rows, and total rows that use the same fields.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=8942 round=1 by=alexei-dobriansky at=2026-07-03 lastSha=1fee8f73a375ae6714897bd31d5e1dc43b24f40d suggestions=S1,S2

AndersLarsenMicrosoft (approved):

dcenic (approved):

PR #8944, opened 2026-06-30
Bug 637264: Reset posting codeunits transaction state after Commit in cost adjustment

alexei-dobriansky:
Fixes [AB#637264](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/637264)

### Problem
During item-by-item cost adjustment, an explicit `Commit()` releases the G/L Entry table lock while `NextEntryNo` is still cached, allowing a concurrent session to read the same last entry number and hit a duplicate-key error.

### Fix
- `Gen. Jnl.-Post Line`: new `ResetTransactionState` that clears the transaction-scoped caches so the next post re-takes the G/L Entry lock and re-reads the last entry number.
- `Inventory Posting To G/L` / `Item Jnl.-Post Line`: `ResetGLPostingState` plumbing to forward the reset.
- `Inventory Adjustment.CheckAndCommit`: calls the reset after each `Commit()` and emits feature telemetry (event ID `0000MEQ`).
- Parallel-run test coverage.

### Localization
The W1 change is propagated to all affected country layers (APAC, BE, CH, ES, FI, FR, IT, NA, NO, RU) via miapp.


attilatoury (approved):

src/Layers/APAC/BaseApp/Inventory/Costing/InventoryAdjustment.Codeunit.al
```diff
@@ -2971,13 +2976,30 @@ codeunit 5895 "Inventory Adjustment" implements "Inventory Adjustment", "Cost Ad
...
-                Commit();
+        ItemsSinceLastCommit += 1;
+        if ItemsBeingAdjusted.Count() > 0 then
+            exit;
 
-            ItemApplicationTrace.Reset();
-            ItemApplicationTrace.DeleteAll();
+        if CommitAdjustedItems then begin
+            Commit();
+            ItemJnlPostLine.ResetGLPostingState();
```
ChethanT, line 2985: Shouldn't this be before Commit() to ensure other transaction do not read state between Commit() and ResetGLPostingState call?

alexei-dobriansky: I don't see the difference. The problem was that I commit but still keep in memory the next G/L entry number I'll use. Another session comes into the G/L that's now unlocked and gets the same entry number. With the ResetGLPostingState I throw away the next number so that I will get a new one later, when my time to post comes.

ChethanT (approved):

PR #8950, opened 2026-07-01
[Main] [all-e]Unbalanced G/L Entries if you apply a Payment and a Posted Purchase Invoice having both different Posting Groups and using the "Applies-to Document No." on the line instead of using Apply Entries in the Spanish version.

sanjmaurya:
I have create new PR


v-ankitgoyal (dismissed):

alexei-dobriansky:

<details>

This PR narrows the ES-specific vendor application account override when a payment is applied through `Applies-to Doc. No.`. The new check uses the existing vendor ledger entry to tell the payment-side detailed entry from the applied invoice-side entry, so the invoice side can use its own vendor posting group.

The change matches the root cause: the detailed buffer `Document Type` alone cannot identify which side of the application is being posted. If the payment ledger entry is not inserted yet, the code still keeps the payment posting group path. The added ES test covers the exact inline `Applies-to Doc. No.` scenario with different posting groups and asserts both payables accounts net to zero.
</details>

<details>

**S1** - *PR text overstates unapply coverage*
The PR text says the change is mirrored in `PostDtldVendLedgEntryUnapply`, but the diff only changes `PostDtldVendLedgEntry`. Please either update the description, or add the same guard and a test if unapply must be covered.
</details>

<details>

**Risk:** This is high-sensitivity posting code in the ES BaseApp vendor application path. The code change is narrow, adds no public API or event signature change, and does not depend on a new BaseApp publisher. W1 and other country BaseApp layers use the existing vendor ledger entry posting group path directly and do not have this ES Cartera special case, so the ES-only scope is appropriate.

**Necessity:** The ADO bug is clear and important: applying a posted purchase invoice through `Applies-to Doc. No.` can leave payables accounts unbalanced when the payment uses a different vendor posting group. The new regression test makes the sensitive path repeatable and checks the resulting G/L balance. Without the change, users must avoid the inline apply path and use Apply Entries instead.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=8950 round=1 by=alexei-dobriansky at=2026-07-03 lastSha=4f2da8f970e29a66f636ee13dc45fc74dc12493c suggestions=S1

AndersLarsenMicrosoft (dismissed):

dcenic (dismissed):

alexei-dobriansky:

# **Recommendation: Accept**

<details>

The new commit adds the same `IsApplnEntryForAppliedVendorDoc(...)` guard to `PostDtldVendLedgEntryUnapply` that the apply path already uses. This removes the mismatch called out in round 1 between the PR description and the unapply code.

The added code is narrow and matches the apply-side account selection shape: the payment-side application entry can still use the payment posting group's payables account, while an applied document entry can fall back to `GetVendDtldCVLedgEntryBufferAccNo(...)` and use its own posting group. In the normal unapply entry point, the `Gen. Journal Line` is rebuilt for unapply and does not carry `Applies-to Doc. No.`, so this change does not broaden standard unapply behavior.
</details>

<details>

| S1 | PR text overstates unapply coverage | Addressed | No written reply; commit `3c41da9c6d` adds the unapply guard. |
</details>

<details>

*None - the only author-owned change since round 1 is the unapply guard, and I did not find a new issue in that change.*
</details>

<details>

**Risk:** This remains sensitive ES vendor posting code because it controls payables account G/L entries during apply and unapply with multiple vendor posting groups. The round-2 code change is only three added lines in the ES `Gen. Jnl.-Post Line` unapply path, adds no public API or event signature change, and does not affect W1 or other country layers. There is still no direct unapply regression test, but I am not treating that as blocking because the change is narrow and the normal unapply setup does not use the inline `Applies-to Doc. No.` field.

**Necessity:** The original bug is valid and important: users can get unbalanced payables accounts when applying a payment to a posted purchase invoice through the inline `Applies-to Doc. No.` path with different vendor posting groups. The round-2 change is justified because it keeps the ES apply and unapply account-selection logic aligned with the PR description and avoids leaving the unapply path as a future divergence point.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=8950 round=2 by=alexei-dobriansky at=2026-07-10T11:04:05.441Z lastSha=3c41da9c6da06abef4028dbc9ec4d8d8cc98c0a3 suggestions=S1:addressed parentRound=1

aholstrup1:
Heads up, these build failures aren't from your code changes. The special character (a stray `]`) in the branch name trips up our pipeline. Recreating the branch without the bracket and repointing the PR should fix it. Sticking to letters, numbers, `/`, `-`, and `_` in branch names avoids this going forward.

PR #8951, opened 2026-07-01
[Master]-Time zone issue to the G/L report in Business Central

neeleshsinghal:
Work Item: [Bug 641014](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/641014): [ALL-E]-[Master]-Time zone issue to the G/L report in Business Central

Fixes [AB#641014](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/641014)

ISSUE:- Time zone issue to the G/L report in Business Central

CAUSE:- The Type Helper codeunit does not account for Daylight Saving Time (DST) in the GetUserTimezoneOffset function.

SOLUTION:- Correct the function


v-ankitgoyal (approved):

WaelAbuSeada (approved):
LGTM

darjoo (approved):

dominicstarkl:
Hi @neeleshsinghal, @v-ankitgoyal, @darjoo 

We have confirmed a downstream regression caused by the change in this PR.

The automated review had already identified the same inconsistency before the merge: `GetUserTimezoneOffset()` now returns a DST-aware offset, while `ConvertDateTimeFromUTCToTimeZone()` and `ConvertDateTimeFromInputTimeZoneToClientTimezone()` still combine that value with the DST-unaware `TimeZoneInfo.BaseUtcOffset`.

As a result, **approval workflow notifications can be scheduled one hour too early or too late while daylight-saving time is active, depending on the conversion direction used**.

Example using `ConvertDateTimeFromInputTimeZoneToClientTimezone()` for Switzerland on 24 September 2026:

Assume that both the input time zone and the client time zone are set to Switzerland. Therefore, converting the scheduled time between these time zones should not change the local time.

- Scheduled local time: 07:00 CEST (UTC+02:00)
- DST-aware client offset returned by `GetUserTimezoneOffset()`: +02:00
- DST-unaware input time-zone offset returned by `TimeZoneInfo.BaseUtcOffset`: +01:00
- Current calculation: `07:00 + 02:00 - 01:00 = 08:00`
- Expected result: **07:00**
- Actual result: **08:00**

The one-hour difference is introduced because a DST-aware client offset is combined with a DST-unaware input time-zone offset.

Affected call path:

[`Notification Schedule.CalculateExecutionTime()`](https://github.com/microsoft/BCApps/blob/main/src/Layers/W1/BaseApp/System/Notifications/NotificationSchedule.Table.al#L278)  
→ `CalculateClientTimeFromRecTimezone()`  
→ `TypeHelper.ConvertDateTimeFromInputTimeZoneToClientTimezone()`  
→ `GetUserClientTypeOffset()`  
→ `GetUserTimezoneOffset()`

Could you please advise whether you would prefer a new issue and corrective PR, or whether this should be tracked as a follow-up to the existing work item?

Thanks in advance for taking a look.

darjoo:
@dominicstarkl please open a new issue and corrective PR. I'll approve the issue.

dominicstarkl:
Thanks @darjoo - I'll create a new PR.
Do you know when the fix will be available then?
Who can I assign as the second approver?

darjoo:
@dominicstarkl Lets get it in to master and 29.x I'll try to get it into 29.0 before that's released.
I'll get the second approver once we have the PR

PR #8967, opened 2026-07-01
Ability to archive agents

krupybalu:
## What & why

Adds admin **archiving of agents** (soft delete) from the Agent List and Agent Card. Agents created via the SDK can't be deleted, so disabled or broken ones accumulate and clutter the role center and agent/task views — partners asked for a way to remove them.

Archiving an agent:
- sets `Substate = Archived` and disables it, so it no longer appears in the role center or agent views, while the agent and its existing tasks and logs stay in the database as read-only for auditing;
- is gated by a confirmation page where the admin types the agent's display name to confirm, and shows a notification on an archived agent's card;
- logs usage telemetry.

Exposed via new additive `Agent.Archive` / `Agent.IsArchived` methods over the platform `Agent.Substate` field. Archiving is intentionally **one-way** (no restore) in this iteration; the confirmation dialog and tooltips state this.

## Linked work

[AB#620767](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/620767)

## How I validated this

## Risk & compatibility

- **Platform dependency:** requires the platform `Agent.Substate` field (`Substate::Archived`) introduced by the same slice ([AB#620767](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/620767)). The app won't build or run against an older platform that lacks it.
- **Irreversible, no data loss:** archiving is a soft delete — there is no restore and nothing is deleted. Existing agents default to non-archived and are unaffected; an archived agent becomes disabled, read-only and hidden, with its tasks and logs retained for auditing.
- **Additive only:** no changes to existing Agent API signatures, no new interface, and no upgrade or data-migration step.
- **Telemetry:** the Agent app now takes a dependency on the System Application **Telemetry** module (added to `app.json`) to log archive usage.
- **Deferred (not in scope):** agent restore and partner-controlled block/allow events — these can follow if partners ask for them.


src/System Application/App/Agent/Interaction/Internal/AgentTaskList.Page.al
```diff
@@ -97,6 +97,13 @@ page 4300 "Agent Task List"
                         TaskPane.ShowAgent(Rec."Agent User Security ID");
                     end;
                 }
+                field(AgentSubstate; Rec."Agent Substate")
+                {
+                    Caption = 'Agent Substate';
+                    ToolTip = 'Specifies whether the agent that owns the task is archived.';
```
krupybalu: Maybe something like "Specifies the substate of the agent' would be better so if we introduce new substates in the future, we dont need to change it?

src/System Application/App/Agent/Setup/AgentList.Page.al
```diff
@@ -47,6 +47,12 @@ page 4316 "Agent List"
                 {
                     Caption = 'State';
                 }
+                field(Substate; Rec.Substate)
+                {
+                    Caption = 'Substate';
+                    ToolTip = 'Specifies whether the agent is archived.';
```
krupybalu: Maybe something like "Specifies the substate of the agent' would be better so if we introduce new substates in the future, we dont need to change it?

src/System Application/App/Agent/Setup/AccessControls/ViewAgentPermissions.Page.al
```diff
@@ -80,6 +80,7 @@ page 4334 "View Agent Permissions"
                 Caption = 'Edit agent permissions';
                 Image = Edit;
                 ToolTip = 'Edit the permission sets assigned to this agent.';
+                Enabled = not AgentIsArchived;
```
qutreson, line 83: There's always some discussion about which pattern to follow for this. Using enabled property disables the action but the user does not necessarily know why they cannot click that action. Raising an error allows to provide that contains. Maybe we could have a notification raised on the page (on the part if that works, or on the parent page) to tell the user they are viewing an archived agent?

src/System Application/App/Agent/Setup/AgentArchiveConfirmation.Page.al
```diff
@@ -0,0 +1,88 @@
...
+        }
+    }
+
+    var
+        ExpectedDisplayName: Text[80];
+        EnteredDisplayName: Text[80];
+        ConfirmationPrompt: Text;
+        Confirmed: Boolean;
+        NameMismatchErr: Label 'The name you entered does not match the agent''s display name.';
+        ConfirmationInstructionLbl: Label 'You are about to archive the agent "%1". This removes the agent from active use and cannot be undone.\\To confirm, type the agent''s display name below.', Comment = '%1 = agent display name';
```
qutreson: Was this prompt checked with UX? I'm unsure about the term 'remove' it since it can lead to think it is removed completely. It might be good to state the impact in more clear terms:
- the agent tasks and logs remain visible for auditing
- but the agent cannot be used anymore to process new tasks
- the agent cannot be restored

src/System Application/App/Agent/Setup/AgentList.Page.al
qutreson: I would keep it more generic and I would rely on the tooltip defined on the platform table level instead of overriding it here

src/System Application/App/Agent/Setup/AgentList.Page.al
```diff
@@ -167,6 +199,9 @@ page 4316 "Agent List"
                 actionref(AgentSetup_Promoted; AgentSetup)
                 {
                 }
+                actionref(ArchiveAgent_Promoted; ArchiveAgent)
```
qutreson: Do we really want this to be promoted that way? Archiving an agent is a very rare operation so I'd wonder whether it needs to be the second action the user sees. Keeping it hidden is probably safer since it's a rare operation.

src/System Application/App/Agent/Setup/AgentList.Page.al
```diff
@@ -78,6 +84,32 @@ page 4316 "Agent List"
                     CurrPage.Update(false);
                 end;
             }
+            action(ArchiveAgent)
+            {
+                ApplicationArea = All;
+                Caption = 'Archive';
+                ToolTip = 'Archive the agent. Archiving removes the agent from active use and cannot be undone.';
```
qutreson: Same comment as elsewhere about the 'remove' term.

src/System Application/App/Agent/Setup/AgentList.Page.al
```diff
@@ -78,6 +84,32 @@ page 4316 "Agent List"
...
+
+                trigger OnAction()
+                var
+                    Agent: Codeunit Agent;
+                    ArchiveConfirmation: Page "Agent Archive Confirmation";
+                begin
+                    if Rec.IsEmpty() then
+                        Error(NoAgentSetupErr);
+
+                    Rec.TestField("Display Name");
```
qutreson, line 105: Is the TestField call needed?

src/System Application/App/Agent/Setup/AccessControls/ViewAgentPermissions.Page.al
krupybalu: added a page-level notification banner on the Agent Card when archived: "This agent is archived and can no longer be modified.  Its tasks and logs remain available for reference."

src/System Application/App/Agent/Setup/AgentList.Page.al
krupybalu: Done

src/System Application/App/Agent/Setup/AgentList.Page.al
krupybalu: Good point. Removed from promoted

src/System Application/App/Agent/Setup/AgentList.Page.al
krupybalu: Display Name isn't mandatory (the SDK's Agent.Create/SetDisplayName and the platform both allow it to be blank), and a blank name makes the type-to-confirm archive dialog impossible to confirm

src/System Application/App/Agent/Setup/AgentArchiveConfirmation.Page.al
```diff
@@ -0,0 +1,88 @@
...
+// ------------------------------------------------------------------------------------------------
+
+namespace System.Agents;
+
+page 4337 "Agent Archive Confirmation"
+{
+    Extensible = false;
+    InherentEntitlements = X;
+    InherentPermissions = X;
+    PageType = StandardDialog;
```
nikolakukrika, line 13: Minor - we have confimation dialog page type, I believe this one would be better for gathering confirmations

src/System Application/App/Agent/Setup/Agent.Codeunit.al
```diff
@@ -57,6 +57,31 @@ codeunit 4321 Agent
         AgentImpl.Deactivate(AgentUserSecurityID);
     end;
 
+    /// <summary>
+    /// Archives the agent. Archiving removes the agent from active use and cannot be undone. The agent must be inactive (deactivated) before it can be archived.
+    /// </summary>
+    /// <param name="AgentUserSecurityID">The user security ID of the agent.</param>
+    procedure Archive(AgentUserSecurityID: Guid)
```
nikolakukrika, line 64: Minor - Instead of adding to this codeunit more methods, we could create an Archive codeunit and Archive Impl codeunit. These methods will not be used much by developers, the standard implementation will most likely cover all the need.

src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
```diff
@@ -56,6 +56,38 @@ codeunit 4301 "Agent Impl."
...
+
+        if Agent.Substate = Agent.Substate::Archived then
+            exit; // Archiving is terminal; idempotent no-op avoids the platform "archived agent cannot be modified" error on re-archive.
+
+        if Agent.State <> Agent.State::Disabled then
+            Error(DeactivateBeforeArchivingErr);
+
+        Agent.Substate := Agent.Substate::Archived;
+        Agent.Modify(true);
+    end;
```
nikolakukrika, line 74: I would recommend commiting here

nikolakukrika: If you run any code in the platform for the change commit will ensure that the state is matching the platform state. It is also an important operation that should not be rolled back.

src/System Application/App/Agent/Setup/Agent.Codeunit.al
```diff
@@ -57,6 +57,31 @@ codeunit 4321 Agent
...
+        FeatureAccessManagement.AgentManagementAllowed(true);
+        AgentImpl.Archive(AgentUserSecurityID);
+    end;
+
+    /// <summary>
+    /// Checks if the agent is archived.
+    /// </summary>
+    /// <param name="AgentUserSecurityID">The user security ID of the agent.</param>
+    /// <returns>True if the agent is archived; otherwise false.</returns>
+    procedure IsArchived(AgentUserSecurityID: Guid): Boolean
```
nikolakukrika, line 77: How did we protect against giving tasks to the archived agents or doing any other operations?
There is some code that requires the agent to be reactivated, we have confirmation dialogs to make agent active again. Do we need any additional code to ensure that unsupported actions should not trigger reactivate agent code / should we block the code that is allowed on deactivated agents?

src/System Application/Test/Agent/src/SDK/AgentTest.Codeunit.al
```diff
@@ -965,4 +967,549 @@ codeunit 133961 "Agent Test"
...
+        Assert.IsTrue(Agent.IsActive(AgentId), 'Agent should be active initially');
+
+        // [WHEN] Archiving the active agent
+        // [THEN] An error is raised asking to deactivate first
+        asserterror Agent.Archive(AgentId);
+        Assert.ExpectedError('Deactivate the agent before archiving it.');
+    end;
+
+    [Test]
+    procedure IsArchivedFalseForNewAgent()
```
nikolakukrika, line 1034: Can we add a test it is not possible to activate the agent again?
Second 2 tests should be 
It is not possible to give it tasks nor to give additional messages to the agent task

nikolakukrika (approved):

nikolakukrika (dismissed):

thloke (commented):
In the platform changes, task modifications for archived agents will throw. I don't see an equivalent change here to disable any actions on tasks for archived agents.

src/System Application/App/Agent/Interaction/Internal/AgentTaskList.Page.al
```diff
@@ -211,6 +217,34 @@ page 4300 "Agent Task List"
...
+                begin
+                    ShouldShowAllAgents := true;
+                    SetAgentSubstateFilter();
+                end;
+            }
+            action(ShowActiveAgents)
+            {
+                ApplicationArea = All;
+                Caption = 'Show active agents';
+                ToolTip = 'Show only tasks from active agents that have not been archived.';
```
darjoo: Has the text gone through review?

`... that have not been archived` sounds redundant.

krupybalu: Not yet

src/System Application/App/Agent/Setup/AgentList.Page.al
```diff
@@ -47,6 +47,11 @@ page 4316 "Agent List"
                 {
                     Caption = 'State';
                 }
+                field(Substate; Rec.Substate)
+                {
+                    Caption = 'Substate';
+                    Editable = false;
+                }
```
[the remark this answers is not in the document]
krupybalu: The page field inherits the ToolTip defined on the platform table field (Agent.Substate / Agent Task "Agent Substate"), so AA0218 is satisfied without a page-level override. This also matches the earlier resolved decision to rely on the table-level tooltip rather than overriding it on the page.

src/System Application/App/Agent/Interaction/Internal/AgentTaskList.Page.al
```diff
@@ -97,6 +97,12 @@ page 4300 "Agent Task List"
                         TaskPane.ShowAgent(Rec."Agent User Security ID");
                     end;
                 }
+                field(AgentSubstate; Rec."Agent Substate")
+                {
+                    Caption = 'Agent Substate';
+                    Editable = false;
+                    Visible = ShouldShowAllAgents;
+                }
```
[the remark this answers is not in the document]
krupybalu: The page field inherits the ToolTip defined on the platform table field (Agent.Substate / Agent Task "Agent Substate"), so AA0218 is satisfied without a page-level override. This also matches the earlier resolved decision to rely on the table-level tooltip rather than overriding it on the page.

src/System Application/App/Agent/Setup/Agent.Codeunit.al
krupybalu: Archived agents are protected at the platform level: the Agent virtual-table provider rejects any modification once Substate = Archived (so an archived agent cannot be reactivated), and the Agent Task provider blocks creating new tasks for it (see  AgentRecordService.ThrowIfAgentArchived(agentDataRecord, this.MetaTable.TableCaptionSafe); ).

src/System Application/Test/Agent/src/SDK/AgentTest.Codeunit.al
krupybalu: added

src/System Application/App/Agent/Setup/AgentArchiveConfirmation.Page.al
krupybalu: The dialog also requires the user to type in the Agent's name, I think an OK/Cancel is better for this purpose.

src/System Application/App/Agent/Setup/Agent.Codeunit.al
krupybalu: I think these two small methods should stay, similarly to Activate/Deactivate they are related to agent lifecycle.

src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
krupybalu: doesnt adding a commit here mean that we also flush any outside caller's pending transaction? We dont have a similar commit for the activate/deactivate.

src/System Application/App/Agent/Setup/AgentList.Page.al
```diff
@@ -78,6 +83,32 @@ page 4316 "Agent List"
                     CurrPage.Update(false);
                 end;
             }
+            action(ArchiveAgent)
+            {
+                ApplicationArea = All;
+                Caption = 'Archive';
+                ToolTip = 'Archive the agent so it can no longer process new tasks. The agent and its existing tasks and logs remain available as read-only. Archiving cannot be undone.';
+                Image = Archive;
+                Enabled = ArchiveActionEnabled;
```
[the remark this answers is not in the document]
krupybalu: Fixed in 339f8bdcf1. Added `and Rec."Can Curr. User Configure Agent"` to `ArchiveActionEnabled` on both the Agent List and Agent Card pages, matching the guard the sibling Configure and Agent User Settings actions already enforce.

I kept this at parity with those sibling actions (client-side enabling), rather than adding a separate server-side check in `AgentImpl.Archive`, since that mirrors the existing pattern for configure-gated actions and the platform additionally freezes the agent record on archive. Happy to add a server-side guard too if we want defense-in-depth here.

src/System Application/Test/Agent/src/SDK/AgentTest.Codeunit.al
```diff
@@ -965,4 +967,587 @@ codeunit 133961 "Agent Test"
...
+
+        // [WHEN] Invoking the Archive action and cancelling the dialog
+        InvokeArchiveActionFromList(AgentId);
+
+        // [THEN] No error is raised and the agent is not archived
+        Assert.IsFalse(Agent.IsArchived(AgentId), 'Agent should not be archived when the confirmation is cancelled');
+    end;
+
+    [ModalPageHandler]
+    procedure AgentArchiveConfirmationModalHandler(var AgentArchiveConfirmation: TestPage "Agent Archive Confirmation")
```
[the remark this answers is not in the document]
krupybalu: Reworked the archive-confirmation tests to follow the enqueue-driven UI-handler pattern from the linked guidance (b7243e6821):

- The tests now `Enqueue` the `(display name, OK/Cancel)` inputs via `Library - Variable Storage`; `AgentArchiveConfirmationModalHandler` `Dequeue`s them (text then boolean) instead of reading codeunit globals.
- Each test ends with `LibraryVariableStorage.AssertEmpty()` to prove the confirmation modal fired exactly once.
- `Initialize()` now calls `LibraryVariableStorage.Clear()` so no queued value can leak across tests.
- `[HandlerFunctions]` lists exactly the one modal handler each scenario triggers.

One deviation worth noting: the guidance's `Assert.ExpectedConfirm`/`ExpectedMessage`/`ExpectedStrMenu` caption asserts are for `[ConfirmHandler]`/`[MessageHandler]`/`[StrMenuHandler]`, which receive a caption string. This is a `[ModalPageHandler]` whose dialog identity is enforced by the strongly-typed `TestPage "Agent Archive Confirmation"` (any other modal aborts with unhandled-UI). The dialog's validation logic is exercised by the name-mismatch/case/whitespace/empty tests, and the enqueue/dequeue/`AssertEmpty` discipline covers the "fired exactly once" guarantee.

src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
darjoo, line 424: Commit here is more important that the archive.

src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
krupybalu: I'd go with no Commit for now. Three main reasons:
1. Archive() can only be called on a deactivated agent. If anyone Reactivates, the Agent.Modify would fail.  
2. The Archive() method is public (via the facade) and if a PTE calls it, I'd not expect a commit to happen right in the middle of my method.
3. We call the archive in two places internally, both are OnAction scopes. At the end of the top level scope (which is the OnAction itself in these cases) the transaction is committed.

src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
krupybalu: I lean towards not adding an explicit commit, see my response to Nikola's comment.

src/System Application/App/Agent/Setup/AgentArchiveConfirmation.Page.al
krupybalu: Not yet, everyone is on vacation 😅

krupybalu: But I revised the msg, hopefully it is better now

src/System Application/App/Agent/Interaction/Internal/AgentTaskList.Page.al
krupybalu: I have revised the user facing messages

src/System Application/App/Agent/Setup/AgentImpl.Codeunit.al
```diff
@@ -56,6 +56,38 @@ codeunit 4301 "Agent Impl."
...
+    var
+        Agent: Record Agent;
+    begin
+        GetAgent(Agent, AgentUserSecurityID);
+
+        exit(Agent.Substate = Agent.Substate::Archived);
+    end;
+
+    // Agent-record changes are frozen by the platform VDP; this guards the other tables.
+    local procedure EnsureNotArchived(AgentUserSecurityID: Guid)
```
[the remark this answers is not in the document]
krupybalu: By design, not a gap. Agent Access Control is authorization metadata (which users may access/manage the agent), not agent business data. This feature freezes agent *data* (the agent record, instructions, tasks, logs) and agent *configuration* (user settings via `UpdateAgentUserSettings`, permission sets via `AssignPermissionSets`), but intentionally leaves the authorization layer administrable so admins can still grant or revoke who may access an archived agent (e.g. offboarding or security response).

That's why `UpdateAgentAccessControl` and the `EditUserAccess` action on View Agent Access Control (page 4320) are not gated, and it's covered by `UpdateAgentAccessControlOnArchivedAgentSucceeds`. `AssignPermissionSets` *is* gated because permission sets define what the agent can do (its capability), which is part of the frozen configuration. The boundary is deliberate, so no guard is added here.

src/System Application/App/Agent/Setup/AgentArchiveConfirmation.Page.al
```diff
@@ -0,0 +1,76 @@
...
+    begin
+        exit(Confirmed);
+    end;
+
+    trigger OnQueryClosePage(CloseAction: Action): Boolean
+    begin
+        if CloseAction <> Action::OK then
+            exit(true);
+
+        // Case-sensitive by design: in-memory AL Text "<>" is ordinal, so the user must type the display name exactly.
```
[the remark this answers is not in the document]
krupybalu: Intentional. This is a deliberate type-to-confirm friction gate (similar to Azure's "type the resource name to delete"), and the case-sensitive comparison is covered by `ArchiveConfirmationWrongCaseDoesNotArchive`. The dialog shows the exact display name to type immediately above the input, so the required casing is already visible to the user. We opted to keep the error message minimal rather than add a case-match hint.

src/System Application/Test/Agent/src/SDK/AgentTest.Codeunit.al
```diff
@@ -965,4 +967,628 @@ codeunit 133961 "Agent Test"
...
+        Agent.Archive(AgentId);
+
+        // [WHEN] Assigning permission sets on the archived agent
+        // [THEN] An error is raised that the archived agent cannot be modified
+        asserterror Agent.UpdateAccessControl(AgentId, TempAccessControlBuffer);
+        Assert.ExpectedError('The agent is archived and cannot be modified.');
+    end;
+
+    [Test]
+    procedure UpdateAgentAccessControlOnArchivedAgentSucceeds()
```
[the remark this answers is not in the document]
krupybalu: fixed

src/System Application/Test/Agent/src/SDK/AgentTest.Codeunit.al
```diff
@@ -965,4 +967,637 @@ codeunit 133961 "Agent Test"
...
+        ProfileAppID := Any.GuidValue();
+
+        // [WHEN] Setting the profile on the archived agent
+        // [THEN] An error is raised that the archived agent cannot be modified
+        asserterror Agent.SetProfile(AgentId, ProfileID, ProfileAppID);
+        Assert.ExpectedError('The agent is archived and cannot be modified.');
+    end;
+
+    [Test]
+    procedure ArchivedAgentUserSettingsActionDisabledOnCard()
```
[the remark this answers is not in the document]
krupybalu: Fixed in b80639dd9d. Both tests open the Agent Card for an archived agent, so `OnAfterGetCurrRecord` runs `SendArchivedNotificationIfNeeded()` -> `ArchivedNotification.Send()`; with no registered handler that uncaptured send raises an unhandled-UI error in headless runs before the assertions execute. Added `[HandlerFunctions('SendArchivedAgentNotificationHandler')]` to both tests plus a minimal `[SendNotificationHandler]` that captures the notification. Agent Test module compiles clean (`BUILD OK`).

qasimikram (changes requested):
Request changes:

S1 - Adding messages to archived-agent tasks is not actually blocked. `AddingMessageToArchivedAgentTaskIsBlocked` expects `AgentTaskMessageBuilder.Create()` to fail, but the current path (`AgentTaskMsgBuilderImpl.Create` -> `AgentTaskImpl.AddMessage`) inserts the message without checking the owning agent's substate. This leaves archived tasks writable and makes the new test fail when executed. Add an archived-agent guard in the message creation path, or remove/correct the scenario if messages are intended to remain writable.

S2 - The Agent Card notification state is keyed only by a page-level Boolean. When a user navigates directly from one archived agent to another, `ArchivedNotificationShown` remains true and the second agent's notification is suppressed. Track the last-notified agent's `User Security ID` instead.

S3 - The PR description has no linked internal work item, validation details, or risk analysis, and the `For Microsoft: Validate link to internal work items` check is failing. Please add the required traceability and concrete validation evidence for this irreversible state transition.

The first two findings are also captured in the existing unresolved inline threads.

krupybalu:
S1 - It is gated on the platform side. Instead of trying to add gate everywhere on the app side, possibly missing some places, these edits are gated on the Agent and Agent Task VDP level.

S2 - fixed

S3 - fixed

src/System Application/App/Agent/Setup/AgentCard.Page.al
```diff
@@ -242,6 +281,25 @@ page 4315 "Agent Card"
     trigger OnAfterGetCurrRecord()
     begin
         UpdateControls();
+        SendArchivedNotificationIfNeeded();
+    end;
+
+    local procedure SendArchivedNotificationIfNeeded()
```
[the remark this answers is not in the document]
krupybalu: fixed

src/System Application/App/Agent/Setup/AgentArchiveConfirmation.Page.al
```diff
@@ -0,0 +1,76 @@
...
+            }
+        }
+    }
+
+    var
+        ExpectedDisplayName: Text[80];
+        EnteredDisplayName: Text[80];
+        ConfirmationPrompt: Text;
+        Confirmed: Boolean;
+        NameMismatchErr: Label 'The name you entered does not exactly match the agent''s display name.';
```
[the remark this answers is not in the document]
krupybalu: fixed

src/System Application/Test/Agent/src/SDK/AgentTest.Codeunit.al
```diff
@@ -965,4 +967,718 @@ codeunit 133961 "Agent Test"
...
+        NameToType := LibraryVariableStorage.DequeueText();
+        CloseWithOK := LibraryVariableStorage.DequeueBoolean();
+        AgentArchiveConfirmation.DisplayNameConfirmation.SetValue(NameToType);
+        if CloseWithOK then
+            AgentArchiveConfirmation.OK().Invoke()
+        else
+            AgentArchiveConfirmation.Cancel().Invoke();
+    end;
+
+    [SendNotificationHandler]
```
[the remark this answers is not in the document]
krupybalu: fixed

src/System Application/Test/Agent/src/SDK/AgentTaskManagementTest.Codeunit.al
```diff
@@ -627,4 +627,202 @@ codeunit 133962 "Agent Task Management Test"
...
+
+        MessageFrom := CopyStr(Any.AlphanumericText(MaxStrLen(MessageFrom)), 1, MaxStrLen(MessageFrom));
+        MessageText := Any.AlphanumericText(2048);
+
+        // [WHEN] Adding a new message to the archived agent's task
+        // [THEN] The operation is rejected because the associated agent is archived
+        AgentTaskMessageBuilder.Initialize(MessageFrom, MessageText);
+        AgentTaskMessageBuilder.SetAgentTask(Task.Id);
+        asserterror AgentTaskMessageBuilder.Create();
+        Assert.ExpectedError('is not allowed because the associated agent is archived');
```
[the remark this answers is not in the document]
krupybalu: blocked by the platform

src/System Application/App/Agent/Setup/Agent.Codeunit.al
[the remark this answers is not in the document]
krupybalu: Same rationale as S1: authorization is enforced at the platform virtual-table layer, not the facade.

`Archive` ends in `Agent.Modify(true)`, which routes through the platform `Agent` virtual table. `AgentVirtualDataProvider.ModifyInternalAsync` calls `EnsureCurrentUserCanManageAgent(agentUserId)` before any state change is persisted, and it also enforces the archived-immutability invariant there. So a caller without per-agent Configure access can't archive by GUID regardless of the page-level `Can Curr. User Configure Agent` hint - the `Modify` throws. `IsArchived` reads via `Agent.Get()` on the same ACL-filtered virtual `Agent` table, so a user who can't see an agent gets "does not exist," not its substate (which is only `None`/`Archived` anyway).

This is deliberate: gate once at the VDP rather than duplicate per-agent checks across every app-side entry point, where we'd risk missing one. It's the identical pattern the pre-existing `Activate`/`Deactivate` already use - `Archive` isn't a new authorization surface, it's the same one, so I'm not adding a redundant app-layer check.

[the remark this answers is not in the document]
krupybalu: gated by the platform

qasimikram (dismissed):
The previously reported blockers are resolved. Archived-task message creation is enforced by the updated platform dependency and covered by the passing runtime test, the Agent Card notification is now keyed to the current agent, and internal work-item validation passes. The full check matrix is successful.

qasimikram (approved):

aholstrup1 (approved):

mazhelez (approved):

thpeder (approved):

PR #9007, opened 2026-07-01
Skip automation on out-of-support branches

mazhelez:
Fixes [AB#641122](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/641122)

## Summary

BCApps automations that open PRs to official branches should not run on branches that are out of support (branches carrying the **"Branch is out of support"** ruleset). Most automations already filter these out via the `GetGitBranches` action, but there were two gaps this PR closes.

## Changes

**1. Defense-in-depth guard in the shared `RunAutomation` action** — `.github/actions/RunAutomation/run.ps1`
Before running any automation, the action now checks the target branch with `Test-IsBranchInSupport` and returns early with a `::Notice::` if the branch is out of support. This covers every current and future `RunAutomation`-based automation, including manual `workflow_dispatch` runs that bypass the `GetGitBranches` filtering.

**2. New automation to keep the AL-Go schedule in sync** — `.github/actions/RunAutomation/UpdateSystemFilesSchedule/run.ps1`
The **Update AL-Go System Files** workflow (`UpdateGitHubGoSystemFiles.yaml`) is an AL-Go–generated file that picks its branches from `workflowSchedule.includeBranches` in `.github/Update AL-Go System Files.settings.json`. Unlike the other automations it does **not** use `GetGitBranches`, so it has no knowledge of the ruleset and can create PRs against out-of-support branches (the list was already stale — e.g. `releases/25.*`).

This new automation rewrites `includeBranches` with the explicit list of every official branch (`main` + `releases/*`) that is currently in support:
- No wildcards — every branch is listed explicitly.
- Recomputed on each run, so out-of-support branches drop off and newly created branches are picked up.
- Deterministic ordering (`main` first, then natural numeric order) to keep diffs clean.

**3. New workflow** — `.github/workflows/UpdateALGoSupportedBranches.yaml`
Runs the new automation on `main` (monthly + `workflow_dispatch`), mirroring the structure of the existing automation workflows.

**4. Docs** — `.github/actions/RunAutomation/HowToAddAnAutomation.md`
Notes that automations are automatically skipped on out-of-support branches, so new automations don't need to handle it.

## Validation

Ran the new automation against the live rulesets. It correctly dropped the out-of-support branches (`releases/25.*`, `releases/26.0`–`26.3`) and produced the explicit in-support list (`main`, `releases/26.4`, `releases/26.5`, `releases/26.x`, `releases/27.0`–`27.5`, `releases/27.x`, `releases/28.0`–`28.3`, `releases/28.x`).

### DIFF
--- a/.github/actions/RunAutomation/UpdateSystemFilesSchedule/run.ps1
+++ b/.github/actions/RunAutomation/UpdateSystemFilesSchedule/run.ps1
@@ -1,0 +1,93 @@
+<#
+.Synopsis
+    Updates the branch list of the "Update AL-Go System Files" scheduled run so it only targets supported branches.
+.Description
+    The "Update AL-Go System Files" workflow (UpdateGitHubGoSystemFiles.yaml) is an AL-Go system file and
+    picks the branches to run on from the 'workflowSchedule.includeBranches' setting in
+    '.github/Update AL-Go System Files.settings.json'. Unlike the other BCApps automations, it does not use
+    the GetGitBranches action and therefore has no knowledge of the "Branch is out of support" ruleset.
+
+    This automation rewrites 'workflowSchedule.includeBranches' with the explicit list of every official
+    branch (main and releases/*) that is currently in support. The list contains no wildcards and is
+    recomputed on every run, so out-of-support branches are dropped and newly created branches are added.
+.Parameter runParameters
+    A hashtable with the parameters passed by the RunAutomation action (Repository and TargetBranch).
+#>
+param (
+    [Parameter(Mandatory = $true)]
+    $runParameters
+)
+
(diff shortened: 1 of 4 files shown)


.github/actions/RunAutomation/UpdateSystemFilesSchedule/run.ps1
```diff
@@ -0,0 +1,93 @@
+<#
```
aholstrup1, line 1: I guess we should delete this one then? 
https://github.com/microsoft/BCApps/blob/main/.github/Update%20AL-Go%20System%20Files.settings.json

spetersenms (approved):

aholstrup1 (approved):

PR #9024, opened 2026-07-02
[Master] - Chance of success % does not validate correctly when using the action type "Update" in Opportunity

DeepsShukla:
Workitem [Bug 640926](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/640926): [master] [ALL-E] Chance of success % does not validate correctly when using the action type "Update" in Opportunity
[AB#640926](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/640926)


v-ankitgoyal (approved):

v-ankitgoyal (approved):

alexei-dobriansky:

<details>

This PR updates `WizardActionTypeValidate2` so changing the action type also copies `Chances of Success %` from the selected sales cycle stage. That matches the bug scenario: the wizard starts on Next, then changing to Update must put the chance value back to the current stage.

The fix is targeted. Each action type already chooses a temp sales cycle stage and updates the stage number and description; copying the chance value from the same temp record keeps those fields aligned. The code after the change only adjusts the cancel-task flag and saves the record, so the new value is not overwritten before the user sees it. No BaseApp event publisher or public signature change is involved.
</details>

<details>

**S1** - *Use fixed different chance values in the test*
The new test gets both stage values from random helpers. Set Stage 1 and Stage 2 to known different values, so the test always proves that `WizardActionTypeValidate2` moved the value.
</details>

<details>

**Risk:** The regression surface is the Opportunity update wizard and any code that calls `WizardActionTypeValidate2` directly. The change affects probability input data, so a wrong value could affect opportunity pipeline values, but the scope is narrow and there is no event or signature compatibility risk. The current GitHub check rollup shows failures, so merge readiness still depends on those checks being resolved.

**Necessity:** Bug 640926 is clear and reproducible: an Opportunity on stage 1 with 10% chance shows 40% for Next, and changing the action type to Update must return to 10%. Without this change, users can keep the wrong chance value unless they correct it manually. The PR changes the right wizard path and includes a focused regression test.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9024 round=1 by=alexei-dobriansky at=2026-07-03T08:17:31Z lastSha=27be72b083cb6392af1752f4e7b7c004ef1be6e1 suggestions=S1

alexei-dobriansky (approved):

v-ankitgoyal (approved):

PR #9027, opened 2026-07-02
Switch AI triage dispatch to PAT-free pull model

JesperSchulz:
## Why

AI issue triage is powered by an agent in the private `microsoft/BCAppsTriage` repo. Today the two repos talk over **two PATs**:

| Secret | Direction | Purpose |
| --- | --- | --- |
| `TRIAGE_DISPATCH_TOKEN` | BCApps → agent | `repository_dispatch` to start the agent |
| `SOURCE_REPO_TOKEN` (on the agent repo) | agent → BCApps | write the comment + team label + issue type back |

The org now caps PAT lifetime at **7 days**, so both tokens expire almost immediately and triage breaks. The agent must also stay in a **separate private repo**, and we **cannot install a GitHub App on BCApps**.

## What this changes

Replaces the single `.github/workflows/ai-triage-dispatch.yml` with a **pull model**. BCApps initiates and applies the result itself; the agent never writes to BCApps.

```mermaid
sequenceDiagram
    participant B as BCApps (public)
    participant A as BCAppsTriage (private)
    B->>B: mint GitHub App token (scoped to BCAppsTriage only)
    B->>A: workflow_dispatch issue-triage.yml (issue_number, run_token)
    A->>B: read issue + code (public, no credential)
    A->>A: upload triage-result.json artifact
    B->>A: download triage-result artifact
    B->>B: apply comment + team label + issue type (native GITHUB_TOKEN)
```

- **No GitHub App is installed on BCApps.** The App is installed only on the private `BCAppsTriage` repo (`Actions: read/write` + `Metadata: read`).
- The App id + private key stored here can **only** start a run and read an artifact in the private repo — they **cannot write to BCApps**.
- Results are applied to the issue with the workflow's built-in `GITHUB_TOKEN` (`issues: write`) — no stored write credential.
- Trigger behavior is unchanged: newly opened issues and issues that get the `ai-triage` label are triaged.

## Required setup before merge takes effect

1. Create a GitHub App in `microsoft` with **Actions: read & write** + **Metadata: read** (no webhook needed).
2. Install it on **`microsoft/BCAppsTriage` only**.
3. Add two repo secrets here:
   - `TRIAGE_APP_ID`
   - `TRIAGE_APP_PRIVATE_KEY`
4. After this is live, the old `TRIAGE_DISPATCH_TOKEN` secret can be deleted (and `SOURCE_REPO_TOKEN` on the agent repo).

## Companion PR

Agent-side changes (artifact output, `workflow_dispatch` trigger) and the full setup guide are in **microsoft/BCAppsTriage#28**.

Fixes: [AB#641237](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/641237)

### DIFF
--- a/.github/workflows/ai-triage-dispatch.yml
+++ b/.github/workflows/ai-triage-dispatch.yml
@@ -1,13 +1,42 @@
-name: Dispatch AI Triage
+name: AI Issue Triage (dispatch)
+
+# Lives in the ISSUES repo (e.g. microsoft/BCApps). It owns nothing secret about
+# the triage agent — it simply:
+#   1. mints a short-lived GitHub App token scoped to the private agent repo,
+#   2. starts the agent (workflow_dispatch) and waits for it,
+#   3. downloads the agent's `triage-result` artifact,
+#   4. applies the comment / team label / issue type using THIS repo's built-in
+#      GITHUB_TOKEN.
+#
+# No PAT is used, and no GitHub App is installed on this repo. The only stored
+# secrets are the App id + private key, which can ONLY start a run and read an
+# artifact in the private agent repo — they cannot write here.
 
 on:
   issues:
+    # Triage newly opened issues and issues that get the `ai-triage` label.
     types: [opened, labeled]
 
-permissions: {}
+permissions:
+  issues: write      # native GITHUB_TOKEN posts the comment + labels + type here
+  contents: read
+
+concurrency:
+  group: ai-triage-${{ github.event.issue.number }}
+  cancel-in-progress: true
+
+env:
+  # The private agent repository (owner/repo, for `gh -R`) and the workflow file
+  # to invoke there.
+  AGENT_REPO: microsoft/BCAppsTriage
+  AGENT_WORKFLOW: issue-triage.yml
+  # Bare repository name — the create-github-app-token `repositories` input takes
+  # names without the owner prefix when `owner` is set.
+  AGENT_REPO_NAME: BCAppsTriage
 
 jobs:
-  dispatch:
+  triage:
+    # Run for a newly opened issue, or when `ai-triage` is added to an open issue.
     if: >-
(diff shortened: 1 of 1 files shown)


.github/workflows/ai-triage-dispatch.yml
```diff
@@ -1,26 +1,184 @@
...
-      - name: Trigger triage on BCAppsTriage
+      - name: Correlation token
+        id: token
+        run: echo "run_token=$(uuidgen)" >> "$GITHUB_OUTPUT"
+
+      - name: Mint token for the agent repo
+        id: app-token
+        uses: actions/create-github-app-token@v2
+        with:
+          app-id: ${{ secrets.TRIAGE_APP_ID }}
```
aholstrup1: Let's make this a Repo variable instead since it isn't really a secret. Then you reference it like this instead
https://github.com/microsoft/BCApps/blob/358e683c13979c23d8fdc1af3bf3fc32669a19ef/.github/workflows/UpdateALGoProjects.yaml#L54

.github/workflows/ai-triage-dispatch.yml
```diff
@@ -1,26 +1,184 @@
...
+          private-key: ${{ secrets.TRIAGE_APP_PRIVATE_KEY }}
```
aholstrup1: This secret really should live inside a branch-protected GitHub environment. In order to make that happen we need to do two things: 
1. Move the secret to a GitHub environment
2. Reference the GitHub environment on the job like this
https://github.com/microsoft/BCApps/blob/358e683c13979c23d8fdc1af3bf3fc32669a19ef/.github/workflows/CopilotPRReviewRunner.yaml#L26

aholstrup1 (approved):

PR #9032, opened 2026-07-02
Use BCPlatform version from Packages.json in NewDevEnv

mazhelez:
## What & why

NewDevEnv was not honoring the `BCPlatform` version pinned in `build/Packages.json`. The reason is that `build/scripts/DevEnv/NewDevEnv.psm1` creates the dev container (`New-BcContainer`) and the compiler folder (`New-BcCompilerFolder`) **directly**, which bypasses the AL-Go override scripts (`NewBcContainer.ps1` / `PreNewBcCompilerFolder.ps1`) that inject the `platformArtifactUrl` derived from the configured BCPlatform version. As a result a NewDevEnv container/compiler always used the platform shipped with the artifact instead of the pinned platform.

This change makes NewDevEnv apply the same platform version that the CI/AL-Go path already uses.

## Linked work

Fixes #

## How I validated this

- Added and ran Pester tests for the new `Get-BCPlatformArtifactUrl` helper in `build/scripts/tests/PlatformHelper.Test.ps1` (full version, major.minor resolution, and the "not configured" case). All 15 PlatformHelper tests pass locally.
- Parse-checked all four modified PowerShell files with the PowerShell language parser; no syntax errors.
- End-to-end container creation was not exercised locally (requires Docker + BC artifacts).

## Risk & compatibility

Backward compatible. The new behavior only sets `platformArtifactUrl` when a `BCPlatform` version is configured in `Packages.json`; when it is absent the helper returns `$null` and container/compiler creation is unchanged. The two AL-Go override scripts were refactored to call the shared helper, so their behavior is functionally identical (deduplication only).


spetersenms (dismissed):

mazhelez:
@copilot+claude-opus-4.8

Please fix this alert:
The cmdlet 'Get-BCPlatformArtifactUrl' returns an object of type 'System.String' but this type is not declared in the OutputType attribute.

mazhelez:

build/scripts/PlatformHelper.psm1
```diff
@@ -156,6 +158,35 @@ function Resolve-PlatformVersion {
...
+    param()
+
+    $platformVersion = (Get-ConfigValue -Key "BCPlatform" -ConfigType Packages).Version
+    if (-not $platformVersion) {
+        return $null
+    }
+
+    $platformVersion = Resolve-PlatformVersion -Version $platformVersion
+    $platformUrl = Get-PlatformVersionUrl -Version $platformVersion
+    return "$platformUrl/platform"
```
aholstrup1: Why wouldn't Get-PlatformVersionUrl not just return the $platformUrl/platform ? It seems like you'd always need to add this suffix anyway

aholstrup1 (dismissed):

build/scripts/PlatformHelper.psm1
mazhelez: Good point - since every caller needs the suffix, I've folded it into `Get-PlatformVersionUrl` so it now returns the full artifact URL (.../platform/$Version/platform) directly. `Get-BCPlatformArtifactUrl` just returns the resolved URL now. Done in e9634bb400.

mazhelez:

build/scripts/DevEnv/NewDevEnv.psm1
```diff
@@ -326,7 +347,21 @@ function CreateCompilerFolder {
...
-    New-BcCompilerFolder -artifactUrl $bcArtifactUrl -cacheFolder $compilerFolder | Out-Null
+
+    $newCompilerFolderParameters = @{
+        artifactUrl = $bcArtifactUrl
+        cacheFolder = $compilerFolder
+    }
+
+    # Use the platform version from Packages.json when specified
+    Import-Module "$PSScriptRoot\..\PlatformHelper.psm1" -DisableNameChecking
+    $platformArtifactUrl = Get-BCPlatformArtifactUrl
```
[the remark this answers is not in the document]
mazhelez: Good catch. `CreateCompilerFolder` now uses the resolved platform artifact URL as part of the cache key: it writes a `.platformArtifactUrl` marker file recording the URL used to build the compiler folder, and on subsequent runs it only reuses the cached folder when the marker matches the currently configured `BCPlatform` version. If they differ (or the pin was added/removed), the cached folder is removed and recreated with the new platform. Fixed in 2116e5ac30.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

PR #9051, opened 2026-07-03
[639592][ALAppExtensions #30277][Event Request] Codeunit 426 "Payment Tolerance Management" OnPmtTolGenJnlOnAfterCheckConditions

Explorer986:
## What & why

Adds a by-reference variant of `OnPmtTolGenJnlOnAfterCheckConditions` so
subscribers can change the general journal line before payment tolerance
processing continues into the sales or purchase path. The existing event passes
the line by value, and changing its signature would break current subscribers, so
this adds a parallel event that exposes the line as var. Scoped to the NL layer.

## Linked work

Fixes [AB#639592](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/639592)

## How I validated this

The new event is raised immediately after the existing one, over the same
temporary line, so nothing changes when there is no subscriber. Existing
subscribers to the by-value event are untouched. No test added since this only
adds an extension point.

## Risk & compatibility

The event lives in the NL layer only, so it is available to NL builds where the
requesting partner needs it. Non-breaking and additive.

### DIFF
--- a/src/Layers/NL/BaseApp/Finance/ReceivablesPayables/PaymentToleranceManagement.Codeunit.al
+++ b/src/Layers/NL/BaseApp/Finance/ReceivablesPayables/PaymentToleranceManagement.Codeunit.al
@@ -227,6 +227,7 @@ codeunit 426 "Payment Tolerance Management"
             exit(true);
 
         OnPmtTolGenJnlOnAfterCheckConditions(TempGenJnlLine, SuppressCommit, Result);
+        OnPmtTolGenJnlOnAfterCheckConditionsGenJnlLineByRef(TempGenJnlLine, SuppressCommit, Result);
 
         case true of
             (TempGenJnlLine."Account Type" = TempGenJnlLine."Account Type"::Customer) or
@@ -3059,6 +3060,18 @@ codeunit 426 "Payment Tolerance Management"
     /// <param name="Result">Result of tolerance condition checking</param>
     [IntegrationEvent(false, false)]
     local procedure OnPmtTolGenJnlOnAfterCheckConditions(GenJournalLine: Record "Gen. Journal Line"; var SuppressCommit: Boolean; var Result: Boolean)
+    begin
+    end;
+
+    /// <summary>
+    /// Integration event raised after checking conditions for payment tolerance in general journal processing.
+    /// Exposes the general journal line by reference so subscribers can modify it before standard payment tolerance processing continues.
+    /// </summary>
+    /// <param name="GenJournalLine">General journal line being processed, passed by reference</param>
+    /// <param name="SuppressCommit">Whether database commit should be suppressed</param>
+    /// <param name="Result">Result of tolerance condition checking</param>
+    [IntegrationEvent(false, false)]
+    local procedure OnPmtTolGenJnlOnAfterCheckConditionsGenJnlLineByRef(var GenJournalLine: Record "Gen. Journal Line"; var SuppressCommit: Boolean; var Result: Boolean)
     begin
     end;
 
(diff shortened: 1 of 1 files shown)


src/Layers/NL/BaseApp/Finance/ReceivablesPayables/PaymentToleranceManagement.Codeunit.al
```diff
@@ -227,6 +227,7 @@ codeunit 426 "Payment Tolerance Management"
             exit(true);
 
         OnPmtTolGenJnlOnAfterCheckConditions(TempGenJnlLine, SuppressCommit, Result);
+        OnPmtTolGenJnlOnAfterCheckConditionsGenJnlLineByRef(TempGenJnlLine, SuppressCommit, Result);
```
mynjj: this is a bit of a weird one, does adding an extra parameter as `var` would work for what they need? also, I'm not a fan of editing the GenJnlLine being validated on the fly, at what point does this happen?

Explorer986: does adding an extra parameter as var would work for what they need?
-- Not on the existing event: its record parameter is by value, and neither flipping it to var nor inserting a new parameter can be done without breaking every current subscriber 

The existing OnPmtTolGenJnlOnAfterCheckConditions passes the line by value, so we can't switch it to var (or add a var param) without breaking existing subscribers so the separate GenJnlLineByRef event. Note the record here is TempGenJnlLine, a temporary copy of the incoming NewGenJnlLine, so subscribers can influence how tolerance is evaluated for this call, but nothing is written back to the source line or the DB. It fires right after the early-exit condition checks and before the customer/vendor dispatch.

ventselartur (dismissed):

src/Layers/NL/BaseApp/Finance/ReceivablesPayables/PaymentToleranceManagement.Codeunit.al
Explorer986: Updated as suggested by adding additional parameter as `var`

alexei-dobriansky:

<details>

This PR changes the NL BaseApp publisher `OnPmtTolGenJnlOnAfterCheckConditions` in codeunit 426 so subscribers can modify the temporary general journal line before customer or vendor payment tolerance processing continues. The call site is after the existing guard checks and before the customer/vendor `case`, so the timing is right, and there is no runtime behavior change when no subscriber changes the line.

The issue is compatibility. The diff changes the existing integration event signature by adding a fourth parameter. This is a public event contract, and the other in-repo layer versions still have the old signature.
</details>

<details>

**S1** - *Keep the existing event signature stable*
Keep `OnPmtTolGenJnlOnAfterCheckConditions` with its current three parameters. Add a separate new event with a new name for the `var Gen. Journal Line` case. This avoids breaking existing subscribers while still giving new subscribers a way to change the line.
</details>

<details>

**Risk:** The changed code is in payment tolerance handling, so an incorrect extension point can affect payment application decisions. The runtime path is narrow and additive when unsubscribed, but changing a published integration event signature is a compatibility risk for existing NL subscribers. The publisher and call site were verified in `src\Layers\NL\BaseApp\Finance\ReceivablesPayables\PaymentToleranceManagement.Codeunit.al`; W1, CH, FR, and IT still expose the old three-parameter event.

**Necessity:** AB#639592 / ALAppExtensions #30277 describes a valid event request: a subscriber needs to adjust the temporary general journal line before the standard payment tolerance flow enters the customer or vendor branch. A pure publisher addition does not need a unit test here, but the public event contract must stay backward compatible.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9051 round=1 by=alexei-dobriansky at=2026-07-09 lastSha=e94cd40fde5dd345f421e7ac3545f0c61d20d1ec suggestions=S1

AleksanderGladkov (approved):

mynjj (approved):

PR #9053, opened 2026-07-03
[ALAppExtensions #30245][W1][Report][1701][Deferral Summary - Sales] Add integration event OnBeforeReturnSalesDocTypeString in ReturnSalesDocTypeString Procedure

Explorer986:
## What & why

Adds an IsHandled-style event to `ReturnSalesDocTypeString` so extensions can
replace the document type text the report shows. The procedure is local and maps
each type with a hardcoded case, so there was no way to override the result. With
this event a subscriber can, for example, derive the text from the
"Sales Comment Document Type" enum instead.

## Linked work

Fixes [AB#639936](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/639936)

## How I validated this

When no subscriber sets IsHandled, the original case statement runs unchanged and
returns the same labels as before. I confirmed the return value is now surfaced
through a named return so the event can override it cleanly. No test added, the
default path is identical to today's behavior.

## Risk & compatibility

None. Additive event, default output unchanged.

### DIFF
--- a/src/Layers/W1/BaseApp/Finance/Deferral/DeferralSummarySales.Report.al
+++ b/src/Layers/W1/BaseApp/Finance/Deferral/DeferralSummarySales.Report.al
@@ -481,11 +481,11 @@ report 1701 "Deferral Summary - Sales"
         DocumentFilter := NewDocumentNoFilter;
     end;
 
-    local procedure ReturnSalesDocTypeString(SalesDocType: Integer) Result: Text
+    local procedure ReturnSalesDocTypeString(SalesDocType: Integer): Text
     var
+        Result: Text;
         IsHandled: Boolean;
     begin
-        IsHandled := false;
         OnBeforeReturnSalesDocTypeString(SalesDocType, Result, IsHandled);
         if IsHandled then
             exit(Result);
--- a/src/Layers/W1/BaseApp/Finance/Deferral/DeferralSummarySales.Report.al
+++ b/src/Layers/W1/BaseApp/Finance/Deferral/DeferralSummarySales.Report.al
@@ -481,8 +481,15 @@ report 1701 "Deferral Summary - Sales"
         DocumentFilter := NewDocumentNoFilter;
     end;
 
-    local procedure ReturnSalesDocTypeString(SalesDocType: Integer): Text
+    local procedure ReturnSalesDocTypeString(SalesDocType: Integer) Result: Text
+    var
+        IsHandled: Boolean;
     begin
+        IsHandled := false;
+        OnBeforeReturnSalesDocTypeString(SalesDocType, Result, IsHandled);
+        if IsHandled then
+            exit(Result);
+
         case SalesDocType of
             0:
                 exit(QuoteLbl);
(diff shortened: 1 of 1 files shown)


Explorer986:

AndreasHans (dismissed):

src/Layers/W1/BaseApp/Finance/Deferral/DeferralSummarySales.Report.al
```diff
@@ -481,8 +481,15 @@ report 1701 "Deferral Summary - Sales"
         DocumentFilter := NewDocumentNoFilter;
     end;
 
-    local procedure ReturnSalesDocTypeString(SalesDocType: Integer): Text
+    local procedure ReturnSalesDocTypeString(SalesDocType: Integer) Result: Text
+    var
+        IsHandled: Boolean;
     begin
+        IsHandled := false;
```
mynjj: unnecessary, by default initialized

src/Layers/W1/BaseApp/Finance/Deferral/DeferralSummarySales.Report.al
```diff
@@ -481,8 +481,15 @@ report 1701 "Deferral Summary - Sales"
         DocumentFilter := NewDocumentNoFilter;
     end;
 
-    local procedure ReturnSalesDocTypeString(SalesDocType: Integer): Text
+    local procedure ReturnSalesDocTypeString(SalesDocType: Integer) Result: Text
```
mynjj: unnecessary, you are always exiting with explicit `exit` anyway

mynjj (commented):
.

ventselartur (approved):

Groenbech96 (approved):

alexei-dobriansky:

# **Recommendation: Accept**

<details>

This PR adds `OnBeforeReturnSalesDocTypeString` in report 1701 "Deferral Summary - Sales" before `ReturnSalesDocTypeString` maps the sales document type integer to a label. A subscriber can set `Result` and `IsHandled` to replace the text shown by the report.

The publisher is in the in-repo BaseApp file `src/Layers/W1/BaseApp/Finance/Deferral/DeferralSummarySales.Report.al`. The event is raised before the existing `case` statement, uses `[IntegrationEvent(false, false)]`, passes the input value plus `var Result` and `var IsHandled`, and keeps the current labels when no subscriber handles it. The naming also matches the procedure and the standard OnBefore/IsHandled pattern.
</details>

<details>

*No blocking or optional suggestions.*
</details>

<details>

**Risk:** This adds a public event, so the signature becomes part of the extension surface. The runtime behavior is otherwise unchanged without subscribers. A subscriber can only change the document type text used by this report output; it does not receive records, amounts, or posting state. The matching purchasing report still has the same local hardcoded mapping and no event, which is consistent with this sales-only request.

**Necessity:** AB#639936 and ALAppExtensions #30245 describe an extensibility gap: `ReturnSalesDocTypeString` is local and its result cannot be replaced after it exits. The scope is narrow and placed at the only useful point before the mapping returns, so a test is not required for the unchanged default path.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9053 round=1 by=alexei-dobriansky at=2026-07-08 lastSha=34b4bfa7428714774ccbe04146f1f54bf37cc317 suggestions=none

PR #9054, opened 2026-07-03
[639939][ALAppExtensions #30250] codeunit 9200 "Matrix Management" - Add integration events for custom rounding

Explorer986:
## What & why

Adds two events so extensions can customize rounding in analysis and financial
reports. Customers picking the "1000" rounding option still see a decimal place,
and they want whole numbers. `OnRoundAmountOnAfterRound` lets a subscriber adjust
the rounded amount, and `OnFormatRoundingFactorOnAfterSetAmountDecimal` lets them
change the decimal formatting for the chosen factor.

## Linked work

Fixes [AB#639939](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/639939)

## How I validated this

`RoundAmount` now assigns the rounded value and raises the event before returning
instead of exiting inside each case. The math is identical, so with no subscriber
the returned amount and formatting match today's output. No test added because the
default behavior is unchanged and the events are extension points.

## Risk & compatibility

None expected. RoundAmount was refactored from early exits to a single exit; the
computed values are the same. Worth a quick reviewer glance at that refactor.

### DIFF
--- a/src/Layers/W1/BaseApp/Finance/Analysis/MatrixManagement.Codeunit.al
+++ b/src/Layers/W1/BaseApp/Finance/Analysis/MatrixManagement.Codeunit.al
@@ -613,15 +613,16 @@ codeunit 9200 "Matrix Management"
 
         case RoundingFactor of
             RoundingFactor::"1":
-                exit(Round(Amount, 1));
+                Amount := Round(Amount, 1);
             RoundingFactor::"1000":
-                exit(Round(Amount / 1000, 0.1));
+                Amount := Round(Amount / 1000, 0.1);
             RoundingFactor::"1000000":
-                exit(Round(Amount / 1000000, 0.1));
+                Amount := Round(Amount / 1000000, 0.1);
             else
                 OnRoundAmountOnElse(Amount, RoundingFactor);
         end;
 
+        OnRoundAmountOnAfterRound(Amount, RoundingFactor);
         exit(Amount);
     end;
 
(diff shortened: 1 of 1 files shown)


Explorer986:

AndreasHans (approved):

src/Layers/W1/BaseApp/Finance/Analysis/MatrixManagement.Codeunit.al
```diff
@@ -613,15 +613,16 @@ codeunit 9200 "Matrix Management"
...
-                exit(Round(Amount / 1000, 0.1));
+                Amount := Round(Amount / 1000, 0.1);
             RoundingFactor::"1000000":
-                exit(Round(Amount / 1000000, 0.1));
+                Amount := Round(Amount / 1000000, 0.1);
             else
                 OnRoundAmountOnElse(Amount, RoundingFactor);
         end;
 
+        OnRoundAmountOnAfterRound(Amount, RoundingFactor);
```
mynjj, line 625: why is this event needed when there's th option of having your own in the `else` above?

Explorer986: Many of our customers wonder why, when they select the "1,000" rounding option in analyses and financial reports, the amounts still include a decimal place. They want the amounts to be rounded to whole numbers.

That's why they wanted a new Event so that they can get whole number after rounding

    // add event here to allow changing Amount based on RoundingFactor - begin
    OnRoundAmountAfterRound(Amount, RoundingFactor);
    // add event here to allow changing Amount based on RoundingFactor - end

ventselartur (approved):

PR #9059, opened 2026-07-03
Resolve test app names from app.json to prevent silent CI test skips

aholstrup1:
## What & why

BCApps CI was silently skipping entire test apps. The parallel test dispatcher decided which test apps to run by matching each project's `projects.json` key (surfaced as `BuildMetadata.ApplicationName`, and also its `groups.json` name) against the app name the container reports (`app.json` `"name"`). Those two strings are *required* to be identical, but when they drift the app is dropped from the dispatch set and **all of its tests are silently skipped while the build stays green**.

This is how `WIPTransferRemainderCreatedWhenOpenLineQuantityReduced` (and the rest of the Subcontracting suite) failed in another build system yet never even ran in GitHub CI: `Subcontracting-Tests` (projects.json key) vs `Subcontracting Test` (app.json name).

A full scan found **7 affected test apps** with this drift: Subcontracting Test (~22 test codeunits), Subscription Billing Test (~30), OnPrem Permissions Test, Statistical Accounts Test, Dynamics SL Migration Tests, Tests for MX DIOT Extension, and Basic Experience Tests. All were installed in the container but never dispatched.

## Approach

Rather than hand-editing the metadata for all 7 (which fixes the symptom but leaves the mechanism fragile), this makes the dispatcher robust by using the authoritative source of truth. `Get-InstalledTestAppNames` now resolves each test app's name from its `app.json` (via `AppJsonPath`) using a new `Get-AppNameFromMetadata` helper, instead of trusting the projects.json key. That is the same string the container reports, so both the installed-name intersection and the downstream name-to-extensionId dispatch map line up, and any future key/name drift self-heals.

When the projects.json key and app.json name disagree, it emits a `::warning::` naming the offending app so the drift is visible instead of silent, while still dispatching the tests correctly.

Verified safe: all 40 legacy-bucket apps and every other test app already have `key == app.json name`, so switching the source of truth changes behavior only for the 7 that were broken.

## Linked work

Fixes #

Related to [AB#641388](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/641388) (Subcontracting `WIPTransferRemainderCreatedWhenOpenLineQuantityReduced` failure surfaced and disabled by this change)
Related to [AB#641478](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/641478) (Subscription Billing IT `DeferralsReleaseSucceedsWhenGLAccountHasDefaultDeferralTemplateAndJournalTemplNotMandatory` failure surfaced and disabled by this change)

## How I validated this

- This is a build-script (PowerShell) change, not an AL app change, so there is no BC app to build or run.
- Added `build/scripts/tests/ParallelTestExecution.Test.ps1` (Pester). 5 tests, all green locally with Pester 5.7.1, covering: name resolved from app.json when it differs from the projects.json key, the aligned case, and both fallbacks (missing app.json path, empty path). The drift test also asserts `Get-InstalledTestAppNames` dispatches the app under its real installed name.
- Confirmed against the real failing run (28391219041): extensionId `32b0d4d1` (Subcontracting Test) was installed but never appeared in any `Dispatching '...'` line in the W1 or US IntegrationTests jobs; no codeunit 149911 header and no test-function output were present. This change makes that app resolve into the dispatch set.

## Risk & compatibility

- No metadata files (`projects.json`, `groups.json`, `app.json`) are changed; this is purely a resolution-logic change in `ParallelTestExecution.psm1`.
- Once merged, the ~7 previously-skipped test suites will start running in CI. Some may surface pre-existing real failures (for example the Subcontracting WIP transfer test), which is the intended outcome. Reviewers should expect newly-executing tests, not regressions from this change itself.
- Two such pre-existing failures found so far have been temporarily added to their apps' DisabledTests JSON and tracked by [AB#641388](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/641388) and [AB#641478](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/641478).
- Fallback preserves today's behavior for any app whose app.json cannot be read.
- Follow-up worth considering (not in this PR): a guard test asserting `projects.json key == groups.json name == app.json "name"` for every test project, so drift is caught at PR time.

### DIFF
--- a/build/scripts/ParallelTestExecution.psm1
+++ b/build/scripts/ParallelTestExecution.psm1
@@ -48,11 +48,50 @@ function Get-CachedTestRunResult {
 
 <#
 .SYNOPSIS
+    Resolves the authoritative app name for a project from its app.json "name" field.
+.DESCRIPTION
+    Get-ApplicationGroup exposes the projects.json key as ApplicationName. That key is required to
+    match the app.json "name", but the two can drift (typos, "-Tests" vs " Test", trailing dots).
+    When they drift, the container only ever knows the app.json "name", so any match keyed off
+    ApplicationName silently drops the app from the test dispatch set and skips ALL of its tests
+    while the build stays green. Reading the name straight from app.json (via AppJsonPath) makes
+    the dispatch robust against that drift. Falls back to ApplicationName if app.json is missing
+    or unreadable.
+.OUTPUTS
+    [string] The app.json "name", or ApplicationName as a fallback.
+#>
+function Get-AppNameFromMetadata {
+    param(
+        [Parameter(Mandatory=$true)]
+        $BuildMetadata
+    )
+
+    $appJsonPath = $BuildMetadata.AppJsonPath
+    if (-not [string]::IsNullOrWhiteSpace($appJsonPath) -and (Test-Path $appJsonPath -PathType Leaf)) {
+        try {
+            $appName = (Get-Content $appJsonPath -Raw | ConvertFrom-Json).name
+            if (-not [string]::IsNullOrWhiteSpace($appName)) {
+                if ($appName -ne $BuildMetadata.ApplicationName) {
+                    Write-Host "::warning::Test app projects.json key '$($BuildMetadata.ApplicationName)' does not match its app.json name '$appName'. Using the app.json name for test dispatch. Align the projects.json key (and build/groups.json name) with the app.json name to avoid confusion."
+                }
+                return $appName
+            }
+        } catch {
+            Write-Host "WARNING: Could not read app name from '$appJsonPath' ($($_.Exception.Message)); falling back to projects.json key '$($BuildMetadata.ApplicationName)'."
+        }
+    }
+    return $BuildMetadata.ApplicationName
+}
+
+<#
+.SYNOPSIS
     Returns the names of test apps that are both expected for a country and installed in the container.
 .DESCRIPTION
     Combines the project metadata in build/projects.json (via Get-ApplicationGroup) with
     Get-BcContainerAppInfo so we only ever try to dispatch apps that actually exist in the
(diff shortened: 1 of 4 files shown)


spetersenms (dismissed):

spetersenms (approved):

build/scripts/ParallelTestExecution.psm1
```diff
@@ -46,13 +46,52 @@ function Get-CachedTestRunResult {
...
+        $BuildMetadata
+    )
+
+    $appJsonPath = $BuildMetadata.AppJsonPath
+    if (-not [string]::IsNullOrWhiteSpace($appJsonPath) -and (Test-Path $appJsonPath -PathType Leaf)) {
+        try {
+            $appName = (Get-Content $appJsonPath -Raw | ConvertFrom-Json).name
+            if (-not [string]::IsNullOrWhiteSpace($appName)) {
+                if ($appName -ne $BuildMetadata.ApplicationName) {
+                    Write-Host "::warning::Test app projects.json key '$($BuildMetadata.ApplicationName)' does not match its app.json name '$appName'. Using the app.json name for test dispatch. Align the projects.json key (and build/groups.json name) with the app.json name to avoid confusion."
```
mazhelez, line 75:
```suggestion
                    Write-Host "::warning::Test app projects.json key '$($BuildMetadata.ApplicationName)' does not match its app.json name '$appName'. Using the app.json name for test dispatch. Align the projects.json key (and groups.json name) with the app.json name to avoid confusion."
```

mazhelez (approved):

PR #9079, opened 2026-07-03
Bug 631472: Cost Shares Breakdown (WIP) does not filter by Item

alexei-dobriansky:
Fixes [AB#631472](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/631472)

Under **Print Cost Share = WIP Inventory** the Production Order / Capacity Ledger Entry dataitems enumerated every released production order regardless of the request-page Item filter. Skip capacity entries whose produced item is not in the Item filter, using the same temp-item ``CopyFilters``/``IsEmpty`` idiom already used for the WIP buffer output in this report. Adds a WIP test asserting only the filtered item is reported.

### DIFF
--- a/src/Layers/W1/BaseApp/Manufacturing/Reports/CostSharesBreakdown.Report.al
+++ b/src/Layers/W1/BaseApp/Manufacturing/Reports/CostSharesBreakdown.Report.al
@@ -66,7 +66,17 @@ report 5848 "Cost Shares Breakdown"
                 DataItemTableView = sorting("Order Type", "Order No.", "Order Line No.") where("Order Type" = const(Production));
 
                 trigger OnAfterGetRecord()
+                var
+                    TempItem: Record Item temporary;
                 begin
+                    if Item.Get("Capacity Ledger Entry"."Item No.") then begin
+                        TempItem := Item;
+                        TempItem.Insert();
+                        TempItem.CopyFilters(Item);
+                        if TempItem.IsEmpty() then
+                            CurrReport.Skip();
+                    end;
+
                     InsertCapLedgEntryCostShare("Capacity Ledger Entry");
                 end;
 
--- a/src/Layers/W1/Tests/SCM/SCMInventoryReportsIV.Codeunit.al
+++ b/src/Layers/W1/Tests/SCM/SCMInventoryReportsIV.Codeunit.al
@@ -539,6 +539,63 @@ codeunit 137351 "SCM Inventory Reports - IV"
     end;
 
     [Test]
+    [HandlerFunctions('ProductionJournalPageHandler,ConfirmHandler,MessageHandler,CostSharesBreakdownRequestPageHandler')]
+    [Scope('OnPrem')]
+    procedure CostSharesBreakdownReportForWIPInventoryFiltersByItem()
+    var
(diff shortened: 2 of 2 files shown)


alexei-dobriansky:

# **Recommendation: Accept**

<details>

This PR fixes the WIP Inventory mode of report 5848, Cost Shares Breakdown, so capacity ledger entries are not inserted for production items that do not match the request-page Item filter. The linked bug and the [AI-REPRO] comment describe a valid scenario where an unrelated production order appears even though the filtered item is not part of that order.

The fix is targeted. It applies the same temporary Item + CopyFilters(Item) + IsEmpty() pattern that the report already uses later when printing WIP buffer rows, but it applies it before InsertCapLedgEntryCostShare() creates capacity cost rows. That matches the analogous use of Capacity Ledger Entry.Item No. filters in manufacturing costing/statistics code, and the added test covers two released production orders where only the filtered produced item should remain in the report.
</details>

<details>

*None.*
</details>

<details>

**Risk:** The changed area is a manufacturing cost report, so wrong filtering can hide or show financial cost details for WIP. The code change is narrow: it only skips non-matching capacity ledger entries before adding them to the temporary report buffer, and it does not change public API, table schema, events, or posting data. Build checks were successful where completed at review time; some unit-test jobs were still in progress.

**Necessity:** The change is needed because the WIP Inventory report could show unrelated production orders under an Item filter, which makes the report misleading. The scope is appropriate for the bug: the existing Item-filter logic already handled WIP buffer rows, and this PR extends the same rule to capacity-ledger rows that were previously bypassing it.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9079 round=1 by=alexei-dobriansky at=2026-07-06T09:01:58Z lastSha=546ba95551078d310a8d27c11bc42d2d8fdda325 suggestions=none

attilatoury (approved):

src/Layers/W1/BaseApp/Manufacturing/Reports/CostSharesBreakdown.Report.al
```diff
@@ -66,7 +66,17 @@ report 5848 "Cost Shares Breakdown"
                 DataItemTableView = sorting("Order Type", "Order No.", "Order Line No.") where("Order Type" = const(Production));
 
                 trigger OnAfterGetRecord()
+                var
+                    TempItem: Record Item temporary;
                 begin
+                    if Item.Get("Capacity Ledger Entry"."Item No.") then begin
+                        TempItem := Item;
+                        TempItem.Insert();
+                        TempItem.CopyFilters(Item);
```
JakovljevicDusan, line 75: Just to check, can this insert fail (duplicate)?

JakovljevicDusan (approved):

src/Layers/W1/BaseApp/Manufacturing/Reports/CostSharesBreakdown.Report.al
alexei-dobriansky: Don't think so, TempItem is local and clean on every OnAfterGetRecord.

PR #9081, opened 2026-07-03
Bug 637846: Item Age Composition '...before' quantity reflected only last entry

alexei-dobriansky:
Fixes [AB#637846](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/637846)

Report 5808 set the quantity buckets ``InvtQty[1..5]``/``TotalInvtQty`` per item ledger entry in ``CalcRemainingQty`` (clearing then setting one bucket) but output them once after the ILE loop, so quantities reflected only the last ILE. Value columns were correct because they accumulate with ``+=``. Accumulate the quantities into ``InvtQtyTotal[]`` and bind the quantity columns (total to ``RemainingQty``). Column names unchanged so layout bindings are preserved. Extends the report test.

### DIFF
--- a/src/Layers/W1/BaseApp/Inventory/Reports/ItemAgeCompositionValue.Report.al
+++ b/src/Layers/W1/BaseApp/Inventory/Reports/ItemAgeCompositionValue.Report.al
@@ -236,7 +236,10 @@ report 5808 "Item Age Composition - Value"
 
                         TotalInvtValueRTC += UnitCost * Abs(TotalInvtQty);
                         InvtValueRTC[i] += UnitCost * Abs(InvtQty[i]);
-                    end
+                    end;
+
+                    for j := 1 to 5 do
+                        InvtQtyTotal[j] += InvtQty[j];
                 end;
 
                 trigger OnPostDataItem()
@@ -256,6 +259,7 @@ report 5808 "Item Age Composition - Value"
                     for i := 1 to 5 do
                         InvtValue[i] := 0;
                     RemainingQty := 0;
+                    Clear(InvtQtyTotal);
                 end;
             }
             dataitem("Integer"; "Integer")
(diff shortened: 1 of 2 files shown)


alexei-dobriansky:

# **Recommendation: Accept**

<details>

This PR fixes Report 5808 so the quantity columns are accumulated across all open item ledger entries for the item, instead of showing the bucket values left by the last entry. The change matches the root cause in Bug 637846: CalcRemainingQty() clears and sets one per-entry bucket, while the dataset row is emitted after the item ledger entry loop.

The new InvtQtyTotal[1..5] is cleared per item and accumulated after each item ledger entry. TotalInvtQty is also bound to RemainingQty, which is already the per-item sum. This matches the obsolete quantity report pattern, keeps dataset column names stable for the Excel layout, and the added test would fail on the old last-entry behavior.
</details>

<details>

*None.*
</details>

<details>

**Risk:** The change is limited to Report 5808 dataset quantities and its test code. It does not change public signatures, events, or layout field names. The extra work is a fixed five-bucket addition per item ledger entry, so the performance risk is very small.

**Necessity:** The linked bug has clear repro steps and shows a real customer-visible Excel quantity error for items with more than one remaining item ledger entry in the same aging bucket. Without this change, the report can continue to show only the last entry in the ...before quantity column, while the obsolete quantity report shows the correct sum.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9081 round=1 by=alexei-dobriansky at=2026-07-06T09:01:35.0142190Z lastSha=6a6e3642ac70e0203b567aa0f67f937e24ca7495 suggestions=none

JakovljevicDusan (approved):

attilatoury (approved):

PR #9086, opened 2026-07-03
Bug 641085: Document Check factbox does not update live on Purchase Invoice/Cr.Memo/Return Order

alexei-dobriansky:
Fixes [AB#641085](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/641085)

The subform ``OnModifyRecordEvent`` subscribers for Purchase Invoice, Purchase Credit Memo and Purchase Return Order all called ``GlobalPurchaseOrderPage.RunBackgroundCheck()`` instead of their own document's global page, so line edits never refreshed the open page's Document Check factbox (only F5 did). Point each subscriber at the matching global page, matching the already-correct Purchase Order and Sales subform handlers (three one-token fixes).

### DIFF
--- a/src/Layers/W1/BaseApp/Utilities/DocumentErrorsMgt.Codeunit.al
+++ b/src/Layers/W1/BaseApp/Utilities/DocumentErrorsMgt.Codeunit.al
@@ -356,7 +356,7 @@ codeunit 9070 "Document Errors Mgt."
     begin
         if BackgroundValidationEnabled() then begin
             SetFullDocumentCheck(true);
-            GlobalPurchaseOrderPage.RunBackgroundCheck();
+            GlobalPurchaseInvoicePage.RunBackgroundCheck();
         end;
     end;
 
@@ -365,7 +365,7 @@ codeunit 9070 "Document Errors Mgt."
     begin
         if BackgroundValidationEnabled() then begin
             SetFullDocumentCheck(true);
-            GlobalPurchaseOrderPage.RunBackgroundCheck();
+            GlobalPurchaseCreditMemoPage.RunBackgroundCheck();
         end;
     end;
 
@@ -374,7 +374,7 @@ codeunit 9070 "Document Errors Mgt."
     begin
         if BackgroundValidationEnabled() then begin
             SetFullDocumentCheck(true);
-            GlobalPurchaseOrderPage.RunBackgroundCheck();
+            GlobalPurchaseReturnOrderPage.RunBackgroundCheck();
         end;
     end;
 


alexei-dobriansky:

<details>

This PR changes the purchase invoice, purchase credit memo, and purchase return order subform modify subscribers in `Document Errors Mgt.` to run the background check on the matching open purchase page instead of `GlobalPurchaseOrderPage`. This matches the Sales subform pattern and the existing purchase header/factbox routing, and it targets the stale Document Check FactBox problem described in Bug 641085.

The code change itself is the right shape: after a line modify, `SetFullDocumentCheck(true)` still runs and the call now reaches the factbox part on the page the user is editing. I did not see a side effect on the already-correct purchase order path, and no public signatures or events change. The remaining issue is validation: the exact live-refresh path that failed is not covered by a regression test.
</details>

<details>

**S1** - *Add live-refresh regression tests*
Please add tests in `Doc. Backgr. Error Handling` for the changed purchase invoice, credit memo, and return order line-edit paths. Existing tests already cover the factbox on page open, but this bug is about the same open page refreshing after a line change. Without that test, the one-token routing errors can come back without being caught.
</details>

<details>

**Risk:** Low runtime risk because this only changes page variables used in three UI event subscribers and keeps the same background validation call. The affected scenario is visible to users who enable Document Check on purchase invoices, credit memos, and return orders. No API, schema, event signature, or data persistence behavior changes.

**Necessity:** The ADO bug has a clear repro (`[AI-REPRO]` score 9/10) and real customer impact, because stale messages make Document Check unreliable and force an F5 refresh. The scope is right for the root cause, but a regression test is needed before merge because existing coverage did not catch this class of mistake.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9086 round=1 by=alexei-dobriansky at=2026-07-06T09:00:35Z lastSha=99f99879e7434c010b6c1c6856876e88d8dd4ad2 suggestions=S1

alexei-dobriansky:
Re **S1** (add live-refresh regression tests): acknowledged. The product change is a 3-token mirror of the already-correct Purchase Order / Sales subform handlers, and the existing DocBackgrErrorHandling tests cover the factbox. A true live-refresh test requires driving the background-validation task through the open TestPage after a subpart line edit, which isn't reliably reproducible headlessly. Given the low risk and the correctness of the fix, deferring the extra test rather than adding a flaky one.

attilatoury (approved):

JakovljevicDusan (approved):

PR #9093, opened 2026-07-04
Add shared CompanyMetadata data source for report layouts

v-mortenr:
Adds a shared **CompanyMetadata** data source so report layouts can bind one shared company block instead of each report hand-coding Company Information fields.

### What
- `ReportManagement` subscribes to the platform `GetCompanyMetadata` business event and populates the shared company block (name, formatted address repeater, phone/fax/email/home page, bank details, VAT/registration/giro, logo) from `Company Information`.
- New `Company Metadata Builder` codeunit (`Access = Internal`): typed setters centralize the wire keys so the subscriber never hand-writes raw key strings; includes the address-line repeater and the paired field captions.
- `CompanyDisplayName` is sourced from the Company record (`CompanyProperty.DisplayName()`, falling back to `CompanyName()`) to match how the platform builds `ReportRequest` — not Company Information "Name 2".
- Empty-safe: absent Company Information fields emit present-but-empty values.
- Adds `Tests-Report` coverage for values, captions, the display-name source, empty-safe behavior, and the address repeater.

### Why draft
This consumes the `GetCompanyMetadata` business event on `Reporting Triggers`. The AL will compile in CI once BCApps uptakes a BCPlatform build that includes that event (current pin `29.0.51074.0` predates it). Kept as a draft until the platform uptake lands so the AL can be reviewed in the meantime.

### DIFF
--- a/src/Layers/W1/BaseApp/Foundation/Reporting/CompanyMetadataBuilder.Codeunit.al
+++ b/src/Layers/W1/BaseApp/Foundation/Reporting/CompanyMetadataBuilder.Codeunit.al
@@ -1,0 +1,189 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+namespace Microsoft.Foundation.Reporting;
+
+/// <summary>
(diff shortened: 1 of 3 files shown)


v-mortenr:
Superseded by #9094 — branch renamed to features/640703/company-metadata-subscriber to match work item AB#640703.

PR #9116, opened 2026-07-06
[28.x] Subscription Billing] Infinite loop in CalculateBillingPeriod when HarmonizeNextBillingTo caps at stale "Next Billing To" date

sanjmaurya:
[Bug 637042](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/637042): [ALL-E][Subscription Billing] Infinite loop in CalculateBillingPeriod when HarmonizeNextBillingTo caps at stale "Next Billing To" date

Fixes [AB#637042](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/637042)

**Issue:-** CalculateBillingPeriod in BillingProposal.Codeunit.al enters an infinite loop (hang / no response) when generating a billing proposal for a harmonized customer subscription contract whose "Next Billing To" date is stale (older than the current billing period being calculated).
**Cause :**- Root Cause
The while loop advances the billing period end date by repeatedly calling CalculateNextBillingToDateForServiceCommitment:

For a harmonized contract, CalculateNextBillingToDateForServiceCommitment calls HarmonizeNextBillingTo, which caps the computed date at CustomerContract."Next Billing To":

When the contract's "Next Billing To" is a stale date earlier than BillingDate, every iteration gets capped back to that same stale value. So BillingPeriodEnd never advances past it, the loop condition BillingPeriodEnd < BillingDate stays true forever, and the loop never terminates.
**Solution**
Add a progress guard that breaks the loop as soon as BillingPeriodEnd stops advancing (i.e., is capped to the same/earlier date). A PreviousBillingPeriodEnd tracker is introduced, and the loop only continues while BillingPeriodEnd > PreviousBillingPeriodEnd:

### DIFF
--- a/src/Apps/W1/Subscription Billing/App/Billing/Codeunits/BillingProposal.Codeunit.al
+++ b/src/Apps/W1/Subscription Billing/App/Billing/Codeunits/BillingProposal.Codeunit.al
@@ -483,6 +483,7 @@ codeunit 8062 "Billing Proposal"
     local procedure CalculateBillingPeriod(ServiceCommitment: Record "Subscription Line"; BillingDate: Date; BillToDate: Date; var BillingPeriodStart: Date; var BillingPeriodEnd: Date)
     var
         UsageDataBilling: Record "Usage Data Billing";
+        PreviousBillingPeriodEnd: Date;
     begin
         BillingPeriodEnd := 0D;
         BillingPeriodStart := ServiceCommitment."Next Billing Date";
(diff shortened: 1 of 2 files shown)


AndersLarsenMicrosoft (approved):

v-ankitgoyal (approved):

AndersLarsenMicrosoft (approved):

aholstrup1:
Heads up, these build failures aren't from your code changes. The special character (a stray `]`) in the branch name trips up our pipeline. Recreating the branch without the bracket and repointing the PR should fix it. Sticking to letters, numbers, `/`, `-`, and `_` in branch names avoids this going forward.

PR #9124, opened 2026-07-06
[E-Document] [Payables Agent] Users can personalize-in the Name column on the draft page

Explorer986:
## What & why

This change adds a read-only Name column to the E-Document purchase draft lines subform (page 6183 E-Doc. Purchase Draft Subform). The column shows the name of the matched entity such as Item, G/L Account, or Resource, so reviewers can confirm a line was matched to the right record. It stays hidden by default and users personalize it into view. The value reflects the matched record rather than the extracted invoice description and cannot be edited.

## Linked work

Fixes [AB#641033](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/641033)

## How I validated this

What I tested and the outcome

- Built E-Document Core locally with no errors and no new analyzer warnings.
- Published to a local sandbox and opened the Purchase Document Draft page.
- Personalized the Name column into the Lines part and confirmed it shows the matched entity name, stays read-only, and updates when Type or No. change.
- Confirmed the column is blank for lines with no match and that it reflects the matched record rather than the edited Description.
- No automated tests added yet. The change is a small refactor of the existing SetDescriptionFromLineTypeNo behavior. I can add a codeunit test asserting GetMatchedEntityName returns the right value per line type if reviewers prefer.

## Risk & compatibility

- Low risk. No table schema change and no new stored fields.
- The Name value binds to a page variable computed in OnAfterGetRecord, so there is no data upgrade impact.
- Existing description defaulting behaves the same because it now calls the shared GetMatchedEntityName helper.
- The new OnAfterGetMatchedEntityName integration event is additive for connector apps.

### DIFF
--- a/src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/EDocumentPurchaseLine.Table.al
+++ b/src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/EDocumentPurchaseLine.Table.al
@@ -287,23 +287,41 @@ table 6101 "E-Document Purchase Line"
 
         case Rec."[BC] Purchase Line Type" of
             "Purchase Line Type"::Item:
-                if Item.Get(Rec."[BC] Purchase Type No.") then
-                    EntityName := Item.Description;
+                begin
+                    Item.SetLoadFields(Description);
+                    if Item.Get(Rec."[BC] Purchase Type No.") then
+                        EntityName := Item.Description;
+                end;
             "Purchase Line Type"::"G/L Account":
-                if GLAccount.Get(Rec."[BC] Purchase Type No.") then
-                    EntityName := GLAccount.Name;
+                begin
+                    GLAccount.SetLoadFields(Name);
+                    if GLAccount.Get(Rec."[BC] Purchase Type No.") then
+                        EntityName := GLAccount.Name;
+                end;
             "Purchase Line Type"::"Allocation Account":
-                if AllocationAccount.Get(Rec."[BC] Purchase Type No.") then
-                    EntityName := AllocationAccount.Name;
+                begin
+                    AllocationAccount.SetLoadFields(Name);
+                    if AllocationAccount.Get(Rec."[BC] Purchase Type No.") then
+                        EntityName := AllocationAccount.Name;
+                end;
             "Purchase Line Type"::"Fixed Asset":
-                if FixedAsset.Get(Rec."[BC] Purchase Type No.") then
-                    EntityName := FixedAsset.Description;
+                begin
+                    FixedAsset.SetLoadFields(Description);
+                    if FixedAsset.Get(Rec."[BC] Purchase Type No.") then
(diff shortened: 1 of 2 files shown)


src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/EDocumentPurchaseLine.Table.al
```diff
@@ -271,29 +282,36 @@ table 6101 "E-Document Purchase Line"
...
         ItemCharge: Record "Item Charge";
     begin
-        if Rec.Description <> '' then
-            exit;
+        if Rec."[BC] Purchase Type No." = '' then
+            exit('');
 
         case Rec."[BC] Purchase Line Type" of
             "Purchase Line Type"::Item:
                 if Item.Get(Rec."[BC] Purchase Type No.") then
```
ventselartur: Add item.SetLoadFields() call to only get necessary field. Do the same for other records.

ventselartur (changes requested):

ventselartur (approved):

AndersLarsenMicrosoft (approved):

Groenbech96 (approved):

alexei-dobriansky:

<details>

This PR adds a hidden, read-only `Name` column to page 6183, `E-Doc. Purchase Draft Subform`, and calculates it from the matched purchase line type and number. The helper keeps the existing Description defaulting behavior by reusing the same lookup logic, and the new page field is available for personalization without adding table fields or stored data.

The change targets the right Payables Agent draft lines page and uses the right page mechanism for a column that is hidden by default. The lookup covers the purchase line types that the old description defaulting handled, and the additive integration event is raised from the new helper inside the same app. I did not find a BaseApp dependency.
</details>

<details>

**S1** - *Clear the new page variable*
The page already clears `LineAmount` in `OnNewRecord`, but it does not clear `MatchedEntityName`. If the Name column is personalized in, a blank new row can show a stale name from the previous line. Clear `MatchedEntityName` there too.
</details>

<details>

**Risk:** The regression surface is limited to the E-Document purchase draft lines page and the refactored lookup used when `[BC] Purchase Type No.` is validated. The new event is additive, and there is no schema or upgrade impact. The main risk is UI confusion if a computed page variable is stale on a new row; the extra per-row lookup is small and uses `SetLoadFields`.

**Necessity:** The linked work item is a Slice, not a Bug, and there is no `[AI-REPRO]` comment. The feature request is still clear: users need to see the actual matched Item, G/L Account, Resource, or similar record name instead of only the invoice description. The scope is narrow and fits the requested hidden, read-only personalized column.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9124 round=1 by=alexei-dobriansky at=2026-07-08 lastSha=2748e5f062d5acd662e06a163916614f41ccd1c3 suggestions=S1

PR #9168, opened 2026-07-07
[QM] Fix internal put-away from Quality Inspection sourcing wrong bin

6CRIPT:
**What**
Fixes the Create Internal Put-away disposition from a Quality Inspection, which failed or produced misleading results in bin-mandatory (directed put-away & pick) locations.

**Why it failed**
The disposition resolved the source location/bin from the originating document via field-name conventions (GetFromDetailsFromInspectionSource) instead of the item's actual Bin Content. For non-item-tracked items this returned a stale bin — the receive bin the goods first landed in, or a bin that no longer holds the item — leading to:

Receive must be equal to 'No' in Bin Type: Code=RECEIVE (platform error),
a 0 PCS ... could not be created notification, and
From Bin Code ... cannot be found in the related table (Bin Content).

**Changes**
Resolve the source bin from real Bin Content for bin-mandatory locations (TryResolveBinsFromBinContent): considers only bins with positive quantity and excludes receive and adjustment bins. This is the non-item-tracked counterpart to GetCurrentLocationOfTrackedInventory. Falls back to the previous naming-convention logic only when the location is not bin-mandatory or the item has no current stock. It reports success even when the bin was already resolved by a previous source record, so the receive-bin fallback is never added alongside the real bin.
Validate all resolved source bins for receive bins up front in Qlty. Disp. Internal Put-away, before any document is created, so we never emit a "created" notification for a put-away that is then rolled back. If the item is still in the receive bin the user gets the actionable "put the items away first" message instead of the cryptic platform error.
Show the intended quantity instead of 0 in the document-creation-failed notification (GetDisplayQuantityForFailure), falling back to the failed/passed/sample/source quantity when the instruction quantity has not been resolved yet.
Extracted IsReceiveBin as a reusable helper (shared by the guard and the bin resolution).

**Extra**
Fixes [AB#641562](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/641562)

Needs a backport to 28.x to fix the original Work Item.

### DIFF
--- a/src/Apps/W1/Quality Management/app/src/Integration/Inventory/QltyInventoryAvailability.Codeunit.al
+++ b/src/Apps/W1/Quality Management/app/src/Integration/Inventory/QltyInventoryAvailability.Codeunit.al
@@ -24,6 +24,7 @@ codeunit 20445 "Qlty. Inventory Availability"
         NoSamplesToMoveErr: Label 'No samples meet the condition specified.', Locked = true;
         SerialQuantityGreaterThanOneErr: Label '%1 (%2) cannot be greater than 1 when New Serial No. is requested.', Comment = '%1=quantity behavior, %2=quantity';
         ZeroQuantityErr: Label 'Unable to use the disposition %1 on the inspection %2 for the item %3 because the quantity is zero.', Comment = '%1=the inspection, %2=the inspection, %3=the item';
+        InsufficientInventoryToAllocateErr: Label 'Unable to move the entire requested quantity for inspection %1 for item %2. There is not enough inventory available in eligible bins (short by %3). Put the remaining inventory away or reduce the quantity, and then try again.', Comment = '%1=the inspection, %2=the item, %3=the missing quantity';
         SupplyFromLocationCodeNameLbl: Label 'Supply-from Location Code', Locked = true;
         FromLocationCodeNameLbl: Label 'From Location Code', Locked = true;
         LocationCodeNameLbl: Label 'Location Code', Locked = true;
@@ -526,6 +527,9 @@ codeunit 20445 "Qlty. Inventory Availability"
                     TempQuantityQltyDispositionBuffer.Insert(false);
                 end;
             until TempExistingInventoryBinContent.Next() = 0;
+
+            if AllocateAcrossBins and (RemainingQuantityToAllocate > 0) then
+                Error(InsufficientInventoryToAllocateErr, QltyInspectionHeader."No.", QltyInspectionHeader."Source Item No.", RemainingQuantityToAllocate);
         end;
 
         OnAfterPopulateBinContentBuffer(QltyInspectionHeader, TempInstructionQltyDispositionBuffer, TempQuantityQltyDispositionBuffer, TempExistingInventoryBinContent);
--- a/src/Apps/W1/Quality Management/app/src/Integration/Inventory/QltyInventoryAvailability.Codeunit.al
+++ b/src/Apps/W1/Quality Management/app/src/Integration/Inventory/QltyInventoryAvailability.Codeunit.al
@@ -163,11 +163,6 @@ codeunit 20445 "Qlty. Inventory Availability"
                 LocationCode := QltyInspectionHeader."Location Code";
                 GetFromLocationAndBinBasedOnNamingConventions(RecordRefToSearch, LocationCode, BinCode, QuantityBaseValue);
                 if LocationCode <> '' then
-                    // In bin mandatory locations (including directed put-away and pick) the bin derived from the source
-                    // document naming conventions can be stale - for example the receive bin the goods first landed in,
-                    // or a bin that no longer holds the item. Resolve the actual current bin(s) from Bin Content instead,
-                    // so the put-away is created from a bin that really contains the inventory. Only fall back to the
-                    // naming convention bin when the location is not bin mandatory or the item is not currently in stock.
                     if not TryResolveBinsFromBinContent(QltyInspectionHeader, LocationCode, TempToMoveBinContent) then begin
                         TempToMoveBinContent.Reset();
                         TempToMoveBinContent.SetRange("Location Code", LocationCode);
@@ -193,15 +188,6 @@ codeunit 20445 "Qlty. Inventory Availability"
         end;
     end;
 
-    /// <summary>
-    /// Resolves the actual bin(s) that currently hold the inspection's item at a bin mandatory location by reading Bin Content.
-    /// Only bins with positive quantity are considered, and receive and adjustment bins are excluded because inventory cannot be
-    /// moved from them with an internal put-away. This is the non item tracked counterpart to GetCurrentLocationOfTrackedInventory.
-    /// </summary>
-    /// <param name="QltyInspectionHeader">The inspection providing the item, variant and any tracking to look up.</param>
-    /// <param name="LocationCode">The location to resolve the bins at.</param>
-    /// <param name="TempToMoveBinContent">The temporary bin content buffer that resolved bins are added to.</param>
-    /// <returns>True when at least one qualifying bin was added, false otherwise so the caller can fall back to other logic.</returns>
     local procedure TryResolveBinsFromBinContent(QltyInspectionHeader: Record "Qlty. Inspection Header"; LocationCode: Code[10]; var TempToMoveBinContent: Record "Bin Content" temporary): Boolean
     var
         Location: Record Location;
@@ -231,9 +217,6 @@ codeunit 20445 "Qlty. Inventory Availability"
             repeat
                 if BinContent."Quantity (Base)" > 0 then
                     if not IsReceiveBin(BinContent."Location Code", BinContent."Bin Code") then begin
-                        // A valid non-receive bin with stock exists for this location. Report success even when the bin was
-                        // already added by a previous source record, otherwise the caller would fall back to the naming
-                        // convention bin (which can be the receive bin) and add it to the buffer alongside the real bin.
                         Added := true;
                         if not TempToMoveBinContent.Get(BinContent."Location Code", BinContent."Bin Code", BinContent."Item No.", BinContent."Variant Code", BinContent."Unit of Measure Code") then begin
                             TempToMoveBinContent.Init();
--- a/src/Apps/W1/Quality Management/app/src/Integration/Inventory/QltyInventoryAvailability.Codeunit.al
+++ b/src/Apps/W1/Quality Management/app/src/Integration/Inventory/QltyInventoryAvailability.Codeunit.al
@@ -164,7 +164,7 @@ codeunit 20445 "Qlty. Inventory Availability"
                 LocationCode := QltyInspectionHeader."Location Code";
                 GetFromLocationAndBinBasedOnNamingConventions(RecordRefToSearch, LocationCode, BinCode, QuantityBaseValue);
                 if LocationCode <> '' then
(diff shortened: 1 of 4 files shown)


alexei-dobriansky:

<details>

This PR changes Quality Management internal put-away creation so non-tracked inventory at bin-mandatory locations is sourced from current Bin Content instead of the source document bin. It also validates receive bins before document creation and improves the failed-notification quantity.

The direction matches the linked bug: after a purchase receipt is put away, the receive bin is stale and the internal put-away must use the bin that now holds the inventory. The receive-bin precheck is also placed before document creation, which avoids the old created-then-rolled-back notification. However, the new Bin Content resolver can return more than one bin, and the existing quantity-buffer logic applies the full requested quantity to each returned bin. That makes the fix unsafe when the item exists in multiple valid bins.
</details>

<details>

**S1** - *Allocates the requested quantity once*
TryResolveBinsFromBinContent can add several bins. PopulateQuantityBuffer then gives the full failed or specific quantity to each bin, so one inspection can create too much put-away quantity. Allocate the requested quantity across the bins, or choose one bin that can cover it.

**S2** - *Adds regression coverage for this flow*
This change has no test, but the area is sensitive and existing Quality Management disposition tests can create internal put-aways. Add a test for the WHITE receipt-put-away-fail scenario and assert the From Bin Code is not RECEIVE and the total quantity is the failed quantity.
</details>

<details>

**Risk:** This touches warehouse source-bin selection for internal put-aways. A wrong result can create a put-away from the wrong bin or for too much quantity. No public API signature or BaseApp event publisher dependency was introduced, but the behavior change is in sensitive warehouse flow logic.

**Necessity:** The linked Bug is clear and important. The reported WHITE scenario cannot create the internal put-away for the failed quantity because the flow uses a receive or stale bin after the receipt put-away. A fix is needed, but it must preserve the requested quantity across multi-bin inventory.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9168 round=1 by=alexei-dobriansky sha=7da827c00027a76cc10daf786f4eae50a81eb9d9

src/Apps/W1/Quality Management/app/src/Integration/Inventory/QltyInventoryAvailability.Codeunit.al
```diff
@@ -162,19 +162,25 @@ codeunit 20445 "Qlty. Inventory Availability"
...
-                    if BinCode <> '' then
-                        TempToMoveBinContent.SetRange("Bin Code", BinCode);
-                    if not TempToMoveBinContent.FindFirst() then begin
-                        TempToMoveBinContent.Init();
-                        TempToMoveBinContent."Location Code" := LocationCode;
-                        TempToMoveBinContent."Bin Code" := BinCode;
-                        TempToMoveBinContent."Min. Qty." := QuantityBaseValue;
-                        TempToMoveBinContent.Insert(false);
+                if LocationCode <> '' then
+                    // In bin mandatory locations (including directed put-away and pick) the bin derived from the source
```
alexei-dobriansky: don't leave so many comments, simply put everything to PR description.

alexei-dobriansky (changes requested):
Review the suggestions

src/Apps/W1/Quality Management/app/src/Integration/Inventory/QltyInventoryAvailability.Codeunit.al
```diff
@@ -417,6 +463,18 @@ codeunit 20445 "Qlty. Inventory Availability"
                 TempExistingInventoryBinContent.FindSet();
             end;
 
+            AllocateAcrossBins := MultipleBins and (TempInstructionQltyDispositionBuffer."Quantity Behavior" in [
```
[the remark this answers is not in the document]
6CRIPT: There is indeed a final check now. And this is duplicated on top of that.

alexei-dobriansky:

# **Recommendation: Accept**

<details>

Since round 1, this PR changed `PopulateQuantityBuffer` so Specific, Sample, Failed, and Passed quantities are allocated across multiple eligible bins instead of copied in full to each bin. It also now errors if the eligible bins cannot cover the full requested quantity, and it adds a regression test for the WHITE failed-quantity internal put-away flow.

The rework addresses the unsafe multi-bin case from round 1. Each bin now receives at most its available quantity, the remaining quantity is reduced after each inserted buffer line, and the final shortfall check prevents a silent partial put-away. The new test covers the linked bug shape by creating a directed warehouse receipt/put-away inspection, using Failed Quantity, asserting no line is sourced from a receive bin, and checking the total put-away quantity equals the failed quantity.
</details>

<details>

| S1 | Allocates the requested quantity once | Addressed | No direct reply; the new allocation path splits the requested quantity across bins and the final commit adds a shortfall error. |
| S2 | Adds regression coverage for this flow | Addressed | No direct reply; `InternalPutaway_FailedQuantity_SourcedFromStorageBinNotReceive` now covers the WHITE failed-quantity scenario. |
</details>

<details>

</details>

<details>

**Risk:** This is still sensitive warehouse logic because a wrong source bin or quantity can create incorrect internal put-away lines. The updated allocation logic narrows that risk, and no public API signature or BaseApp event dependency appears to be introduced.

**Necessity:** The linked Bug is clear and important. After a purchase receipt is put away in WHITE, the internal put-away for failed inspection quantity must use current storage bin content and must move exactly the failed quantity.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9168 round=2 by=alexei-dobriansky sha=e65b7954642418b2df99481c4336acc0d3e8e66c suggestions=S1:addressed,S2:addressed parentRound=1

alexei-dobriansky (dismissed):

src/Apps/W1/Quality Management/app/src/Integration/Inventory/QltyInventoryAvailability.Codeunit.al
```diff
@@ -187,6 +189,48 @@ codeunit 20445 "Qlty. Inventory Availability"
...
+        BinContent.SetAutoCalcFields("Quantity (Base)");
+        if BinContent.FindSet() then
+            repeat
+                if not IsReceiveBin(BinContent."Location Code", BinContent."Bin Code") then begin
+                    Added := true;
+                    if not TempToMoveBinContent.Get(BinContent."Location Code", BinContent."Bin Code", BinContent."Item No.", BinContent."Variant Code", BinContent."Unit of Measure Code") then begin
+                        TempToMoveBinContent.Init();
+                        TempToMoveBinContent := BinContent;
+                        TempToMoveBinContent."Min. Qty." := BinContent."Quantity (Base)";
+                        TempToMoveBinContent.Insert(false);
```
JakovljevicDusan, line 226: Is it possible to have duplicate of Bin Content, so TempToMoveBinContent.Insert(false); fails?

src/Apps/W1/Quality Management/app/src/Integration/Inventory/QltyInventoryAvailability.Codeunit.al
```diff
@@ -187,6 +189,48 @@ codeunit 20445 "Qlty. Inventory Availability"
         end;
     end;
 
+    local procedure TryResolveBinsFromBinContent(QltyInspectionHeader: Record "Qlty. Inspection Header"; LocationCode: Code[10]; var TempToMoveBinContent: Record "Bin Content" temporary): Boolean
```
JakovljevicDusan: Maybe remove Try, since it is not Try procedure.

src/Apps/W1/Quality Management/app/src/Integration/Inventory/QltyInventoryAvailability.Codeunit.al
6CRIPT: I don't think so, because just above we have
 `if not TempToMoveBinContent.Get(BinContent."Location Code", BinContent."Bin Code", BinContent."Item No.", BinContent."Variant Code", BinContent."Unit of Measure Code") then begin`
those 5 parameters are the exact PK so there can't be 2 exactly the same and even if we call the function several time, if it was already found before, it just skips the insert. In addition, the same pattern is present in `ProcessBinMandatoryLocations ` in the same file

alexei-dobriansky:

# **Recommendation: Accept**

<details>

Since round 2, the only AL change is renaming `TryResolveBinsFromBinContent` to `ResolveBinsFromBinContent` and updating its call site.

This removes the misleading `Try` prefix but does not change behavior. The bin-sourcing paths remain intact: item-tracked inventory still uses `GetCurrentLocationOfTrackedInventory`; non-item-tracked inventory still resolves positive Bin Content for bin-mandatory locations while excluding receive and adjustment bins; the fallback to source-record naming conventions still happens only when no eligible Bin Content is found. The multi-bin allocation and shortfall error from round 2 are unchanged.
</details>

<details>

| S1 | Allocates the requested quantity once | Addressed | Still addressed; the round-3 commit did not change the allocation logic. |
| S2 | Adds regression coverage for this flow | Addressed | Still addressed; the regression test remains unchanged. |
</details>

<details>

*None - commit `c9bfa1a2` only removes the misleading `Try` prefix from the local Bin Content resolver name.*
</details>

<details>

**Risk:** The round-3 change is a local rename only, so it does not add new warehouse behavior or compatibility risk. The remaining risk is the same as round 2: this is sensitive source-bin and quantity logic for Quality Management internal put-aways.

**Necessity:** The linked work item is a Bug. The fix is still needed because failed inspection quantity in a directed warehouse location must be sourced from the current storage bin, not from the stale receive bin.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9168 round=3 by=alexei-dobriansky at=2026-07-13T11:04:41Z lastSha=c9bfa1a2962fc9c16a2e4425d97530b7d8e6eaa2 suggestions=S1:addressed,S2:addressed parentRound=2

qasimikram (approved):
**Approve with suggestions**

The fix is well-reasoned and internally consistent. `ResolveBinsFromBinContent` correctly mirrors the tracked-inventory counterpart (positive qty filtered at the query, receive/adjustment bins excluded, dedupe on the true PK). The receive-bin guard runs before any document is created (loop, then `FindSet()` re-anchor), so no misleading "created" notification precedes a rollback. The cross-bin allocation caps each bin at its `Min. Qty.`, guards the `<= 0` case, and errors on remaining shortfall — fixing the previous per-bin over-allocation. Earlier review findings (zero-qty allocation, silent shortfall, filter pushdown, `Try` naming) are all addressed, and CI is green across locales/buckets.

Non-blocking suggestions:

**S1 — Test coverage gaps (Medium).** Only one new test is added, and it appears to exercise a single storage bin. Consider adding coverage for: (a) the receive-bin actionable guard when stock is still only in the receive bin, (b) the `InsufficientInventoryToAllocateErr` shortfall path, (c) a genuine allocation split across **multiple** eligible bins (asserting the per-line split, not just the total), and (d) the `GetDisplayQuantityForFailure` fallback quantity in the failure notification. These would lock the fix against regression.

**S2 — `GetDisplayQuantityForFailure` for Specific Quantity (Low).** The `else` branch returns `Source Quantity (Base)`, so a "Specific Quantity" behavior with an unresolved instruction qty would display the source qty rather than the requested specific qty. Display-only and low impact, but worth an explicit case or a comment.

Risk is low and the change is scoped to QM dispositions. Reminder from the description: this needs the 28.x backport to fix the original work item.

JakovljevicDusan (approved):

PR #9180, opened 2026-07-07
Copy record links to posted Phys. Invt. Order and Recording headers

Franco111000:
## What & why

Record links (the standard Notes and Links factboxes) attached to a Phys. Invt. Order Header or a Phys. Invt. Recording Header were lost on posting: codeunit 5884 "Phys. Invt. Order-Post" never called `"Record Link Management".CopyLinks`, unlike the sales, purchase, and warehouse posting flows, which all carry record links over to their posted history tables.

This change copies the links in the two places where the posted headers are inserted: `RecordLinkManagement.CopyLinks(PhysInvtOrderHeader, PstdPhysInvtOrderHdr)` after the insert in `InsertPostedHeader()`, and `RecordLinkManagement.CopyLinks(PhysInvtRecordHeader, PstdPhysInvtRecordHdr)` after each recording header insert in `InsertPostedRecordings()`. Both calls sit inside the existing `if not IsHandled then` blocks, so subscribers that replace the standard insert keep full control. A regression test posts an order with a note on both the order and the recording and verifies the links arrive on the posted documents.

This is a re-submission of microsoft/BusinessCentralApps#1900 (pilot repository, retired). The fix and the test were reviewed there; the regression test was added at the reviewer's request. The pilot PR was closed unmerged by the repository retirement.

## Linked work

Fixes #9099

Original pilot review: microsoft/BusinessCentralApps#1900 (fixing microsoft/BusinessCentralApps#1610).

## How I validated this

- [x] I read the full diff and it contains only changes I intended.
- [ ] I built the affected app(s) locally with no new analyzer warnings.
- [ ] I ran the change in Business Central and confirmed it behaves as expected.
- [x] I added or updated tests for the new behavior, or explained below why none are needed.

**What I tested and the outcome**

- Verified against current main that neither `InsertPostedHeader()` nor `InsertPostedRecordings()` references `CopyLinks` or `Record Link Management` anywhere in the codeunit, so the defect is still present.
- New test `CopyLinksOnPostingPhysInvtOrder` in PhysInvtItemTracking.Codeunit.al mirrors the existing `DropShipmentCopyLinksOnPostingReceiptShipment` pattern: it attaches a record link to both the Phys. Invt. Order Header and the Phys. Invt. Recording Header, posts the order, and asserts both posted headers carry the link. The test was reviewed in the pilot PR.
- Verified codeunit 5884 has no country layer copies (W1 only), so no localization propagation is needed.
[...]

Issue #9097, opened 2026-07-05
[Bug]: Pay-to Name 2 is editable on other conditions than all other Pay-to fields


Franco111000:
### Describe the issue

This is the purchase-side counterpart of #8312.

On the Purchase Quote, Purchase Order, and Purchase Invoice pages, the field "Pay-to Name 2" is only editable when the Pay-to option is "Another Vendor":

```al
Editable = PayToOptions = PayToOptions::"Another Vendor";
Enabled = PayToOptions = PayToOptions::"Another Vendor";
```

Every other field in the pay-to address group (Address, Address 2, City, County, Post Code, Country/Region Code, Contact No., Contact) uses:

```al
Editable = (PayToOptions = PayToOptions::"Custom Address") or (Rec."Buy-from Vendor No." <> Rec."Pay-to Vendor No.");
```

So with "Custom Address" selected, the whole pay-to address group can be filled in except "Pay-to Name 2". Like its sales counterpart, the field appears to have been missed when the PayToOptions editability logic was applied to the pay-to address group. "Pay-to Name 2" on the Purchase Header is a plain text field with no validation logic and no table relation.

The Blanket Purchase Order, Purchase Credit Memo, and Purchase Return Order pages have no PayToOptions-based conditions on this field and are not part of this issue.

### Expected behavior

When selecting "Custom Address" in the Pay-to section of a purchase document, the "Pay-to Name 2" field should be editable under the same conditions as all other pay-to address fields.

### Steps to reproduce

1. Create a purchase document (e.g. Purchase Order)
2. Select "Pay-to" -> "Custom Address"
3. Try to enter a different Pay-to Name 2

### Additional context

The sales side is being fixed in #9096. The same correction can be applied here: give "Pay-to Name 2" the same Editable and Enabled expression as its sibling fields on the three affected pages, propagate to the localization layers that carry copies of these pages, and extend the existing CustomAddressPayToOption tests.

### I will provide a fix for a bug

- [x] I will provide a fix for a bug

PR #9181, opened 2026-07-07
Filter the source record before FindFirst in Prices Overview caption

### LINKED ISSUE #9101: [Bug]: Prices Overview data caption shows the first record's name instead of the filtered record
### Describe the issue

Page 7024 "Prices Overview" builds its data caption in `GetFilterDescription()` by assigning the source number to a record field and then calling `FindFirst()` without any filter:

```al
PriceSource."Source Type"::Customer:
    begin
        SalesSrcTableName := ObjTranslation.TranslateObject(ObjTranslation."Object Type"::Table, 18);
        Cust."No." := CopyStr(SourceNoFilter, 1, MaxStrLen(Cust."No."));
        if Cust.FindFirst() then
            Description := Cust.Name;
    end;
```

Assigning to `Cust."No."` does not filter the record, so `FindFirst()` returns the first customer in the table and the caption shows an unrelated customer's name next to the correct number. The Customer Price Group and Campaign branches have the same defect.

Verified still present in current main: `src/Layers/W1/BaseApp/Pricing/PriceList/PricesOverview.Page.al`, `GetFilterDescription()`. The APAC layer copy of the page (`src/Layers/APAC/BaseApp/Pricing/PriceList/PricesOverview.Page.al`) contains the same pattern.

The fix is to replace the field assignments with `SetFilter`/`SetRange` on the key field before `FindFirst()` in all three branches.

### Expected behavior

The page caption shows the name of the record actually selected in the Source No. filter.

### Steps to reproduce

1. Create a page action in AL that runs Prices Overview modally.
2. Run the action in Business Central.
3. Choose Source Type "Customer" and as Source No. any customer except the first one in the table.
4. The caption shows the chosen customer's number but the first customer's name.

### Additional context

This is a copy of microsoft/BusinessCentralApps#1371 (labeled Approved/SCM in the contribution pilot). It was being fixed in microsoft/BusinessCentralApps#1911, which was reviewed with regression tests added, but was closed unmerged with a note about test failures when the repository was retired. The likely cause of those failures has been identified (the tests opened the page in view mode on an empty Price List Line table, where the platform suppresses the DataCaptionExpression, see microsoft/AL#6388) and will be corrected in the re-submitted fix.

Franco111000:
## What & why

Page 7024 "Prices Overview" builds its data caption in `GetFilterDescription()` by assigning the source number to a record field and then calling `FindFirst()` without a filter. Assignment is not filtering, so `FindFirst()` returns the first record of the table and the caption shows an unrelated record's name next to the correct number. The Customer, Customer Price Group, and Campaign branches all have the defect.

This change replaces the assignments with `SetFilter` on the key field in all three branches, so the caption looks up the record actually selected in the Assign-to Filter. The APAC layer carries a full copy of the page with identical code and receives the same three-line change, following the practice of propagating W1 fixes to layer copies.

Three regression tests cover the three branches: each creates a record whose name is not derived from its number, opens Prices Overview, sets the type and number filters, and asserts the caption contains the record's name. The tests open the page in edit mode deliberately: the platform suppresses `DataCaptionExpression` on a page opened in view mode with no records (microsoft/AL#6388), and Prices Overview starts on an empty Price List Line table in the test setup. This is also the most plausible cause of the internal test failures reported on the pilot PR, which used view mode.

This is a re-submission of microsoft/BusinessCentralApps#1911 (pilot repository, retired). The fix was reviewed there and closed unmerged with a note about test failures when the repository was retired; the view-mode issue described above is corrected in this version.

## Linked work

Fixes #9101

Original pilot review: microsoft/BusinessCentralApps#1911 (fixing microsoft/BusinessCentralApps#1371).

## How I validated this

- Verified against current main that the assignment plus unfiltered `FindFirst()` pattern is still present in `GetFilterDescription()`, byte-identical in the W1 page and the APAC layer copy, and that no other layer carries a copy.
- The three new tests (T217, T218, T219 in PriceListsUI.Codeunit.al) each verify that the caption shows the name of the selected record and not the first record in the table. They use `OpenEdit()` for the reason described above; the page has no page-level `Editable = false`, so edit mode is valid.
- Test numbering continues the codeunit's existing T-series (T216 was the highest in use).
- I did not build the Base Application locally; the change is three `SetFilter` replacements per page copy plus tests and relies on CI for compilation and the test run.

## Risk & compatibility

- Caption-only change: no persisted data, posting, or price calculation is affected.
- `SourceNoFilter` is a filter value entered in a filter field, so passing it to `SetFilter` is the intended semantics; when the filter matches several records the caption shows the first match, which is strictly better than showing an unrelated record.
- No schema, permission, or signature changes.

Fixes [AB#648173](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/648173)


Franco111000:
Merge conflict resolved, the branch is up to date with main.

The conflict was limited to `PriceListsUI.Codeunit.al`: #9571 added T220-T222 to the same test codeunit, so both blocks now sit in numeric order alongside T217-T219. No test numbers clash and nothing from main was dropped. The two `PricesOverview.Page.al` files merged clean.

@alexei-dobriansky following up on the labelling ask from #9180: this PR and #9182 came out with `Other` and `Ownership: Needs Review` rather than an area label, while my other open PRs got SCM. Both linked issues (#9101 here, #9097 on #9182) are labelled SCM, so that looks like the right area for both. I still don't have label permissions on the repo, so correct me if I'm wrong on how contributors are meant to set these.

alexei-dobriansky:

# **Recommendation: Accept**

<details>

This PR fixes a caption bug in page 7024 "Prices Overview": `GetFilterDescription()` was assigning the source number to a record field instead of calling `SetFilter`, so `FindFirst()` always returned the first record in the table and the caption showed an unrelated name. The fix replaces all three assignments (Customer, Customer Price Group, Campaign) with `SetFilter` on the key field.

The fix is correct and complete. `SetFilter` restricts the record set before `FindFirst()`, so the caption now looks up the record that matches the filter the user entered. If the filter matches no records, `FindFirst()` returns `false` and `Description` stays blank — the same safe outcome as before. The same three-line change is applied identically to both the W1 and APAC layer copies of the page, leaving no divergence between layers.

</details>

<details>

No suggestions. The fix is targeted and clean.

</details>

<details>

**Risk:** Caption-only change in `GetFilterDescription()` — no data is persisted, no posting or price calculation is affected, and no public API or event signature is changed. The only observable difference is the page title, which will now show the correct record name. Users who never looked at the caption, or who always filtered to a single record whose name equals its number, will see no change. Regression surface is minimal.

**Necessity:** The bug is clearly described, reproducible in current main, and has been tracked since the contribution pilot. Showing an unrelated customer's name next to the correct number in a pricing page is a visible, confusing defect. The three tests (T217–T219) cover each branch with a direct assertion on the caption content, using `OpenEdit()` to avoid the platform suppression documented in microsoft/AL#6388. The fix is sized appropriately for a bug fix.

</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=1 system=github pr=9181 round=1 by=alexei-dobriansky at=2026-08-11T10:13:42Z lastSha=1e5adf6f57f8460b0b726e30293e48f4aed94d8b reviewKey=aba2eb3351ea642d9f972a9dc03629d67b174affc3322c14aea0df47d9fbd715 suggestions=

alexei-dobriansky (dismissed):
Good!

Franco111000:
> Good!

Thanks !

alexei-dobriansky:
## Good Sense Reviewer - Reply

Thanks for confirming. No open review points from my side.

---
> [AI-PR-REPLY] version=1 promptVersion=4 system=github pr=9181 reviewRound=1 by=alexei-dobriansky at=2026-08-26T00:10:38Z respondsTo=2026-08-11T12:00:47Z reviewKey=aba2eb3351ea642d9f972a9dc03629d67b174affc3322c14aea0df47d9fbd715

src/Layers/W1/Tests/ERM/PriceListsUI.Codeunit.al
```diff
@@ -4845,6 +4846,102 @@
...
+        PriceSource: Record "Price Source";
+        PricesOverview: TestPage "Prices Overview";
+        DistinctName: Text[100];
+    begin
+        // [FEATURE] [Prices Overview] [Caption]
+        // [SCENARIO 9101] Setting "Assign-to Type Filter" = Customer and "Assign-to Filter" = a customer No. shows that customer's Name in the page caption (not the first customer in the table).
+        Initialize(true);
+
+        // [GIVEN] A customer whose Name is distinct from its No.
+        LibrarySales.CreateCustomer(Customer);
```
PredragMaricic: **[S1] Make these tests fail against the old unfiltered lookup.** This test creates only the target customer, so if it is the first or only customer, the old `FindFirst()` returns the same record and the test passes without the fix. Please create an unrelated record guaranteed to sort first (and do the equivalent for price group and campaign), or otherwise assert the selected target is not first.

PredragMaricic (dismissed):
The production `SetFilter` change correctly fixes the caption lookup and is consistently mirrored in W1 and APAC.

**S1 - Strengthen the regression setup.** Each test currently creates only the target source record. If that record is first or alone, the old unfiltered `FindFirst()` returns it and the test passes without the production fix. Please create an unrelated record guaranteed to sort first (or otherwise prove the target is not first) in each scenario. The actionable suggestion is tracked in the unresolved inline thread.

src/Layers/W1/Tests/ERM/PriceListsUI.Codeunit.al
Franco111000: Good catch, thanks. Fixed in 05e1f98.

You are right that the setup did not force the point: with a single created record, if it happened to be first (or the only one, which is the realistic case for Campaign in a clean company), the unfiltered `FindFirst()` returned that same record and the assertion passed against unfixed code.

Rather than create a decoy that has to be guaranteed to sort first, which puts the test at the mercy of collation and of whatever the table already holds, each test now creates two records and checks both in turn as the "Assign-to Filter" on the same page. An unfiltered lookup returns the same record for either filter value, so it necessarily fails on whichever of the two is not first, and the test never has to assume an order. Same shape for customer, customer price group, and campaign.

One thing I should be straight about: the build has not run on this branch, so these are reviewed and not yet executed.

alexei-dobriansky:

# **Recommendation: Accept**

<details>

This PR fixes the Prices Overview caption so Customer, Customer Price Group, and Campaign descriptions are looked up from the actual source filter before `FindFirst()` runs. Since round 1, the production diff is unchanged and the test changes make each caption test use two records, so an unfiltered lookup cannot satisfy both assertions.

The fix still matches the root cause: assigning a value to a record field did not filter the record set, while `SetFilter` applies the same filter text that the page uses. The updated tests strengthen the regression coverage for all three affected branches without widening the product change.
</details>

<details>

Round 1 raised no suggestions.
</details>

<details>

*None - the new commit only strengthens the existing regression tests.*
</details>

<details>

**Risk:** The product change is confined to caption construction for three source-filter branches in both page copies. It does not persist data, change price calculation, posting, schema, permissions, public APIs, or event signatures. The round-2 delta is test-only.

**Necessity:** The change is needed because the page could show a correct source number with an unrelated record name. The scope is targeted, and the strengthened tests now prove that the caption follows the selected filter rather than the first record in the table.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=9181 round=2 by=alexei-dobriansky at=2026-08-29T18:11:30Z lastSha=05e1f98a7c24f42e9879f9f782bfc458707b1132 reviewKey=ef99bb0154ebe85d6b32224b8b13d6a5ed74b7e3344046ddada6a8302b14f7f2 suggestions= parentRound=1

PredragMaricic (approved):
Reviewed the caption lookup fix and strengthened regression coverage. `SourceNoFilter` is validated and already used as filter text for the page records, so applying the same filter before `FindFirst()` correctly resolves the selected Customer, Customer Price Group, or Campaign. W1 and APAC remain aligned, and the two-record tests now fail against the old unfiltered behavior. No blocking issues found.

JakovljevicDusan (approved):

PR #9193, opened 2026-07-07
Lock Code labels on default Transformation Rule records

### LINKED ISSUE #9192: [Bug]: Transformation Rule default Codes are translated (not Locked), causing inconsistent Code values across language installations
### Describe the issue

Table 1237 "Transformation Rule" (namespace System.IO) seeds its default rules in CreateDefaultTransformations() using labels for the Code field (UPPERCASETxt, LOWERCASETxt, TITLECASETxt, TRIMTxt, ...).

None of these labels are marked Locked = true, so the literal text stored in the Code field which is the table's primary key gets translated according to the environment's installation/UI language.

### Expected behavior

Code values must stay constant regardless of installation language only Description 
should be translated. These labels should be marked Locked = true, since Code is a 
primary key referenced by other objects/configurations as a fixed identifier.

### Steps to reproduce

1. Install on an en-US environment and on a non-English one.
2. Open the "Transformation Rules" page on both.
3. Compare the Code column for the same rule (e.g. Uppercase) the value differs 
   between environments instead of staying constant.

### Additional context

_No response_

yematsmartit:
Fixes #9192

## What & why

The default Transformation Rule records created by `CreateDefaultTransformations()` use
`Label` declarations for the `Code` field without `Locked = true`. Since `Code` is this
table's primary key and is referenced by other objects/configurations as a fixed
identifier, translating it causes the same logical rule (e.g. "Uppercase") to have a
different `Code` value depending on the environment's installation/UI language or an
installed translation/language extension — breaking portability of configurations
between differently-localized environments.

This PR adds `Locked = true` to the 12 Code-producing labels, following the same pattern
already used for `AppendSpaceTxt` in this file. The `...DescTxt` labels used for
`Description` are left translatable, since only the Description should be localized.

## Linked work

Fixes #9192

## How I validated this

- Built the Base Application module locally with the change applied — no new analyzer
  warnings introduced.
- Installed a translation/language extension targeting these labels and confirmed that,
  after the fix, the Code-producing labels (UPPERCASETxt, LOWERCASETxt, TITLECASETxt,
  TRIMTxt, etc.) can no longer be overridden by the translation — Locked = true correctly
  keeps the Code value fixed, while the corresponding Description labels remain
  translatable as before.
- No automated tests added: this is a metadata-only change (Locked = true on existing
  Label declarations) with no new branching logic to cover.

## Risk & compatibility

Environments where these default Transformation Rule records were already inserted with
a translated (non-English) Code — prior to this fix — will not have their existing rows
renamed. Since `InsertRec` only inserts a rule if `Get(NewCode)` does not find it, an
upgrade/reinstall that re-runs `CreateDefaultTransformations()` on such an environment
would insert a new row with the English Locked Code alongside the existing
translated-Code row, rather than replacing it. Maintainers may want to consider whether
an upgrade codeunit is needed to clean up/rename pre-existing translated rows, or whether
this is acceptable as-is since the old rows would simply become unused duplicates.

### DIFF
--- a/src/Layers/W1/BaseApp/System/DataExchange/TransformationRule.Table.al
+++ b/src/Layers/W1/BaseApp/System/DataExchange/TransformationRule.Table.al
@@ -286,31 +286,31 @@ table 1237 "Transformation Rule"
     end;
 
     var
-        UNIXTIMESTAMPTxt: Label 'UNIXTIMESTAMP', Comment = 'Assigned to Transformation.Code field for Unix Timestamp';
+        UNIXTIMESTAMPTxt: Label 'UNIXTIMESTAMP', Locked = true, Comment = 'Assigned to Transformation.Code field for Unix Timestamp';
         UNIXTimeStampDescTxt: Label 'Transforming UNIX timestamp to text format.';
-        UPPERCASETxt: Label 'UPPERCASE', Comment = 'Assigned to Transformation.Code field for Upper case';
+        UPPERCASETxt: Label 'UPPERCASE', Locked = true, Comment = 'Assigned to Transformation.Code field for Upper case';
         UpperCaseDescTxt: Label 'Upper Case Text';
-        LOWERCASETxt: Label 'LOWERCASE', Comment = 'Assigned to Transformation.Code field for Lower case';
+        LOWERCASETxt: Label 'LOWERCASE', Locked = true, Comment = 'Assigned to Transformation.Code field for Lower case';
         LowerCaseDescTxt: Label 'Lower Case Text';
-        TITLECASETxt: Label 'TITLECASE', Comment = 'Assigned to Transformation.Code field for Title case';
+        TITLECASETxt: Label 'TITLECASE', Locked = true, Comment = 'Assigned to Transformation.Code field for Title case';
         TitleCaseDescTxt: Label 'Title Case Text';
-        TRIMTxt: Label 'TRIM', Comment = 'Assigned to Transformation.Code field for Trim';
+        TRIMTxt: Label 'TRIM', Locked = true, Comment = 'Assigned to Transformation.Code field for Trim';
         TrimDescTxt: Label 'Trim Text';
-        FOURTH_TO_SIXTH_CHARTxt: Label 'FOURTH_TO_SIXTH_CHAR', Comment = 'Assigned to Transformation.Code field for getting the 4th to 6th characters in a string';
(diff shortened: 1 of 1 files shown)


yematsmartit:

alexei-dobriansky:
## AI PR Review — Round 1

**Recommendation:** Accept

**Risk assessment:** The change only locks invariant default transformation rule code labels so they are no longer translated into persisted codes.

### Findings / suggestions

*None.*

---
Review mode: autonomous conversation comment (`interactive:false`). A human should make the final approve/request-changes decision.

> [AI-PR-REVIEW] by=alexei-dobriansky round=1 sha=79819ef21d0fd564d774c7bd9d05b761a39ff1be promptVersion=good-sense-reviewer-2026-08 reviewKey=c6d658998bb87787 suggestions=[]

alexei-dobriansky:

<details>

The change locks the labels that are used as Code values for the default transformation rules, while leaving the description labels translatable. This targets the reported bug where default rule codes can become language-dependent even though Code is the primary key.

The main code path is consistent: CreateDefaultTransformations() passes these labels into InsertRec() and InsertFindAndReplaceRule(), and the public Get...Code() helpers return the same labels. Locking the labels makes future seeded default codes stable without changing the transformation logic itself.
</details>

<details>

No previous suggestions were recorded in round 1.
</details>

<details>

**S1** (🟠 Moderate): *Clarify existing localized rows*
Existing companies can still have default rules saved under translated Code values. Please either add an upgrade/cleanup path for those rows, or document why leaving old translated rows and possible duplicate English rows is acceptable.

**S2** (🟡 Minor): *Update the locked label comment*
The comment on DeleteNOTPROVIDEDTxt still says DELETE should be translated, but the label is now locked. Please update the comment so the label metadata matches the fixed-code behavior.
</details>

<details>

**Risk:** The functional risk is low because the diff only changes label metadata for existing default identifiers. The main remaining risk is upgrade behavior: companies that already seeded translated codes may keep those old primary keys, and recreating defaults can add English-code rows beside them.

**Necessity:** The change is needed because transformation rule codes are fixed identifiers and are used by other objects and configurations. The scope is narrow and matches the requested behavior for future default records.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=9193 round=2 by=alexei-dobriansky at=2026-08-31T22:17:55Z lastSha=79819ef21d0fd564d774c7bd9d05b761a39ff1be reviewKey=95a658b6c7ac5a2cefe2adc7e24206f94731dcde155fb3e768077391c4e9924e suggestions=S1@7cec031b:new,S2@3aac53b9:new parentRound=1

PR #9196, opened 2026-07-07
[Master] - What If Impact on Planning and Supply

v-rohangarg20:
Workitem Slice 636114: [master] [all-e] What-if analysis — Analyze negative impact of supply changes on demand

Fixes [AB#636114](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/636114)


PredragMaricic (changes requested):
Request Changes

S1 - The simulation does not apply a functional quantity/date change (High)

`UpdatePurchaseLine`, `UpdateTransferLine`, and the production-order subscriber assign Quantity and date fields directly and call `Modify()`. This bypasses the field validation logic that recalculates Quantity (Base), outstanding/remaining quantities, planning dates, production components, and reservation entries.

When a source line already has reservations, `RunWhatIfAnalysis` then skips planning entirely and returns every existing outbound reservation. That result is independent of whether the entered quantity reduction actually creates a shortage or whether the delayed receipt still precedes demand. The current tests consequently verify reservation connectivity rather than the requested negative-impact analysis.

Please apply the simulated values through the appropriate validation/calculation path and test cases where changing the what-if value changes which demand is actually infeasible.

S2 - Date-driven planning suggestions are discarded (High)

`SimulatePlanningAndFindReservations` deletes every requisition line whose Action Message is not `Change Qty.` before carry-out. This removes `Reschedule`, `Resched. & Chg. Qty.`, `New`, and `Cancel` suggestions, including the date-only rescheduling results that the What-If Date scenario is specifically intended to identify.

Please preserve and process the relevant planning actions, with a successful unreserved date-delay test that proves the affected downstream document is reported.

S3 - Rollback safety is caller-dependent (High)

`RunWhatIfAnalysis` writes real purchase, transfer, production, and requisition records. `[CommitBehavior(CommitBehavior::Ignore)]` suppresses explicit commits but does not roll back those writes. The only rollback guarantee is the page caller raising `Error('')` after the procedure returns. Any other internal caller - particularly the future agent integration described by the linked work item - can accidentally persist the simulated changes.

Please move the rollback guarantee behind a safe simulation boundary rather than requiring every caller to know and reproduce this hidden contract.

The added tests all use Purchase Lines as scenario sources. None proves a successful simulate/carry-out/rollback path, and Transfer Line and Prod. Order Line source behavior is untested. Add coverage that verifies the generated impact and confirms that no source documents, generated orders, reservations, or requisition lines remain changed after simulation.

alexei-dobriansky:

<details>

This is Phase 1 of the "Supply What-If" feature (Slice 636114). It lets a planner open a purchase, transfer, or production order line, change the quantity or the receipt/due date, and preview which downstream production and transfer demand would be impacted - without changing the real data. A new engine codeunit "Supply What-If Planning Engine" (5455) plus a manufacturing companion "Mfg. Supply What-If Plan. Eng." (99000848) drive the flow. Two temporary tables ("Supply What-If Scenario" 5440 and "What-If Impact" 5441) and two pages (5440, 5443) show the input and the impact list. Existing planning objects (codeunit 5431, codeunit 99000854, report 99001020, codeunit 333) get a new `SimulationMode` flag that gates their `Commit()` calls and a few worksheet-template lookups.

The design is coherent as a feature increment. The impact detection works by reading reservation entries that link the changed supply to downstream demand, and only falls back to a full planning simulation when the line has no reservations. The safety model is layered: `RunWhatIfAnalysis` carries `[CommitBehavior(CommitBehavior::Ignore)]`, the touched planning objects skip `Commit()` when `SimulationMode` is on, and the page action ends with `Error('')` to roll back. The `SimulationMode` gates all default to `false`, so the normal planning path is unchanged - this keeps regression risk on shipping code low. The main open questions are around the correctness of applying the change with `Modify()` (no `Validate`) and the robustness of the rollback, covered below.
</details>

<details>

**S1** - *Live documents are changed then rolled back*
The engine writes to real Purchase, Transfer, and Prod. Order lines and real Requisition Lines, then relies on `[CommitBehavior(CommitBehavior::Ignore)]` plus the final `Error('')` in the page action to undo everything. This is high blast radius on shipping planning code. Please confirm no `Commit()` in the deep planning call tree can escape the ignore scope, and confirm `Error('')` reliably rolls back the live-document writes. A dedicated preview/rollback wrapper would make the intent and the safety clearer than an empty `Error`.

**S2** - *Quantity and date set without Validate*
`UpdatePurchaseLine`, `UpdateTransferLine`, and `OnBeforeUpdateTableRecord` set `Quantity` and the date field with a plain `Modify()`. This does not recalculate dependent fields such as Outstanding Quantity (Base) or reservation quantities, which the planning engine reads. Please confirm the simulation actually reflects the new quantity/date, or use `Validate` so the outstanding amounts update.

**S3** - *UserId used as worksheet template name*
`CopyStr(UserId(), 1, 10)` is used as the Requisition "Worksheet Template Name" filter for `DeleteAll(true)` and for the simulation. Two users whose UserId shares the first 10 characters would map to the same name, so one what-if run could delete or read another user's scratch lines. Consider a value that is unique per session, or per user without truncation risk.

**S4** - *Impact list may not react to the size of the change*
When the changed line already has reservations, the engine lists the reserved downstream demand from existing reservation entries. A small quantity reduction and a full delete then show the same impacted documents. If Phase 1 intends to show impact regardless of change size this is fine, but please confirm it matches the intended behavior and note it for the user.

**S5** - *Add tests for guards and non-purchase entry points*
The engine tests are solid and use exact-count assertions. They do not cover the guard errors (`WhatIfQuantityTooLargeErr`, `WhatIfDateTooEarlyErr`, `NoChangesToSimulateErr`) or the transfer-order and production-order entry points and the BOM traversal in codeunit 99000848. Adding a few small tests here is mechanical and would protect this logic.

**S6** - *Minor: page feedback and field captions*
The page action ends with `Error('')`, so the user gets no message when the run finishes; a short confirmation or status would help. Also several fields on table 5441 "What-If Impact" have a `ToolTip` but no `Caption` (for example "Entry No.", "Document No."), which can show blank headers. These are small polish items.
</details>

<details>

**Risk:** The highest risk is that the feature modifies live Purchase/Transfer/Prod. Order lines and Requisition Lines inside the user's transaction and depends on the rollback to undo them (S1). It also changes four core, widely-used planning objects (codeunit 5431 "Calc. Item Plan - Plan Wksh.", codeunit 99000854 "Inventory Profile Offsetting", report 99001020 "Carry Out Action Msg. - Plan.", codeunit 333 "Req. Wksh.-Make Order"). These changes are guarded by a `SimulationMode` flag that defaults to `false`, so the normal planning path is unchanged and regression risk on existing behavior is low. The new action is added consistently across all country layers (W1, NA, ES, FR, IT, APAC, RU). A correctness risk remains that setting `Quantity` without `Validate` (S2) may make the simulation ignore the actual change. New object IDs (tables 5440/5441, pages 5440/5443, codeunits 5455/99000848) should be confirmed as reserved for BaseApp.

**Necessity:** The scenario is well described in the work item and is important - planners today cannot preview downstream impact of a supply change, and this is the analytical foundation for the planned Purchase Communication Agent. The scope matches a Phase 1 slice (negative changes, 1-level impact). The change is large but that is expected for a feature slice, not a bug fix, so the size is appropriate for this channel.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9196 round=1 by=alexei-dobriansky at=2026-08-03T17:22:16Z lastSha=bc682a6dfdaa46bfe74a8c82c990f867258e8b4f suggestions=S1,S2,S3,S4,S5,S6

v-rohangarg20:
> Request Changes
> 
> S1 - The simulation does not apply a functional quantity/date change (High)
> 
> `UpdatePurchaseLine`, `UpdateTransferLine`, and the production-order subscriber assign Quantity and date fields directly and call `Modify()`. This bypasses the field validation logic that recalculates Quantity (Base), outstanding/remaining quantities, planning dates, production components, and reservation entries.
> 
> When a source line already has reservations, `RunWhatIfAnalysis` then skips planning entirely and returns every existing outbound reservation. That result is independent of whether the entered quantity reduction actually creates a shortage or whether the delayed receipt still precedes demand. The current tests consequently verify reservation connectivity rather than the requested negative-impact analysis.
> 
> Please apply the simulated values through the appropriate validation/calculation path and test cases where changing the what-if value changes which demand is actually infeasible.
> 
> S2 - Date-driven planning suggestions are discarded (High)
> 
> `SimulatePlanningAndFindReservations` deletes every requisition line whose Action Message is not `Change Qty.` before carry-out. This removes `Reschedule`, `Resched. & Chg. Qty.`, `New`, and `Cancel` suggestions, including the date-only rescheduling results that the What-If Date scenario is specifically intended to identify.
> 
> Please preserve and process the relevant planning actions, with a successful unreserved date-delay test that proves the affected downstream document is reported.
> 
> S3 - Rollback safety is caller-dependent (High)
> 
> `RunWhatIfAnalysis` writes real purchase, transfer, production, and requisition records. `[CommitBehavior(CommitBehavior::Ignore)]` suppresses explicit commits but does not roll back those writes. The only rollback guarantee is the page caller raising `Error('')` after the procedure returns. Any other internal caller - particularly the future agent integration described by the linked work item - can accidentally persist the simulated changes.
> 
> Please move the rollback guarantee behind a safe simulation boundary rather than requiring every caller to know and reproduce this hidden contract.
> 
> The added tests all use Purchase Lines as scenario sources. None proves a successful simulate/carry-out/rollback path, and Transfer Line and Prod. Order Line source behavior is untested. Add coverage that verifies the generated impact and confirms that no source documents, generated orders, reservations, or requisition lines remain changed after simulation.

Thanks for the detailed review. Please see our responses below:

S1 – Simulation does not apply functional quantity/date changes
We intentionally do not use Validate on the Purchase Line because we need to retain the existing reservation entries. Using Validate would update the reservation entries when the quantity changes, which is not desired for this simulation scenario. Therefore, UpdatePurchaseLine updates the quantity/date directly without triggering the reservation update logic.

S2 – Date-driven planning suggestions are discarded
We are reverting the deleted requisition lines by raising an error, which rolls back the transaction. This ensures that the changes made during the simulation are not persisted.

S3 – Rollback safety is caller-dependent
We have fixed this issue by moving the rollback handling to a safe simulation boundary, so it no longer depends on the caller to handle the rollback.

alexei-dobriansky:

<details>

Phase 1 of the Supply What-If feature: preview downstream planning impact of a supply change without committing real data. Since round 1 the author reworked the rollback: the simulation now runs inside a nested `Codeunit.Run()` (`OnRun` -> `RunSimulationWithinRollbackBoundary`, marked `[CommitBehavior(CommitBehavior::Ignore)]`) that captures the impact into a temporary record, then raises `Error(SimulationRolledBackErr)` so all live-document writes are rolled back; `RunWhatIfAnalysis` commits the clean pre-state first, then reads the results back. The page action no longer relies on an empty `Error('')` and the promoted action was moved to an `area(Promoted)`/`actionref` block. The rollback design is now much clearer. One correctness issue from round 1 remains open and is the blocker below.
</details>

<details>

| ID | Title | Status | Notes |
|----|-------|--------|-------|
| S1 | Live documents changed then rolled back | Addressed | Rollback moved inside the engine via a nested `Run()` with `CommitBehavior::Ignore` and a terminating `Error`; impact is returned through a temporary record and the empty page `Error('')` was removed. This is the dedicated preview/rollback wrapper that was requested. |
| S2 | Quantity and date set without Validate | Not addressed | `UpdatePurchaseLine` and `UpdateTransferLine` still set `Quantity`/date and call `Modify()` with no `Validate`. On the no-reservation path the simulation runs planning, which reads `Outstanding Qty. (Base)`; that field is not recalculated, so the what-if quantity change is not reflected and the impact list can be wrong. See blocker below. |
| S3 | UserId used as worksheet template name | Not addressed | Still `CopyStr(UserId(), 1, 10)`. The new rollback boundary now undoes any `DeleteAll` of another user's requisition lines, so the data-loss blast radius is reduced, but the truncation collision remains. Consider a per-session unique value. |
| S4 | Impact list may not react to change size | Not addressed | When the line has reservations, existing reserved demand is listed regardless of the change magnitude. Please confirm this matches the intended Phase 1 behavior. |
| S5 | Add tests for guards and non-purchase entry points | Not addressed | Guard errors and the transfer/production entry points are still uncovered. Non-blocking (the feature persists nothing), but worth adding, especially after the engine was refactored. |
| S6 | Page feedback and field captions | Partially addressed | Promotion modernized and the blank `Error('')` removed, so the finish behavior is cleaner. The missing `Caption` values on table 5441 `What-If Impact` fields are still open. |
</details>

<details>

**S2** - *Simulation does not apply the quantity or date change*
`UpdatePurchaseLine` and `UpdateTransferLine` set `Quantity` and the date field and call `Modify()` without `Validate`. Planning reads `Outstanding Qty. (Base)`, which is only recalculated by field validation, so on the no-reservation path the simulated plan still uses the original quantity and the impact result does not reflect the what-if change. Please use `Validate` on the quantity and date fields (or otherwise refresh the outstanding base quantities) so the simulation reflects the actual change.
</details>

<details>

The rollback rework is a clear improvement and the error handling is sound: a real failure inside the run leaves `SimulationCompleted` false and `RunWhatIfAnalysis` re-raises `GetLastErrorText()`, so genuine errors still surface while only the rollback marker is swallowed.

Informational (no change requested): `RunWhatIfAnalysis` now calls `Commit()` before the simulation run. In the modal what-if flow the state should be clean, but be aware this commits any pending changes in the caller's transaction before the boundary is established.
</details>

<details>

**Risk:** The rollback is now well-bounded (nested `Run` + `CommitBehavior::Ignore` + terminating `Error`), and the `SimulationMode` gates on the four shared planning objects still default to `false`, so regression risk on the normal planning path stays low. The remaining correctness risk is S2: on the no-reservation path the preview can show impact computed from the unchanged quantity, which is misleading for the planner the feature is meant to help.

**Necessity:** The scenario is well described and important - planners cannot currently preview downstream impact of a supply change - and the slice scope (negative changes, one-level impact) is appropriate for this channel. The blocker is not the feature's value but the fidelity of its core output, which S2 must fix.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=9196 round=2 by=alexei-dobriansky at=2026-08-20T12:18:01.2027507Z lastSha=13608c18f48eaf1ce5004077a571e5012d86a601 reviewKey=6b8279dcf0f02a96b5a0573802ae942f75387d28db4036c11c0be70b5dfb9ae0 suggestions=S1@d21ae293:addressed,S2@089b39b2:notaddressed,S3@8a1909b3:notaddressed,S4@d15fd997:notaddressed,S5@3939fe7c:notaddressed,S6@5898fe82:notaddressed parentRound=1

PredragMaricic (changes requested):
S1 (High): The result lists reservation links, not negatively impacted demand.

In `SupplyWhatIfPlanningEngine.Codeunit.al:123`, any existing positive reservation skips planning and every paired outbound reservation is returned. A PO reduced from 100 to 80 with only 70 reserved is reported as impacting that 70-unit demand even though it remains fully covered. With multiple reservations, all are reported rather than only the uncovered portion. This contradicts work item 636114's goal to analyze negative impact, and the new tests currently encode the false-positive behavior. Please calculate which demand actually loses coverage (including the affected quantity/date) and update the tests accordingly.

S2 (High): Simulated document changes bypass field validation.

`SupplyWhatIfPlanningEngine.Codeunit.al:237-254` and `MfgSupplyWhatIfPlanEng.Codeunit.al:106-109` assign Quantity/date fields directly and call `Modify()`. This leaves dependent base/outstanding quantities, planning dates, production components, and reservation handling stale, so the fallback planning run does not simulate the scenario the user entered. Please use `Validate(...)` for the changed fields and add tests proving dependent fields and calculated impacts change correctly.

alexei-dobriansky:

<details>

This PR adds phase 1 of Supply What-If planning, so a planner can preview downstream demand impact from a reduced supply quantity or later supply date. Since round 2, the PR head only advanced by merging main; I did not find any author-owned code change or reply after the round 2 review.

The rollback design from round 2 still looks correct: the simulation runs inside a nested engine Run() and the rollback marker error keeps live document changes from being committed. The remaining blocker is unchanged. The simulation still writes the what-if quantity and date directly to purchase and transfer lines before planning, so the calculated impact can still be based on the original base quantity/date values instead of the user's what-if values.
</details>

<details>

| S1 | Live documents changed then rolled back | Addressed | No new response. The nested rollback boundary is still present. |
| S2 | Quantity and date set without Validate | Not addressed | No new response. UpdatePurchaseLine and UpdateTransferLine still assign fields directly and call Modify(). |
| S3 | UserId used as worksheet template name | Not addressed | No new response. The code still uses CopyStr(UserId(), 1, 10). |
| S4 | Impact list may not react to change size | Not addressed | No new response. The reservation path still lists existing reserved demand without checking the change size. |
| S5 | Add tests for guards and non-purchase entry points | Not addressed | No new response. The added tests still do not cover the guard errors and all entry points. |
| S6 | Page feedback and field captions | Not addressed | No new response. Several What-If Impact fields still have no Caption. |
</details>

<details>
<summary><strong>Current suggestions</strong></summary>

**S2** - *Simulation does not apply the quantity or date change*
UpdatePurchaseLine and UpdateTransferLine still set Quantity and the date field directly and then call Modify(). Planning reads calculated/base values such as Outstanding Qty. (Base), and those are refreshed by validation, not by direct assignment. Please use Validate for the simulated quantity and date, or otherwise refresh the dependent planning fields, so the impact result matches the what-if input.
</details>

<details>

*None.* The only new PR commit since round 2 is a merge from main, so I did not attribute any new code hunk to this PR round.
</details>

<details>

**Risk:** Normal planning remains protected by the simulation-mode gates, and the rollback boundary still limits live-data risk. The core feature risk remains high for this slice because the preview can show the wrong impact when the no-reservation path runs planning from stale line values.

**Necessity:** The feature remains useful and the scope is valid for phase 1. It should not merge until the simulation applies the user's what-if quantity/date through the same validation path that normal planning depends on.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=9196 round=3 by=alexei-dobriansky at=2026-08-21T18:16:09.1991753Z lastSha=a348409b305ae92a71dc4367d73c0ae3ecf9bfab reviewKey=c6d229c06ba4b8ab34ba7f332b9234cc582a6fb5a2aaf0497bb561d227bf5615 suggestions=S1@d21ae293:addressed,S2@089b39b2:notaddressed,S3@8a1909b3:notaddressed,S4@d15fd997:notaddressed,S5@3939fe7c:notaddressed,S6@5898fe82:notaddressed parentRound=2

alexei-dobriansky (changes requested):
Please review the suggestions

alexei-dobriansky:
## Good Sense Reviewer - Reply

Thanks for the update. There is no technical review point to agree with or hold against here. The latest comment only says the build status check is stale, so it does not change the previous review findings; the build still needs to be refreshed separately.

---
> [AI-PR-REPLY] version=1 promptVersion=4 system=github pr=9196 reviewRound=3 by=alexei-dobriansky at=2026-08-25T14:20:32.7829063Z respondsTo=2026-08-25T00:30:05Z reviewKey=c6d229c06ba4b8ab34ba7f332b9234cc582a6fb5a2aaf0497bb561d227bf5615

PredragMaricic (approved):

alexei-dobriansky (approved):
Please respond to the suggestions.

PR #9206, opened 2026-07-08
Added new privacy policy for BizChat depending on Microsoft Copilot 365 with EEA environments behavior changes

Kimi310:
Slice: [AB#626609](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/626609)

New Microsoft Copilot privacy policy added for Chat V2 rollout. Behavior changes in EEA environments including disabling chat functionality by default for European users and leaving the default ON behavior for other locations. Wiring Platform functionality for a new shortcut to AI and agents capabilities page.


src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
```diff
@@ -156,6 +165,13 @@ codeunit 1565 "Privacy Notice Impl."
             Session.LogMessage('0000GKF', StrSubstNo(UserPrivacyApprovalStateTelemetryTxt, "Privacy Notice Approval State"::Agreed, PrivacyNoticeId), Verbosity::Normal, DataClassification::SystemMetadata, TelemetryScope::ExtensionPublisher, 'Category', TelemetryCategoryTxt);
             exit("Privacy Notice Approval State"::Agreed); // If user clicked no, they will still be notified until admin makes a decision
         end;
+
+        // Apply the runtime default policy (e.g. non-EUDB tenants auto-approve Copilot) without persisting an admin decision.
+        if ShouldApproveByDefault(PrivacyNoticeId) then begin
+            Session.LogMessage('', StrSubstNo(PrivacyNoticeAutoApprovedByDefaultTelemetryTxt, PrivacyNoticeId), Verbosity::Normal, DataClassification::SystemMetadata, TelemetryScope::ExtensionPublisher, 'Category', TelemetryCategoryTxt);
```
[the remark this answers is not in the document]
Kimi310: Added telemetry tags

src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
```diff
@@ -346,14 +372,43 @@ codeunit 1565 "Privacy Notice Impl."
...
+                exit(false);
+            if (SystemPrivacyNoticeReg.TryGetIsNotWithinEUDB(IsNotWithinEUDB)) then
+                exit(IsNotWithinEUDB);
+        end;
+
         exit(false);
     end;
 
+    // Returns true if the SaaS-registered "Chat" Copilot capability exists and is currently marked Inactive.
+    local procedure IsChatCapabilityInactive(): Boolean
```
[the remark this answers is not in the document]
Kimi310: Fixed in the new commit

src/System Application/App/Privacy Notice/src/PrivacyNotice.Codeunit.al
```diff
@@ -238,4 +238,17 @@ codeunit 1563 "Privacy Notice"
...
+    /// <summary>
+    /// Raised while resolving the runtime default approval policy for a privacy notice.
+    /// Modules that own an integration can subscribe to decide whether the notice should be approved by default,
+    /// without persisting an explicit admin or user decision.
+    /// </summary>
+    /// <param name="IntegrationID">The identifier of the privacy notice / integration being evaluated.</param>
+    /// <param name="ShouldApproveByDefault">Set by a subscriber to indicate whether the notice should be approved by default.</param>
+    /// <param name="Handled">Set to true by the subscriber that resolves the policy for the given integration.</param>
+    [IntegrationEvent(false, false)]
+    procedure OnCheckPrivacyNoticeApprovalDefault(IntegrationID: Text; var ShouldApproveByDefault: Boolean; var Handled: Boolean)
```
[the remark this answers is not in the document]
Kimi310: Fixed

src/System Application/App/AI/src/Copilot/CopilotCapabilityInstall.Codeunit.al
```diff
@@ -62,7 +62,7 @@ codeunit 7760 "Copilot Capability Install"
         if CopilotCapability <> Enum::"Copilot Capability"::Chat then
             exit;
 
-        if not RequiredPrivacyNotices.Contains(SystemPrivacyNoticeReg.GetMicrosoftLearnID()) then
-            RequiredPrivacyNotices.Add(SystemPrivacyNoticeReg.GetMicrosoftLearnID());
+        if not RequiredPrivacyNotices.Contains(SystemPrivacyNoticeReg.GetMicrosoftCopilotID()) then
```
msft-sam: it makes sense that we don't check learn with v2, but I wonder if we need to handle the case when we do not load v2 but show chat v1

maybe it is safe for now to still check learn + copilot, then once we no longer show chat v1 then we can remove learn? (basically once we leave private preview)

src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
```diff
@@ -343,15 +394,43 @@ codeunit 1565 "Privacy Notice Impl."
...
+        // Microsoft Learn keeps the standard prompting flow; being in geo only provides a default-approve hint.
+        if CheckIntegrationIDEquality(SystemPrivacyNoticeReg.GetMicrosoftLearnID(), IntegrationID) then begin
+            if SystemPrivacyNoticeReg.TryGetMicrosoftLearnInGeoSupport(MicrosoftLearnServiceInGeo) then
+                DefaultApproval := MicrosoftLearnServiceInGeo;
+            exit(false);
+        end;
+
+        // Let integrations (for example Copilot in the AI SDK module) resolve their own runtime default approval policy.
+        // When handled, the integration fully governs the no-decision state through this default.
+        PrivacyNotice.OnCheckPrivacyNoticeApprovalDefault(IntegrationID, DefaultApproval, Handled);
```
msft-sam: I don't think we need to set up this event flow, wouldn't it be easier to just add Copilot like we add Learn above?

msft-sam: Then you can follow the same pattern for the geo support like we do for Learn etc.

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
```diff
@@ -508,6 +508,54 @@ codeunit 7774 "Copilot Capability Impl"
...
+        if Handled then
+            exit;
+
+        if CopyStr(UpperCase(IntegrationID), 1, 50) <> UpperCase(SystemPrivacyNoticeReg.GetMicrosoftCopilotID()) then
+            exit;
+
+        Handled := true;
+
+        // Trial / evaluation tenants: unconditionally auto-approve; no EUDB split and no legacy admin gates to inherit.
+        if TenantLicenseState.IsTrialMode() or TenantLicenseState.IsEvaluationMode() then begin
```
msft-sam: I don't think we have checked this for similar integrations, but instead looked at if the company itself is an evaluation company, is this intentional (and should we be doing the same for other integrations?)

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
```diff
@@ -508,6 +508,54 @@ codeunit 7774 "Copilot Capability Impl"
...
+            ShouldApproveByDefault := true;
+            exit;
+        end;
+
+        // If the pre-BizChat Chat capability was turned off, honor that choice instead of auto-approving.
+        if IsChatCapabilityInactive() then begin
+            ShouldApproveByDefault := false;
```
msft-sam: I don't think that is needed, the capability will still be off, so even if this privacy notice is true/undefined, the capability is still disabled

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
```diff
@@ -508,6 +508,54 @@ codeunit 7774 "Copilot Capability Impl"
...
+            exit;
+        end;
+
+        ShouldApproveByDefault := not IsWithinEUDB();
```
msft-sam: So really this is the only thing we need to check, the framework will skip this if the company is an evaluation company already

src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
```diff
@@ -149,13 +172,31 @@ codeunit 1565 "Privacy Notice Impl."
         if SkipCheckInEval and Company.Get(CompanyName()) and Company."Evaluation Company" then
             exit("Privacy Notice Approval State"::Agreed); // Auto-agree for evaluation companies if admin has not explicitly disagreed
 
+        // Integrations that fully govern the "Let Users Decide" state (for example Copilot) resolve their runtime
+        // default here. The default is never persisted; when it is off the state stays undecided ("Not set").
+        if ResolveDefaultApproval(PrivacyNoticeId, DefaultApproval) then begin
```
msft-sam: We should respect the user decision before resolving the default approval (I don't think we need this logic)

src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
```diff
@@ -149,13 +172,31 @@ codeunit 1565 "Privacy Notice Impl."
...
         // Check if user made a decision and if so, return that
         PrivacyNotice.SetRange("User SID Filter", UserSecurityId());
         PrivacyNotice.CalcFields(Enabled);
         if PrivacyNotice.Enabled then begin
             Session.LogMessage('0000GKF', StrSubstNo(UserPrivacyApprovalStateTelemetryTxt, "Privacy Notice Approval State"::Agreed, PrivacyNoticeId), Verbosity::Normal, DataClassification::SystemMetadata, TelemetryScope::ExtensionPublisher, 'Category', TelemetryCategoryTxt);
             exit("Privacy Notice Approval State"::Agreed); // If user clicked no, they will still be notified until admin makes a decision
         end;
+
+        // Apply the runtime default-approve hint (e.g. Microsoft Learn in geo) without persisting an admin decision.
+        if DefaultApproval then begin
```
msft-sam: Would this case be hit if an event subscriber from `ResolveDefaultApproval` did set the `DefaultApproval` but did not set `Handled` to `true`? Is that a bug?

I think in this case we should just flow through to `Not set` (admin + user have not set a value) -- then we can flow through to the defaulting logic (this is how it works for other scenarios)

src/System Application/App/AI/src/Copilot/CopilotCapabilityInstall.Codeunit.al
Kimi310: You are right, I think even after the v2 rollout there are some locations where v2 wont be available (from what I think Blazej told me). I restored the logic for it

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
Kimi310: You are right, this is a leftover of my previous logic that altered the privacy notice table itself. I dont think we need it either. Deleted it.

src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
Kimi310: That depends on the other comment that you have written, because right now the Copilot Privacy Notice is dependant on more complex logic than Learn fx. the check for trial user environment which I would same favor current approach. If the trial check proves redundant I think I can re-do it to the learn approach

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
Kimi310: That depends on if the trial environment can make non-eval companies? If it can than I guess it is better to leave it as it is becasue then the matrix assumptions that Blazej did will not be met. If they cannot and basically trial environment = eval companies only then we can drop this and re do what I have done to more micorsoft learn approach

src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
Kimi310: However also, from what I can see, Learn gets persisted to the table as "Agreed for everyone" when it is in geo for it. From the last talk that I had with you I think for Copilot we want to have it to "Let users decide" if no decision have been made and resolve the notice temporarly at a runtime. Correct me if I am wrong

src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
```diff
@@ -99,6 +101,20 @@ codeunit 1565 "Privacy Notice Impl."
...
+        // policy (such as a country/geo matrix). For those, the default is applied at runtime without ever being persisted:
+        // only users who can approve for the organization are prompted, and everyone else is asked to contact their admin.
+        if ResolveDefaultApproval(PrivacyNoticeId, DefaultApproval) then begin
+            if DefaultApproval then begin
+                Session.LogMessage('0000UHZ', StrSubstNo(PrivacyNoticeAutoApprovedByDefaultTelemetryTxt, PrivacyNoticeId), Verbosity::Normal, DataClassification::SystemMetadata, TelemetryScope::ExtensionPublisher, 'Category', TelemetryCategoryTxt);
+                exit(true);
+            end;
+            if CanCurrentUserApproveForOrganization() then
+                exit(ShowPrivacyNotice(PrivacyNotice)); // Eligible user can change the notice, so let them decide
+            Message(AdminDisabledIntegrationMsg, PrivacyNotice."Integration Service Name");
```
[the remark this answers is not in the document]
Kimi310: Changed the logic to allow per user acceptence as in other notice flows

src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
Kimi310: Managed to change the behavior to closer match the one microsoft learn does now

src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
Kimi310: Revised it and now the user decission is before deafult approval

src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
Kimi310: Changed the design to a closer one to Learn so this is no longer a threat

src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
```diff
@@ -107,6 +113,12 @@ codeunit 1565 "Privacy Notice Impl."
             exit(true); // If user clicked no, they will still be notified until admin makes a decision
         end;
 
+        // Apply the runtime default-approve hint (e.g. Microsoft Learn in geo).
+        if DefaultApproval then begin
```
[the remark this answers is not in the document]
Kimi310: Fixed

src/System Application/App/Privacy Notice/app.json
```diff
@@ -32,6 +38,11 @@
       "id": "7c2785ec-13a5-470b-87a4-6838a2c19a3a",
       "name": "Privacy Notice Tests",
       "publisher": "Microsoft"
+    },
+    {
+      "id": "d3433b68-4901-445f-9547-fdfeca57575a",
+      "name": "AI SDK",
```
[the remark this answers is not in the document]
Kimi310: Deleted the redundant reference

src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
```diff
@@ -99,6 +101,11 @@ codeunit 1565 "Privacy Notice Impl."
         if SkipCheckInEval and Company.Get(CompanyName()) and Company."Evaluation Company" then
             exit(true); // Auto-agree for evaluation companies if admin has not explicitly disagreed
 
+        // Resolve the runtime default (e.g. Copilot's geo/EUDB matrix or Microsoft Learn in geo). This is never persisted;
+        // it is only applied below after an explicit user approval. Per-user declines are not stored (see SetApprovalState),
+        // so a declined user is re-evaluated on every check; when the default is off, the user is prompted to decide.
+        ResolveDefaultApproval(PrivacyNoticeId, DefaultApproval);
```
juliamakulec: Nit: If the DefaultApproval is acted on after the lines below, wouldn't it be easier to read the flow/logic if this was moved down as well?

src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
```diff
@@ -149,13 +163,25 @@ codeunit 1565 "Privacy Notice Impl."
         if SkipCheckInEval and Company.Get(CompanyName()) and Company."Evaluation Company" then
             exit("Privacy Notice Approval State"::Agreed); // Auto-agree for evaluation companies if admin has not explicitly disagreed
 
+        // Resolve the runtime default (e.g. Copilot's geo/EUDB matrix or Microsoft Learn in geo). This is never persisted;
+        // it is only applied below after an explicit user approval. Per-user declines are not stored (see SetApprovalState),
+        // so a declined user is re-evaluated on every check; when the default is off, the state stays undecided ("Not set").
+        ResolveDefaultApproval(PrivacyNoticeId, DefaultApproval);
```
juliamakulec: Nit: If the `DefaultApproval` is acted on after the lines below, wouldn't it be easier to read the flow/logic if this was moved down as well?

src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
```diff
@@ -107,6 +114,12 @@ codeunit 1565 "Privacy Notice Impl."
             exit(true); // If user clicked no, they will still be notified until admin makes a decision
         end;
 
+        // Apply the runtime default-approve hint (e.g. Microsoft Learn in geo).
+        if DefaultApproval then begin
```
juliamakulec: I think this is still unresolved, unless the new version is not pushed? It only checks here what is the DefaultApproval value, not if there's PrivacyNotice.Disabled

Kimi310: The admin PrivacyNotice.Disabled check is made before that on top of the method so the admin cannot be overridden by the default however for all the other privacy notices we don't store per user disagree. We only check if the user agreed before. I have moved the default check so it happens after that, but I don't think that we should change the pre existing logic for not storing the disagrees

src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
Kimi310: Moved down

src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
Kimi310: Moved down

src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
```diff
@@ -156,6 +169,17 @@ codeunit 1565 "Privacy Notice Impl."
             Session.LogMessage('0000GKF', StrSubstNo(UserPrivacyApprovalStateTelemetryTxt, "Privacy Notice Approval State"::Agreed, PrivacyNoticeId), Verbosity::Normal, DataClassification::SystemMetadata, TelemetryScope::ExtensionPublisher, 'Category', TelemetryCategoryTxt);
             exit("Privacy Notice Approval State"::Agreed); // If user clicked no, they will still be notified until admin makes a decision
         end;
+
+        // Resolve and apply the runtime default (e.g. Copilot's geo/EUDB matrix or Microsoft Learn in geo) without persisting
```
msft-sam: Do we still need this logic? Can't we just let this return `Not set` (as the state is `Not set`) -- then the calling logic can decide what to do with the `Not set` state (same for above).

Edit: If you fix the comment below and persist the default state, we would inherit the default state and return `Approved` or `Not Set` in the default case (so then `Not set` would mean the feature is off), and in the procedure above we would return `Approved` or prompt for approval.

src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
juliamakulec: When the user disagreed before, they would get prompted every time. But here we don't want the singular user to be able to decide, even if the admin set it to "let users decide?
I'm not saying that the code is wrong, I'm just trying to fully understand the intended design

src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
```diff
@@ -237,12 +261,22 @@ codeunit 1565 "Privacy Notice Impl."
     var
         PrivacyNoticeApproval: Codeunit "Privacy Notice Approval";
     begin
-        if ShouldApproveByDefault(PrivacyNotice.ID) then begin
+        if ShouldApproveByDefault(PrivacyNotice.ID) and ShouldPersistDefaultApproval(PrivacyNotice.ID) then begin
```
msft-sam: The intent behind `ShouldApproveByDefault` is that it is approved and persisted, so it becomes a bit tricky to follow to then have a separate function to maintain another list of integrations.

For regions where we are enabled by default, it would make sense to populate it with the approved state, so we don't need the `ShouldPersistDefaultApproval`.

I guess I am asking, why aren't we persisting Copilot?

src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
juliamakulec: With the moved down code block and comment now this is better explained 😄

juliamakulec:
Nit: You could change the comment/docs on the method in the PrivacyNotice codeunit to include the default approving scenario https://github.com/microsoft/BCApps/blob/f7f34ab8b52abac48c1ec844fc7676c101bad4b9/src/System%20Application/App/Privacy%20Notice/src/PrivacyNotice.Codeunit.al#L96

src/System Application/App/AI/app.json
```diff
@@ -100,6 +100,12 @@
       "name": "Client Type Management",
       "publisher": "Microsoft",
       "version": "29.0.0.0"
+    },
```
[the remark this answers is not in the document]
Kimi310: Removed

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
msft-sam: Yeah lets confirm the intent here, it would be good to be consistent with the framework (license states can also change, which could be problematic), I think it would be safer to limit to eval companies

src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
juliamakulec: Agree.
The admin can still go to the privacy notices and override the default decision we have made,
So is the non-persisting necessary?

src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
juliamakulec: As I understand it now:
- If there's not specific reasons, we should persist the default state
- Then, checking the privacy notice (the code here) is not relevant anymore, as the setting of the notice took care of the correct value
- If This is a new privacy notice, newly required - we won't have a case that an upgraded environment in EEA had this notice already accepted and is showing the chat, when it's not supposed to. Instead, the default approval will mark it as "not set", meaning that the Chat capability will be turned off :)

So with the above, the changes could be much simplified, unless we are missing something about the persisting the defaults state

src/System Application/App/Privacy Notice/src/SystemPrivacyNoticeReg.Codeunit.al
```diff
@@ -85,11 +90,34 @@ codeunit 1566 "System Privacy Notice Reg."
...
     end;
+
+    [TryFunction]
+    internal procedure TryGetMicrosoftCopilotDefaultApproval(var ShouldApproveByDefault: Boolean)
+    var
+        TenantLicenseState: Codeunit "Tenant License State";
+        ALCopilotFunctions: DotNet ALCopilotFunctions;
+    begin
+        // Trial / evaluation tenants: unconditionally auto-approve; no EUDB split and no legacy admin gates to inherit.
+        if TenantLicenseState.IsTrialMode() or TenantLicenseState.IsEvaluationMode() then begin
```
juliamakulec: Is there a reason we are checking the tenant state directly for the Copilot privacy notice?

If it's an eval company, this line takes care of it for us before we even get to default approvals I believe https://github.com/microsoft/BCApps/pull/9206/changes#diff-384897525cd81c4de7e231d5c4825ff54851c9ae26fd38cfe9c04a38aaa48bb7L149

src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
Kimi310: I thought that we can't really change the table state without the admin decision and ShouldApproveByDeafult does that however I think you are right and if Learn already does it this way Copilot should too. Changed the behavior so that it matches Learn and auto-approve persists by default now

src/System Application/App/Privacy Notice/src/SystemPrivacyNoticeReg.Codeunit.al
Kimi310: Changed it so that Copilot also relies on the previous check for eval company instead of the custom one

src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
Kimi310: Changed it to the simpler version similar to Learn. Now not set basically means privacy notice not accepted by default -> chat off.

juliamakulec (dismissed):

src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
```diff
@@ -101,7 +101,7 @@ codeunit 1565 "Privacy Notice Impl."
 
         // Check if user made a decision and if so, return that
         PrivacyNotice.SetRange("User SID Filter", UserSecurityId());
-        PrivacyNotice.CalcFields(Enabled, Disabled);
+        PrivacyNotice.CalcFields(Enabled);
```
juliamakulec: nit: unrelated to your changes, could leave as-is

src/System Application/App/Privacy Notice/src/PrivacyNotice.Codeunit.al
```diff
@@ -60,6 +60,7 @@ codeunit 1563 "Privacy Notice"
     /// User:
     ///     If admin has agreed or disagreed, that will be returned
     ///     If user has agreed, that will be returned
+    ///     If the integration is enabled by default for the environment's region (for example Microsoft Learn in a supported geo, or Microsoft Copilot outside the EU Data Boundary), it is pre-approved for the organization when the notice is registered; an admin can change this on the Privacy Notices page.
```
msft-sam: Minor: I would remove the examples from the comments (they may change over time)

msft-sam (dismissed):

src/System Application/App/Privacy Notice/src/PrivacyNotice.Codeunit.al
Kimi310: Removed the examples

src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
Kimi310: Reverted to previous state

src/System Application/App/Privacy Notice/src/SystemPrivacyNoticeReg.Codeunit.al
```diff
@@ -38,6 +39,9 @@ codeunit 1566 "System Privacy Notice Reg."
         TempPrivacyNotice.ID := BingTxt;
         TempPrivacyNotice."Integration Service Name" := BingTxt;
         if not TempPrivacyNotice.Insert() then;
+        TempPrivacyNotice.ID := MicrosoftCopilotTxt;
```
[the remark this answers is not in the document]
Kimi310: Fixed

src/System Application/App/AI/src/Copilot/CopilotCapabilityImpl.Codeunit.al
```diff
@@ -518,6 +518,13 @@ codeunit 7774 "Copilot Capability Impl"
         IsEnabled := AzureOpenAI.IsEnabled(CopilotCapability, Silent, AppId);
     end;
 
+    [EventSubscriber(ObjectType::Codeunit, Codeunit::"Navigation Bar Subscribers", 'OnBeforeDefaultOpenCopilotCapabilities', '', false, false)]
+    local procedure OpenCopilotCapabilities(var Handled: Boolean)
+    begin
+        Page.Run(Page::"Copilot AI Capabilities");
+        Handled := true;
+    end;
```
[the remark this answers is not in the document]
Kimi310: Fixed

src/System Application/App/Navigation Bar Subscribers/src/NavigationBarSubImpl.Codeunit.al
```diff
@@ -47,6 +47,17 @@ codeunit 1818 "Navigation Bar Sub. Impl."
...
+    [EventSubscriber(ObjectType::Codeunit, Codeunit::"System Action Triggers", OpenCopilotCapabilities, '', false, false)]
+    local procedure DefaultOpenCopilotCapabilities()
+    var
+        NavigationBarSubscribers: Codeunit "Navigation Bar Subscribers";
+        Handled: Boolean;
+    begin
+        NavigationBarSubscribers.OnBeforeDefaultOpenCopilotCapabilities(Handled);
+        if not Handled then
+            Error(SettingsNotAvailableErr);
+    end;
```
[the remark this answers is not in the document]
Kimi310: Fixed

src/System Application/App/AI/src/Copilot/CopilotCapabilityInstall.Codeunit.al
```diff
@@ -62,7 +64,13 @@ codeunit 7760 "Copilot Capability Install"
         if CopilotCapability <> Enum::"Copilot Capability"::Chat then
             exit;
 
-        if not RequiredPrivacyNotices.Contains(SystemPrivacyNoticeReg.GetMicrosoftLearnID()) then
-            RequiredPrivacyNotices.Add(SystemPrivacyNoticeReg.GetMicrosoftLearnID());
+        // In environments where the BizChat (Microsoft 365 Copilot) experience is enabled, the Chat capability
+        // relies on the Microsoft Copilot privacy notice; otherwise it relies on the Microsoft Learn privacy notice.
+        if ALCopilotFunctions.IsBizChatEnabled() then begin
```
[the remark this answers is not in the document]
Kimi310: Fixed

src/System Application/App/Privacy Notice/src/SystemPrivacyNoticeReg.Codeunit.al
```diff
@@ -85,11 +89,28 @@ codeunit 1566 "System Privacy Notice Reg."
...
     begin
         HasInGeoSupport := ALMicrosoftLearnFunctions.HasInGeoSupport()
     end;
+
+    [TryFunction]
+    internal procedure TryGetMicrosoftCopilotDefaultApproval(var ShouldApproveByDefault: Boolean)
+    var
+        ALCopilotFunctions: DotNet ALCopilotFunctions;
+    begin
+        ShouldApproveByDefault := not ALCopilotFunctions.IsWithinEUDB();
```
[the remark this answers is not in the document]
Kimi310: There is error handling if the alcopilot functions won't respond plus the same pattern is already implmented in Microsoft Learn privacy notice.

src/System Application/App/AI/src/Copilot/CopilotCapabilityInstall.Codeunit.al
```diff
@@ -62,7 +64,22 @@ codeunit 7760 "Copilot Capability Install"
         if CopilotCapability <> Enum::"Copilot Capability"::Chat then
             exit;
 
-        if not RequiredPrivacyNotices.Contains(SystemPrivacyNoticeReg.GetMicrosoftLearnID()) then
-            RequiredPrivacyNotices.Add(SystemPrivacyNoticeReg.GetMicrosoftLearnID());
+        // In environments where the BizChat (Microsoft 365 Copilot) experience is enabled, the Chat capability
+        // relies on the Microsoft Copilot privacy notice; otherwise it relies on the Microsoft Learn privacy notice.
+        // The BizChat check is a DotNet interop call that can raise; if it fails we degrade gracefully to the Microsoft Learn notice.
```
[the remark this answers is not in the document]
Kimi310: Changed to deafult to Microsoft Copilot notice instead

src/System Application/App/AI/src/Copilot/CopilotCapabilityInstall.Codeunit.al
```diff
@@ -62,10 +65,26 @@ codeunit 7760 "Copilot Capability Install"
...
+        // In environments where the BizChat (Microsoft 365 Copilot) experience is enabled, the Chat capability
+        // relies on the Microsoft Copilot privacy notice; otherwise it relies on the Microsoft Learn privacy notice.
+        // The BizChat check is a DotNet interop call that can raise. Fail closed: if the status cannot be
+        // determined we require the Microsoft Copilot notice (which governs the Microsoft 365 Copilot data
+        // transfer), so that consent for the Microsoft Learn experience can never implicitly authorize BizChat.
+        BizChatStatusKnown := TryGetBizChatEnabled(BizChatEnabled);
 
-        if not RequiredPrivacyNotices.Contains(SystemPrivacyNoticeReg.GetMicrosoftCopilotID()) then
-            RequiredPrivacyNotices.Add(SystemPrivacyNoticeReg.GetMicrosoftCopilotID());
+        if (not BizChatStatusKnown) or BizChatEnabled then begin
```
msft-sam, line 70: Is it correct that we assume BizChat if the call fails? Wouldn't that be more indicative of chat v1?

src/System Application/App/AI/src/Copilot/CopilotCapabilityInstall.Codeunit.al
```diff
@@ -62,10 +65,26 @@ codeunit 7760 "Copilot Capability Install"
         if CopilotCapability <> Enum::"Copilot Capability"::Chat then
             exit;
 
-        if not RequiredPrivacyNotices.Contains(SystemPrivacyNoticeReg.GetMicrosoftLearnID()) then
-            RequiredPrivacyNotices.Add(SystemPrivacyNoticeReg.GetMicrosoftLearnID());
+        // In environments where the BizChat (Microsoft 365 Copilot) experience is enabled, the Chat capability
```
msft-sam: Minor: I don't think we need this comment, the code is fairly simple to follow (will likely go out of sync as this changes in the future)

src/System Application/App/AI/src/Copilot/CopilotCapabilityInstall.Codeunit.al
Kimi310: Yes this is correct. There is always a trade off here, I assumed the copilot one would be a better default for the future as more and more customers would switch to the new copilot experience however if you think it would be best to default to Learn I can also change this

src/System Application/App/AI/src/Copilot/CopilotCapabilityInstall.Codeunit.al
Kimi310: Removed

vesakr (approved):

krupybalu (approved):

juliamakulec (approved):

aholstrup1 (approved):

PR #9219, opened 2026-07-08

neeleshsinghal:


alexei-dobriansky:

<details>

The PR keeps `AddAllSourceType` in the source list, but changes `Price List Management` so `Price List Line Review` uses native filters instead of `Mark` / `MarkedOnly`. That targets the valid performance bug where opening Sales Prices from Customer Card can materialize a very large All Customers result set.

The optimization does not preserve the old exact source matching. The old code marked rows for each `(Source Type, Parent Source No., Source No.)` entry separately. The new code ORs each field independently, which can admit combinations that were never in the source list. It also moves the `OnBuildSourceFiltersOnBeforeFindLines` event before the source filters are set, which changes the extension contract.
</details>

<details>

**S1** - *Preserve source tuple filtering*
`BuildSourceFilters` now builds independent OR filters for `Source Type`, `Parent Source No.`, and `Source No.`. That can return a Cartesian product of values from different `PriceSourceList` entries, so unrelated price list lines can pass the filter. Build a filter that preserves each exact source tuple, or keep a safe per-source path.

**S2** - *Keep event timing compatible*
`OnBuildSourceFiltersOnBeforeFindLines` used to run after the source filters were set and just before `FindSet`. It now runs before those filters are set, and the normal path no longer does a per-source find. Keep the old event contract or use a new event, because extensions may depend on the filtered record.

**S3** - *Add regression and performance coverage*
This PR changes which price list lines can be shown, but no test file changed. Add a regression test with Customer, Customer Price Group, Customer Disc. Group, All Customers, and an overlapping non-matching source code. Also include a small performance measurement or test note for the large All Customers case.
</details>

<details>

**Risk:** High. This is a pricing UI path used from Customer, Vendor, Contact, Campaign, and Job flows. If the filter is too broad, users can see or change prices and discounts that do not belong to the selected source. The changed integration event timing also creates compatibility risk for extensions.

**Necessity:** The change is needed because the work item describes a clear and severe performance problem: about 220 seconds to open the page on large datasets. The scope should stay focused on avoiding expensive materialization while keeping the exact same rows and event behavior.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9219 round=1 by=alexei-dobriansky sha=af13e0fee690984935eb2391e6e4eac8db4f8422

alexei-dobriansky:

<details>

The new commit tries to address the source tuple issue by using a `SystemId` filter when the source list contains an entry with `Parent Source No.`. It also moves `OnBuildSourceFiltersOnBeforeFindLines` after the source filters are set again.

This still does not fully fix the customer-card scenario. Customer, Customer Price Group, Customer Disc. Group, and All Customers entries can all have blank `Parent Source No.`, so `HasCorrelatedSourceEntries` is false and the code still builds independent OR filters for `Source Type`, `Parent Source No.`, and `Source No.`. That can still admit a value combination that was never present in `PriceSourceList`. The event is also still not fully compatible on this path, because filters added by the event are not used when the final OR filters are built.
</details>

<details>

| S1 | Preserve source tuple filtering | Not addressed | No reply. The new `SystemId` path only handles entries with nonblank `Parent Source No.`, so the main customer path can still mix source tuple values. |
| S2 | Keep event timing compatible | Not addressed | No reply. The event timing is partly restored, but the non-correlated path still ignores filters that the event adds before the final filter is built. |
| S3 | Add regression and performance coverage | Not addressed | No reply. The PR still changes only `PriceListManagement.Codeunit.al`; no regression or performance test was added. |
</details>

<details>

*None - the remaining issues are the open round-1 blockers.*
</details>

<details>

**Risk:** High. This code controls which price list lines are shown in pricing review pages. If the filter is too broad, users can see or edit unrelated customer prices or discounts. The event behavior is also compatibility-sensitive for extensions.

**Necessity:** The work item is a valid bug. Opening the page can take about 220 seconds on large datasets, so the performance fix is needed. The fix must keep the same row set and event contract while avoiding the expensive `MarkedOnly` materialization.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9219 round=2 by=alexei-dobriansky sha=129c962b7071fd6940224f3d59d4411eb7738631

PR #9259, opened 2026-07-08
[Quality Management] [28.x] Enforce rules and restrictions with indirect and implicit permissions

JakovljevicDusan:
## What & why
Implement InherentPermissions so users without Quality Management permissions would not silently skip mandatory Quality Inspection creation or transaction restrictions.

Automatic Quality Inspection creation is now handled like this:
1. If there is no Quality Inspection Setup or there is no Generation Rule triggered on specific event, Quality Management code is skipped, and user does not need any Quality Management permission set. InherentPermissions for these two tables are used.

2. However, if there is trigger, then user has to have Quality Management permissions. Otherwise, user gets standard permission error, such as this:
image

3. Minimum is new permission set "QltyMgmt - I. Create"/'Quality Inspection - Create' which grants indirect permissions for creating inspections, without accessing or processing them further.
- This permission set grants permission to run creation inspection manually as well, via "Create Quality Inspection" actions. But, due to AccessByPermission, View actions are hidden if user has just that Quality Management permission set.

## Linked work
Fixes [AB#641763](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/641763)

### DIFF
--- a/src/Apps/W1/Quality Management/app/src/Configuration/GenerationRule/QltyInspecGenRuleMgmt.Codeunit.al
+++ b/src/Apps/W1/Quality Management/app/src/Configuration/GenerationRule/QltyInspecGenRuleMgmt.Codeunit.al
@@ -15,6 +15,10 @@ using Microsoft.QualityManagement.Utilities;
 /// </summary>
 codeunit 20405 "Qlty. Inspec. Gen. Rule Mgmt."
 {
+    Permissions =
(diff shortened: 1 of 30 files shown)


JakovljevicDusan:
Won't backport at the moment. It will be included in v29. Closing PR.

PR #9263, opened 2026-07-08
[SOA]: Bugbash for releases 28.x - Contact unable to find due to Qasim map to Megan, different names

tomasevicst:
**Issue**

_Bug #640657: Agent unable to find mapped contacts in Contact List_

## What & why

Fixes contact lookup and reply routing for Sales Order Agent tasks where the sender email is mapped to a different contact.

When an unknown sender is handled with "Use another contact once", the selected contact is stored in `SOA Task Contact Override`. The Contact List now participates in that flow through the new Contact List find-record event so the agent session can surface the mapped contact while browsing/searching contacts.

When an unknown sender is handled with "Use another contact always", the sender email is stored in the contact's `E-Mail 2` field. The Contact List shows `E-Mail 2` only during SOA agent sessions so agents can identify alternate email mappings without changing the standard non-agent view.

Reply handling now respects mapped contacts. If an input message has a mapped contact override, the reply stays in the original email thread but is addressed to the mapped contact, while preserving CC/BCC recipients from the originating incoming email. If there is no mapped contact, the existing Reply All behavior is used.

The SOA contact search subscriber is manually bound from `SOA Session Events`, matching the item search pattern, so the Contact List event is registered only for SOA agent sessions.

## Linked work

Fixes [AB#640657](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/640657)

## How I validated this

Validated the affected AL files with focused diagnostics:
- `SOAFiltersImpl.Codeunit.al`: no errors
- `SOASendReplies.Codeunit.al`: no errors
- `SOASessionEvents.Codeunit.al`: no errors

Also ran `git diff --check`; no whitespace issues were reported.

Known local validation note: `SOAContactSearchImpl.Codeunit.al` depends on the new base-app `Contact List.OnBeforeFindRecord` event. In a workspace without the updated base app symbols, diagnostics still report that event as missing. That is expected until the base app symbols include the new event.

Scenarios covered by inspection/focused validation:
- "Use another contact always" updates `E-Mail 2` and uses clearer confirmation text when replacing an existing value.
- Contact List displays `E-Mail 2` only in SOA agent sessions.
- SOA contact search events are manually bound only for SOA agent sessions.
- Mapped-contact replies use the mapped contact as the `To` recipient and preserve original CC/BCC recipients where the originating email message is still available.
- Non-mapped replies continue using the existing Reply All path.

## Risk & compatibility

Non-agent users keep the standard Contact List experience because the additional `E-Mail 2` column is visible only during SOA agent sessions.

The contact search event is manually bound from SOA session initialization, so the subscriber is not globally active outside SOA agent sessions.

Reply routing changes only when a mapped contact override exists. In that case, the reply is intentionally sent to the mapped contact instead of the original sender, while preserving CC/BCC recipients from the original incoming email when available. Existing Reply All behavior remains unchanged when there is no mapped contact.


src/Apps/W1/SalesOrderAgent/app/src/PageExtention/SOAContactListExt.PageExt.al
```diff
@@ -0,0 +1,75 @@
...
+    begin
+        if not IsAgentSessionInitialized then begin
+            IsAgentSession := SOAKPITrackAll.IsOrderTakerAgentSession(AgentTaskID);
+            IsAgentSessionInitialized := true;
+        end;
+
+        PopulateMappedEmail();
+    end;
+
+    local procedure PopulateMappedEmail()
```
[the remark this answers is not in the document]
tomasevicst: Applied.

src/Apps/W1/SalesOrderAgent/app/src/PageExtention/SOAContactListExt.PageExt.al
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    trigger OnAfterGetRecord()
+    var
+        SOAKPITrackAll: Codeunit "SOA - KPI Track All";
+    begin
+        if not IsAgentSessionInitialized then begin
+            IsAgentSession := SOAKPITrackAll.IsOrderTakerAgentSession(AgentTaskID);
+            IsAgentSessionInitialized := true;
+        end;
```
AleksandricMarko: is it necessary selected part to be executed for every record? Could it be done OnOpenPage?

src/Apps/W1/SalesOrderAgent/app/src/Integration/SOAContactSearchImpl.Codeunit.al
```diff
@@ -0,0 +1,43 @@
...
+        FindRecordContact(Rec, Which, Found, IsHandled);
+    end;
+
+    procedure FindRecordContact(var Rec: Record Contact; Which: Text; var Found: Boolean; var IsHandled: Boolean)
+    var
+        SOATaskContactOverride: Record "SOA Task Contact Override";
+        SOAKPITrackAll: Codeunit "SOA - KPI Track All";
+        AgentTaskID: BigInteger;
+        IsAgentSession: Boolean;
+    begin
```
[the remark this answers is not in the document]
tomasevicst: We can't mapp each message to different contact. This is happening once, so in this case Task Message ID is irrelevant, and that's why i used FindFirst().

src/Apps/W1/SalesOrderAgent/app/src/Validation/SOAFiltersImpl.Codeunit.al
```diff
@@ -250,14 +250,14 @@ codeunit 4305 "SOA Filters Impl."
         if ContactList.RunModal() <> Action::LookupOK then
             exit;
         ContactList.GetRecord(SelectedContact);
-        if SelectedContact."E-Mail" <> '' then
```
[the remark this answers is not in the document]
tomasevicst: We expose now "E-Mail 2" field on Contact List page, so agent will be aware of that field and will be able to find Contact that match field "E-Mail 2".

src/Apps/W1/SalesOrderAgent/app/src/Integration/SOAContactSearchImpl.Codeunit.al
```diff
@@ -0,0 +1,43 @@
...
+namespace Microsoft.Agent.SalesOrderAgent;
+
+using Microsoft.CRM.Contact;
+
+codeunit 4419 "SOA Contact Search Impl"
+{
+    Access = Internal;
+    InherentEntitlements = X;
+    InherentPermissions = X;
+
```
qasimikram, line 15: We should have manual binding of these events, like we do for item search. So only if you have SOA then these events will be registered

`EventSubscriberInstance = Manual;`

also check `codeunit 4304 "SOA Session Events"` how we manually bind events.

src/Apps/W1/SalesOrderAgent/app/src/Integration/SOASendReplies.Codeunit.al
```diff
@@ -89,13 +90,40 @@ codeunit 4581 "SOA Send Replies"
...
-        AddMessageAttachments(EmailMessage, OutputAgentTaskMessage);
 
-        exit(Email.ReplyAll(EmailMessage, SOASetup."Email Account ID", SOASetup."Email Connector"));
+        // Check if there's a mapped contact
+        MappedContactEmail := GetMappedContactEmail(InputAgentTaskMessage."Task ID", InputAgentTaskMessage.ID);
+
+        if MappedContactEmail <> '' then begin
+            // Reply in same thread but send to mapped contact
+            ToRecipients.Add(MappedContactEmail);
+            EmailMessage.CreateReply(ToRecipients, Subject, Body, true, InputAgentTaskMessage."External ID");
```
qasimikram: Should we also add cc and bcc from the origin incoming email?

qasimikram (changes requested):

**S2 - Key-usage inconsistency between the two override consumers.** `SOA Task Contact Override` has PK `(Task ID, Task Message ID)`. `SOAContactSearchImpl.FindRecordContact` filters on **Task ID only** + `FindFirst()`, returning an arbitrary (first) message's override, while `SOASendReplies.GetMappedContactEmail` uses `Get(TaskID, TaskMessageID)` for the specific message. For a task with multiple mapped messages, the Contact List can surface a different contact than the one the reply is sent to. Please align both on the same lookup semantics.

**S4 - Reply routing switches from ReplyAll to Reply, dropping other recipients.** When a mapped email exists the code now sends only to the mapped contact via `CreateReply`/`Reply`, silently dropping any original CC/other To recipients that `ReplyAll` previously included. Please confirm this is intended, add coverage for it, and call it out under Risk & compatibility.

**S6 - Stale PR description.** The description documents a superseded design ("Mapped Email" column, the SOAFiltersImpl rationale) that no longer matches the final diff (`E-Mail 2` field, OnFindRecord approach). Please update it and fill in the empty "What & why" / "How I validated" sections so reviewers are not misled. Minor: the `ContactAlreadyHasEmail2Qst` label text ("email address 2 %2") reads awkwardly.

Additionally, this is a non-trivial behavior change (email routing + a base-app integration event) with no tests included; adding coverage for the routing scenarios would strengthen confidence.

src/Apps/W1/SalesOrderAgent/app/src/Integration/SOAContactSearchImpl.Codeunit.al
tomasevicst: Applied.

src/Apps/W1/SalesOrderAgent/app/src/Integration/SOASendReplies.Codeunit.al
tomasevicst: Applied.

alexei-dobriansky:

<details>

This PR makes SOA use manual contact mappings and `E-Mail 2` mappings when it resolves the sender contact, filters contacts/customers, and sends replies. This is the right problem to solve for the bug: the agent must honor the contact selected by the user instead of searching only for the sender display name.

The fix covers many paths, but two paths now choose the first matching contact without proving that the match is unique and safe. That can turn a valid email-based lookup into a false-positive contact match, especially when `E-Mail 2` is duplicated or when one task has more than one contact override.

</details>

<details>

**S1** - *Handle duplicate E-Mail 2 matches*
Do not route replies from an `E-Mail 2` match unless exactly one contact matches that address. `SOASendReplies.GetMappedContactEmail` and `SOATaskMessage.GetSentMessageToAddress` ignore `ContactCount`, so `FindFirst()` can choose the wrong contact and send the reply to that contact's primary email.

**S2** - *Keep Contact List matching unambiguous*
Do not make `SOAContactSearchImpl.FindRecordContact` select the first override for the whole task. Overrides are stored per task message, and this handler also clears the current Contact List filters with `Rec.Reset()`. Keep the page filters and only handle the event when the mapping is unambiguous.

**S3** - *Add mapped-contact regression coverage*
Add regression coverage for the mapped-contact paths. Cover `Use another contact once`, `Use another contact always`, and duplicate `E-Mail 2` matches so the tests prove the agent uses the mapped contact without widening access to another customer. This is reply-routing and customer/contact selection logic, so a wrong match is a real customer-impact risk.

</details>

<details>

**Risk:** This code controls which contacts and customers the agent can see, which customer is used for quote creation, and which recipient gets an email reply. The new Contact List event exists in the changed BaseApp pages and the SOA subscriber is manually bound like the item-search pattern, but the subscriber clears filters and reply routing uses first-match `E-Mail 2` lookups. A wrong match can expose customer data or send a reply to the wrong contact.

**Necessity:** The bug is valid and important. The work item shows that the manual contact override is stored, but the agent does not use it. A fix is needed, but it should be narrowed so only unique, task-message-safe mappings are used.

</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9263 round=1 by=alexei-dobriansky at=2026-07-14T17:07:40Z lastSha=66c6a4509dc0ac4395b633a057461e4d3f96accf suggestions=S1,S2,S3

PR #9265, opened 2026-07-09

neeleshsinghal:


alexei-dobriansky:

<details>

This PR tries to speed up Price List Line Review opened from Customer Card by using `IsEmpty()` in search mode and `SetLoadFields("Price List Code", "Line No.")` before marking matching source lines. It also adds a UI test for `All Customers` price lines.

The change preserves the exact source tuple behavior: each `(Source Type, Parent Source No., Source No.)` is still filtered separately, and `OnBuildSourceFiltersOnBeforeFindLines` still runs after those filters are set. This avoids the independent OR-filter bug seen in PR 9219/9269. However, the main performance root cause remains: the non-search path still iterates and marks every matching `All Customers` line, then applies `MarkedOnly(true)`.

Relationship to the other PRs: PR 9219 was closed after round-2 Request Changes. PR 9265 is a different replacement/alternative, not the same diff. PR 9269 is still open for the same work item and is closer to the OR-filter approach from PR 9219, so it is a competing duplicate alternative.
</details>

<details>

**S1** - *Still materializes all source lines*
The non-search path still calls `FindSet()` and `Mark(true)` for every matching `All Customers` line, and `SetPriceListLineFilters` still applies `MarkedOnly(true)`. `SetLoadFields` reduces loaded columns, but it does not remove the 61k-row marking work that caused the 220s load. Replace this with an approach that avoids materializing the broad source set while still preserving each exact source tuple.

**S2** - *Test does not prove the fix*
`T220_SalesPriceLinesFromCustomerCardManyAllCustomersLinesAreNotMarked` only creates 10-25 `All Customers` lines and counts them. The old code already returns those lines, so the test can pass without this fix. Add a regression or performance test that fails on the old marking path and proves the large-source path is no longer materialized.

**S3** - *Close the duplicate alternative PR*
PR 9219 is closed, and PR 9269 is still open for the same ADO bug with a different OR-filter approach. Keep one replacement PR and close the other alternative. This avoids merging a conflicting fix for the same pricing path.
</details>

<details>

**Risk:** This is a pricing UI path, so incorrect filtering can show the wrong price list lines. This PR keeps the exact tuple filters and keeps the event contract, which is good. The remaining risk is that the performance bug is not actually fixed because the broad `All Customers` set is still fully marked.

**Necessity:** ADO bug 641061 is valid and important. Opening Sales Prices from Customer Card can take about 220 seconds on large datasets, so a fix is needed. The scope should stay in this filtering path, but it must remove the expensive materialization, not only reduce loaded fields.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9265 round=1 by=alexei-dobriansky sha=a86bd0d73b22e13940929511ccb740037e673f49

neeleshsinghal:
s1

Addressed

alexei-dobriansky:

<details>

The new commit adds a source pre-check so `SetPriceListLineFilters` only applies `MarkedOnly(true)` when more than one source tuple returns rows. This helps only when a single source tuple has data.

It does not fully fix the reported Customer Card scenario. If both customer-specific rows and `All Customers` rows exist, the pre-check sets `MarkingIsUsed` to true, and `BuildSourceFilters` still calls `FindSet()` and `Mark(true)` over the matching rows before applying `MarkedOnly(true)`. The test and duplicate PR state were not updated.
</details>

<details>

| S1 | Still materializes all source lines | Not addressed | "s1 Addressed", but the new code still marks all rows when more than one source tuple has data. |
| S2 | Test does not prove the fix | Not addressed | No test change since round 1. |
| S3 | Close the duplicate alternative PR | Not addressed | PR 9269 is still open; PR 9219 is closed. |
</details>

<details>

**S4** - *Source filter event now runs twice*
The new source pre-check calls `OnBuildSourceFiltersOnBeforeFindLines`, and the real build pass calls it again. Subscribers can now run twice in the non-search path. Use a separate pre-check hook or avoid raising the existing event during the count pass.
</details>

<details>

**Risk:** This is a pricing UI path, so incorrect filters can show wrong price list lines. The remaining performance risk is high because large `All Customers` rows can still be marked when another source tuple also has rows. The new pre-check also changes event behavior by calling the source filter event twice.

**Necessity:** ADO bug 641061 is valid and important. The reported Customer Card Sales Prices path can take about 220 seconds on large datasets, so a fix is needed. The fix must remove the broad materialization while keeping exact source tuple semantics, and one duplicate PR should be closed.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9265 round=2 by=alexei-dobriansky sha=363e2e3947ad3d611676274b6e3b3e41f6d2bba4

qasimikram:
## Review: this PR vs. #9269, and a suggestion to make it a complete perf fix

Both PRs target the same root cause: `SetPriceListLineFilters` -> `BuildSourceFilters` marks every matching line one-by-one and then calls `MarkedOnly(true)`, materializing the entire "All Customers" set (~61k lines) -> ~220s. Marking exists because a source list is a set of `(Source Type, Parent Source No., Source No.)` tuples that cannot be expressed as a single contiguous filter range.

Note this codeunit **already** contains the exact fix pattern for the sibling asset path (`BuildAssetFilters` + `CheckIfPriceListLineMarkingIsNeeded` + a `MarkingIsUsed` flag): skip marking / `MarkedOnly` when it is not needed.

### This PR (#9265) - mirror the existing asset-filter pattern
Adds `CheckIfPriceListLineMarkingIsNeededForSources`, counting non-empty source tuples. Marking is skipped only when exactly **one** source has lines (a single `SetRange` is then exact); falls back to marking for 0 or >=2.

- Structurally identical to the already-shipped `BuildAssetFilters` fix -> consistent, low review risk, proven.
- **Always correct**: the non-marking path is a single exact `SetRange`; the multi-source path keeps exact per-tuple marking.
- **Partial perf fix**: if the customer has its own lines *and* "All Customers" lines (>=2 non-empty sources), it still marks all 61k -> still slow. It fully fixes only the clean single-source case.

### #9269 - OR-combined filter
Collects non-empty tuples and builds combined `SetFilter("Source Type",'A|B')`, `SetFilter("Parent Source No.",...)`, `SetFilter("Source No.",...)`; 1 -> `SetRange`, >=2 -> OR filter. Also adds a `SetLoadFields` micro-opt in the asset loop.

- Eliminates `MarkedOnly` in **all** non-empty cases, including >=2 sources -> broader perf win.
- **Correctness risk (cartesian product)**: three independent `SetFilter`s match the cross-product of the value sets, not the exact tuple list. Because customer numbers, price-group codes, etc. share the `Code[20]` space, a collision (e.g. a customer numbered `GOLD` and a price group `GOLD`) makes lines like `(Customer,'','GOLD')` match even though that tuple was never a source. Marking / this PR never over-match.

### Verdict
This PR (#9265) is the safer, more maintainable choice: provably correct, reuses the established in-file convention, near-zero regression risk. #9269 is the more complete perf fix but introduces a latent, data-dependent correctness defect (cross-type `Source No.` collisions). Its `SetLoadFields` optimization is worth cherry-picking regardless.

---

## Suggestion: make #9265 a complete (not partial) perf fix

The "partial" limitation is fundamental: with >=2 non-empty sources this PR marks everything, including the 61k "All Customers" bucket, so the reported worst case (a customer that has its own lines *and* "All Customers" lines) is still slow.

The fix without inheriting #9269's cartesian-collision bug leans on an asymmetry in the data:

- The only huge bucket is always a **broad "All" source**, whose `Source No.` is `''`.
- **Specific** sources (the customer, its price/disc groups) always have a **non-empty** `Source No.`.

Empty vs non-empty `Source No.` perfectly segregates All-type lines from specific-type lines, so a combined OR-filter that unions the broad source with the specific ones is **collision-free for the broad source** - the part that actually causes the 220s. Phantoms like `(Customer,'','')` or `(All Customers,'','C10')` never match real data (customer lines always carry a source no; All-type lines never do).

### Enhancement
Keep this PR's structure and its marking fallback, but add a fast combined-filter path for the >=2 case, gated by a cheap in-memory collision check over the (small) source list:

```al
local procedure BuildSourceFilters(var PriceListLine: Record "Price List Line"; PriceSourceList: Codeunit "Price Source List"; var MarkingIsUsed: Boolean)
var
    PriceSource: Record "Price Source";
    TypeFilter, ParentFilter, NoFilter : TextBuilder;
    Sep: Text[1];
    NonEmptyCount: Integer;
begin
    // Safe to combine when no two tuples with different (Source Type, Parent Source No.)
    // share the same NON-EMPTY Source No. All-type sources have Source No = '' and never
    // collide with specific sources, so the huge bucket is always covered.
    if SearchIfPriceExists or not CanCombineSourceFilters(PriceListLine, PriceSourceList, NonEmptyCount) then begin
        MarkingIsUsed := (NonEmptyCount <> 1);
        BuildSourceFiltersByMarking(PriceListLine, PriceSourceList, MarkingIsUsed);  // current #9265 body
        exit;
    end;

    // Fast path: one combined filter, no MarkedOnly.
    MarkingIsUsed := false;
    if PriceSourceList.First(PriceSource, 0) then
        repeat
            PriceListLine.SetRange("Source Type", PriceSource."Source Type");
            PriceListLine.SetRange("Parent Source No.", PriceSource."Parent Source No.");
            PriceListLine.SetRange("Source No.", PriceSource."Source No.");
            OnBuildSourceFiltersOnBeforeFindLines(PriceListLine, PriceSource);
            if not PriceListLine.IsEmpty() then begin
                TypeFilter.Append(Sep + Format(PriceSource."Source Type"));
                ParentFilter.Append(Sep + GetFilterText(PriceSource."Parent Source No."));
                NoFilter.Append(Sep + GetFilterText(PriceSource."Source No."));
                Sep := '|';
            end;
        until not PriceSourceList.Next(PriceSource);
    ClearSourceFilters(PriceListLine);
    PriceListLine.SetFilter("Source Type", TypeFilter.ToText());
    PriceListLine.SetFilter("Parent Source No.", ParentFilter.ToText());
    PriceListLine.SetFilter("Source No.", NoFilter.ToText());
end;
```

`CanCombineSourceFilters` reuses the emptiness pass this PR already does, plus an O(k^2) loop over the k collected tuples (k is tiny) that returns false if two tuples with different `(Source Type, Parent Source No.)` share the same non-empty `Source No.`.

### What it buys
- **Bug scenario fully fixed**: customer + groups + "All Customers" with distinct codes -> combined filter, no marking, no 61k `MarkedOnly` materialization.
- **Correct**: the broad bucket is provably collision-free; the guard catches the specific-vs-specific code collision that #9269 misses.
- **Consistent**: reuses the existing `GetFilterText` OR-filter idiom already used in this codeunit (`SetHeadersFilters`), and keeps marking as the safe fallback.

### Honest caveat
One residual over-match cannot be eliminated cheaply: if a *different, unlisted* entity shares a code with one of the specific sources (e.g. a customer literally numbered `GOLD` while `GOLD` is also this customer's price group), the cartesian filter could surface that entity's lines - the same class as #9269, narrowed to specific-vs-specific. If provable correctness is preferred over the multi-specific-source perf edge, marking is the only exact method; #9265-as-is is then the floor and this enhancement is a pragmatic upgrade for the dominant case.

alexei-dobriansky:

<details>

Since round 2, the PR adds a combined source-filter path in `Price List Management`. It scans the source tuples that have data, then applies OR filters on `Source Type`, `Parent Source No.`, and `Source No.` so `MarkedOnly` can be skipped.

This removes the broad marking work for the common `Customer` + `All Customers` case. However, the three fields are combined independently. That can match price list lines for source tuples that were never in the source list.
</details>

<details>

| S1 | Still materializes all source lines | Addressed | Changed in `e6619b98`; the common path now skips `MarkedOnly`. See S5 for the new correctness issue. |
| S2 | Test does not prove the fix | Not addressed | No test change since round 2. |
| S3 | Close the duplicate alternative PR | Not addressed | PR 9269 is still open. |
| S4 | Source filter event now runs twice | Not addressed | The new combined path still raises `OnBuildSourceFiltersOnBeforeFindLines` during the scan and again during the final filter build. |

</details>

<details>

**S5** - *Combined source filters can over-match*
The new code builds separate OR filters for `Source Type`, `Parent Source No.`, and `Source No.`. This can return a cross-product, for example a `Customer` line whose customer number equals a customer price group code in the source list. Keep tuple-exact filtering, or prove each combined tuple exists before skipping marking.

</details>

<details>

**Risk:** The changed area controls which sales, purchase, job, contact, and campaign price lines are shown. A wrong match can show prices for the wrong source in the review page. The performance risk is lower than before for the common case, but the new tuple-combination path adds a pricing correctness risk.

**Necessity:** The bug is important: the linked work item reports about 220 seconds to open Sales Prices from Customer Card on large datasets. The fix is needed, but it must keep the source tuple filtering exact while removing the expensive marking path.

</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9265 round=3 by=alexei-dobriansky at=2026-07-10T11:08:18Z lastSha=e6619b988e84a5a5fb216207c77ed4f8b3af952e suggestions=S1:addressed,S2:notaddressed,S3:notaddressed,S4:notaddressed,S5:new parentRound=2

alexei-dobriansky:
## Agentic PR Review - Round 4

<details>

Since round 3, the source-filter logic was restructured. `BuildSourceFilters` now calls a new guard `CanCombineSourceFilters`. It only uses the combined OR-filter path when it is safe, and otherwise falls back to `BuildSourceFiltersByMarking`. The duplicate `OnBuildSourceFiltersOnBeforeFindLines` calls in the redundant scan loops were removed.

This is a real improvement. `CanCombineSourceFilters` builds a map of `Source No. -> (Source Type | Parent Source No.)`. If the same `Source No.` maps to a different type or parent, it returns false and the code marks instead of combining. This removes the common tuple-mixing case and keeps the fast path for the normal Customer Card sources.
</details>

<details>

| S1 | Still materializes all source lines | Addressed | The common case now uses server-side filters and skips `MarkedOnly`. |
| S2 | Test does not prove the fix | Not addressed | No test was added. A mixed-source regression test is now more important, because the combine guard is subtle. |
| S3 | Close the duplicate alternative PR | Not addressed | PR 9269 is still open for the same work item. One of the two should be closed. |
| S4 | Source filter event now runs twice | Addressed | The duplicate `OnBuildSourceFiltersOnBeforeFindLines` calls were removed from the extra scan loops. It now runs once per source in the active path. |
| S5 | Combined source filters can over-match | Addressed (with a small residual) | The new guard prevents combining when a `Source No.` maps to different type/parent, so the common mixing case now falls back to marking. See S6 for the remaining edge. |
</details>

<details>
<summary><strong>New observations (commits since round 3)</strong></summary>

**S6** - *Residual over-match for multi-parent sources*
The combine guard keys only on `Source No.`. When sources share the same type but different `Parent Source No.` (for example Job or Contact task pricing), the three OR filters can still cross-match, for example `Parent = JOB1` with a task code that belongs to `JOB2`. Also, `CanCombineSourceFilters` checks `IsEmpty` without raising `OnBuildSourceFiltersOnBeforeFindLines`, so its decision can differ from the event-filtered result. Consider covering the multi-parent case in the guard or a test.
</details>

<details>

**Risk:** The change is in shared price list line review filtering (sales, purchase, job, contact, campaign). The main correctness risk from round 3 is now much smaller because unsafe source sets fall back to marking. A narrow over-match risk remains for multi-parent Job/Contact pricing, and there is still no test that proves the combine path is tuple-safe.

**Necessity:** AB#641061 is clear and important, because opening Sales Prices from Customer Card can take about 220 seconds on large data. This fix now keeps the fast path safe for the common case. Please add a mixed-source regression test and close the duplicate PR 9269 before merge.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9265 round=4 by=alexei-dobriansky at=2026-07-10T17:04:19Z lastSha=e567961b6e4583dbaa6ce95fb0182f072907e062 suggestions=S1:addressed,S2:notaddressed,S3:notaddressed,S4:addressed,S5:addressed,S6:new parentRound=3

alexei-dobriansky:
## Agentic PR Review - Round 5

<details>

Since round 4, `CanCombineSourceFilters` was changed to allow the combined filter path only when all non-empty sources have the same `Parent Source No.`. This fixes the multi-parent part of S6, because Job or Contact sources with different parents now fall back to marking.

The change is still not tuple-safe. Same-parent sources with different `Source Type` and `Source No.` can still be mixed by the final three OR filters, so the review page can show price lines for source tuples that were never in the source list. `CanCombineSourceFilters` also still checks `IsEmpty()` without `OnBuildSourceFiltersOnBeforeFindLines`, so its decision can differ from the event-filtered source set.
</details>

<details>

| S1 | Still materializes all source lines | Addressed | The common Customer Card path still uses server-side filters and skips `MarkedOnly`. |
| S2 | Test does not prove the fix | Not addressed | No test was added; the diff since round 4 only changes `PriceListManagement.Codeunit.al`. |
| S3 | Close the duplicate alternative PR | Not addressed | PR 9269 is still open as the competing duplicate for the same work item. |
| S4 | Source filter event now runs twice | Addressed | The new commits do not reintroduce the duplicate event scan. |
| S5 | Combined source filters can over-match | Not addressed | The new guard only compares `Parent Source No.`. For same-parent sources, the final OR filters can still mix `Source Type` and `Source No.` values, for example `Customer` with a customer price group code. |
| S6 | Residual over-match for multi-parent sources | Not addressed | The different-parent part was fixed, but `CanCombineSourceFilters` still does not raise `OnBuildSourceFiltersOnBeforeFindLines`, so the combine decision can be based on a different source set than the final filtered result. |
</details>

<details>
<summary><strong>New observations (commits since round 4)</strong></summary>

*None as new S7 items. The new commits only change the prior combine guard, and the remaining problems are carried forward as S5 and S6.*
</details>

<details>

**Risk:** This code controls which price list lines are shown for sales, purchase, job, contact, and campaign sources. A tuple-mixing bug can show price lines for the wrong source. The event-filter mismatch also keeps a compatibility risk for subscribers that narrow or change the source filters.

**Necessity:** AB#641061 is still important because opening Sales Prices from Customer Card can take about 220 seconds on large data. The fix is needed, but it must keep source tuple filtering exact, add a mixed-source regression test, and avoid merging both this PR and duplicate PR 9269.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9265 round=5 by=alexei-dobriansky at=2026-07-13T05:40:40Z lastSha=60d7b8c6172121622f65eb755214bb447df4d3f0 suggestions=S1:addressed,S2:notaddressed,S3:notaddressed,S4:addressed,S5:notaddressed,S6:notaddressed parentRound=4

qasimikram (changes requested):
## Review

Performance fix replaces O(n) record-marking with a combined-filter fast path in `SetPriceListLineFilters(PriceSourceList,...)`, falling back to marking otherwise. Real perf win, all CI green, good reuse of the existing asset-side pattern, and a new test (T220). A few concerns before merge:

### S1 - Cross-product false matches in the combined path (High/Medium confidence)
`CanCombineSourceFilters` only guards that all non-empty sources share the same `Parent Source No.`, but the fast path then applies three independent OR filters: `Source Type in {..}` AND `Parent = ..` AND `Source No. in {..}`. With multiple distinct `(Source Type, Source No.)` pairs sharing a parent, the AND-of-ORs matches cross-product combinations that are not actual sources. In the customer-card case the types (Customer / All Customers / Cust. Price Group / Cust. Disc. Group) have practically disjoint `Source No.` domains, so it usually works, but it is not provably safe - e.g. a Customer No. equal to a Disc. Group code could surface an unintended line. The original per-source `SetRange` (exact) had no such risk. Please either tighten the guard (require a single `Source Type`, or only combine the field that actually varies) or document why cross-type combining is safe here.

### S2 - Extensibility regression on `OnBuildSourceFiltersOnBeforeFindLines` (Medium)
The original `BuildSourceFilters` raised this integration event immediately before `FindSet`/`Mark`. The new `BuildSourceFiltersByMarking` loop does not raise it - it now fires only inside `CheckIfPriceListLineMarkingIsNeededForSources` and the combined path. A subscriber that adds filters will see inconsistent behavior between the marking-needed check and the actual line selection, and the event contract changes for the marking path. Please restore the event inside the marking loop.

### S3 - Filter-character safety (Low/Medium)
The combined path builds filters by concatenating `Source No.` via `GetFilterText`, which only escapes the empty string, not filter metacharacters (`|`, `&`, `*`, `..`, `<>`). Codes containing such characters could broaden or break the filter; `SetRange` was immune. There is precedent in `SetSourceFilters`, so lower priority, but worth a look.

### Test gap
T220 covers only the single-source-type ("many All Customers") happy path - exactly the case that combines cleanly. Please add coverage for multiple source types sharing a parent (exercises S1) and the mixed-parent fallback path.

Recommendation: Request Changes, primarily on S1 (correctness on a core pricing path) and S2 (event regression). Both can be resolved quickly, or downgraded if the domain guarantees are confirmed and the event is restored.

alexei-dobriansky:
## Agentic PR Review - Round 6

<details>

Since round 5, `BuildSourceFilters` was changed to combine only sources that share the same `Source Type` and `Parent Source No.`, then OR only the `Source No.` values. That makes the fast path tuple-safe for the normal case, because it no longer builds independent OR filters across type, parent, and number.

The new commit also adds a mixed-source UI test and closes the duplicate PR 9269. However, the test only proves that valid mixed sources are shown; it does not add a negative row that would have failed the old cross-product bug. `CanCombineSourceFilters` still checks `IsEmpty()` without `OnBuildSourceFiltersOnBeforeFindLines`, and the marking fallback now raises that event in both the pre-check and the marking pass.
</details>

<details>

| S1 | Still materializes all source lines | Addressed | The common path still uses server-side filters and skips `MarkedOnly`. |
| S2 | Test does not prove the fix | Not addressed | `T221` was added, but it has no wrong-source row and would also pass with the earlier cross-product filter. |
| S3 | Close the duplicate alternative PR | Addressed | PR 9269, the competing duplicate, is now closed. |
| S4 | Source filter event now runs twice | Not addressed | The new marking fallback calls the event in `CheckIfPriceListLineMarkingIsNeededForSources` and again in `BuildSourceFiltersByMarking`. |
| S5 | Combined source filters can over-match | Addressed | The fast path now requires the same `Source Type` and `Parent Source No.`, and only combines `Source No.` values. |
| S6 | Residual over-match for multi-parent sources | Not addressed | The multi-parent part is fixed, but `CanCombineSourceFilters` still makes its decision without raising `OnBuildSourceFiltersOnBeforeFindLines`. |

</details>

<details>
<summary><strong>New observations (commits since round 5)</strong></summary>

*None as new S7 items. The new commit fixes S5 and closes the duplicate PR, but it leaves S2 and S6 open and reintroduces the S4 event issue in the marking fallback.*
</details>

<details>

**Risk:** This code controls which price list lines are shown for sales, purchase, job, contact, and campaign sources. The tuple-mixing risk is now much lower, but event subscribers can still see inconsistent behavior. If a subscriber narrows filters, `CanCombineSourceFilters` can choose the fast path from a broader raw set than the set used later.

**Necessity:** AB#641061 is still important because opening Sales Prices from Customer Card can take about 220 seconds on large data. The performance fix is needed and the duplicate PR is closed, but the event behavior and regression test gap should be fixed before merge.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9265 round=6 by=alexei-dobriansky at=2026-07-13T17:04:30Z lastSha=bf4e3bb2c779edd90519102ef07eebd0704322f0 suggestions=S1:addressed,S2:notaddressed,S3:addressed,S4:notaddressed,S5:addressed,S6:notaddressed parentRound=5

qasimikram (changes requested):
## Review: Request Changes

The optimization approach is sound - within a shared Source Type + Parent Source No., filtering `Source No.` by a value list is equivalent to the union of per-source filters, so avoiding per-record marking is a valid way to fix the ~220s regression. However there is a blocking issue plus a few correctness concerns.

### S1 (High, blocking) - The newly added test fails in CI across all localizations
`LegacyTestsBucket1` fails deterministically in every country build with:

> `T221_SalesPriceLinesFromCustomerCardMultipleSourceTypesSharingEmptyParent` - Assert.AreEqual failed. Expected:<4> (Integer). Actual:<3> (Integer). Not all applicable price lines are shown for the customer.

The mixed-source-type scenario (Customer / All Customers / Customer Price Group / Customer Disc. Group) returns only 3 of the 4 expected lines. The uniform failure across ~21 localizations rules out flakiness - this is a real functional gap. A likely cause is that the Customer Disc. Group line is a discount line (Amount Type = Discount) and is filtered out by the Amount Type filter applied when the review page opens, meaning either the production code drops it or the test's expectation of 4 is incorrect. Please resolve this before merge. (The other legacy failures in the bucket - VAT49Pct, GLTransfer, PullServiceLine - are unrelated to pricing and appear to be pre-existing bucket noise.)

### S2 (Medium) - Event not fired consistently during the combine decision
`CanCombineSourceFilters` determines emptiness and the shared type/parent **without** firing `OnBuildSourceFiltersOnBeforeFindLines`, whereas both the combine path in `BuildSourceFilters` and the marking path in `BuildSourceFiltersByMarking` **do** fire it. An extension that alters the line filters in that event could cause the combine decision (shared type/parent selection, non-empty determination) to disagree with the line set actually built, producing wrong or missing lines. Fire the event consistently in `CanCombineSourceFilters`, or document why it is intentionally skipped.

### S3 (Medium) - Empty combined filter can over-match
In the combine branch, if `NoFilter` ends up empty (possible when the event empties sources that `CanCombineSourceFilters`, which does not fire the event, counted as non-empty), `PriceListLine.SetFilter("Source No.", NoFilter.ToText())` with an empty string clears the filter and returns **all** lines for that Source Type + Parent Source No. Please guard against an empty `NoFilter` (fall back to marking, or apply an impossible filter).

### S4 (Low) - Increased event invocation count
`OnBuildSourceFiltersOnBeforeFindLines` now fires additional times (in `CheckIfPriceListLineMarkingIsNeededForSources` and the extra passes). Subscribers with side effects or assumptions about invocation count may be affected - worth noting in the PR description.

Once T221 passes (with the intended semantics confirmed) and S2/S3 are addressed or justified, this should be in good shape.

alexei-dobriansky:
## Agentic PR Review - Round 7

<details>

Since round 6, commit `17eeb187` adds `OnBuildSourceFiltersOnBeforeFindLines` inside `CanCombineSourceFilters`, before the `IsEmpty()` check. This fixes the earlier mismatch where the combine decision used raw filters while the final build used event-adjusted filters.

The event behavior is still not safe. The code now raises the same event in the combine pre-check and again in the real build path. If the code falls back to marking, the event can also run in `CheckIfPriceListLineMarkingIsNeededForSources` and again in `BuildSourceFiltersByMarking`. No test file changed since round 6.
</details>

<details>

| S1 | Still materializes all source lines | Addressed | Unchanged; the normal path still uses server-side filters and skips `MarkedOnly`. |
| S2 | Test does not prove the fix | Not addressed | No reply after round 6; no negative-row mixed-source test was added. |
| S3 | Close the duplicate alternative PR | Addressed | Unchanged; PR 9269 is closed. |
| S4 | Source filter event now runs twice | Not addressed | Commit `17eeb187` adds the event to `CanCombineSourceFilters`, but the same event still runs again in the final build or marking path. |
| S5 | Combined source filters can over-match | Addressed | Unchanged; the fast path still requires the same `Source Type` and `Parent Source No.`, and only ORs `Source No.`. |
| S6 | Residual over-match for multi-parent sources | Addressed | `CanCombineSourceFilters` now raises `OnBuildSourceFiltersOnBeforeFindLines` before `IsEmpty()`, so its decision uses the event-filtered set. |
</details>

<details>
<summary><strong>New observations (commits since round 6)</strong></summary>

*None as new S7 items. The new commit addresses S6, but S2 and S4 remain open.*
</details>

<details>

**Risk:** This code controls which price list lines are shown for sales, purchase, job, contact, and campaign sources. The tuple-mixing risk is now much lower. The remaining risk is compatibility and side effects for subscribers to `OnBuildSourceFiltersOnBeforeFindLines`, because the same source can still publish the event more than once.

**Necessity:** AB#641061 is still important because opening Sales Prices from Customer Card can take about 220 seconds on large data. The performance fix is needed, but the event behavior and regression-test gap should be fixed before merge.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9265 round=7 by=alexei-dobriansky at=2026-07-14T11:03:05Z lastSha=17eeb1870aa13754934ed4c8174f86e369b298e6 suggestions=S1:addressed,S2:notaddressed,S3:addressed,S4:notaddressed,S5:addressed,S6:addressed parentRound=6

qasimikram (dismissed):
**Summary:** Replaces the per-record `Mark(true)` loop in `BuildSourceFilters` with a combined `SetFilter("Source No.", ...)` when all matching price sources share the same Source Type + Parent Source No., falling back to marking (`BuildSourceFiltersByMarking`) for mixed sources. This mirrors the existing `BuildAssetFilters`/`CheckIfPriceListLineMarkingIsNeeded` pattern and addresses the ~220s slowdown from marking large "All Customers" datasets one record at a time.

**Assessment:** Correct and well-structured. The combine path is set-equivalent to marking because `CanCombineSourceFilters` only returns true when every non-empty source shares Source Type + Parent, so the union reduces cleanly to `(Type, Parent, Source No. IN {...})`. `MarkingIsUsed` plumbing and the single-source short-circuit faithfully follow the proven asset-filter code. `GetSafeFilterText` correctly escapes embedded quotes, and blank Source No. maps to `''` without triggering a false fallback.

**Suggestions (non-blocking):**

- **S1 (Medium):** `OnBuildSourceFiltersOnBeforeFindLines` side-effects are discarded in the combine path. The event fires per source during the emptiness scans, but its filters are cleared before the combined `SetRange`/`SetFilter` is applied. An extension subscribing to this event to restrict lines would be bypassed in the combine path, potentially showing more lines than intended - a behavioral divergence from the marking path, which honors the event. Please confirm this is acceptable given BC extensibility compatibility, or apply the shared portion of the event filters to the combined query.
- **S2 (Low):** The combine path scans sources twice (`CanCombineSourceFilters` + build loop) and the fallback adds another scan, so side-effecting event subscribers now fire more often. Still far cheaper than per-record marking.
- **S3 (Low):** Test gaps - no coverage for the single-source non-"All Customers" short-circuit (`MarkingIsUsed = false`) or for a Source No. containing a quote (`GetSafeFilterText` escaping). T220 (combine path) and T221 (mixed-type fallback) are good targeted additions.
- **S4 (CI):** An earlier run showed `Unit Tests (LegacyTestsBucket1)` failing uniformly across all countries - the bucket that contains the modified `ERM/PriceListsUI` tests. A fresh run is currently pending; please confirm it goes green before merge since the failure was in the exact area changed.

Logic and performance approach are sound; approving with the above suggestions.

qasimikram:
> ## Agentic PR Review - Round 7
> # **Recommendation: Request Changes**
> **What this PR does**
> Since round 6, commit `17eeb187` adds `OnBuildSourceFiltersOnBeforeFindLines` inside `CanCombineSourceFilters`, before the `IsEmpty()` check. This fixes the earlier mismatch where the combine decision used raw filters while the final build used event-adjusted filters.
> 
> The event behavior is still not safe. The code now raises the same event in the combine pre-check and again in the real build path. If the code falls back to marking, the event can also run in `CheckIfPriceListLineMarkingIsNeededForSources` and again in `BuildSourceFiltersByMarking`. No test file changed since round 6.
> 
> **Status of previous suggestions**
> ID	Title	Status	Author response
> S1	Still materializes all source lines	Addressed	Unchanged; the normal path still uses server-side filters and skips `MarkedOnly`.
> S2	Test does not prove the fix	Not addressed	No reply after round 6; no negative-row mixed-source test was added.
> S3	Close the duplicate alternative PR	Addressed	Unchanged; PR 9269 is closed.
> S4	Source filter event now runs twice	Not addressed	Commit `17eeb187` adds the event to `CanCombineSourceFilters`, but the same event still runs again in the final build or marking path.
> S5	Combined source filters can over-match	Addressed	Unchanged; the fast path still requires the same `Source Type` and `Parent Source No.`, and only ORs `Source No.`.
> S6	Residual over-match for multi-parent sources	Addressed	`CanCombineSourceFilters` now raises `OnBuildSourceFiltersOnBeforeFindLines` before `IsEmpty()`, so its decision uses the event-filtered set.
> **New observations (commits since round 6)**
> _None as new S7 items. The new commit addresses S6, but S2 and S4 remain open._
> 
> **Risk assessment and necessity**
> **Risk:** This code controls which price list lines are shown for sales, purchase, job, contact, and campaign sources. The tuple-mixing risk is now much lower. The remaining risk is compatibility and side effects for subscribers to `OnBuildSourceFiltersOnBeforeFindLines`, because the same source can still publish the event more than once.
> 
> **Necessity:** AB#641061 is still important because opening Sales Prices from Customer Card can take about 220 seconds on large data. The performance fix is needed, but the event behavior and regression-test gap should be fixed before merge.
> 
> > [AI-PR-REVIEW] version=1 system=github pr=9265 round=7 by=alexei-dobriansky at=2026-07-14T11:03:05Z lastSha=17eeb1870aa13754934ed4c8174f86e369b298e6 suggestions=S1:addressed,S2:notaddressed,S3:addressed,S4:notaddressed,S5:addressed,S6:addressed parentRound=6

Also review the 2 non addressed suggestions here.

qasimikram (commented):
I found two extensibility concerns that remain in the current head:

**S1 - Combined path drops subscriber filters**

`OnBuildSourceFiltersOnBeforeFindLines` affects the preliminary `IsEmpty()` checks, but the filter ultimately returned by the combined path is rebuilt using only `Source Type`, `Parent Source No.`, and `Source No.`. Any subscriber that adds another constraint, such as currency or validity dates, has that constraint silently removed from the final result. This can show price lines the subscriber intended to exclude. The existing discussion at `PriceListManagement.Codeunit.al:781` is resolved but not outdated, and the current implementation still exhibits the reported behavior.

**S2 - Published event now fires during probe passes**

The same event is invoked from `CanCombineSourceFilters` and `CheckIfPriceListLineMarkingIsNeededForSources`, then again during the actual build pass. This changes subscriber-visible firing semantics and can repeat subscriber side effects. The analogous asset path uses a separate `OnCheckIfPriceListLineMarkingIsNeededOnBeforeFindLines` event for its probe phase; the source path should preserve the same separation.

The added tests cover the base-app combined and marking paths, but neither exercises an event subscriber, so these regressions remain uncovered.

neeleshsinghal:
> ined path is rebuilt using only `Source Type`, `Parent Source No.`, and `Source No.`. Any subscriber that adds another constraint, such as currency or validity dates, has that constraint silently removed from the final result. This can show price lines the subscriber intended to exclude. The existing discussion at `PriceListMana`

Hi are we allowed to add new events

PR #9272, opened 2026-07-09
Fix duplicate in-transit location in localized Contoso demo data

attilatoury:
## Problem
In non-English Contoso demo data (confirmed Italian), two near-duplicate 'Own Logistics' in-transit locations are created (e.g. `LOG PROP.` and `LOG. PROP.`) instead of one.

## Root cause
The same logical in-transit location code was defined independently in two demo-data modules with separate translatable labels:
- `Create Location` (Inventory): `OwnLogLocationTok: Label 'OWN LOG.'`
- `Create Whse Location` (Warehousing): `TransitLocationTok: Label 'OWN LOG.'`

In English both resolve to `OWN LOG.` so the `Location.Get(Code)` duplicate guard skips the second insert. When translators produce even slightly different abbreviations for the two trans-units, the guard misses and two records are created.

## Fix
`Create Whse Location` now reuses `Create Location.OwnLogLocation()` as the single source of truth for the code, and the redundant `TransitLocationTok` label is removed. Only the location code is unified (the duplicate-guard key); the fix is safe regardless of module run order because the getter returns a constant and `InsertLocation` is insert-if-not-exists.

## Validation
- Confirmed `OwnLogLocation()` is an existing public `Code[10]` procedure in the same app.
- `TransitLocation()` has no callers outside this codeunit.
- Forward-fix only: does not clean up databases that already contain the duplicate.

Fixes [AB#640939](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/640939)


JakovljevicDusan (approved):

JakovljevicDusan (dismissed):

src/Apps/W1/ContosoCoffeeDemoDataset/app/DemoData/Warehousing/2.Master Data/CreateWhseLocation.Codeunit.al
```diff
@@ -273,7 +274,7 @@ codeunit 4787 "Create Whse Location"
 
     procedure TransitLocation(): Code[10]
     begin
-        exit(TransitLocationTok);
+        exit(CreateInvtLocation.OwnLogLocation());
```
qasimikram, line 277: S1 (non-blocking): Could we add an Italian-localized regression test confirming that the Inventory and Warehousing demo modules produce only one own-logistics in-transit location? The implementation correctly establishes a single source of truth; this would protect the localization-sensitive behavior from regressing.

qasimikram (dismissed):
The change correctly routes the localized value through the canonical `Create Location.OwnLogLocation()` source, and the existing helper retains duplicate protection. This should fix fresh generation and keep reruns idempotent.

S1 is a non-blocking suggestion for localized regression coverage.

**Recommendation: Approve with Suggestions**

PredragMaricic (dismissed):

alexei-dobriansky:

<details>

This PR makes the Warehousing demo-data in-transit location code come from Create Location.OwnLogLocation() instead of from its own translatable label. That removes the second translation unit that caused LOG PROP. and LOG. PROP. to be created as separate Italian demo-data locations.

The change addresses the root cause. Both Inventory transfer routes and Warehousing setup now point to the same in-transit location code, and both InsertLocation helpers are insert-if-not-exists when overwrite is off. This keeps the module order safe: if Inventory runs first, Warehousing reuses the existing code; if Warehousing runs first, Inventory later uses the same code and does not create a second record.
</details>

<details>

**S1** - *Add localized demo-data regression coverage*
Please add a small regression test if the test app can load the localized labels. It should prove that Inventory and Warehousing create one own-logistics in-transit location, or at least that TransitLocation() returns the same code as OwnLogLocation().
</details>

<details>

**Risk:** The code change is narrow and stays inside Contoso demo data. It changes the location code used by Warehousing setup to the Inventory code, but the existing Inventory transfer routes and transfer orders already use that same code, so the change does not orphan those references. Existing databases that already have the duplicate are not cleaned up by this PR, so this is a forward fix only.

**Necessity:** The linked Bug is clear and reproducible for Italian Contoso demo data. The change is needed because two separate translatable labels for the same logical in-transit location can diverge in any localization, while one shared source prevents that class of duplicate.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9272 round=1 by=alexei-dobriansky at=2026-08-03T11:23:12Z lastSha=a503aa280444cb74dac7337923ced9c6841d88c5 suggestions=S1

alexei-dobriansky (dismissed):

alexei-dobriansky:

# **Recommendation: Accept**

<details>

This PR fixes localized Contoso demo data by making the Warehousing in-transit location code come from Create Location.OwnLogLocation() instead of a separate translatable label. The ADO repro and PR description describe the issue clearly: in Italian, two translations for the same English label created LOG PROP. and LOG. PROP. as different in-transit locations.

The change addresses the root cause because there is now only one translatable code source for the duplicate guard key. Create Whse Location.TransitLocation() is still the Warehousing getter, and Create Location.OwnLogLocation() is a public, side-effect-free getter in the same app. The insert-if-not-exists behavior remains the same for English data and now uses the same key for localized data.
</details>

<details>

*None.*
</details>

<details>

**Risk:** The regression surface is limited to Contoso demo data setup for Inventory and Warehousing. The changed getter returns a different code only in languages where the two old labels diverged, which is the intended fix. It does not change a signature, posting, amounts, or BaseApp event flow, and the PR checks are green.

**Necessity:** The change is needed because localized demo data can persist duplicate in-transit locations and confuse transfer routes or location selection. The scope is right: it removes the duplicate code label and keeps the existing display name label in Warehousing.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=9272 round=1 by=alexei-dobriansky at=2026-08-12T08:15:49Z lastSha=0d68f51a17f47f0df93b3c2547736de64e2f92f8 reviewKey=c7b16f40989e25a7f05ffc187981068f4f74a918565840b493e87bbc0925a991 suggestions=

alexei-dobriansky (approved):

gggdttt (approved):

PR #9280, opened 2026-07-09
Project Task Billing Method reverts to original value in project card without validation or error

tomasevicst:
Fixes an issue where changing the "Task Billing Method" field on the Project Card from "One Customer" to "Multiple Customers" would visually revert to the original value without any error message.

The root cause is a race condition between the OnValidate trigger and the platform page refresh: after validation completes successfully, the platform issues a page refresh (GetPage) before the new value is committed to the database, causing the old value to overwrite the in-memory change.

The fix adds a **Modify()** call at the end of the OnValidate trigger to persist the record immediately after validation and **InitCustomerOnJobTasks()** complete, ensuring the updated value is in the database before the platform refreshes the page.

Added 3 tests to JobsMultipleCustomers.Codeunit.al for scenario 639677:
1. **TaskBillingMethodIsPersistedAfterValidateWithExistingTasks** — Core fix validation. Changes billing method on a project with tasks, then re-reads from DB to confirm the value was persisted by the **Modify()** call in the trigger.

2. **TaskBillingMethodIsNotPersistedWhenNoTasksExist** — Verifies the early-exit path still works: when no tasks exist, the trigger exits before **Modify()**, so the DB value is unchanged (page framework handles the save).

3. **TaskBillingMethodModifyNotCalledWhenValueUnchanged** — Verifies the **xRec** guard: when the value doesn't actually change, no unnecessary **Modify()** is called (checked via Last Date Modified not updating).

## Linked work

Fixes [AB#639677](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/639677/?view=edit)

## How I validated this


src/Layers/W1/BaseApp/Projects/Project/Job/Job.Table.al
```diff
@@ -500,6 +500,9 @@ table 167 Job
                         Error('');
 
                 InitCustomerOnJobTasks();
+
```
JakovljevicDusan, line 503: @tomasevicst
Would this work if page control OnValidate() has CurrPage.Update(true)? Although, new tests are no longer relevant with that solution.

JakovljevicDusan (dismissed):

src/Layers/W1/BaseApp/Projects/Project/Job/Job.Table.al
tomasevicst: Yes, that one works also. The problem is that it can't be reproduced. I was thinking that this is more reliable solution. But only my opinion.

AleksandricMarko (dismissed):

qasimikram (dismissed):
## Review

Adds a guarded `Modify()` at the end of the `"Task Billing Method"` field `OnValidate` in `Job.Table.al` so the value is persisted before the platform page refresh can revert it. Plus 3 tests.

The production fix is minimal, guarded, and mirrors an established precedent in the same table - the `Status` field `OnValidate` uses the identical `if xRec.<field> <> <field> then Modify();` pattern - so persisting inside `OnValidate` here is consistent with existing code. Approving, with suggestions to address before/after merge:

### S1 - Test `TaskBillingMethodModifyNotCalledWhenValueUnchanged` does not test what it claims (Medium/High)
Production `Modify()` uses `RunTrigger = false`, so the `OnModify` trigger (which sets `"Last Date Modified" := Today`) never runs on the forced write. The test asserts the guard prevented `Modify()` by checking `"Last Date Modified"` stayed at yesterday - but that assertion passes whether or not the `<> xRec` guard exists, because `Modify(false)` never updates `"Last Date Modified"` either way. The test gives false confidence and cannot catch a regression. Please verify it actually fails when the guard is removed; if not, switch to a real signal (subscriber-based spy, or observe an actual persisted difference).

### S2 - No-tasks path may be an incomplete fix, and the test codifies that (Medium)
When no tasks exist, `OnValidate` early-exits before `Modify()`. `TaskBillingMethodIsNotPersistedWhenNoTasksExist` asserts the DB is unchanged and treats that as correct, but it only checks the DB value, not that the in-memory/visual value survives the platform refresh (the actual reported symptom). If the refresh race applies to the no-tasks case too, the original revert is unfixed there. Please confirm the no-tasks scenario does not reproduce the revert, ideally with a test asserting the in-memory value is retained.

### S3 - `Modify(false)` skips business logic (Low, has precedent)
The forced write bypasses `OnModify` (`Last Date Modified`, `CheckRemoveFromMyJobsFromModify`, MyJobs maintenance), leaving `Last Date Modified` stale until the next full save. The `Status` field precedent does the same, so likely acceptable, but please confirm this is intended rather than `Modify(true)`.

Note: the build/test suite has not completed on this PR (several checks cancelled, one still in progress), so please rely on local verification for the new tests.

alexei-dobriansky:

<details>

This PR tries to persist `Job."Task Billing Method"` before `JobCard.Page.al` refreshes the page with `CurrPage.Update(false)`. For projects with existing tasks, the new `Modify()` runs after `InitCustomerOnJobTasks()`, while the error and cancel paths still stop before the save.

The fix does not cover all repro paths. The table trigger exits before the new `Modify()` when no project tasks exist, but the linked bug says the value reverts both with and without tasks. Since the page still calls `CurrPage.Update(false)`, the no-task case can still reload the old database value.

</details>

<details>

**S1** - *Early exit still leaves no-task case broken*  
The trigger exits before `Modify()` when there are no project tasks. The work item says the value also reverts in that case, and `JobCard.Page.al` still refreshes with `CurrPage.Update(false)`. Persist the changed billing method before this early exit too, and cover the no-task page path.

**S2** - *Explicit Modify bypasses normal modify logic*  
`Modify()` in the field trigger does not run the table `OnModify` trigger. This can leave `Last Date Modified` and other normal modify logic unchanged when this field is saved only by this trigger. Use the same behavior as a normal page save, or call `Modify(true)` if an explicit table save is still needed.

</details>

<details>

**Risk:** The affected field controls project task billing and invoice behavior, so saving the wrong value can affect later project invoicing. No event publisher or public API signature is changed, so there is no BaseApp event contract to verify. The main regression risk is the explicit save inside table validation, especially if it bypasses normal `OnModify` behavior.

**Necessity:** The linked ADO bug is valid and concrete. Users can select a billing method and see it silently revert, so the change is needed. The current scope is small, but it is incomplete because one documented repro path remains unfixed.

</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9280 round=1 by=alexei-dobriansky at=2026-07-13T11:05:14Z lastSha=a39d4a6e9279f5202b147088ffca6565f212023f suggestions=S1,S2

alexei-dobriansky:

<details>

The round-2 change updates the explicit save in `Job.Table.al` from `Modify()` to `Modify(true)`. This now runs the table `OnModify` trigger when the billing method is saved from the field validation path.

The no-task path is still unchanged. `JobTask.IsEmpty()` still exits before any explicit save, and `JobCard.Page.al` still calls `CurrPage.Update(false)` after validation. The ADO bug says the value reverts both with and without task lines, so one documented repro path still appears open.

</details>

<details>

| S1 | Early exit still leaves no-task case broken | Not addressed | No reply found after round 1. |
| S2 | Explicit Modify bypasses normal modify logic | Addressed | Fixed in `2e354873` by changing `Modify()` to `Modify(true)`. |

</details>

<details>

*None - no new issue found in commit `2e354873`.*

</details>

<details>

**Risk:** The remaining risk is the same as round 1. Projects without task lines still return before the explicit save, then the page refresh can reload the old database value. The field controls project task billing, so a silent revert can affect later invoicing setup. The change to `Modify(true)` reduces risk for the task-line path because normal `OnModify` logic now runs.

**Necessity:** The linked ADO work item is a Bug and gives a concrete repro. It explicitly says the behavior is the same with or without existing task lines. Persisting the changed value before the page refresh is still needed for both paths.

</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9280 round=2 by=alexei-dobriansky at=2026-07-14T05:45:35Z lastSha=cac0fe7534ecea718525569ffcc3dd996f55eb34 suggestions=S1:notaddressed,S2:addressed parentRound=1

AleksandricMarko (approved):

JakovljevicDusan (approved):

PR #9281, opened 2026-07-09
Add read isolation support to Isolated Storage Management 

jhejlsberg:
## What & why

Adds `Get` overloads to the **Isolated Storage Management** codeunit that accept an `IsolationLevel` for both `Text` and `SecretText` values.

This allows callers to explicitly control read isolation while preserving the existing key-length validation and output-clearing behavior.

## Linked work

Fixes [AB#630565](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/630565)

## How I validated this

No dedicated tests were added because the overloads directly delegate to the existing platform `IsolatedStorage.Get` implementation without introducing new storage logic. 

## Risk & compatibility

None


darjoo (approved):

ehenriksen9 (approved):

src/Layers/W1/BaseApp/System/IsolatedStorage/IsolatedStorageManagement.Codeunit.al
```diff
@@ -55,5 +55,20 @@ codeunit 1293 "Isolated Storage Management"
...
+
+    [NonDebuggable]
+    [Scope('OnPrem')]
+    procedure Get("Key": Text; Datascope: DataScope; IsolationLevel: IsolationLevel; var Value: Text): Boolean
+    begin
+        Value := '';
+        exit(ISOLATEDSTORAGE.Get(CopyStr(Key, 1, 200), Datascope, IsolationLevel, Value));
+    end;
+
+    [Scope('OnPrem')]
```
AndersLarsenMicrosoft: Should there be a   [NonDebuggable] here aswell?

PR #9295, opened 2026-07-09
Fixes #9144 - External File Attachment - Migrating files fails

### LINKED ISSUE #9144: [Bug]: External File Attachment - Migrating files fails for Tenant Media referenced more than once.
### Describe the issue

When migrating Document Attachment records to external storage, multiple Document Attachment records can reference the same Tenant Media record. If the migration deletes the Tenant Media record immediately after processing the first attachment, subsequent attachments referencing the same media fail because the underlying media no longer exists.

The migration processes Document Attachment records individually.

For each attachment it:
1. Reads the file from Document Reference ID.
2. Uploads the file to Azure Blob Storage.
3. Marks the attachment as stored externally.
4. Deletes the associated Tenant Media record.

The problem is that Document Reference ID is a Media field, and multiple Document Attachment records may reference the same Tenant Media record.

A common example is when a document is posted. The attachment on the posted document frequently references the same Tenant Media record as the original document rather than creating a duplicate media record.

As a result:
- Attachment A is migrated successfully.
- The shared Tenant Media record is deleted.
- Attachment B, which references the same media, can no longer retrieve the file.
- The migration fails for Attachment B.

### Expected behavior

The migration should recognise that multiple Document Attachment records can reference the same Tenant Media record.
A Tenant Media record should only be deleted after every referencing Document Attachment has been successfully migrated.

### Steps to reproduce

1. Create a Purchase Invoice.
2. Add a document attachment.
3. Post the document.
4. Verify that both the original and posted Document Attachment records reference the same Tenant Media record (Document Reference ID.Id).
5. Run the external storage migration.
6. Observe that the first attachment is migrated successfully.
7. Observe that subsequent attachments referencing the same Tenant Media record fail because the media has already been deleted.

### Additional context

_No response_

dorianwoolger:
## What & why

Fixes a bug in the "DA External Storage Sync" migration (report 8752) where the shared Tenant Media record behind multiple Document Attachment rows could be deleted after the first attachment was migrated, causing subsequent attachments referencing the same media to fail.

DeleteFromInternalStorage now takes a pre-built Guid → reference count map and only physically deletes a Tenant Media record once its count reaches zero, i.e. once every Document Attachment referencing it has been migrated. The map is built with a single pass over Document Attachment (BuildMediaReferenceCounts) before each batch runs, rather than re-querying per record — Media fields can't be filtered or passed as parameters directly, so this avoids an O(n²) scan while staying correct across partial runs and Maximum Records to Process limits.

## Linked work

Fixes #9144 

## How I validated this

- [Y] I read the full diff and it contains only changes I intended.
- [Y] I built the affected app(s) locally with no new analyzer warnings.
- [Y] I ran the change in Business Central and confirmed it behaves as expected.
- [Y]I added or updated tests for the new behavior, or explained below why none are needed.

I started with a fresh Cronus company. With External file storage disabled.
I attached some documents to an un-posted invoice and posted it. Verified that the new attachments on the posted invoice and the posted receipt were sharing the same Tenant Media Record. 
I then enabled External File storage and ran the Sync with Destination to external storage and action of Move.
Let the report run.
The report ran without an failed files.
Checked that all attachments migrated correctly and that nothing was left un migrated.
Existing test updated to include the new MediaReferenceCounts Dictionary.

## Risk & compatibility

This change only affects the migration of document attachments from internal to external storage.

### DIFF
--- a/src/Apps/W1/External Storage - Document Attachments/Test/src/DAExtStorageImplTests.Codeunit.al
+++ b/src/Apps/W1/External Storage - Document Attachments/Test/src/DAExtStorageImplTests.Codeunit.al
@@ -122,6 +122,7 @@ codeunit 136820 "DA Ext. Storage Impl. Tests"
     var
         DocumentAttachment: Record "Document Attachment";
         DAExternalStorageImpl: Codeunit "DA External Storage Impl.";
+        MediaReferenceCounts: Dictionary of [Guid, Integer];
         Result: Boolean;
     begin
         // [SCENARIO] Delete from internal should succeed after file is uploaded externally
@@ -135,8 +136,11 @@ codeunit 136820 "DA Ext. Storage Impl. Tests"
         DocumentAttachment.SetRecFilter();
         DocumentAttachment.FindFirst();
 
+        // [GIVEN] A reference count map reflecting the current single reference to the media
+        DAExternalStorageImpl.BuildMediaReferenceCounts(MediaReferenceCounts);
+
         // [WHEN] Delete from internal is attempted
-        Result := DAExternalStorageImpl.DeleteFromInternalStorage(DocumentAttachment);
+        Result := DAExternalStorageImpl.DeleteFromInternalStorage(DocumentAttachment, MediaReferenceCounts);
 
         // [THEN] Delete should succeed
         Assert.IsTrue(Result, 'Delete from internal should succeed');
@@ -324,6 +328,7 @@ codeunit 136820 "DA Ext. Storage Impl. Tests"
     var
         DocumentAttachment: Record "Document Attachment";
         DAExternalStorageImpl: Codeunit "DA External Storage Impl.";
+        MediaReferenceCounts: Dictionary of [Guid, Integer];
         Result: Boolean;
     begin
         // [SCENARIO] Delete from internal storage should succeed for externally stored docs
@@ -334,8 +339,11 @@ codeunit 136820 "DA Ext. Storage Impl. Tests"
         DocumentAttachment."Stored Externally" := true;
         DocumentAttachment.Modify();
 
+        // [GIVEN] A reference count map reflecting the current single reference to the media
+        DAExternalStorageImpl.BuildMediaReferenceCounts(MediaReferenceCounts);
+
         // [WHEN] Delete from internal is attempted
-        Result := DAExternalStorageImpl.DeleteFromInternalStorage(DocumentAttachment);
+        Result := DAExternalStorageImpl.DeleteFromInternalStorage(DocumentAttachment, MediaReferenceCounts);
 
         // [THEN] Delete should succeed
         Assert.IsTrue(Result, 'Delete from internal should succeed');
@@ -351,6 +359,7 @@ codeunit 136820 "DA Ext. Storage Impl. Tests"
     var
         DocumentAttachment: Record "Document Attachment";
         DAExternalStorageImpl: Codeunit "DA External Storage Impl.";
+        MediaReferenceCounts: Dictionary of [Guid, Integer];
         Result: Boolean;
     begin
         // [SCENARIO] Delete from internal should fail for non-external document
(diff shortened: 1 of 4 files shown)


dorianwoolger:

duichwer:
I like the now used approach but wouldn't it be an alternative solution to simple clear the media Id on the the document attachment an then later use the platform capability to delete orphaned media ?

What is the advantage of deleting it right away?

dorianwoolger:
> I like the now used approach but wouldn't it be an alternative solution to simple clear the media Id on the the document attachment an then later use the platform capability to delete orphaned media ?
> 
> What is the advantage of deleting it right away?

I had not considered the FindOrphans method.  That could also be a good way to solve the bug.  The only thing there is it seems to be a blanket across the whole environment. Although removing orphaned media should be harmless, I could be removing media not just orphaned by the migration? Is it the job of the migration to also clear other orphaned media?

dorianwoolger:
Is there anything extra I need to do on this one to get it through?

alexei-dobriansky:
**Problem-solution fit: Partial**

This PR addresses a real bug in the move-to-external migration for shared Tenant Media records, and the reference-count approach is reasonable. The current patch is not ready to merge because the signature change was not carried through all callers and it also changes an existing public procedure in a breaking way.

### Suggestions
- **S1** (🔴 High) Fix stale callers after the signature change - This patch changes DeleteFromInternalStorage to require a dictionary, but the page action and the OnAfterInsert subscriber still call the old signature, and the page calls BuildMediaReferenceCounts with two arguments. As written, the app will not compile. Update every caller before merge.
- **S2** (🔴 High) Preserve backward compatibility for the public procedure - DeleteFromInternalStorage is a public procedure today, so making a new required parameter mandatory is a breaking API change for dependent extensions. Keep the old signature and forward it to the new logic, or add a compatible overload and obsolete the old form first.
- **S3** (🟠 Moderate) Add a direct shared-media regression test - The updated tests still cover only single-reference cases. Please add one test where two Document Attachment rows share the same Tenant Media record and the move path deletes the media only after the last attachment is processed.

### Risk & necessity
Not merging leaves the original bug in place: move-to-external migration can still fail when multiple attachments share one media record, so some files stay unmigrated. Merging as-is is riskier because the patch has compile-time and compatibility problems.

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=9295 round=1 by=alexei-dobriansky at=2026-08-26T22:29:50Z lastSha=663e9ad76699dc8d16b862ede8195b81c7664958 reviewKey=9caf244afa42387d819e08a8ad1d540961df98cb3770a38c87b7c3ce9a3869b6 suggestions=S1@1f4c9e2a,S2@7b2d5c41,S3@9a6e3b7d

alexei-dobriansky:

# **Recommendation: Accept**

<details>

This round updates every stale caller, restores the original public signature as an obsolete forwarding overload, and adds a direct regression test for two attachments sharing one Tenant Media record. The batch page now builds the reference-count map once and passes it through each deletion.
</details>

<details>

| S1 | Fix stale callers after the signature change | Addressed | The page and insert subscriber now call valid overloads. |`n| S2 | Preserve backward compatibility for the public procedure | Addressed | The original signature remains as an obsolete forwarding overload. |`n| S3 | Add a direct shared-media regression test | Addressed | A test now verifies the media survives the first deletion and is removed after the last reference. |
</details>

<details>

*None - the product and test changes directly address all previous findings.*
</details>

<details>

**Risk:** The migration path deletes shared media only after the final reference, and compatibility is preserved for dependent extensions. The package-version update is repository infrastructure rather than runtime behavior.

**Necessity:** The fix is needed so shared attachments can migrate to external storage without leaving files behind or deleting shared media too early.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=9295 round=2 by=alexei-dobriansky at=2026-09-14T22:31:13Z lastSha=ab59cc844c7caf9c064e7c0402ceccec5f4c9696 reviewKey=9756c42d9a29c611e5d56df664e81ae8a3fb415f9fdde24f553c9f57494d8211 suggestions=S1@1f4c9e2a:addressed,S2@7b2d5c41:addressed,S3@9a6e3b7d:addressed parentRound=1

PR #9296, opened 2026-07-09
Add test stability mode to the test tool

nikolakukrika:
## Summary

Adds a **test stability mode** to the test tool. It re-runs an existing AL test suite under one or more **presets** (individually or combined) to surface flaky, order-dependent and data-dependent tests, and captures the outcome of **every** test method (pass and fail) — including the preset combination, seed, WorkDate, execution order, error message and call stack — so failures are easy to reproduce and troubleshoot.

Runnable from both the **UI** (AL Test Tool) and **PowerShell / CI** (Command Line Test Tool).

## Presets

Configured as `+`-separated CODE tokens (one string = one combination):

| Token | Effect |
| --- | --- |
| `BASELINE` | No preset |
| `SEED-<n>` | Force pseudo-random seed n in `Any` and `Library - Random` (even when a test hardcodes its seed) |
| `WORKDATEFUTURE-<n>YEAR` / `-<n>MONTH` | Shift `WorkDate` into the future |
| `ONEBYONE` | Run each test method in isolation |
| `REVERSE-CODEUNITS` / `REVERSE-METHODS` | Reverse execution order |

Example: `SEED-2+WORKDATEFUTURE-1YEAR+REVERSE-METHODS`. A default, editable set is created per suite.

## What's included

- New `Stability` folder in the **Test Runner** module (minimal-change, Code-Coverage-style): single-instance context, preset parser, **Stability Run Configuration** + **Stability Run Result** tables, event subscribers (`OnBeforeTestMethodRun`/`OnAfterTestMethodRun`), orchestrator, two pages, and page extensions on the **AL Test Tool** and **Command Line Test Tool**.
- Seed override lives in the **Any** app (new `Any Seed Override` codeunit); `Any` and `Library - Random` consult it, so `SEED-<n>` is effective even for tests that hardcode a seed. The Test Runner app takes a dependency on `Any`.
- PowerShell entry point `build/scripts/StabilityTests/RunStabilityTestsInBcContainer.ps1` drives a run and writes a JSON artifact (shape mirrors `unstable-tests.json`).
- `README.md` documenting the feature, tokens and grammar.

## Order & WorkDate handling

Execution order is realized purely by `Line No.` numbering when cloning the suite (no Test Runner change). The WorkDate offset is re-applied before every test method and coordinates with `ALTestRunner Reset Environment` (which restores WorkDate after each codeunit).

## Notes

- Complementary to the existing **Test Tolerance** feature (this *discovers* instability; tolerance *tolerates* known-unstable tests).
- Wiring stability mode into a GitHub workflow is intentionally deferred to a follow-up PR.
- Draft: opened for early review; CI provides the authoritative AL build.


src/Layers/W1/Tests/ApplicationTestLibrary/LibraryRandom.Codeunit.al
```diff
@@ -93,18 +93,27 @@ codeunit 130440 "Library - Random"
     end;
 
     procedure SetSeed(Val: Integer): Integer
+    var
+        AnySeedOverride: Codeunit "Any Seed Override";
     begin
         // Set the random seed to reproduce pseudo random sequence
+        if AnySeedOverride.IsActive() then
```
nikolakukrika: If tests are setting seeds automatically we should not override this setting. It should be kept. 
Only if the seed is not set we should use override.

src/Tools/Test Framework/Test Libraries/Any/src/Any.Codeunit.al
```diff
@@ -232,7 +232,11 @@ codeunit 130500 "Any"
     /// </summary>
     /// <param name="NewSeed">New seed to be used.</param>
     procedure SetSeed(NewSeed: Integer)
+    var
+        AnySeedOverride: Codeunit "Any Seed Override";
     begin
+        if AnySeedOverride.IsActive() then
+            NewSeed := AnySeedOverride.GetSeed();
```
nikolakukrika: Same comment here, we should not override if it is set. 
We should update the libraries if they use set seed concept.

src/Tools/Test Framework/Test Libraries/Any/src/AnySeedOverride.Codeunit.al
```diff
@@ -0,0 +1,59 @@
...
+namespace System.TestLibraries.Utilities;
+
+/// <summary>
+/// Single instance store used by the test stability tools to force a specific pseudo-random seed.
+/// When an override is active, both "Any" and "Library - Random" use the override seed even when a
+/// test hardcodes its own seed (for example LibraryRandom.SetSeed(1) inside Initialize).
+/// This makes it possible to re-run a suite with different random values in order to surface
+/// data-dependent (flaky) tests.
+/// </summary>
+codeunit 130501 "Any Seed Override"
```
nikolakukrika: Ideally, we should not have an override but a system that sets the seed if possible. Please list all of the users of the setseed that we have in the libraries.

src/Tools/Test Framework/Test Runner/src/Stability/StabilityALTestTool.PageExt.al
```diff
@@ -0,0 +1,90 @@
...
+                        StabilityRunResults: Page "Stability Run Results";
+                        SuiteName: Code[10];
+                    begin
+                        SuiteName := GetCurrentSuite();
+                        StabilityTestMgt.RunStabilityTests(SuiteName);
+                        StabilityRunResults.Run();
+                    end;
+                }
+
+                action(StabilityConfiguration)
```
nikolakukrika: Ideally this should be a field called configuration. If it is blank then we have not set anything.

src/Tools/Test Framework/Test Runner/src/Stability/StabilityCmdLineTestTool.PageExt.al
```diff
@@ -0,0 +1,78 @@
...
+{
+    layout
+    {
+        addlast(content)
+        {
+            group(StabilityGroup)
+            {
+                Caption = 'Stability';
+
+                field(StabilitySuiteName; StabilitySuiteName)
```
nikolakukrika: Same comment here, we should use configuration instead of the suite name and actions.

src/Tools/Test Framework/Test Runner/src/Stability/StabilityPreset.Codeunit.al
```diff
@@ -0,0 +1,127 @@
...
+/// <list type="bullet">
+/// <item>BASELINE - no preset (the empty string is equivalent).</item>
+/// <item>SEED-&lt;n&gt; - force pseudo-random seed n in Any and Library - Random.</item>
+/// <item>WORKDATEFUTURE-&lt;n&gt;YEAR / WORKDATEFUTURE-&lt;n&gt;MONTH - shift WorkDate into the future.</item>
+/// <item>ONEBYONE - run each test method in isolation.</item>
+/// <item>REVERSE-CODEUNITS - run the test codeunits in reverse order.</item>
+/// <item>REVERSE-METHODS - run the test methods within a codeunit in reverse order.</item>
+/// </list>
+/// </summary>
+codeunit 130469 "Stability Preset"
```
nikolakukrika: lets refactor everything to use Test Configurations.
Test configurations should be extensible, each stability test should add their own things to the test configuration.

nikolakukrika:
Reworked based on the review feedback. Force-pushed a follow-up commit that reworks both areas.

### Seed — no longer hijacks `SetSeed`
- Reverted `Any` and `Library - Random` to `main` (no override logic).
- Added a passive **`Configured Random Seed`** store (codeunit 130501, in the `Any` app) that only holds a value.
- **`Reset State Before Test Run`** (130301) — which already runs before every test method and did `LibraryRandom.SetSeed(1)` — now reads the store and, when set, seeds both libraries via the **existing** `SetSeed` methods. Default `SetSeed(1)` is preserved when unset, and a test that sets its own seed still wins because that call runs after the reset.
- The provider only writes the store; no library behavior changes.

Infrastructure that seeds the random libraries (for reference): `Reset State Before Test Run` (130301), `Any` / `Library - Random` (`SetSeed`), `CALTestRunnerPublisher` / `CALTestRunner` / `CALCommandLineTestRunner`, and the internal `TestRunner` / `SnapTestRunner`. Individual tests calling `LibraryRandom.SetSeed(...)` in their own `Initialize()` are intentionally left untouched — they now override the configured seed rather than being overridden.

### Test Configuration — extensible provider framework
Replaced the `+`-separated config string, the separate config table and the suite-name-driven actions with an extensible framework modeled on DataDrivenTest:
- **`ITest Configuration Provider`** interface + extensible **`Test Configuration Provider`** enum.
- **`Test Configuration`** (header) + **`Test Configuration Line`** (a provider + its JSON settings) — each preset contributes its own part; an empty configuration applies nothing.
- Context / orchestrator / runner codeunits + 5 providers (Seed, WorkDateFuture, OneByOne, ReverseCodeunits, ReverseMethods), result table, pages, and page extensions.
- Adding a new behavior = implement the interface + add an enum value + add a line. No existing configuration has to change.

README updated under `src/Tools/Test Framework/Test Runner/src/TestConfiguration/`.

PR #9297, opened 2026-07-10
[Main]- Error The Bin Content does not exist.Not able to post an Inventory Pick for Assemble to Order

DeepsShukla:
Workitem-[Bug 642202](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/642202): [master] [ALL-E] Error: "The Bin Content does not exist...." Not able to post an Inventory Pick for Assemble to Order items

Fixes [AB#642202](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/642202)

**Issue:** "The Bin Content does not exist...." Not able to post an Inventory Pick for Assemble to Order items
**Cause:**  CheckQuantityInBinContentForTracking iterates all pick lines and calls WMSMgt.CheckWhseJnlLine (which validates existing bin content) on ATO lines — even though ATO output bin content is only created during assembly posting, not before the pick.
**Solution:**  Skip ATO pick lines (if not WarehouseActivityLine2."Assemble to Order") in the bin-content validation loop


alexei-dobriansky:

<details>

This PR changes `Whse.-Activity-Post` so `CheckQuantityInBinContentForTracking` skips warehouse activity lines marked `Assemble to Order`. It also adds a regression test for a sales order that mixes a normal lot-tracked item with an ATO output item on a new bin-mandatory location.

The direction is correct. Other warehouse code already treats ATO inventory pick lines differently, and the later posting path still checks the warehouse journal line after `Sales-Post` can create the assembly output bin content. But the procedure still exits before the loop when the first current activity line is an ATO item. That keeps the validation dependent on line order.

</details>

<details>

**S1** - *Remove the first-line ATO exit*  
`CheckQuantityInBinContentForTracking` still exits before the loop when the first activity line is an ATO item. Remove that header-level ATO exit and rely on the per-line `"Assemble to Order"` check. Otherwise, a mixed pick with the ATO line first skips bin-content validation for normal tracked lines; add a test for that order.

</details>

<details>

**Risk:** This is warehouse posting logic for inventory picks, item tracking, bin content, and ATO assembly output. The change is narrow and does not change public APIs or events, but the remaining line-order dependency can bypass the intended pre-post validation for non-ATO lines before source document posting. BCApps is single-layer, so there is no country-layer ripple.

**Necessity:** The bug is valid and important. Without a fix, users can be blocked from posting an Inventory Pick for an ATO sales order on a new bin-mandatory location because the ATO output bin content is not available until assembly posting creates it. The scope is right, but the implementation should make the skip per-line only.

</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9297 round=1 by=alexei-dobriansky at=2026-07-13T05:43:26Z lastSha=4f35dcaf7c80ec2afcbcf42775c2afebeaea99f1 suggestions=S1

v-ankitgoyal (dismissed):

qasimikram (changes requested):
## Review

**S1 - Remove the first-line ATO exit**

`CheckQuantityInBinContentForTracking` still exits before scanning the activity lines when the current first line's item is Assemble-to-Order. This leaves the new per-line skip dependent on line order: if the ATO sales line is first, the procedure returns and normal tracked lines never receive the bin-content validation.

Please remove the procedure-level ATO exit and rely on `WarehouseActivityLine2."Assemble to Order"` inside the loop. Add or reorder a regression test so the ATO line precedes the normal tracked line and verify that the normal line is still validated.

This is blocking because warehouse posting can bypass validation for non-ATO tracked lines based only on document line order.

DeepsShukla:
> ## Review
> **S1 - Remove the first-line ATO exit**
> 
> `CheckQuantityInBinContentForTracking` still exits before scanning the activity lines when the current first line's item is Assemble-to-Order. This leaves the new per-line skip dependent on line order: if the ATO sales line is first, the procedure returns and normal tracked lines never receive the bin-content validation.
> 
> Please remove the procedure-level ATO exit and rely on `WarehouseActivityLine2."Assemble to Order"` inside the loop. Add or reorder a regression test so the ATO line precedes the normal tracked line and verify that the normal line is still validated.
> 
> This is blocking because warehouse posting can bypass validation for non-ATO tracked lines based only on document line order.

Updated the same.

qasimikram (changes requested):
## Review

**S1 - Evaluate Order Tracking Policy per line**

`CheckQuantityInBinContentForTracking` still reads `WarehouseActivityLine."Item No."` before iterating and exits when that first item's `Order Tracking Policy` is `None`. For a mixed Inventory Pick where the first line is untracked and a later non-ATO line uses `Tracking & Action Msg.`, the later line never receives the bin-content validation. This preserves the same line-order dependency that the ATO change is intended to remove.

Please move the policy check into the per-line loop (or otherwise evaluate each line independently), alongside the per-line ATO condition.

**S2 - Make the ATO-first test detect the old behavior**

`InvtPickForATOLineBeforeNormalItemStillValidatesNormalLine` only verifies that posting succeeds and that a warehouse entry exists. The pre-change implementation also exits early and then posts the normal line, so this test would pass before the fix and does not prove that the normal line's pre-post bin-content validation ran.

Please make the normal line fail specifically when that validation executes, or add another observable assertion that distinguishes the validation path from ordinary successful posting.

This is blocking because warehouse posting validation remains dependent on document line order, and the added test does not protect the corrected line-order behavior.

DeepsShukla:
> ## Review
> **S1 - Evaluate Order Tracking Policy per line**
> 
> `CheckQuantityInBinContentForTracking` still reads `WarehouseActivityLine."Item No."` before iterating and exits when that first item's `Order Tracking Policy` is `None`. For a mixed Inventory Pick where the first line is untracked and a later non-ATO line uses `Tracking & Action Msg.`, the later line never receives the bin-content validation. This preserves the same line-order dependency that the ATO change is intended to remove.
> 
> Please move the policy check into the per-line loop (or otherwise evaluate each line independently), alongside the per-line ATO condition.
> 
> **S2 - Make the ATO-first test detect the old behavior**
> 
> `InvtPickForATOLineBeforeNormalItemStillValidatesNormalLine` only verifies that posting succeeds and that a warehouse entry exists. The pre-change implementation also exits early and then posts the normal line, so this test would pass before the fix and does not prove that the normal line's pre-post bin-content validation ran.
> 
> Please make the normal line fail specifically when that validation executes, or add another observable assertion that distinguishes the validation path from ordinary successful posting.
> 
> This is blocking because warehouse posting validation remains dependent on document line order, and the added test does not protect the corrected line-order behavior.

Updated the suggested code.

v-ankitgoyal (approved):

PredragMaricic (approved):

src/Layers/W1/Tests/SCM-Assembly/SCMWhseAsmToOrder.Codeunit.al
```diff
@@ -2210,6 +2210,133 @@ codeunit 137914 "SCM Whse.-Asm. To Order"
...
+        OutputSalesLine: Record "Sales Line";
+        SalesHeader: Record "Sales Header";
+        WarehouseEmployee: Record "Warehouse Employee";
+        WhseActivityHeader: Record "Warehouse Activity Header";
+        WhseActivityLine: Record "Warehouse Activity Line";
+        ComponentLotNo: Code[20];
+        NormalItemLotNo: Code[20];
+        OutputLotNo: Code[20];
+    begin
+        // [FEATURE] [AI test 0.4]
```
qasimikram, line 2237: S1 - Replace the generated test tag

`// [FEATURE] [AI test 0.4]` is an unedited placeholder and does not identify the tested feature area. Please replace it with the established feature tags for this scenario, for example `[SCM] [Assembly] [Assembly to Order] [Bin] [Item Tracking]`.

qasimikram (approved):
## Review

The production change correctly removes both first-line dependencies and now evaluates each non-ATO activity line independently. The regression test reproduces Bug 642202, and the change is narrowly scoped to warehouse bin-content validation.

**S1 - Replace the generated test tag**

The new test still uses `[FEATURE] [AI test 0.4]`, which appears to be a generated placeholder rather than the established feature-area tags. I left an inline suggestion for this non-blocking cleanup.

**Risk:** Low-to-moderate. This affects warehouse posting validation, but ATO lines are excluded explicitly while normal tracked lines retain the existing validation.

PR #9303, opened 2026-07-10
Inventory Valuation - Excel Layout

### LINKED ISSUE #9300: [BC Idea]: Excel Layout for the Inventory Valuation report.
### BC Idea Link

Request by @KennieNP

### Description

- Create an Excel Layout
- Retain the RDLC layout
- Apply optimisations

microsoft/BusinessCentralApps#1705

daniel-williams-fenwick:
#### Summary 
For this report, we have:
- Created an Excel layout
- Applied some optimisations to significantly reduce report generation time
- Retained the RDLC layout

Outputs prior to changes:
[OG Inventory Valuation (RDLC) - No Expected Cost.pdf](https://github.com/user-attachments/files/24869569/OG.Inventory.Valuation.RDLC.-.No.Expected.Cost.pdf)
[OG Inventory Valuation (RDLC) - Expected Cost.pdf](https://github.com/user-attachments/files/24869570/OG.Inventory.Valuation.RDLC.-.Expected.Cost.pdf)

Outputs:
[Inventory Valuation (RDLC) - Expected Cost (4).pdf](https://github.com/user-attachments/files/24869581/Inventory.Valuation.RDLC.-.Expected.Cost.4.pdf)
[Inventory Valuation (RDLC) - No Expected Cost (4).pdf](https://github.com/user-attachments/files/24869582/Inventory.Valuation.RDLC.-.No.Expected.Cost.4.pdf)
[Inventory Valuation (Excel) - Expected Costs.xlsx](https://github.com/user-attachments/files/24869583/Inventory.Valuation.Excel.-.Expected.Costs.xlsx)
[Inventory Valuation (Excel) - No Expected Costs.xlsx](https://github.com/user-attachments/files/24869584/Inventory.Valuation.Excel.-.No.Expected.Costs.xlsx)

Tests (did not use preprocessor symbols as we're retaining the RDLC for this one, so we haven't used any CLEAN tags)
[Inventory Valuation - Pre Changes - AL Test Tool.xlsx](https://github.com/user-attachments/files/24869562/Inventory.Valuation.-.Pre.Changes.-.AL.Test.Tool.xlsx)
[Inventory Valuation - WITHOUT preprocessor symbols - AL Test Tool.xlsx](https://github.com/user-attachments/files/24869566/Inventory.Valuation.-.WITHOUT.preprocessor.symbols.-.AL.Test.Tool.xlsx)

#### Work Item(s) 
Fixes #9300 
Fixes [AB#616682](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/616682)

microsoft/BusinessCentralApps#1744


alexei-dobriansky:

<details>

This PR adds an Excel layout for report 1001 "Inventory Valuation", keeps the RDLC layout in the rendering section, and changes the report calculation to use Item FlowFields for better performance. The feature is useful, but the current diff is not safe to merge. The W1 report now reads FlowFields that are added only in the CH Item table, existing integration events no longer control the main value calculation, and the Excel layout does not define number formats for the visible amount and quantity pivot values.
</details>

<details>

**S1** - *Add FlowFields to the W1 Item table*
The W1 report uses fields such as "Opening Bal. ILE Qty.", "Increases ILE Qty.", and "Cost Posted To G/L", but the PR adds these fields only to src/Layers/CH/BaseApp/Inventory/Item/Item.Table.al. Add the FlowFields to the W1 Item table, or stop using them from the W1 report. As written, the W1 report cannot compile.

**S2** - *Keep existing event filter behavior*
OnItemOnAfterGetRecordOnAfterValueEntrySetInitialFilters and OnCalculateItemOnBeforeAssignDecreaseAmounts used to let subscribers change the Value Entry filters before the sums were calculated. The new FlowField sums ignore those subscriber changes for opening, increase, decrease, and G/L amounts. Keep the old extension behavior or add a compatible replacement before changing the calculation path.

**S3** - *Format Excel amounts and quantities*
The visible Excel pivot tables have no 
umFmtId on their value fields, and the visible report sheets use General formatting for amount and quantity cells. Add explicit number formats for LCY amounts and quantities so the Excel layout shows stable decimals and separators across cultures. This report is financial, so raw General formatting is too risky.

**S4** - *Do not make public Item procedures local*
UpdateMyItem in the CH Item table changes from a public procedure to a local procedure. That is a breaking API change for any extension that calls it. Keep the procedure public unless there is a documented obsoletion and replacement path.
</details>

<details>

**Risk:** high. Inventory Valuation is a financial reconciliation report, and wrong or unformatted amounts can mislead users. The RDLC layout is still present, but the W1 report object is currently coupled to fields that exist only in CH, and public event/API behavior changes can break extensions.

**Necessity:** the linked work item is a Deliverable for modernizing inventory reports, so an Excel layout is a valid feature increment. The scope is justified, but it needs to compile in W1, preserve extension behavior, and format financial values before merge.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9303 round=1 by=alexei-dobriansky at=2026-08-03T11:19:07Z lastSha=469734ff90d1919ed549e0583b389bc82cebc939 suggestions=S1,S2,S3,S4

alexei-dobriansky (changes requested):
Please review the suggestions

daniel-williams-fenwick:
@alexei-dobriansky

S1 - Fixed.
S2 - Fixed.
S3 - Will not change. After lengthy discussions with Kennie and the MS team, it was agreed formatting would come from the BC report object. Excel is not smart enough to apply region agnostic formats.
S4 - Fixed.

alexei-dobriansky:

<details>

Since round 1, the PR adds 17 FlowFields and 3 FlowFilters to the W1 Item table so the Inventory Valuation report can use them without depending on country-layer fields. The report's `OnPreDataItem` trigger sets filter ranges on the new FlowFilter fields and calls `SetAutoCalcFields`, replacing three separate `ValueEntry.CalcSums` loops with a single per-item FlowField read. Ending values (invoiced, expected, expected-not-invoiced) are now computed as discrete variables instead of inline column expressions. A new `SkipZeroLines` request page option is added. The test codeunit drops the old Excel-validation tests (which relied on `LibraryReportValidation`) and replaces them with `LibraryReportDataset` equivalents that work with the XML dataset output.

The core calculation is correct: opening balance, increases, and decreases are read from the FlowFields that were missing in round 1 and are now properly added. The FlowFilter date ranges (`Opening Bal. Date Filter`, `Inv. Val. Period Date Filter`, `Closing Bal. Date Filter`) are set correctly in `OnPreDataItem`. Transfer entries are still handled through the existing `ValueEntry` loop, which is the right approach. The `SkipZeroLines` guard correctly applies after the `IsEmptyLine` check.
</details>

<details>

| S1 | Add FlowFields to the W1 Item table | Addressed | Fields 123–139 added to `src/Layers/W1/BaseApp/Inventory/Item/Item.Table.al` |
| S2 | Keep existing event filter behavior | Disputed | "Fixed" — `OnCalculateItemOnAfterAssignDecreaseAmounts` now passes all increase/decrease amounts as `var` parameters, letting subscribers adjust final values. However, `OnCalculateItemOnBeforeAssignDecreaseAmounts` passes `Item` by value, so subscribers can no longer change the FlowFilter fields on Item to alter what FlowFields return. See new observation S5. |
| S3 | Format Excel amounts and quantities | Disputed | "After lengthy discussions with Kennie and the MS team, it was agreed formatting would come from the BC report object. Excel is not smart enough to apply region agnostic formats." |
| S4 | Do not make public Item procedures local | Addressed | The CH layer file (`Item.Table.al` in CH) is no longer part of the PR diff. |
</details>

<details>

**S5** - *Document the changed contract of `OnCalculateItemOnBeforeAssignDecreaseAmounts`*
The `Before` event fires before the new FlowField reads, but `Item` is passed by value, so a subscriber that changes FlowFilter fields on `Item` inside this handler will not affect the FlowField results. The `After` event has `var` parameters for all computed amounts and is the correct place for subscribers to adjust values. Add a comment on the `Before` event stub (or in its documentation) noting that ValueEntry filter changes no longer affect the main calculation, so existing subscribers are aware that any FlowFilter adjustment must be done before `OnPreDataItem`. This avoids silent behavior breaks for extensions that wrote handlers for the old contract.
</details>

<details>

**Risk:** medium. The core correctness concerns from round 1 are addressed. The remaining risk is the silent contract change on `OnCalculateItemOnBeforeAssignDecreaseAmounts` — any extension that subscribed to this event to change ValueEntry filters will now silently have no effect on opening, increase, and decrease amounts. The `After` event with `var` params provides an alternative, but existing subscribers will not know to switch. The `SkipZeroLines` feature is additive and safe.

**Necessity:** unchanged from round 1. This is a Deliverable-level feature for modernizing the Inventory Valuation report with an Excel layout and a FlowField-based calculation that is significantly faster on large datasets.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=1 system=github pr=9303 round=2 by=alexei-dobriansky at=2026-08-11T10:10:56Z lastSha=369c7375bca5782e19d0038c40508d1e6e7bc309 reviewKey=f4b48d26482dcb898b4083d822c4b3b2bf24c814fb7657aec93eca04c9ae4429 suggestions=S1@:addressed,S2@:disputed,S3@:disputed,S4@:addressed,S5@new parentRound=1

alexei-dobriansky (approved):

PredragMaricic (changes requested):
## Recommendation: Request Changes

The Excel layout is a useful addition, but the current head is not ready to merge.

**S1 - Run the required localization propagation**

`Verify App Changes` fails because the W1 Item fields and report changes have not been integrated into the corresponding localized copies. The workflow lists these missing files:

- `src/Layers/APAC/BaseApp/Inventory/Item/Item.Table.al`
- `src/Layers/CH/BaseApp/Inventory/Item/Item.Table.al`
- `src/Layers/ES/BaseApp/Inventory/Item/Item.Table.al`
- `src/Layers/GB/BaseApp/Inventory/Item/Item.Table.al`
- `src/Layers/IT/BaseApp/Inventory/Item/Item.Table.al`
- `src/Layers/NA/BaseApp/Inventory/Item/Item.Table.al`
- `src/Layers/NA/BaseApp/Inventory/Reports/InventoryValuation.Report.al`
- `src/Layers/NA/BaseApp/Inventory/Reports/InventoryValuation.rdlc`
- `src/Layers/RU/BaseApp/Inventory/Item/Item.Table.al`

Run `Invoke-Miapp` and include the generated changes. The functional build jobs were skipped after this validation failure.

**S2 - Preserve the public calculation and event contracts**

`CalculateItem(var Item)` remains public, but it now reads the new Item FlowFields directly. Those FlowFields are configured with date ranges and `SetAutoCalcFields` only in the report dataitem's `OnPreDataItem`. A caller that invokes `CalculateItem` directly bypasses that setup and can receive zero or stale totals.

The refactor also changes existing event behavior without changing the signatures. `OnItemOnAfterGetRecordOnAfterValueEntrySetInitialFilters` and `OnCalculateItemOnBeforeAssignDecreaseAmounts` still expose a mutable `ValueEntry`, but its filters no longer drive the opening, increase, decrease, or G/L sums. Existing subscribers that customize those filters can silently stop affecting the report. Move the required FlowFilter/FlowField setup into the calculation path and preserve the existing subscriber behavior, or introduce a versioned replacement contract.

**S3 - Cover the new default rendering and zero-line behavior**

This PR makes Excel the default layout, yet the affected tests were changed from workbook-level assertions to dataset-only assertions. That verifies the dataset but cannot catch broken workbook formulas, sheets, captions, or formatting. Keep at least one test that renders the new default Excel layout and validates its output.

The new `Skip Zero Lines` behavior is also untested: `InvtValuationRequestPageHandler` never sets `RequestSkipZeroLines`, so all current scenarios execute only the default false path. Add true/false scenarios proving a zero-balance item is skipped only when requested.

**S4 - Avoid decrementing the minimum date**

In `OnPreDataItem`, `StartDate = 00000101D` satisfies `StartDate > 0D`, so `CalcDate('<-1D>', StartDate)` attempts to move before the minimum date. The report already treats `00000101D` as a sentinel in `UpdateRequestPageFilterValues`. Use the same guard when setting `Opening Bal. Date Filter`.

**Risk:** High. This is a financial reconciliation report. The current head fails required source propagation, can return incorrect values through a public procedure and existing extensibility points, and lacks automated coverage for both the new default workbook and the new skip branch.

PR #9306, opened 2026-07-10
[Main] Unbalanced G/L Entries if you apply a Payment and a Posted Purchase Invoice having both different Posting Groups and using the "Applies-to Document No." on the line instead of using Apply Entries in the Spanish version.Initial commit

sanjmaurya:
[Bug 640410] Unbalanced G/L Entries if you apply a Payment and a Posted Purchase Invoice having both different Posting Groups and using the "Applies-to Document No." on the line instead of using Apply Entries in the Spanish version.

Fixes [AB#640410](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/640410)

Issuse :- Bug 640410: [Main] [all-e]Unbalanced G/L Entries if you apply a Payment and a Posted Purchase Invoice having both different Posting Groups and using the "Applies-to Document No." on the line instead of using Apply Entries in the Spanish version.

Cause :-

n GenJnlPostLine.Codeunit.al:

In CalcApplication, an application produces two detailed buffer entries (one for the payment ledger entry, one for the applied invoice). Both inherit "Document Type" = GenJnlLine."Document Type" (= Payment); they differ only by "CV Ledger Entry No.".
A Cartera special-case in PostDtldVendLedgEntry (added in PR 200509 / WI 557674) forced every such application entry to the payment line's posting-group payables account whenever "Applies-to Doc. No." <> '' and buffer "Document Type" = Payment. That condition wrongly matched the invoice-side entry too, so the invoice's closing posted to account 410 instead of 400 → 400 never cleared → imbalance.
"Apply Entries" sets Applies-to ID (not Applies-to Doc. No.), so the special-case was skipped — which is why only the inline path failed.

Solution :-
Added helper IsApplnEntryForAppliedVendorDoc that detects when an application detailed entry belongs to the applied document (its existing vendor ledger entry has a different Document Type than the payment line). The override now applies only to the payment's own entry (or an entry whose ledger entry isn't inserted yet), letting the applied invoice post to its own posting-group payables account via GetVendDtldCVLedgEntryBufferAccNo. Mirrored in PostDtldVendLedgEntryUnapply for apply/unapply symmetry.

### DIFF
--- a/src/Layers/ES/BaseApp/Finance/GeneralLedger/Posting/GenJnlPostLine.Codeunit.al
+++ b/src/Layers/ES/BaseApp/Finance/GeneralLedger/Posting/GenJnlPostLine.Codeunit.al
@@ -5631,7 +5631,9 @@ codeunit 12 "Gen. Jnl.-Post Line"
             if DetailedCVLedgEntryBuffer."Initial Document Type" = DetailedCVLedgEntryBuffer."Initial Document Type"::Bill then
                 AccNo := VendPostingGr."Payables Account"
             else
-                if (GenJournalLine."Document Type" = GenJournalLine."Document Type"::Payment) and (GenJournalLine."Applies-to Doc. No." <> '') and (DetailedCVLedgEntryBuffer."Document Type" = DetailedCVLedgEntryBuffer."Document Type"::Payment) then
+                if (GenJournalLine."Document Type" = GenJournalLine."Document Type"::Payment) and (GenJournalLine."Applies-to Doc. No." <> '') and (DetailedCVLedgEntryBuffer."Document Type" = DetailedCVLedgEntryBuffer."Document Type"::Payment) and
+                   (not IsApplnEntryForAppliedVendorDoc(GenJournalLine, DetailedCVLedgEntryBuffer))
+                then
                     AccNo := VendPostingGr."Payables Account"
                 else
                     AccNo := GetVendDtldCVLedgEntryBufferAccNo(GenJournalLine, DetailedCVLedgEntryBuffer)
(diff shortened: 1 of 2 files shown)


v-ankitgoyal (approved):

alexei-dobriansky:

# **Recommendation: Accept**

<details>

This PR narrows the ES-specific vendor application account override when a payment is applied through `Applies-to Doc. No.`. The guard uses the existing vendor ledger entry to detect the applied invoice-side detailed entry, so that side can use the invoice vendor posting group's payables account instead of the payment line posting group.

The code change matches the root cause from AB#640410. It keeps the payment-side fallback for the not-yet-inserted payment ledger entry, lets the applied document side use `GetVendDtldCVLedgEntryBufferAccNo(...)`, and adds the same guard shape to the unapply path. PR #8950 is closed but not merged, and `origin/main` still does not contain this fix. This PR is effectively the same accepted fix as #8950 final, apart from whitespace.
</details>

<details>

None.
</details>

<details>

**Risk:** This is sensitive ES vendor posting code because it controls payables account G/L entries during application and unapplication with multiple vendor posting groups. The diff is narrow, adds no public API or event signature change, and does not depend on a BaseApp event publisher. W1 uses the existing vendor ledger entry posting group path directly, so keeping the change in the ES Cartera special case is appropriate.

**Necessity:** The ADO bug is clear and important. A posted purchase invoice can leave account 400 open while the payment account 410 is also left open when users apply through inline `Applies-to Doc. No.` with different vendor posting groups. Apply Entries already works, so the change is needed to make the inline path post balanced G/L entries too. The added regression test covers the exact inline apply path and checks that both payables accounts net to zero.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9306 round=1 by=alexei-dobriansky at=2026-07-15T05:42:40Z lastSha=590dce807dc06dcf68ef7df5627e03264c8501e5 suggestions=none

AndersLarsenMicrosoft (approved):

PR #9312, opened 2026-07-10
Migrate obsoleted "Object Name" field to "Name" and fix silent truncation

SBalslev:
> [!IMPORTANT]
> **BLOCKED — do not merge yet.** This PR migrates references from the obsoleted `"Object Name"` field to the new `Name` field on the platform virtual tables **AllObj (2000000038)**, **AllObjWithCaption (2000000058)**, and **System Object (2000000029)**.
>
> CI currently fails all app builds with `AL0132 / AL0186 'Name' does not exist` because the new `Name` field has **not yet shipped** into the platform symbols/artifacts that BCApps compiles against. It exists only in the platform source branch.
>
> **This PR can only go green once the platform change adding the `Name` field is released into the BCApps build artifacts.** The migration code itself is correct — the missing-field errors are the only compile failures. Parked as draft until the platform dependency lands.

---
## What & why

The `"Object Name"` field (`Text[30]`) on the system virtual tables **AllObj** (2000000038), **AllObjWithCaption** (2000000058), and **System Object** (2000000029) was obsoleted on the platform and replaced by a new `Name` field (`Text[100]`). This PR migrates every reference in BCApps and eliminates the silent truncation the widened type would otherwise cause.

- **Migrate references** — all `"Object Name"` usages on these three tables to `Name` (144 files).
- **Widen persisted fields / FlowFields** that receive object names to `Text[100]` (non-breaking): `ObjectTranslation`, `ExportProtocol`, `CALTestResult`, `ChangeGlobalDimLogEntry`, `GenericChartSetup`, `TaxEntity`, `EntityAttributeMapping`, `MigrationTableMapping`.
- **Widen byref (`var`) parameter chains and their callers** to `Text[100]` — AL requires byref arguments to match the parameter type exactly (including length), so `Text[100]` cannot flow into `var Text`. Affects the INTaxEngine object/table lookup cascade (`AppObjectHelper`, `TaxTypeObjectHelper`), `GenericChartMgt`, and the `ScriptSymbolLookupDialog` publishers and `TaxTypeObjectHelper` subscribers.
- **Fix a JSON round-trip regression** — import read object names via `JToken2Text30` (truncating to 30) before an exact-match lookup; switched the object-name sites to the unbounded `JToken2Text` so long names resolve.

**Convention applied:** by-value params & return values to unbounded `Text`; persisted table fields & FlowFields to `Text[100]`; byref params and everything passed into them to `Text[100]`. Increasing an AL text length is non-breaking (only shortening needs a data upgrade).

## Linked work

Fixes #

> An approved issue still needs to be linked here before merge.

## How I validated this

- [x] Behavior is covered by existing tests (INTaxEngine `AppObjectHelperTests` / `TaxTypeObjectHelperTests`, test-runner suites); no new tests added since this is a field rename + width-safety change.

- Audited every sink where a `Name` value flows: all `:= <table>.Name` assignments land in unbounded `Text`, `Text[100]`, or `Text[128]`; all `exit(...Name)` return sites return unbounded `Text` / `Text[256]`.
- Verified integration event publisher/subscriber signatures still match after widening.
- **Not yet built locally** — this environment can't compile AL. Relying on CI here to confirm the build; happy to iterate on any analyzer/compile findings.

## Risk & compatibility

- All schema changes are **field-length increases** (`Text[30]` to `Text[100]`), which are non-breaking and need no data upgrade.
- No behavioral changes beyond removing truncation of names longer than 30 characters.
- Obsoleted-field references on the three virtual tables are fully removed; same-named `"Object Name"` fields on unrelated tables (`Tenant Permission`, `Metadata Permission`, `Expanded Permission`, `Application Object Metadata`) are intentionally untouched.


src/Apps/W1/EDocumentConnectors/Avalara/App/src/Extensions/TransformationRule.TableExt.al
```diff
@@ -28,7 +28,7 @@ tableextension 6374 "Transformation Rule" extends "Transformation Rule"
         }
         field(6371; "Lookup Table Name"; Text[30])
```
pri-kise: I know that the PR is now only draft, but I was just wanted to confirm:
will those field length be updated to the new length (This is how I understood the PR description)?

Do we need obsoletion process like it was done for the home page field in the Base App? 
That would require 3 Releases and the CalcForumula couldn't be changed right now. 

Or will the platform ignore overflow of Lookups?

SBalslev: This pr is to understand the impact. As no data will exceed 30 char in the first release overflow will not happen but absolutely something to be aware of.

PR #9333, opened 2026-07-13
[Main]-Bin content Block Movement does not prevent outbound posting for negative adjustments and sales orders

DeepsShukla:
Workitem [Bug 639575](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/639575): [All-E] Bin content Block Movement does not prevent outbound posting for negative adjustments and sales orders
Fixes [AB#639575](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/639575)


v-ankitgoyal (dismissed):

alexei-dobriansky:

<details>

This PR blocks outbound postings from a Bin Content when `Block Movement` is `Outbound` or `All`. It does this for non-directed bin-mandatory item journal decreases, and it adds the same warehouse journal validation before sales warehouse journal lines are posted.

For the linked repro, the fix addresses the root cause. The new check runs before `WhseJnlPostLine.Run`, so the blocked warehouse entry is not posted. The new tests cover both `Outbound` and `All` for negative adjustments and sales shipments. No public event or API signature is changed.
</details>

<details>

**S1** - *Check the purchase return path*  
The same outbound movement can happen through purchase return shipment, but `PurchPost.PostItemJnlLineWhseLine` still posts the warehouse journal line without `WMSMgt.CheckWhseJnlLine`. If purchase returns must obey Bin Content `Block Movement`, add the same validation there or document why that document type is outside this bug.

**S2** - *Add an inbound control test*  
The new tests cover the blocked paths. Please also add a small control check that `Block Movement = Outbound` still allows an inbound positive adjustment, so the new decrease check cannot accidentally block inbound movement later.
</details>

<details>

**Risk:** The risk is in warehouse posting. `WMS Management` is a shared codeunit, so the new decrease check can affect item-journal-source outbound flows that already call `CheckWhseJnlLine`. The sales change also adds a new validation point before posting warehouse journal lines in the Sales-Post copies. This is the right place for the linked sales scenario, but it can block postings that previously bypassed this rule.

**Necessity:** The bug is valid and important. Without this change, a bin content marked as blocked for outbound movement can still be depleted by negative adjustments and sales shipments. The scope is mostly right for the linked repro, and the tests cover the main reported cases.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9333 round=1 by=alexei-dobriansky at=2026-07-13T17:05:45Z lastSha=0197474a2f3feddadde0c7592376b8ba581846ac suggestions=S1,S2

v-ankitgoyal (dismissed):

qasimikram (changes requested):
S1: The new unconditional `WMSMgt.CheckWhseJnlLine(...)` call in `PostItemJnlLineWhseLine` breaks sales posting when the temporary warehouse journal line has no location. `CheckWhseJnlLine` immediately requires `Location Code`, and the current CI failures reproduce this broadly across localizations with `Location Code must have a value in Warehouse Journal Line`, with the stack pointing to the added call.

Please guard this validation so it only runs for the bin/location scenario it is intended to protect, and add a regression test covering sales posting without a location. The targeted blocked-bin tests cover the bug scenario but not this newly broken path.

DeepsShukla:
> S1: The new unconditional `WMSMgt.CheckWhseJnlLine(...)` call in `PostItemJnlLineWhseLine` breaks sales posting when the temporary warehouse journal line has no location. `CheckWhseJnlLine` immediately requires `Location Code`, and the current CI failures reproduce this broadly across localizations with `Location Code must have a value in Warehouse Journal Line`, with the stack pointing to the added call.
> 
> Please guard this validation so it only runs for the bin/location scenario it is intended to protect, and add a regression test covering sales posting without a location. The targeted blocked-bin tests cover the bug scenario but not this newly broken path.

Updated the changes

v-ankitgoyal (approved):

PredragMaricic (approved):

JakovljevicDusan (approved):

qasimikram (approved):
The previously identified regression is addressed: the warehouse journal validation is guarded when the location is blank, and a regression test now covers sales posting without a location.

The outbound Bin Content validation correctly blocks both Outbound and All movement for negative adjustments and sales shipments. The implementation is propagated consistently across the Sales-Post localization copies, and the targeted tests cover the reported scenarios. I found no remaining blocking issues.

PR #9342, opened 2026-07-13
Preserve existing users' custom Role Center on plan re-sync (Bug 641534)

gggdttt:
## Problem

After a plan re-sync (for example running **Retrieve Users**, or opening a same-tenant copy-acc sandbox), **all** existing users' previously customized Role Centers are silently reset to **Business Manager** (Bedrijfsmanager). Customers reported every user in the environment being affected, with no customization involved.

## Root cause

`Permission Manager.AssignDefaultRoleCenterToUser` has a long-standing branch (from 2019) that overwrites an existing user's `User Personalization` profile when `IsAllProfileFiltered` is true. Originally that flag was set only for CRONUS evaluation companies (demo experience). PR 213237 ("Pick 'Business Manager' role for users created via Update from MS action") renamed the helper to `FilterProfileToBusinessManager` and made it set `IsFiltered := true` for **all non-evaluation** companies as well. That unintentionally activated the overwrite branch for real customers: any plan re-sync (`OnUpdateUserAccessForSaaS` raised from `AzureADPlanImpl.AddNewlyAssignedUserPlans`) now reset every existing user's custom Role Center.

## Fix (conservative)

Keep PR 213237's intended behaviour - **new** non-evaluation users still get the `Business Manager` Role Center (via the `Insert` branch, which uses the filtered profile). Only stop flagging non-evaluation companies for the *overwrite* path, so **existing** users keep their customized Role Center. CRONUS evaluation/demo behaviour is unchanged (existing users are still aligned to the evaluation role center).

The only functional change is removing `IsFiltered := true;` from the non-evaluation branch of `FilterProfileToBusinessManager`.

## Changes

- `src/Layers/W1/BaseApp/Modules/System/PermissionSets/PermissionManager.Codeunit.al`
- `src/Layers/W1/Tests/SINGLESERVER/UserAccessinSaaSTests.Codeunit.al` - regression test `TestUpdateUserAccessForSaaSDoesNotOverwriteExistingRoleCenter` (real/non-eval company, existing user with a custom Role Center is preserved after `UpdateUserAccessForSaaS`).

## Risk

Low. New-user role assignment is unchanged; CRONUS eval behaviour is unchanged. Only the overwrite of existing real-customer users is removed.

Fixes Bug 641534.

### DIFF
--- a/src/Layers/W1/BaseApp/Modules/System/PermissionSets/PermissionManager.Codeunit.al
+++ b/src/Layers/W1/BaseApp/Modules/System/PermissionSets/PermissionManager.Codeunit.al
@@ -172,6 +172,10 @@ codeunit 9002 "Permission Manager"
             UserPersonalization.Validate(Scope, AllProfile.Scope);
             UserPersonalization.Insert();
         end else
+            // Only overwrite an existing user's role center in CRONUS evaluation companies (demo experience).
+            // For real (non-evaluation) companies the user may have chosen a custom role center, which must
+            // not be reset during a plan re-sync (regression from PR 213237). IsAllProfileFiltered is only
+            // set for the CRONUS evaluation case below.
             if IsAllProfileFiltered then begin
                 UserPersonalization.Validate("Profile ID", AllProfile."Profile ID");
                 UserPersonalization.Validate("App ID", AllProfile."App ID");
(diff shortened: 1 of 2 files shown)


alexei-dobriansky:

# **Recommendation: Accept**

<details>

This change keeps existing users' customized Role Centers when SaaS user access is re-synced in a normal company. It does that by keeping the Business Manager profile filter for new users, but no longer marking the non-evaluation path as an overwrite case for existing user personalization.

The branch logic matches the intended split: new users still receive Business Manager from the filtered profile, while existing users are updated only in the CRONUS evaluation case. The added regression test sets a non-evaluation company, creates an existing user with a different enabled profile, runs the SaaS update path, and checks that the profile is preserved.
</details>

<details>

**Fit:** Strong

The reported scenario is a real customer-impacting reset of user personalization during plan re-sync. The code change is narrow and directly removes the flag that made existing non-evaluation users enter the overwrite branch, while keeping the new-user behavior intact.
</details>

<details>

*None.*
</details>

<details>

**Risk:** Low to moderate. The touched code is in permission and user personalization setup, so an incorrect branch could affect many users, but the actual change narrows overwrites instead of broadening them and keeps CRONUS evaluation behavior explicit.

**Necessity:** Justified. Without this change, routine user-plan synchronization can reset customized Role Centers for existing users, and the scope is limited to the branch that caused that regression.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=9342 round=1 by=alexei-dobriansky at=2026-08-26T22:24:34Z lastSha=11fc078c58057ff3333506e184dd5669083407ff reviewKey=574974f70f7cdf4b8d7eacf9dbef140b824f62ad8489556f936568ecd59cc323 suggestions=none

PR #9351, opened 2026-07-13
[master]Subscription Billing: O(N²) contract-renewal detection in Sales-Post subscribers slows posting (~28% of batch CPU)

AndreasHans:
Fixes [AB#640070](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/640070)

### DIFF
--- a/src/Apps/W1/Subscription Billing/App/Contract Renewal/Codeunits/SubContrRenewalSubcribers.Codeunit.al
+++ b/src/Apps/W1/Subscription Billing/App/Contract Renewal/Codeunits/SubContrRenewalSubcribers.Codeunit.al
@@ -124,7 +124,7 @@ codeunit 8001 "Sub. Contr. Renewal Subcribers"
     [EventSubscriber(ObjectType::Codeunit, Codeunit::"Sales-Post", OnPostSalesLineOnAfterTestSalesLine, '', false, false)]
     local procedure PerformRenewalForSalesLine(var SalesLine: Record "Sales Line"; var SalesHeader: Record "Sales Header"; var WhseShptHeader: Record "Warehouse Shipment Header"; WhseShip: Boolean; PreviewMode: Boolean; var CostBaseAmount: Decimal)
     begin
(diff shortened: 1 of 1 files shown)


src/Apps/W1/Subscription Billing/App/Contract Renewal/Codeunits/SubContrRenewalSubcribers.Codeunit.al
```diff
@@ -419,7 +419,56 @@ codeunit 8001 "Sub. Contr. Renewal Subcribers"
...
+    end;
+
+    // Cached wrapper for SalesHeader.HasOnlyContractRenewalLines(), keyed by document (SystemId).
+    // The result is stable during a posting run, so caching avoids re-evaluating every posted line
+    // (turning the repeated per-line checks from O(N^2) into O(N)). Cache is cleared at start/end of posting.
+    local procedure HasOnlyContractRenewalLines(SalesHeader: Record "Sales Header"): Boolean
+    var
+        Result: Boolean;
+    begin
+        if HasOnlyContractRenewalLinesCache.Get(SalesHeader.SystemId, Result) then
```
mynjj, line 449: how does the instance of `HasOnlyContractRenewalLinesCache` remains the same across the different events? What happens if two postings occur at the same time? can that happen? would the cache of one interfere with the other?

AndreasHans: how does the instance of HasOnlyContractRenewalLinesCache remains the same across the different events?
codeunit is single instance

What happens if two postings occur at the same time? 
The cache would contain results for both

can that happen?
dont know

would the cache of one interfere with the other?
Believe no. We use guid as key and we assume those are unique, so the cache would in that case contain both.

mynjj (approved):

djukicmilica (approved):

PR #9364, opened 2026-07-13
Slice 622414: [SOA] [Pioneer] Enhancing Sales Order Agent with Item Variant : availability and prices

tomasevicst:
## What & why

This PR adds end-to-end Item Variant support to the Sales Order Agent. The agent can now resolve a customer request to a specific item and variant, use that variant when checking availability and calculating prices, and carry the resolved Variant Code through to the sales quote or order.

Previously, item search operated primarily at the item level. Exact item matches could bypass the Item Selector, variant intent expressed in natural language could be lost, and availability checks could fall back to the unqualified item. This could produce quote lines with an empty or incorrect Variant Code.

The updated flow:
 - Runs the Item Selector consistently for exact, standard, and broader search results.
 - Includes known item variants and relevant supporting context from the incoming task message and non-ignored attachments during selection.
 - Returns an optional **variant_code** together with explicit **variant_match** and independent **variant_substitution_safety** classifications.
 - Validates returned items and item-variant combinations against the original candidate set.
 - Retains concrete variant alternatives only when they are explicitly classified as safe.
 - Allows alternatives for presentation-only differences, such as color, finish, or pattern, when the change preserves form, fit, function, safety, compatibility, and intended use, or when the customer explicitly permits the change.
 - Discards unsafe, uncertain, or unclassified substitutions.
 - Treats suitability-changing differences, such as size, fit, compatibility, capacity, technical specification, safety classification, or intended user, as unsafe unless the customer explicitly permits that change.
 - Treats a valid empty selection as an authoritative result and does not fall back to the original candidates.
 - Distinguishes valid empty decisions from selector, AOAI, or response-parsing failures.
 - Checks availability for matching item-variant combinations before considering concrete alternatives.
 - Uses the resolved variant for availability, pricing, shipment date, translations, and quote or order creation.
 - Leaves Variant Code empty when no variant was requested.
 - When no requested variant or safe alternative qualifies, replies to the customer that the requested item and variant are unavailable, does not request internal assistance, and does not create a quote.
 - Emits non-sensitive selector telemetry using stable failure categories, status codes, booleans, and aggregate counts. Customer content, item and variant values, and raw error text are not logged.
 - Adds ten active accuracy scenarios covering explicit, implicit, broad, missing, unavailable, prohibited-substitution, and multi-item variant requests.

The availability page remains item-based. When the requested variant is unavailable and the selector returns multiple safe, interchangeable alternatives, the initial implementation presents the first available concrete alternative for that item. It does not silently substitute the variant or create a quote until the customer confirms the alternative.

The mandatory-variant-without-request scenario remains deferred because **Variant Mandatory if Exists** is not currently included in the Item Selector candidate metadata.

## Linked work

Fixes [AB#622414](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/622414/?view=edit)

## How I validated this

## What I tested and the outcome

**Validation executed**
 - Full multi-project AL build with CodeCop enabled: Passed with no errors or warnings.
 - Item Selector tool schema parsed with **ConvertFrom-Json**: Passed.
    - **selected_items** is the sole required top-level result field.
    - Each selected item requires **item_no, confidence, variant_match, variant_substitution_safety**, and **reason**.
    - **variant_code** remains optional.
    - **unresolved_variant_requests** is not present.
 - Focused diagnostics for the changed AL and prompt files: No errors found.
 - Selector telemetry privacy verification: Passed.
    - Telemetry contains stable failure categories, status codes, booleans, and aggregate counts.
    - Raw AOAI errors, exception messages, customer content, item numbers, and variant values are not written to custom dimensions.

**Accuracy coverage prepared**

The following ten active scenarios are defined in [SOA-QUOTE_ITEMS_HANDLING_VARIANTS.yaml](vscode-file://vscode-app/c:/Users/stomasevic/AppData/Local/Programs/Microsoft%20VS%20Code/fc3def6774/resources/app/out/vs/code/electron-browser/workbench/workbench.html):

 - **QUOTE_5_ITEMS_VARIANTS_01**: Explicit variant codes resolve to the expected quote-line Variant Codes.
 - **QUOTE_5_ITEMS_VARIANTS_02**: Variants expressed within natural-language item descriptions resolve correctly.
 - **QUOTE_3_ITEMS_VARIANTS_03**: Semantic and broad variant wording, including requests such as “any color,” resolves to suitable variants.
 - **QUOTE_3_ITEMS_VARIANTS_04**: Items without a requested variant retain an empty Variant Code.
 - **QUOTE_1_ITEM_VARIANT_ALTERNATIVE_05**: A missing interchangeable color offers one safe, available same-item variant and does not create a quote before customer confirmation.
 - **QUOTE_1_ITEM_VARIANT_ALTERNATIVE_06**: An existing but unavailable shoe size does not cause available sizes 42 or 46 to be offered.
 - **QUOTE_1_ITEM_VARIANT_ALTERNATIVE_07**: A non-existent shoe size produces no unsafe size substitution and no quote.
 - **QUOTE_1_ITEM_VARIANT_ALTERNATIVE_08**: An explicit prohibition against color substitution is honored and no quote is created.
 - **QUOTE_1_ITEM_VARIANT_ALTERNATIVE_10:** An unavailable requested color can offer one safe, available same-item color, but no quote is created before customer confirmation.
 - **QUOTE_2_ITEMS_VARIANT_CONTEXT_ISOLATION_11**: Variant intent for one requested item does not leak to another item in the same request.
 - **QUOTE_1_ITEM_VARIANT_ALTERNATIVE_09**, covering an item where a variant is mandatory but none was requested, remains deferred. Variant Mandatory if Exists is not currently included in the Item Selector candidate metadata, so this behavior cannot yet be enforced reliably.

**UI implementation verification**

The Variant Code field is visible on the Sales Quote and Sales Order subforms used by the agent. The resolved Variant Code is also applied to availability, pricing, shipment-date, and translation calculations before quote or order creation.

## Risk & compatibility

Risk level: **Medium**. The change affects item search, variant selection, availability, pricing, and quote creation, but remains scoped to the Sales Order Agent workflow.

 - Customers without Item Variant records retain the existing item-level behavior, including an empty Variant Code.
 - No new setup option, permission, data migration, or schema upgrade is required.
 - **Variant Mandatory if Exists** is not currently included in the Item Selector candidate metadata. Mandatory-without-request enforcement and its accuracy scenario remain deferred until that metadata is supplied or deterministic downstream enforcement is added.
 - Exact, standard, and broader searches now invoke the Item Selector consistently. This adds AOAI latency and token usage to paths that could previously bypass model selection.
 - Known variants and supporting task-message and attachment context are included in selector input. This increases prompt size, although attachment input is limited to non-ignored extracted text from the relevant incoming message.
 - Matching item-variant combinations are availability-checked first. Concrete alternatives returned by the selector are considered only when no matching combination remains available.
 - Items with several selected candidate variants can require additional availability calculations.
 - Model output is treated as untrusted. Returned item numbers must belong to the original candidate set, and returned Variant Codes must exist for the selected item.
 - Variant alternatives are retained only when **variant_match** is **alternative** and the independent **variant_substitution_safety** classification is explicitly safe. Unsafe, uncertain, malformed, or unclassified alternatives are discarded.
 - The application does not synthesize alternatives for unresolved variant requests. Alternatives must be returned as concrete selector entries and pass server-side item, variant, and substitution-safety validation.
 - Suitability-changing variants, such as size, fit, compatibility, capacity, technical specification, safety classification, or intended user, are not substituted unless the customer explicitly permits that difference.
 - The availability page remains item-based and supports one displayed variant per item. When several safe alternatives are returned, only the first available concrete alternative is presented.
 - An alternative is not silently substituted and is not used to create a quote until the customer confirms it.
 - Resolved variant state is cleared whenever the search changes, including when the search is cleared, to prevent values from leaking between searches.
 - A valid empty selector result is authoritative and does not fall back to the original candidates. For a variant-specific request, this produces a customer-facing unavailability response rather than an internal assistance request or quote.
 - Selector execution failures and malformed responses remain technical failures and retain the existing best-effort fallback behavior.
 - Selector telemetry distinguishes valid empty decisions from AOAI, parsing, and server-side validation failures. Custom dimensions are limited to stable failure categories, status codes, booleans, candidate counts, and aggregate rejection counts; raw error text, customer content, item numbers, Variant Codes, and descriptions are not emitted in custom dimensions.
 - The change targets version 29 and is not intended for backporting until the broader variant workflow has been validated.


alexei-dobriansky:

<details>

Feature summary: this PR adds item variant support to Sales Order Agent item search, item availability, price calculation, and the sales line instructions. It lets the item selector return `variant_code`, shows a Variant Code column in the availability page, and passes the resolved variant into availability, capable-to-promise, translations, and temporary sales line price calculation.

The specific-variant path is mostly wired through correctly. The open issue is the no-specific-variant path. The Slice says that when the customer does not request a specific variant, the agent should suggest several available variants. The current flow is still keyed by `Item.SystemId` and the page source is still `Item`, so the page can only show one row per item and one resolved variant code per item. For an item-only request, the new fallback also does not add variants, so the agent can return a blank Variant Code instead of variant choices.
</details>

<details>

**S1** - *Show variant choices when none is specified*  
When the query names only an item, return available variant choices instead of a blank Variant Code. The current `HasVariantSignalForItem` / `ResolvedItemVariants` flow does not do that, and the `Item` page can only show one resolved variant per item.

**S2** - *Add tests for variant selection and pricing*  
Please add tests for the new variant behavior. Cover at least a specific variant request, an item-only request where variants must be suggested, and price or availability calculation with the selected Variant Code.
</details>

<details>

**Risk:** this touches item selection, availability, capable-to-promise, and sales price calculation. A wrong variant can show the wrong inventory, wrong shipment date, or wrong customer price, and can also create a sales line that fails when variants are mandatory.

**Necessity:** the feature is useful and matches the Slice direction, especially for customers that sell items with mandatory variants. The scope is right, but the item-only variant path is not complete yet, and the behavior is testable through the Sales Order Agent test app internals.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9364 round=1 by=alexei-dobriansky at=2026-07-14T17:07:47Z lastSha=328958f66529adf8688f69dd225ef7082d1cdc99 suggestions=S1,S2

src/Apps/W1/SalesOrderAgent/app/src/ItemSearch/SOAItemSearch.Codeunit.al
```diff
@@ -219,9 +242,11 @@ codeunit 4591 "SOA Item Search"
                 CountBeforeAvailabilityCheck := ItemFilter.Split('|').Count();
                 ApplyAvailabilityFilter := CheckAvailability and (SOASetup."Search Only Available Items" and not SOASetup."Incl. Capable to Promise");
 
-                // Try item selection only for broader search results with multiple candidates
-                if (SearchType = 'broader_item_search') and (CountBeforeAvailabilityCheck > 1) then
-                    if SelectBestItem(ItemFilter, BuildSearchQueryText(SearchKeyWordsTrimmed), CandidateArray, SelectedMatchingItemFilter, SelectedAlternativeItemFilter) then begin
+                // Run item selection for all candidate payloads so variant resolution is consistent.
```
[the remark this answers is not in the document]
tomasevicst: We run Item Selector even for exact or single-candidate matches because an exact item match identifies the item, not the requested variant, which may still need semantic resolution from wording such as color, size, or “any color.” Guarding only by candidate count would regress variant handling and make behavior inconsistent across search paths, though we could skip the selector when the candidate has no variants.

src/Apps/W1/SalesOrderAgent/app/src/ItemSearch/SOAItemSearch.Codeunit.al
```diff
@@ -208,6 +228,9 @@ codeunit 4591 "SOA Item Search"
         if not GetItemFilters(ItemFilter, SearchKeyWordsTrimmed, SearchType) then //Search for the items using the entity search
             exit;
 
+        if (ItemFilter <> '') and (CandidateArray.Count() = 0) then
```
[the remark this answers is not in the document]
tomasevicst: The item_get fast path resolves only the item identity; it cannot determine which variant the customer requested from semantic wording such as a color description, age group, or “any color,” nor can it rank same-item variant alternatives. Resolving that locally would duplicate semantic selection logic and behave differently from other search paths, so the selector must run for a direct hit when the item has variants, though it can be skipped when the item has none.

src/Apps/W1/SalesOrderAgent/app/src/ItemSearch/SOAItemSearch.Codeunit.al
```diff
@@ -303,53 +337,302 @@ codeunit 4591 "SOA Item Search"
...
+    local procedure NormalizeVariantAlternatives(var SelectedMatchingItemFilter: Text; var SelectedMatchingItemVariants: Dictionary of [Text, Text]; var SelectedAlternativeItemFilter: Text; var SelectedAlternativeItemVariants: Dictionary of [Text, Text]; SearchQuery: Text)
+    begin
+        AddSameItemVariantAlternativesForMissingVariant(SelectedMatchingItemFilter, SelectedMatchingItemVariants, SelectedAlternativeItemFilter, SelectedAlternativeItemVariants, SearchQuery);
+        KeepSameItemVariantAlternativesOnly(SelectedMatchingItemFilter, SelectedAlternativeItemFilter, SelectedAlternativeItemVariants);
+        RemoveMatchingItemsWithBlankVariantAndSameItemAlternatives(SelectedMatchingItemFilter, SelectedMatchingItemVariants, SelectedAlternativeItemVariants);
+    end;
+
+    local procedure AddSameItemVariantAlternativesForMissingVariant(SelectedMatchingItemFilter: Text; SelectedMatchingItemVariants: Dictionary of [Text, Text]; var SelectedAlternativeItemFilter: Text; var SelectedAlternativeItemVariants: Dictionary of [Text, Text]; SearchQuery: Text)
```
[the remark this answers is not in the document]
tomasevicst: Valid finding. Same-item variants are intentionally preferred for confirmed missing-variant requests, but the current fallback infers variant intent from any unmatched token and then destructively replaces all selector alternatives. That can discard legitimate cross-item alternatives. We should preserve the two alternative classes separately and rely on explicit selector output, or a stronger variant-intent signal, before applying same-item priority. We will add coverage for a mixed same-item/cross-item response and retest the existing yellow T-shirt alternative scenario.

src/Apps/W1/SalesOrderAgent/app/src/ItemSearch/SOAItemSearch.Codeunit.al
```diff
@@ -303,53 +337,302 @@ codeunit 4591 "SOA Item Search"
...
+
+        foreach ItemSystemId in ItemFilter.Split('|') do
+            if SelectedItemVariants.ContainsKey(ItemSystemId) then begin
+                VariantCode := SelectedItemVariants.Get(ItemSystemId);
+                if not ResolvedItemVariants.ContainsKey(ItemSystemId) then
+                    ResolvedItemVariants.Add(ItemSystemId, VariantCode);
+            end;
+    end;
+
+    local procedure BuildFilteredItemFilter(SourceItemFilter: Text; var Rec: Record Item; RequiredQuantity: Decimal; InUOMCode: Code[10]; ApplyAvailabilityFilter: Boolean; SelectedItemVariants: Dictionary of [Text, Text]; var AvailableItemVariants: Dictionary of [Text, Text]): Text
```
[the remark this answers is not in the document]
tomasevicst: The reduction is intentional for the current iteration. The selector may return multiple same-item variant candidates so availability filtering can evaluate them in order, but the availability page is Item-based and supports one resolved variant per item row. The first available candidate is surfaced to the agent. Preserving every item+variant pair through the UI would require changing the page to an item+variant buffer and updating the resolved-variant event contract, pricing, and availability flow. This limitation is documented in the design. We will clarify the selector prompt/schema wording so it describes ranked fallback candidates rather than promising multiple displayed alternatives.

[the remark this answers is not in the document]
tomasevicst: The reduction is intentional for the current iteration. The selector may return multiple same-item variant candidates so availability filtering can evaluate them in order, but the availability page is Item-based and supports one resolved variant per item row. The first available candidate is surfaced to the agent. Preserving every item+variant pair through the UI would require changing the page to an item+variant buffer and updating the resolved-variant event contract, pricing, and availability flow. This limitation is documented in the design. We will clarify the selector prompt/schema wording so it describes ranked fallback candidates rather than promising multiple displayed alternatives.

src/Apps/W1/SalesOrderAgent/app/src/ItemAvailability/SOAMultiItemsAvailability.Page.al
```diff
@@ -1016,4 +1033,9 @@ page 4410 "SOA Multi Items Availability"
     local procedure OnAfterInitPage(var CustomerNo: Code[20]; var ContactNo: Code[20]; var LocationFilter: Text; var CalculateEarliestShipmentDate: Boolean)
     begin
     end;
+
+    [InternalEvent(false, false)]
+    local procedure OnGetResolvedVariant(ItemSystemId: Guid; var VariantCode: Code[10])
```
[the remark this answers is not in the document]
tomasevicst: Applied suggestion.

src/Apps/W1/SalesOrderAgent/app/src/ItemSearch/SOAItemSearch.Codeunit.al
```diff
@@ -360,6 +534,54 @@ codeunit 4591 "SOA Item Search"
...
+        AvailableVariantCode := '';
+        if SelectedItemVariants.ContainsKey(ItemSystemId) then
+            VariantCodes := SelectedItemVariants.Get(ItemSystemId);
+
+        if VariantCodes.Count() = 0 then begin
+            Item.SetRange("Variant Filter", '');
+            exit(IsRequiredQuantityAvailable(Item, RequiredQuantity, InUOMCode));
+        end;
+
+        foreach VariantCode in VariantCodes do begin
```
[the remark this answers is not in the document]
tomasevicst: Each candidate variant requires a separate projected-availability calculation because CalcAvailQuantities operates on the current Variant Filter, and a combined filter would only provide aggregate availability rather than identify an available variant. BuildFilteredItemFilter does not perform another full calculation in addition to these calls; it delegates to FindFirstAvailableVariant, which stops at the first available variant. The cache in SOACreateTaskImpl benefits repeated identical lookups, whereas this loop normally evaluates each item/variant pair only once. A cache here would therefore not reduce the common-case number of calculations and a session-level cache could become stale. We should benchmark searches with high variant counts and consider ranking or bounding generated alternatives if this becomes measurable.

src/Apps/W1/SalesOrderAgent/app/src/ItemSearch/SOAItemSearch.Codeunit.al
```diff
@@ -360,6 +625,54 @@ codeunit 4591 "SOA Item Search"
         exit(FilteredItemFilter.TrimEnd('|'));
     end;
 
+    local procedure FindFirstAvailableVariant(var Item: Record Item; ItemSystemId: Text; RequiredQuantity: Decimal; InUOMCode: Code[10]; SelectedItemVariants: Dictionary of [Text, List of [Code[10]]]; var AvailableVariantCode: Code[10]): Boolean
```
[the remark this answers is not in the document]
tomasevicst: This is an intentional consequence of variant-level availability resolution. Each candidate variant must be calculated separately because CalcAvailQuantities operates on the current Variant Filter, while a combined variant filter would only return aggregate availability and would not identify which variant satisfies the requested quantity. Each item/variant pair is normally evaluated once and the loop stops at the first available variant, so caching would not reduce the common-case calculation count. The cache in SOACreateTaskImpl addresses repeated identical lookups and is therefore not analogous. There is currently no public batch API that returns projected availability per variant. We can measure searches with unusually high variant counts and revisit ranking or bounding alternatives if this becomes a demonstrated performance issue.

src/Apps/W1/SalesOrderAgent/app/src/ItemSearch/SOAItemSearch.Codeunit.al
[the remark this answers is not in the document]
tomasevicst: Tests are part of other PR, as test app is in different repository.

src/Apps/W1/SalesOrderAgent/app/src/ItemSearch/SOAItemSearch.Codeunit.al
```diff
@@ -219,9 +241,12 @@ codeunit 4591 "SOA Item Search"
                 CountBeforeAvailabilityCheck := ItemFilter.Split('|').Count();
                 ApplyAvailabilityFilter := CheckAvailability and (SOASetup."Search Only Available Items" and not SOASetup."Incl. Capable to Promise");
 
-                // Try item selection only for broader search results with multiple candidates
-                if (SearchType = 'broader_item_search') and (CountBeforeAvailabilityCheck > 1) then
-                    if SelectBestItem(ItemFilter, BuildSearchQueryText(SearchKeyWordsTrimmed), CandidateArray, SelectedMatchingItemFilter, SelectedAlternativeItemFilter) then begin
+                // Run item selection for all candidate payloads so variant resolution is consistent.
```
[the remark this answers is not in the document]
tomasevicst: The increased selector frequency is intentional and was approved as part of variant resolution. A single candidate only makes the item unambiguous; it does not make the requested variant unambiguous. Exact-item searches such as “MagicToyland Fairy Doll in any color” still require semantic resolution against that item’s variants, including relevant same-item context from the incoming message. Restoring the previous single-candidate shortcut would regress the variant acceptance scenarios that motivated this change.

We agree that this increases AOAI latency, token usage, and the scope of message-content processing. A safe optimization would be to bypass the selector for a single candidate that has no variants, because neither item nor variant resolution requires the model. We should treat further message minimization as a separate privacy change rather than infer it from candidate count.

src/Apps/W1/SalesOrderAgent/app/src/ItemSearch/SOAItemSelector.Codeunit.al
```diff
@@ -28,19 +28,42 @@ codeunit 4417 "SOA Item Selector"
...
+    internal procedure SelectBestMatchingItemWithVariantsAndUnresolvedRequests(SearchQuery: Text; MessageContent: Text; CandidateArray: JsonArray; var MatchingItemFilter: Text; var AlternativeItemFilter: Text; var MatchingItemVariants: Dictionary of [Text, List of [Code[10]]]; var AlternativeItemVariants: Dictionary of [Text, List of [Code[10]]]; var UnresolvedVariantRequests: Dictionary of [Text, Boolean]): Boolean
     begin
         MatchingItemFilter := '';
         AlternativeItemFilter := '';
+        Clear(MatchingItemVariants);
+        Clear(AlternativeItemVariants);
+        Clear(UnresolvedVariantRequests);
 
-        if not TrySelectBestMatchingItem(SearchQuery, CandidateArray, MatchingItemFilter, AlternativeItemFilter) then
+        if not TrySelectBestMatchingItem(SearchQuery, MessageContent, CandidateArray, MatchingItemFilter, AlternativeItemFilter, MatchingItemVariants, AlternativeItemVariants, UnresolvedVariantRequests) then
```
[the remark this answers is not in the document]
tomasevicst: Agreed. The current Boolean conflates a valid empty selection with technical failures. We will preserve the existing fallback behavior, but log thrown errors immediately after the TryFunction returns false. AOAI status failures and missing function-call responses must be handled separately because they exit the TryFunction normally and do not populate last-error context. Telemetry will include only technical status and candidate counts, not customer query or message content.

src/Apps/W1/SalesOrderAgent/app/src/ItemSearch/SOAItemSearch.Codeunit.al
```diff
@@ -303,53 +345,263 @@ codeunit 4591 "SOA Item Search"
...
+
+            if not NewSelectedMatchingItemVariants.ContainsKey(ItemSystemId) then
+                NewSelectedMatchingItemVariants.Add(ItemSystemId, MatchingVariantCodes);
+        end;
+
+        SelectedMatchingItemFilter := NewSelectedMatchingItemFilter;
+        SelectedMatchingItemVariants := NewSelectedMatchingItemVariants;
+    end;
+
+    local procedure BuildCandidateArrayFromItemFilter(ItemFilter: Text; var CandidateArray: JsonArray)
```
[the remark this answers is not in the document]
tomasevicst: Addressed. BuildCandidateArrayFromItemFilter now builds a filter for the complete candidate Item No. set, loads all corresponding Item Variant records with a single FindSet, and groups them by Item No. before constructing the selector payload. The payload loop now reads variants from the in-memory dictionary and no longer performs one Item Variant query per candidate.

src/Apps/W1/SalesOrderAgent/app/.resources/Prompts/ItemSearch/itemselector-tool.md
```diff
@@ -3,41 +3,83 @@
...
             "properties": {
               "item_no": {
                 "type": "string",
                 "description": "The Item No. value"
               },
+              "variant_code": {
+                "type": "string",
+                "description": "The selected Item Variant Code from the candidate's Variants column. Return an empty string when no variant is explicitly or semantically identified by the query or relevant same-item message context. Do not invent variant codes."
+              },
               "confidence": {
```
qasimikram, line 24: Should we now have confidence for item and variant separately? Based on that we can make decisions?

src/Apps/W1/SalesOrderAgent/app/.resources/Prompts/ItemSearch/itemselector-task.md
```diff
@@ -2,12 +2,21 @@ You are an intelligent item selection assistant for a Sales Order system.
 
 Your task is to analyze a search query and a list of item candidates and return relevant matches and optional alternatives.
 
+The payload can also contain `message_content`. Use the **message content** only as supporting context for the current `search_query`. The search query remains the primary selection request.
```
qasimikram: Should we move message content instructions under a heading, so its consistent with this prompt and clear which instruction are related to message_content and which are related to search_query.

PredragMaricic (dismissed):

src/Apps/W1/SalesOrderAgent/app/.resources/Prompts/ItemSearch/itemselector-tool.md
```diff
@@ -3,41 +3,83 @@
...
                 "type": "string",
                 "enum": ["matching", "alternative", "none"],
                 "description": "\"matching\" → high confidence (strong identifier or clear semantic match); \"alternative\" → related but less precise match within the same product family or intent; \"none\" → no meaningful relation to the query (must NOT be returned)"
               },
+              "variant_resolution": {
```
qasimikram: I am not sure about what each option means, can we have similar confidence for variants that based on search query and message content is the variant selected with high confidence, low confidence and with options that no variant could be selected something.

src/Apps/W1/SalesOrderAgent/app/.resources/Prompts/ItemSearch/itemselector-tool.md
```diff
@@ -3,41 +3,83 @@
...
+                "enum": ["not_requested", "resolved", "unresolved_interchangeable", "unresolved_non_interchangeable"],
+                "description": "Whether no variant was requested, variant_code resolves the request, the requested value is absent but safe alternatives are allowed, or the requested value is absent and alternatives would change suitability"
+              },
               "reason": {
                 "type": "string",
                 "description": "Short explanation of how the item was evaluated against the query."
               }
             },
-            "required": ["item_no", "confidence", "reason"]
+            "required": ["item_no", "variant_code", "confidence", "variant_resolution", "reason"]
```
qasimikram: variant_code is mandatory so LLM will be forced to select  one, event if there is none matching. Conflicts with the instructions above to keep it empty.

src/Apps/W1/SalesOrderAgent/app/.resources/Prompts/ItemSearch/itemselector-tool.md
```diff
@@ -3,41 +3,83 @@
...
+          }
+        },
+        "unresolved_variant_requests": {
```
qasimikram: I would prefer a simpler tools call, 3 fields to identify variants looks complicated. I would have variant no and confidence. This looks slight complex tool and instructions that are based on this,

qasimikram (commented):
Reviewed the prompts for now.

src/Apps/W1/SalesOrderAgent/app/.resources/Prompts/ItemSearch/itemselector-tool.md
tomasevicst: Currently, confidence classifies the complete item-and-variant option as matching or alternative, while variant_resolution separately describes whether the variant request was resolved, absent but interchangeable, or absent and non-interchangeable. I would avoid adding independent model-generated item and variant confidence values unless we have a concrete downstream decision that needs them, because those scores would be difficult to calibrate and should not control substitution safety. If the concern is uncertainty between multiple existing variants, an explicit ambiguous or needs_clarification resolution state would give us a more deterministic contract than another confidence field.

tomasevicst: Variant handling already has a separate structured axis:

`"variant_resolution": "not_requested"
                    | "resolved"
                    | "unresolved_interchangeable"
                    | "unresolved_non_interchangeable"`

This is more useful than model-generated variant confidence because our decisions are based on concrete states:

 - Was a variant requested?
 - Does it exist in the candidate data?
 - Is substitution suitable?
 - Is the selected variant available?

A separate variant_confidence such as high/medium/low would be difficult to calibrate and should not determine safety-sensitive substitutions. It would also be unclear what downstream threshold should mean.

src/Apps/W1/SalesOrderAgent/app/.resources/Prompts/ItemSearch/itemselector-task.md
tomasevicst: Agreed. We will group these rules under a dedicated “Search Query and Message Context” heading. This will make it explicit that search_query is the primary selection request and that message_content may only provide supporting context for the same item, particularly in multi-item messages. This is a structural clarification and does not change the current selection behavior.

src/Apps/W1/SalesOrderAgent/app/.resources/Prompts/ItemSearch/itemselector-tool.md
tomasevicst: I agree that the existing option descriptions are difficult to interpret and that we do not currently represent variant-selection uncertainty explicitly. I would avoid modeling this only as high/low/none confidence, because “none” cannot distinguish no variant request, a missing requested value, and ambiguity between existing variants. Those states require different downstream behavior. My proposal is to keep the explicit variant_resolution contract, clarify each option, and add an ambiguous or needs_clarification state. A low-confidence interpretation would then produce no selected variant_code and request clarification instead of allowing an uncertain variant to be used. If we need telemetry about exact versus semantic matching, we can add a separate match-basis field rather than using confidence to drive fulfillment safety.

tomasevicst: Current meanings are:

 - **not_requested** - Neither search_query nor relevant message_content requests a variant
 - **resolved** - One existing variant clearly satisfies the request
 - **unresolved_interchangeable** - Requested value does not exist, but another variant may be a safe substitute
 - **unresolved_non_interchangeable** - Requested value does not exist and substitution is unsafe

tomasevicst: I would not use only high, low, and none, because none would conflate several materially different cases:
 - No variant was requested.
 - A variant was requested but does not exist.
 - Matching variants exist, but the request is ambiguous.
 - The item has no variants.

Also, a low-confidence variant should not be silently selected. It should trigger clarification

src/Apps/W1/SalesOrderAgent/app/.resources/Prompts/ItemSearch/itemselector-tool.md
tomasevicst: required in the tool schema only requires the variant_code property to be present; it does not require a non-empty value. An empty string is valid because the schema does not specify minLength, and the field description explicitly instructs the model to return "" when no variant is identified. We intentionally keep the property required to maintain a consistent response shape and avoid treating both a missing property and an empty value as separate representations of “no variant selected.” We can clarify the description to say “This property is always required, but its value must be an empty string when no variant is selected.”

src/Apps/W1/SalesOrderAgent/app/.resources/Prompts/ItemSearch/itemselector-tool.md
tomasevicst: I partly agree. The current contract duplicates variant state:
1. **variant_code**
2. **variant_resolution**
3. **unresolved_variant_requests**

In the AL parser, variant_resolution values beginning with unresolved_* populate the same dictionary later populated by unresolved_variant_requests. Meanwhile, not_requested and resolved are not used downstream. That makes variant_resolution largely redundant.

However, variant_code plus generic confidence alone is insufficient:
 - An empty code could mean no variant was requested.
 - It could mean a requested variant does not exist.
 - A non-interchangeable missing variant may produce no selected_items entry, so its base item must be reported separately.
 - matching/alternative currently describes the complete item-variant option, not why a variant could not be selected.

I recommend simplifying to:
`{
  "selected_items": [
    {
      "item_no": "L100",
      "variant_code": "BLACK",
      "confidence": "alternative",
      "reason": "Bronze is unavailable; black is a safe color alternative."
    }
  ],
  "unresolved_variant_requests": [
    {
      "item_no": "L100",
      "alternatives_allowed": true
    }
  ]
}`

Remove variant_resolution from selected_items.

The meaning then becomes:
 - **Empty variant_code, no unresolved entry** - No variant was requested or identified
 - **Non-empty variant_code, no unresolved entry** - A concrete existing variant was selected
 - **Unresolved entry with true** - Requested value is absent; safe alternatives may be returned
 - **Unresolved entry with false** - Requested value is absent; do not substitute

This keeps one field for the concrete selection and one separate channel for cases where no concrete selection can safely represent the request.

src/Apps/W1/SalesOrderAgent/app/src/ItemSearch/SOAItemSearch.Codeunit.al
```diff
@@ -303,53 +345,286 @@ codeunit 4591 "SOA Item Search"
...
+                VariantObject.Add('Description 2', ItemVariant."Description 2");
+                VariantArray.Add(VariantObject);
+
+                if ItemVariants.ContainsKey(ItemVariant."Item No.") then
+                    ItemVariants.Set(ItemVariant."Item No.", VariantArray)
+                else
+                    ItemVariants.Add(ItemVariant."Item No.", VariantArray);
+            until ItemVariant.Next() = 0;
+
+        if not Item.FindSet() then
```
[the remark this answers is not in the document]
tomasevicst: Addressed. The initial Item iteration now caches the payload fields in a temporary Item buffer while building the Item No. filter. After the batched Item Variant lookup, the selector payload is assembled from that buffer, removing the second database read of the same Item rows.

src/Apps/W1/SalesOrderAgent/app/src/ItemSearch/SOAItemSearch.Codeunit.al
```diff
@@ -303,53 +345,294 @@ codeunit 4591 "SOA Item Search"
...
+
+        repeat
+            TempItem.Init();
+            TempItem."No." := Item."No.";
+            TempItem.Description := Item.Description;
+            TempItem."Description 2" := Item."Description 2";
+            TempItem.SystemId := Item.SystemId;
+            TempItem.Insert(false, true);
+
+            ItemNoFilterBuilder.SetRange("No.", Item."No.");
```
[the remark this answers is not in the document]
tomasevicst: Addressed. The candidate Item No. filter is now accumulated with a TextBuilder and converted to Text once before the batched Item Variant query, avoiding repeated Text reallocations in the candidate loop.

src/Apps/W1/SalesOrderAgent/app/.resources/Prompts/ItemSearch/itemselector-task.md
```diff
@@ -2,12 +2,25 @@ You are an intelligent item selection assistant for a Sales Order system.
...
+### SEARCH QUERY AND MESSAGE CONTEXT
+
+The `search_query` is the primary selection request.
+
+The payload can also contain `message_content`. Use it only as supporting context for the current `search_query`.
+
+- Use `message_content` to recover intent that clearly applies to the item in the `search_query`, such as "any color", a size, or an age group omitted from the extracted query.
+- A message can request multiple items. Do not select an item or variant merely because it appears elsewhere in `message_content`.
+- Apply a variant signal from `message_content` only when it clearly refers to the same product identified by the `search_query`.
+- If the `search_query` names only an item and `message_content` provides no relevant same-item variant signal, return an empty `variant_code`.
```
qasimikram: What does this last instruction mean? Can you share an example what this is for? maybe the names in the search_query is a variant in the system.

src/Apps/W1/SalesOrderAgent/app/.resources/Prompts/ItemSearch/itemselector-task.md
```diff
@@ -130,6 +143,40 @@ For broad queries (e.g., product families):
 
 ---
 
+### VARIANT SELECTION
+
+When a selected item has available variants in the Variants column, return `variant_code` using these rules:
+
+- Return the exact variant code when the query, or supporting message context for the same item, explicitly specifies a variant code or variant description.
```
qasimikram, line 161: when you use the work query, should we be consisten and always refer query as `search_query`

src/Apps/W1/SalesOrderAgent/app/.resources/Prompts/ItemSearch/itemselector-tool.md
```diff
@@ -3,41 +3,83 @@
...
+            },
+            "required": ["item_no", "alternatives_allowed"]
           }
         }
       },
-      "required": ["selected_items"]
+      "required": ["selected_items", "unresolved_variant_requests"]
     }
   },
   "additional_instructions": [
```
qasimikram: why we need such a big set of additional instructions when we covered most of it in the system prompt? should we have all instructions in one place, easy to read all instructions, and easy to maintain.

src/Apps/W1/SalesOrderAgent/app/.resources/Prompts/ItemSearch/itemselector-task.md
tomasevicst: This instruction covers searches where the product is identified but no specific variant is requested. In that case, we should not arbitrarily select one of the item’s variants. If the search_query itself contains a variant code or description, such as “Contoso bicycle RED,” then the instruction does not apply and RED should be selected. I agree the phrase “names only an item” is ambiguous, so we should reword it to explicitly say that an empty variant code is returned only when neither the search query nor relevant same-item message context identifies a variant from the candidate’s Variants data.

src/Apps/W1/SalesOrderAgent/app/.resources/Prompts/ItemSearch/itemselector-task.md
tomasevicst: Agreed. We should consistently use the exact payload property name search_query whenever referring to the primary search input, and message_content whenever referring to supporting message context. We will reserve terms such as “customer request” and “requested variant” for the inferred business intent. This removes ambiguity around what “query” means and makes the precedence rules easier to follow.

src/Apps/W1/SalesOrderAgent/app/.resources/Prompts/ItemSearch/itemselector-tool.md
tomasevicst: Agreed. The additional instructions were intended to reinforce important constraints at function-call time, but they now duplicate most of the system prompt and create two sources of behavioral truth. We should keep all selection, context, safety, and alternative rules in the system prompt, and reduce the tool definition to the structured response schema with concise field-specific descriptions. Where a constraint can be represented structurally, such as limiting returned confidence values to matching and alternative, we should encode it in the schema instead of repeating it as prose. We should validate the cleanup against the existing accuracy scenarios to ensure removing the duplicated instructions does not reduce model adherence.

src/Apps/W1/SalesOrderAgent/app/src/ItemSearch/SOAItemSearch.Codeunit.al
```diff
@@ -474,7 +686,7 @@ codeunit 4591 "SOA Item Search"
         end;
 
         // Log usage
-        FeatureTelemetry.LogUsage('0000QB0', SOASetup.GetFeatureName(), 'SOA Multi Items Availability: Find Items', TelemetryCustomDimension)
+        FeatureTelemetry.LogUsage('0000QB7', SOASetup.GetFeatureName(), 'SOA Multi Items Availability: Find Items v2', TelemetryCustomDimension)
```
qasimikram: I dont think we should change the tag here. We can know from the app version which item search they are on so no need to change that. This might break some dashboard or monitoring as well.

src/Apps/W1/SalesOrderAgent/app/src/Integration/SOATaskMessageReader.Codeunit.al
```diff
@@ -0,0 +1,61 @@
...
+using System.Agents;
+
+codeunit 4420 "SOA Task Message Reader"
+{
+    Access = Internal;
+    InherentEntitlements = X;
+    InherentPermissions = X;
+    Permissions = tabledata "Agent Task Message" = r;
+
+    internal procedure GetLastIncomingMessageContent(AgentTaskID: BigInteger): Text
```
qasimikram, line 18: I think we should also add attachment content to the message content that is returned, that is our main scenario where people send attachment, just the message might not be enought. My suggestion would be that MessageContent = Last email message + Content of all attachment in last email message.

src/Apps/W1/SalesOrderAgent/app/src/ItemSearch/SOAItemSelector.Codeunit.al
```diff
@@ -66,24 +96,64 @@ codeunit 4417 "SOA Item Selector"
...
+
+    local procedure LogItemSelectorException(CandidateCount: Integer; ErrorText: Text; ErrorCallStack: Text)
+    var
+        FeatureTelemetry: Codeunit "Feature Telemetry";
+        SOASetup: Codeunit "SOA Setup";
+        TelemetryDimensions: Dictionary of [Text, Text];
+    begin
+        TelemetryDimensions.Add('CandidateCount', Format(CandidateCount));
+        TelemetryDimensions.Add('Error', ErrorText);
+        FeatureTelemetry.LogError('0000QB3', SOASetup.GetFeatureName(), 'Item Selector Exception', ItemSelectorExceptionTelemetryMsg, ErrorCallStack, TelemetryDimensions);
```
qasimikram: Make sure new telemetry tags are not AI generated, and generated through the script.

qasimikram: There are other places where new tags were added. Just to double check.

src/Apps/W1/SalesOrderAgent/app/src/ItemSearch/SOAItemSelectorFunc.Codeunit.al
```diff
@@ -32,60 +41,151 @@ codeunit 4416 "SOA Item Selector Func" implements "AOAI Function"
...
+            if (VariantMatch <> MatchingTok) and (VariantMatch <> AlternativeTok) and (VariantMatch <> NotRequestedTok) then
+                exit(SetSelectionResultFailure(RejectedVariantMatchCombinationFailureTok));
+            if ((VariantCode = '') and (VariantMatch <> NotRequestedTok)) or
+               ((VariantCode <> '') and (VariantMatch = NotRequestedTok))
+            then
+                exit(SetSelectionResultFailure(RejectedVariantMatchCombinationFailureTok));
+
+            if not ItemObject.Get('confidence', ResultToken) or not ResultToken.IsValue() then
+                exit(false);
+            Confidence := ResultToken.AsValue().AsText();
```
qasimikram, line 129: Could you verify whether accepting inconsistent `confidence` and `variant_match` values is the desired behavior here? They are validated independently, so a response with `confidence: "matching"` and `variant_match: "alternative"` is accepted and added to `MatchingItemVariants`. Downstream logic can then treat a substitute variant as a match rather than requiring customer confirmation. Should the parser reject inconsistent cross-field combinations?

tomasevicst: Good catch. I updated the parser to reject **confidence: "matching"** combined with **variant_match: "alternative"** using the existing **RejectedVariantMatchCombination** failure category. This prevents substitute variants from entering the matching-item path and bypassing customer confirmation.

**confidence: "alternative"** with **variant_match: "matching"** remains valid because an alternative item may still contain the exact requested variant.

I also clarified the contract in the tool schema and prompt.

alexei-dobriansky:

<details>

Feature summary (Slice 622414): the agent's item selection for variants was reworked. The AOAI selector now returns each item with a `variant_match` value (`matching`, `alternative`, or `not_requested`) next to `confidence`, and the response is validated field by field before use. Variant storage moved from pipe-joined `Text` to `List of [Code[10]]`, so an item can carry several variant codes. A new `SOA Task Message Reader` codeunit passes the last incoming message and its attachment text into the selector as extra context, and new telemetry logs selector exceptions, response failures, and rejected item/variant counts.

Correctness of the changes since round 1: the new variant rules are coherent. When no variant is requested the selector returns `not_requested` and no variant code, which is a cleaner design than forcing variant guesses. When a requested variant is missing, same-item substitutes are offered only for interchangeable (presentation-only) dimensions, and non-interchangeable requests (size, voltage, fit) return an empty selection. Selected variant codes are re-checked against real `Item Variant` records in `AddSelectedItem`, and each returned item is validated for item number, variant code, and `variant_match`/`confidence` consistency. The move of message reading into a scoped codeunit with `tabledata "Agent Task Message" = r` on that codeunit only (not on `SOA Item Search`) is a good change. The main open point is that the availability page still shows one resolved variant per item, so the up-to-3 same-item variant alternatives the selector can produce are not all visible.
</details>

<details>

| ID  | Title                                          | Status        | Author response |
|-----|------------------------------------------------|---------------|-----------------|
| S1  | Show variant choices when none is specified    | Addressed     | Variant logic redesigned. Item-only requests now intentionally return `not_requested` with no variant, and same-item substitutes are offered only when a requested variant is missing and interchangeable. See S3 for the remaining display limit. |
| S2  | Add tests for variant selection and pricing    | Not addressed | No test was added in the commits since round 1. |
</details>

<details>

**S3** - *Availability page shows only one variant per item*
`ResolvedItemVariants` is still keyed by item `SystemId` with a single `Code[10]`, and the availability page returns one variant per item. The selector can now return up to 3 same-item variant alternatives, but only the first available one is shown. Please confirm this is intended, or surface the extra variant alternatives so the customer can see them.

**S4** - *Full message and attachment text sent to the selector*
The selector call now includes the last incoming email body and all attachment text as `message_content`. Please confirm this wider payload is needed for item selection, and consider limiting it to the part that refers to the current item. This reduces the data sent to the model and lowers the chance that unrelated attachment text influences the choice.
</details>

<details>

**Risk:** the changes touch item selection, variant resolution, availability, and price calculation, so a wrong variant can still show wrong stock or a wrong price. The redesign reduces this risk by validating every returned variant code against `Item Variant` and by returning an empty selection for non-interchangeable requests. Two areas still carry risk: there is no automated test for the new validation and classification logic (`ParseSelectedItems`, `AddSelectedItem`, candidate enrichment), and the selector now receives the full message plus attachment text (see S4). The interchangeable-by-default rule for color, finish, and pattern is a judgment call and may substitute a variant the customer did not want.

**Necessity:** the feature is needed and matches the Slice direction for item variant support in availability and pricing. The scope is right for a Slice. Tests remain absent (S2); note there is no dedicated Sales Order Agent test app and the core selection is AOAI-driven, so full coverage is hard, but the deterministic parsing and validation code is testable and worth covering given the sensitive area.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9364 round=2 by=alexei-dobriansky at=2026-08-03T17:16:48Z lastSha=cab664b1a39628d659f988c157c718dae5ab89d8 suggestions=S1:addressed,S2:notaddressed,S3:new,S4:new parentRound=1

src/Apps/W1/SalesOrderAgent/app/src/ItemSearch/SOAItemSearch.Codeunit.al
```diff
@@ -219,9 +244,14 @@ codeunit 4591 "SOA Item Search"
                 CountBeforeAvailabilityCheck := ItemFilter.Split('|').Count();
                 ApplyAvailabilityFilter := CheckAvailability and (SOASetup."Search Only Available Items" and not SOASetup."Incl. Capable to Promise");
 
-                // Try item selection only for broader search results with multiple candidates
-                if (SearchType = 'broader_item_search') and (CountBeforeAvailabilityCheck > 1) then
-                    if SelectBestItem(ItemFilter, BuildSearchQueryText(SearchKeyWordsTrimmed), CandidateArray, SelectedMatchingItemFilter, SelectedAlternativeItemFilter) then begin
+                // Run item selection for all candidate payloads so variant resolution is consistent.
+                if CandidateArray.Count() > 0 then begin
+                    SearchQuery := BuildSearchQueryText(SearchKeyWordsTrimmed);
+                    MessageContent := TaskMessageReader.GetLastIncomingMessageContent(AgentTaskID);
```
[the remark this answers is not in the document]
tomasevicst: This is intentional for the current implementation. The extracted item search query can omit variant intent such as “any color,” particularly when that intent appears in the email body or an attachment. The selector receives the latest message as untrusted supporting context and is instructed to use only signals clearly referring to the item in the current search_query. This scope is documented as an approved design decision. Narrowing the payload requires reliable item-scoped context extraction and will be considered separately to avoid regressing multi-item and attachment scenarios.

alexei-dobriansky:

<details>

The new commits refine Sales Order Agent item selection after round 2. They cache the task message content, add a safety gate for variant substitutions, prefer concrete variant alternatives, validate selected variants, and reduce logged error details.

The direction is sound: unsafe variant substitutions are now discarded, and an unavailable requested variant can produce an empty selection instead of silently choosing a risky substitute. The code still keeps only one resolved variant per item, and the selector still receives the full `message_content` payload, so two earlier concerns remain open.
</details>

<details>

| ID  | Title                                      | Status        | Author response |
|-----|--------------------------------------------|---------------|-----------------|
| S1  | Show variant choices when none is specified | Addressed     | Already addressed in round 2; no regression seen in the new commits. |
| S2  | Add tests for variant selection and pricing | Not addressed | No Sales Order Agent test change was added in the commits since round 2. |
| S3  | Availability page shows only one variant per item | Not addressed | The new code still stores one `Code[10]` per item in `ResolvedItemVariants` and uses the first available or first selected variant. |
| S4  | Full message and attachment text sent to the selector | Not addressed | The payload still includes `message_content`; the new code caches it but does not narrow it. |
</details>

<details>

**S5** - *Prompt examples omit required safety field*  
The tool schema now requires `variant_substitution_safety`, but the concrete examples still omit it. Please update the examples, especially the safe alternative variant example, so the model sees the same shape that the code requires.
</details>

<details>

**Risk:** the risk is still centered on selecting the wrong item variant, showing wrong availability, or calculating a quote from an unintended variant. The new safety gate lowers that risk for unsafe substitutions, but the single resolved variant per item and the wider selector payload remain regression surfaces.

**Necessity:** the Slice remains useful and appropriately scoped for variant-aware Sales Order Agent item selection. The new commits are necessary refinements, but the open suggestions should still be considered before merge.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9364 round=3 by=alexei-dobriansky at=2026-08-04T11:04:17Z lastSha=2b611183 suggestions=S1:addressed,S2:notaddressed,S3:notaddressed,S4:notaddressed,S5:new parentRound=2

qasimikram (approved):

AleksandricMarko (approved):

PR #9369, opened 2026-07-13
[FI] Delocalize VAT Entry EU Service key into VAT Core FI app

djukicmilica:
## Summary

Delocalizes the FI-specific customization of the **VAT Entry** table (254) out of the FI BaseApp layer into a dedicated first-party extension app, **VAT Core FI**.

### Background

The FI layer overrode the *entire* `VAT Entry` table object (1413 lines) solely to append the standard field `"EU Service"` (field 59) to **Key4**. That full-object override is unnecessary and drifts from W1.

### Change

- **New app `VAT Core FI`** (`src/Apps/FI/VATCoreFI`, id range 13421-13425, OnPrem) containing a single `tableextension 13421 "VAT Entry FI"`.
- The extension adds a **supplementary key** `EUServiceFI` = Key4 fields + `"EU Service"`, with the same `SumIndexFields = Base, "Additional-Currency Base"`. Table extensions cannot modify a base key, so the FI-needed SIFT/index is provided as an additional key; the base W1 Key4 is left unchanged.
- **Deletes** the redundant FI override `src/Layers/FI/BaseApp/Finance/VAT/Ledger/VATEntry.Table.al`.
- **Registers** `VAT Core FI` in the BCApps build registry (`build/projects.json`, `build/groups.json`), mirroring `FI Core`.

### Behavior

Behavior-preserving. No fields, procedures, or controls change. No feature-key gating (a key/SIFT index is a passive schema element). No permission set required (key-only table extension).


djukicmilica:
Consolidating the VAT Entry EU Service key delocalization into PR #9362 (FI VIES declaration delocalization) instead of a standalone PR.

PR #9371, opened 2026-07-13
Bug 642180: DE report 11015 Export Business Data fails when scheduled via Job Queue: Client callbacks are not supported

AleksanderGladkov:
**ISSUE:**
The DE report "Export Business Data" (11015) fails when it is scheduled to run via a Job Queue Entry, ending with "Client callbacks are not supported... attempted to issue a client callback to download a file (CodeUnit 419 File Management)".

**CAUSE:**
The report is processing-only and delivers its result by pushing the ZIP straight to the client (a client callback). A Job Queue runs in a background session with no client, so the callback throws and nothing is produced.

**SOLUTION:**
When the report runs in the background (no GUI), the generated ZIP is now stored in the Report Inbox instead of being pushed to the client, so the user can download it from there - the same way scheduled layout reports behave. Running the report interactively still downloads the ZIP directly as before.

**TESTS:**
Tested manually both new and old ways of export, works as expected

Fixes [AB#642180](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/642180)

### DIFF
--- a/src/Layers/DACH/BaseApp/Local/Finance/AuditFileExport/ExportBusinessData.Report.al
+++ b/src/Layers/DACH/BaseApp/Local/Finance/AuditFileExport/ExportBusinessData.Report.al
@@ -920,9 +920,6 @@ report 11015 "Export Business Data"
         TempBlobZipArchive.CreateInStream(ZipArchiveInStream);
         ZipFileName := DataExportRecordDefinition."Data Exp. Rec. Type Code" + '.zip';
 
-        // When the report runs in the background (e.g. scheduled via Job Queue) there is no client
-        // to receive a file download callback. Save the result to the Report Inbox instead so the
-        // user can download it from there.
         if not GuiAllowed() then begin
             SaveZipToReportInbox(ZipArchiveInStream, DataExportRecordDefinition);
             exit;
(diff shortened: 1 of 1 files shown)


alexei-dobriansky:

<details>

This PR changes DE report 11015 "Export Business Data" so non-GUI runs save the generated ZIP to Report Inbox instead of calling the File Management download handlers. Interactive runs still use the old download path.

The fix addresses the root cause of the Job Queue failure: the background path exits before `DownloadHandler` or `DownloadFromStreamHandler` can issue a client callback. The ZIP stream is created after `SaveZipArchive()`, then copied into `Report Inbox` with `Output Type = Zip`, so the generated data is not lost. I did not find a problem in the new stream handling or the `GuiAllowed()` branch.
</details>

<details>

**S1** - *Add a Job Queue regression test*
Add a regression test for the Job Queue path. Run report 11015 in a no-GUI or Job Queue execution path, then assert that a Report Inbox entry for the user has `Output Type = Zip` and a non-empty `Report Output` blob. Keep one interactive-path assertion so the old download behavior stays covered.
</details>

<details>

**Risk:** The code change is narrow and only affects report 11015 in non-GUI sessions. The main regression risk is output delivery: scheduled exports must create a usable ZIP in Report Inbox, while manual exports must still download the ZIP directly. There are no public API or event signature changes.

**Necessity:** The linked bug is active and clear. Scheduling this finance audit export currently fails with a client callback error, and the practical workaround is to run it manually. The scope is right, but this background behavior is important enough to need automated coverage.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9371 round=1 by=alexei-dobriansky at=2026-07-14T05:47:56Z lastSha=724b4f485016eb3a18b405271bae6469acae1117 suggestions=S1

ventselartur (approved):

AndersLarsenMicrosoft (approved):

PR #9398, opened 2026-07-14
[Main]Incorrect Cost Amount (Actual) in Value Entries after posting Assembly Process and calulating the assembly standard costsInitial commit

sanjmaurya:
[Bug 642126](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/642126): [Master] [ALL-E] Incorrect Cost Amount (Actual) in Value Entries after posting Assembly Process and calulating the assembly standard costs

Fixes [AB#642126](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/642126)

Issue :- Incorrect Cost Amount (Actual) in Value Entries after posting Assembly Process and calulating the assembly standard costs

Root cause :-
The item's per-unit indirect cost and the per-output indirect cost are computed with inconsistent rounding, so the amounts diverge once Calc. Assembly Std. Cost re-stamps the item's standard cost with high precision.

Solutions :-
The assembly overhead recompute runs after the if/else, covering both branches
if not InvtAdjmtEntryOrder.Find() or not InvtAdjmtEntryOrder."Completely Invoiced" then
InvtAdjmtEntryOrder.GetCostsFromItem(OutputQty)
else begin
InvtAdjmtEntryOrder.RoundCosts(OutputQty);
InvtAdjmtEntryOrder.CalcDirectCostFromCostShares();
InvtAdjmtEntryOrder.CalcIndirectCostFromCostShares();
InvtAdjmtEntryOrder.CalcUnitCost();
end;

if InvtAdjmtEntryOrder."Order Type" = InvtAdjmtEntryOrder."Order Type"::Assembly then begin
InvtAdjmtEntryOrder.CalcOvhdCost(OutputQty); // overhead from stable matcap × Indirect %
InvtAdjmtEntryOrder.RoundCosts(1);
InvtAdjmtEntryOrder.CalcIndirectCostFromCostShares();
InvtAdjmtEntryOrder.CalcUnitCost();
end;


v-ankitgoyal (dismissed):

qasimikram (changes requested):
S1 - Regression test does not verify the reported cost invariant (High confidence)

In `SCMCalculateAssemblyCost.Codeunit.al`, `CalcAssemblyStdCostDoesNotDistortAssemblyOutputActualCost` only asserts that no Manufacturing Overhead variance entry exists. Bug 642126 specifically requires `Cost Amount (Actual)` / indirect cost to remain **1.12 instead of 1.17**. This test could pass if adjustment produces no variance entry but leaves the output at another incorrect amount.

Please assert the resulting assembly-output cost amount, and preferably the relevant indirect-cost value entry, against the rolled standard cost or the explicit expected value. The production logic is directionally consistent with `CalcActualUsageCosts`, but this high-risk costing change needs an end-state assertion for the customer-visible result.

v-ankitgoyal (dismissed):

qasimikram (changes requested):
S1 - Regression test omits the trigger (High confidence)

The reported bug requires recalculating the assembly standard cost between the two cost-adjustment runs. The linked work item explicitly notes that the result is correct when that intermediate calculation is omitted, but this test posts the item charge and immediately runs the final adjustment without calling `CalculateAssemblyStandardCost` again. As written, the test can pass on the unfixed code and does not prove this production change.

Please add `CalculateAssemblyStandardCost(AssemblyItem."No.");` after `PostFreightChargeToReceipt(...)` and before the final `AdjustCostItemEntries(...)`, then confirm the test fails without the production fix.

v-ankitgoyal (approved):

PredragMaricic (approved):

sanjmaurya:
> S1 - Regression test omits the trigger (High confidence)
> 
> The reported bug requires recalculating the assembly standard cost between the two cost-adjustment runs. The linked work item explicitly notes that the result is correct when that intermediate calculation is omitted, but this test posts the item charge and immediately runs the final adjustment without calling `CalculateAssemblyStandardCost` again. As written, the test can pass on the unfixed code and does not prove this production change.
> 
> Please add `CalculateAssemblyStandardCost(AssemblyItem."No.");` after `PostFreightChargeToReceipt(...)` and before the final `AdjustCostItemEntries(...)`, then confirm the test fails without the production fix.

Please review i have done the suggested changes

src/Layers/W1/Tests/SCM-Assembly/SCMCalculateAssemblyCost.Codeunit.al
```diff
@@ -565,9 +717,25 @@ codeunit 137911 "SCM Calculate Assembly Cost"
         Item.Validate("Costing Method", CostingMethod);
         Item.Validate("Replenishment System", ReplenishmentSystem);
         Item.Validate("Standard Cost", StandardCostAmt);
+        Item.Validate("VAT Prod. Posting Group", VATProdPostingGroup);
```
qasimikram, line 720: S1: `CreateItem` now always validates `"VAT Prod. Posting Group"` from a global populated only by the new test. Test-codeunit globals can outlive the test transaction, while the VAT setup created by that test is rolled back, so another test executed afterward can reuse a stale posting group and become order-dependent. Please scope this assignment to the two items in the new scenario, or parameterize the helper so existing tests retain the posting group created by `LibraryInventory.CreateItem`.

qasimikram (approved):
The production change is narrowly scoped to assembly standard-cost calculation, and the regression test now covers the required intermediate `CalculateAssemblyStandardCost` trigger. The remaining test-isolation concern is captured in S1.

PR #9418, opened 2026-07-14
[ALAppExtensions #30229][Event Request] table 12113 "Tmp Withholding Contribution"

Explorer986:
## What & why

Adds `OnCalculateWithholdingTaxOnBeforeAssignWithholdingTaxValues` (an `IsHandled` event) to table 12113 "Tmp Withholding Contribution" and extracts the standard field assignments in `CalculateWithholdingTax()` into a new `AssignWithholdingTaxValues` local procedure (the reviewer-requested pattern, instead of wrapping an inline block). The negative-taxable-base check stays outside the handled block so it always runs, regardless of who computes the values.

## Linked work

Fixes [AB#641342](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/641342)

## How I validated this

- Diff review: assignment logic moved verbatim into `AssignWithholdingTaxValues`; the `IsHandled` event precedes it; `Error(NegativeTaxableBaseErr)` remains after the guarded block. Default path is behavior-preserving.
- No tests added: extract-method plus opt-in override; standard result and the negative-base validation are unchanged.

## Risk & compatibility

Low. `IsHandled` guards the calculation only; the negative-base validation is intentionally kept outside it so it cannot be bypassed. No existing event signatures changed, no schema/data impact.

### DIFF
--- a/src/Layers/IT/BaseApp/Local/Finance/WithholdingTax/TmpWithholdingContribution.Table.al
+++ b/src/Layers/IT/BaseApp/Local/Finance/WithholdingTax/TmpWithholdingContribution.Table.al
@@ -444,16 +444,16 @@ table 12113 "Tmp Withholding Contribution"
             Error(NegativeTaxableBaseErr);
     end;
 
-    local procedure AssignWithholdingTaxValues(WithholdCodeLine: Record "Withhold Code Line")
+    local procedure AssignWithholdingTaxValues(WithholdCodeLineToAssign: Record "Withhold Code Line")
     begin
         "Withholding Account" := WithholdCode."Withholding Taxes Payable Acc.";
-        "Withholding Tax %" := WithholdCodeLine."Withholding Tax %";
-        "Non Taxable %" := 100 - WithholdCodeLine."Taxable Base %";
+        "Withholding Tax %" := WithholdCodeLineToAssign."Withholding Tax %";
+        "Non Taxable %" := 100 - WithholdCodeLineToAssign."Taxable Base %";
         "Taxable Base" := Round(((
                                   "Total Amount" -
                                   "Base - Excluded Amount" -
                                   "Non Taxable Amount By Treaty") *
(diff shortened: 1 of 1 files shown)


alexei-dobriansky:

# **Recommendation: Accept**

<details>

This PR adds an IsHandled integration event in table 12113 "Tmp Withholding Contribution" before the withholding tax values are assigned in CalculateWithholdingTax(). It also moves the existing standard assignments into AssignWithholdingTaxValues().

The default path is behavior-preserving: WithholdCode and WithholdCodeLine are still loaded first, the same fields are assigned when IsHandled is false, the existing OnCalculateWithholdingTaxOnAfterAssignWithholdingTaxAmount event still runs after the assignment point, and the negative taxable-base validation still always runs. The new publisher uses [IntegrationEvent(false, false)], has a standard OnCalculateWithholdingTaxOnBefore... name, and passes both WithholdCodeLine and ar Rec so subscribers can replace the calculation when needed.
</details>

<details>

No suggestions.
</details>

<details>

**Risk:** The regression surface is small because the standard calculation path is the same unless an extension subscribes and sets IsHandled := true. The new event signature becomes public extension API, so it must be right before merge, but the parameters and timing match the requested customization point. No table schema, persisted data, posting flow, or existing event signature changes.

**Necessity:** The request is valid for an extensibility slice. The existing after-event only exposes the temporary withholding contribution record, so subscribers cannot use WithholdCodeLine at the point where the standard values are calculated. A test is not needed for this pure publisher addition and extract-method refactor because the default product behavior is unchanged.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9418 round=1 by=alexei-dobriansky at=2026-07-14T11:06:24Z lastSha=39bf7278211a9a9fcc347f1b5c825e4ea8dd26db suggestions=none

AndersLarsenMicrosoft (approved):

mynjj (approved):

PR #9421, opened 2026-07-14
[Event Request] codeunit 367 "CheckManagement": add OnFinancialVoidCh…

Explorer986:
## What & why

Adds `OnFinancialVoidCheckOnBeforePostBalanceAccount` (an `IsHandled` event) to codeunit 367 "CheckManagement" so partners can replace the balance-account posting in `FinancialVoidCheck` — the exact extension point the requester needed but could not get from the too-early existing events. The large `case CheckLedgEntry."Bal. Account Type"` block was extracted verbatim into a new `FinancialVoidPostBalanceAccount` local procedure so the guard reads cleanly.

## Linked work

Fixes [AB#640242](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/640242)

## How I validated this

- Diff review: the case body is moved unchanged into `FinancialVoidPostBalanceAccount` (with `ConfirmFinancialVoid.GetVoidType()/GetVoidDate()` passed in as `VoidType`/`VoidDate` params); default path (`IsHandled := false`) calls it, preserving behavior. AL language server reports no errors.

## Risk & compatibility

Low-medium. Behavior is preserved when unsubscribed, but this is core financial-void posting, so it deserves the regression check above. The new `IsHandled` event lets a subscriber skip the standard bal.-account posting - document if the event is treated as public. No schema/data impact.

### DIFF
--- a/src/Layers/RU/BaseApp/Bank/Check/CheckManagement.Codeunit.al
+++ b/src/Layers/RU/BaseApp/Bank/Check/CheckManagement.Codeunit.al
@@ -266,13 +266,38 @@ codeunit 367 CheckManagement
         GenJnlLine2."Allow Zero-Amount Posting" := true;
         CheckUnpostedVendor(CheckLedgEntry, GenJnlLine2);
         OnFinancialVoidCheckOnBeforeCheckBalAccountType(GenJnlLine2, CheckLedgEntry, BankAccLedgEntry3);
+        IsHandled := false;
+        OnFinancialVoidCheckOnBeforePostBalanceAccount(GenJnlLine2, CheckLedgEntry, BankAccLedgEntry3, BalanceAmountLCY, IsHandled);
+        if not IsHandled then
+            FinancialVoidPostBalanceAccount(CheckLedgEntry, ConfirmFinancialVoid.GetVoidType(), ConfirmFinancialVoid.GetVoidDate(), BalanceAmountLCY);
+
+        if ConfirmFinancialVoid.GetVoidDate() = CheckLedgEntry."Check Date" then begin
+            BankAccLedgEntry2.Open := false;
+            BankAccLedgEntry2."Remaining Amount" := 0;
+            BankAccLedgEntry2."Statement Status" := BankAccLedgEntry2."Statement Status"::Closed;
+            BankAccLedgEntry2.Modify();
+        end;
+
+        // rounding error from currency conversion
+        if CheckAmountLCY + BalanceAmountLCY <> 0 then
+            PostRoundingAmount(BankAcc, CheckLedgEntry, ConfirmFinancialVoid.GetVoidDate(), -(CheckAmountLCY + BalanceAmountLCY));
+
+        MarkCheckEntriesVoid(CheckLedgEntry, ConfirmFinancialVoid.GetVoidDate());
+        Commit();
+        UpdateAnalysisView.UpdateAll(0, true);
+
+        OnAfterFinancialVoidCheck(CheckLedgEntry);
+    end;
+
+    local procedure FinancialVoidPostBalanceAccount(var CheckLedgEntry: Record "Check Ledger Entry"; VoidType: Integer; VoidDate: Date; var BalanceAmountLCY: Decimal)
+    begin
         case CheckLedgEntry."Bal. Account Type" of
             CheckLedgEntry."Bal. Account Type"::"G/L Account":
                 FinancialVoidPostGLAccount(GenJnlLine2, BankAccLedgEntry2, CheckLedgEntry, BalanceAmountLCY);
             CheckLedgEntry."Bal. Account Type"::Customer:
                 begin
-                    if ConfirmFinancialVoid.GetVoidType() = 0 then   // Unapply entry
-                        if UnApplyCustInvoices(CheckLedgEntry, ConfirmFinancialVoid.GetVoidDate()) then
+                    if VoidType = 0 then   // Unapply entry
+                        if UnApplyCustInvoices(CheckLedgEntry, VoidDate) then
                             GenJnlLine2."Applies-to ID" := CheckLedgEntry."Document No.";
                     CustLedgEntry.SetCurrentKey("Transaction No.");
                     CustLedgEntry.SetRange("Transaction No.", BankAccLedgEntry2."Transaction No.");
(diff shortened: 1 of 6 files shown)


AndersLarsenMicrosoft (dismissed):

alexei-dobriansky:

<details>

This PR adds `OnFinancialVoidCheckOnBeforePostBalanceAccount` in codeunit 367 so an extension can replace the standard balance-account posting during `FinancialVoidCheck`. It also moves the existing balance-account `case` block into `FinancialVoidPostBalanceAccount`, and the default path still calls that block when `IsHandled` stays false.

The event is in the right part of the flow: after the void bank line is posted and before the balance-account entries are posted. The standard path keeps updating `BalanceAmountLCY`, closing the bank ledger entry, posting rounding, marking the check entries void, and committing as before. The blocker is the handled path: a subscriber can skip the standard posting, but the event does not pass all context needed to safely do an equivalent replacement.
</details>

<details>

**S1** - *Pass posting context to handled subscribers*
Add the current `GenJnlPostLine` instance and the selected void type/date to `OnFinancialVoidCheckOnBeforePostBalanceAccount`. A subscriber that sets `IsHandled := true` must be able to post replacement lines through the same posting instance and match the same unapply decision. Without that context, the later rounding, mark-void, and commit steps can run after an incomplete replacement.
</details>

<details>

**Risk:** This is core financial void posting in W1, NA, APAC, and RU. The unsubscribed path looks behavior-preserving, including `BalanceAmountLCY` feeding the rounding calculation. The risk is for subscribers of the new `IsHandled` event: the standard balance-account posting and unapply block can be skipped, while the procedure still closes ledger state, posts rounding if needed, marks check entries void, commits, and updates analysis views.

**Necessity:** The event request is valid because the existing early event forces extensions to replace too much of `FinancialVoidCheck`. The scope is close to the right extension point, and missing tests are not by itself a blocker for this event request. The event still needs the missing posting context before it is safe as a replacement seam in this posting-adjacent flow.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9421 round=1 by=alexei-dobriansky at=2026-07-14T17:02:29Z lastSha=acf90107064ab1e2e47b0e25a8060adcc9fe0641 suggestions=S1

src/Layers/RU/BaseApp/Bank/Check/CheckManagement.Codeunit.al
```diff
@@ -266,13 +266,38 @@ codeunit 367 CheckManagement
         GenJnlLine2."Allow Zero-Amount Posting" := true;
         CheckUnpostedVendor(CheckLedgEntry, GenJnlLine2);
         OnFinancialVoidCheckOnBeforeCheckBalAccountType(GenJnlLine2, CheckLedgEntry, BankAccLedgEntry3);
```
mynjj: consider using this event and extending the parameter list; one event after another is not a good pattern IMO

Explorer986: Sure, Added two trailing parameters, 1. IsHandled 2. BalanceAmountLCY which will let subscriber bypass case block

Added BalanceAmountLCY so that if a subscriber sets IsHandled = true then there will not be rounding error from currency conversion

`
if CheckAmountLCY + BalanceAmountLCY <> 0 then
            PostRoundingAmount(BankAcc, CheckLedgEntry, ConfirmFinancialVoid.GetVoidDate(), -(CheckAmountLCY + BalanceAmountLCY));
`

alexei-dobriansky:

<details>

The new commit removes the separate `OnFinancialVoidCheckOnBeforePostBalanceAccount` event and changes the existing `OnFinancialVoidCheckOnBeforeCheckBalAccountType` event instead. It adds `var BalanceAmountLCY` and `var IsHandled`, and the surrounding code now skips the standard balance-account `case` block when `IsHandled` is true.

This addresses only the rounding part of the previous concern. A handled subscriber still does not get the current `GenJnlPostLine` instance, and it still does not get the selected void type. The void date can be inferred from the journal line posting date, but the unapply decision cannot be matched safely without the void type.
</details>

<details>

| S1 | Pass posting context to handled subscribers | Not addressed | The response says `IsHandled` and `BalanceAmountLCY` were added so a subscriber can bypass the case block and keep rounding balanced. The new event signature still does not pass `GenJnlPostLine` or the void type. |
</details>

<details>

**S2** - *Do not add IsHandled to existing event*
Keep `OnFinancialVoidCheckOnBeforeCheckBalAccountType` as the existing notification event and add a new dedicated handled event instead. Adding `var IsHandled` changes the contract and meaning of an event that subscribers may already use. This can break subscribers or let one subscriber unexpectedly skip the standard financial void posting for all others.
</details>

<details>

**Risk:** This is still core financial void posting in W1, NA, APAC, and RU. The round-2 change also changes a public integration event signature and turns an existing notification event into an override seam. That is a compatibility risk on top of the original posting risk.

**Necessity:** The event request is still valid because extensions need a narrower point to replace balance-account posting. The replacement seam must be safe for posting and safe for existing subscribers. The current shape is not safe enough because it keeps the handled path under-contextualized and changes an existing event instead of adding a dedicated one.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9421 round=2 by=alexei-dobriansky at=2026-07-15T17:04:30Z lastSha=f36f6719919d30399abc3815a6ad931aa0e8b807 suggestions=S1:notaddressed,S2:new parentRound=1

src/Layers/W1/BaseApp/Bank/Check/CheckManagement.Codeunit.al
```diff
@@ -1142,11 +1144,13 @@ codeunit 367 CheckManagement
...
     /// <param name="BankAccountLedgerEntry">Bank account ledger entry associated with the check</param>
+    /// <param name="BalanceAmountLCY">Running balancing amount in LCY. A subscriber that sets IsHandled must add the LCY amounts it posts so the caller's currency-rounding calculation stays balanced.</param>
+    /// <param name="IsHandled">Set to true to skip the standard balance account type posting logic</param>
     /// <remarks>
     /// Raised from FinancialVoidCheck procedure before processing balance account type specific logic.
     /// </remarks>
     [IntegrationEvent(false, false)]
-    local procedure OnFinancialVoidCheckOnBeforeCheckBalAccountType(var GenJournalLine: Record "Gen. Journal Line"; var CheckLedgerEntry: Record "Check Ledger Entry"; var BankAccountLedgerEntry: Record "Bank Account Ledger Entry")
+    local procedure OnFinancialVoidCheckOnBeforeCheckBalAccountType(var GenJournalLine: Record "Gen. Journal Line"; var CheckLedgerEntry: Record "Check Ledger Entry"; var BankAccountLedgerEntry: Record "Bank Account Ledger Entry"; var BalanceAmountLCY: Decimal; var IsHandled: Boolean)
```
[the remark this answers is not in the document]
mynjj: I don't see how 

```
OnWhatever()
OnWhatever2(IsHandled)
```

would make it any better, or impact any of the subscribers. Consider not making this a hard rule, we can talk about this one.

Already adding an `IsHandled` in general wherever could already be frowned upon, but clearly we are not getting anywhere in that battle of redesigning extensibility to use other patterns, more in old code.

mynjj (dismissed):

AndersLarsenMicrosoft (dismissed):

mynjj (approved):

djukicmilica (approved):

PR #9422, opened 2026-07-14
table 12116 "Withholding Tax": scope Reported/Paid modify guard to st…

Explorer986:
## What & why

The original request asked for an `IsHandled` event in `OnModify` to bypass the `Reported`/`Paid` immutability check - which triage rejected because it lets any subscriber silently disable a compliance guard that official withholding-tax reporting/payment relies on. Instead, this narrows the guard: a modify on a `Reported`/`Paid` entry is still blocked when any standard field changed, but modifications limited to extension fields (added by other apps) are allowed. Partners can maintain their own data without altering officially reported values, and no bypass hook is exposed.

## Linked work

Fixes [AB#640860](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/640860)

## How I validated this

- Diff review: `OnModify` now calls `StandardFieldModified(xRec)`, which compares `Rec` vs `xRec` over base fields (field no. < 50000, `FieldClass::Normal`) via `RecordRef`. Table has only simple field types (no BLOB/Media), so the value comparison is safe. AL language server reports no errors.
- modifying a standard field on a reported entry still errors, modifying only an extension field succeeds. 

## Risk & compatibility

Medium-low. Standard reported values stay fully immutable (no regression there); the only relaxation is allowing extension-field-only edits, which were previously over-blocked. Small per-modify `RecordRef` loop on a low-volume, user-driven table — negligible perf. No schema/data impact. Worth a reviewer/product nod since it touches compliance behavior.

### DIFF
--- a/src/Layers/IT/BaseApp/Local/Finance/WithholdingTax/WithholdingTax.Table.al
+++ b/src/Layers/IT/BaseApp/Local/Finance/WithholdingTax/WithholdingTax.Table.al
@@ -246,10 +246,12 @@ table 12116 "Withholding Tax"
     end;
 
     trigger OnModify()
+    var
+        PreviousWithholdingTax: Record "Withholding Tax";
     begin
-        if Reported or
-           Paid
-        then
+        if not PreviousWithholdingTax.Get(Rec."Entry No.") then
+            exit;
+        if (Reported or Paid) and StandardFieldModified(PreviousWithholdingTax) then
             Error(Text1033);
     end;
 
@@ -343,6 +345,28 @@ table 12116 "Withholding Tax"
     begin
         WithholdingTaxLine.SetRange("Withholding Tax Entry No.", "Entry No.");
         WithholdingTaxLine.DeleteAll(true);
+    end;
+
+    local procedure StandardFieldModified(PreviousWithholdingTax: Record "Withholding Tax"): Boolean
+    var
+        CurrentRecordRef: RecordRef;
+        PreviousRecordRef: RecordRef;
+        CurrentFieldRef: FieldRef;
+        PreviousFieldRef: FieldRef;
+        FieldIndex: Integer;
+    begin
+        CurrentRecordRef.GetTable(Rec);
+        PreviousRecordRef.GetTable(PreviousWithholdingTax);
+        for FieldIndex := 1 to CurrentRecordRef.FieldCount() do begin
+            CurrentFieldRef := CurrentRecordRef.FieldIndex(FieldIndex);
+            // Field numbers below 50000 belong to the standard table; 50000 and above are extension fields.
+            if (CurrentFieldRef.Class = FieldClass::Normal) and (CurrentFieldRef.Number < 50000) then begin
+                PreviousFieldRef := PreviousRecordRef.Field(CurrentFieldRef.Number);
+                if CurrentFieldRef.Value <> PreviousFieldRef.Value then
+                    exit(true);
+            end;
+        end;
+        exit(false);
     end;
 
     [IntegrationEvent(false, false)]


AndersLarsenMicrosoft (dismissed):

alexei-dobriansky:

<details>

This PR narrows the OnModify guard in table 12116 "Withholding Tax". A Reported or Paid record still errors when a standard field changes, but a modify that only changes extension fields is allowed.

The approach is safer than an IsHandled event because it does not add a general bypass. The RecordRef comparison checks normal base fields below 50000 against xRec, so changes to official withholding tax values are still caught. I did not find a direct data-integrity hole in the code, but the new allowed path is not covered by a test.
</details>

<details>

**S1** - *Add tests for scoped modify guard*
Add a test for table 12116 that changes only a test extension field on a Reported or Paid record and verifies Modify(true) succeeds. Also keep or add a check that changing a standard field on the same record still fails, because this guard protects reported withholding tax data.
</details>

<details>

**Risk:** This touches an Italian compliance guard for reported or paid withholding tax records. The code change is narrow and has only a small RecordRef loop on modify, but a wrong comparison would either keep blocking partner data or allow changes to official reported values. There is no public event or schema change, and no BaseApp publisher dependency was introduced.

**Necessity:** The linked request is valid, but the requested IsHandled bypass would be too broad. This narrower fix is the right direction because partners can store their own data without changing official values. A regression test is needed and feasible here; the IT test app already has table 12116 OnModify tests, and this is sensitive data-integrity logic.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9422 round=1 by=alexei-dobriansky at=2026-07-14T17:06:09Z lastSha=c1bb712215486b7377dbc9ff9caac71d1a1944b9 suggestions=S1

src/Layers/IT/BaseApp/Local/Finance/WithholdingTax/WithholdingTax.Table.al
```diff
@@ -247,9 +247,11 @@ table 12116 "Withholding Tax"
...
     trigger OnModify()
     begin
-        if Reported or
-           Paid
-        then
+        // Reported or paid withholding taxes are official values and must stay immutable.
+        // Only standard (base) fields are protected; changes limited to extension fields added
+        // by other apps are allowed, so partners can maintain their own data without altering
+        // the officially reported figures.
+        if (Reported or Paid) and StandardFieldModified(xRec) then
```
mynjj: `xRec` can have some unexpected behavior with API pages and maybe other scenarios (xRec not actually having the values prior to the modification). Is this a valid concern? Are there any APIs that use the table? If not, there could be partner APIs.

if `xRec` has the same values as `Rec` then `StandardFieldModified` will always return false, even when there are standard fields modified. This would mean that the tax record could have relevant field updated if they do it through an API without triggering an error.

What can we do in that case?

My recomendation: check if the xRec behavior that I mentioned is still true for API calls or if this was a weird behavior in the past of the platform that doesn't happen anymore, and check in other places of the codebase, I think there are some  places where we consider xRec and APIs, so you could incorporate that here.

Explorer986: Hi Joshua, thanks for concerns.
Agree, I think xRec gives unexpected behavior in below scenarios

1. When we are working with APIs which works non-interactively or xRec has modified values and that may match with Rec in comparison.
2.  When we insert record or new record on Pages then xRec doesn't work because it is not populated before. 

In this repo only Pages and Cards uses "Withholding Taxes" table but customer can always implement a new Custom API.

Now in onModify() trigger I am fetching last committed record from DB and then if record not found then it is not valid case for modify (there can be no modify call for non-existent record) then simply exit() otherwise compare current record with previous record which is fetched from DB.

AndersLarsenMicrosoft (approved):

mynjj (approved):

PR #9425, opened 2026-07-14
hasTranslations most be specified if there is translation

AndersLarsenMicrosoft:
## Linked work

Fixes #

## How I validated this


AndersLarsenMicrosoft:
a mistake

PR #9439, opened 2026-07-14
[Extensibility Request] issue 30330: add OnCancelInvoiceOnBeforeConfirm event

AleksandricMarko:
## Summary
When a posted sales invoice is corrected or cancelled, an extension can now change what happens to the originating sales order's quantities (via #29949, shipped in 28.2). However, the confirmation dialog shown before cancelling still states that the original quantities will be restored, which becomes inaccurate when that behavior is overridden. The reporter needs a hook to adjust the confirmation text so it matches the actual behavior. This PR adds an integration event that lets extensions override the confirmation question before it is shown.

Source issue repository: microsoft/AlAppExtensions; issue number: 30330

## Changes Made
- `Cancel PstdSalesInv (Yes/No)` (codeunit 1323), `CancelInvoice` - resolve the confirmation text into a `ConfirmQuestion` variable and raise the new `OnCancelInvoiceOnBeforeConfirm` integration event before the `Confirm` call, allowing subscribers to adjust the message; added the event publisher.

Fixes [AB#641733](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/641733)

### DIFF
--- a/src/Layers/W1/BaseApp/Sales/History/CancelPstdSalesInvYesNo.Codeunit.al
+++ b/src/Layers/W1/BaseApp/Sales/History/CancelPstdSalesInvYesNo.Codeunit.al
@@ -37,12 +37,15 @@ codeunit 1323 "Cancel PstdSalesInv (Yes/No)"
         CorrectPostedSalesInvoice: Codeunit "Correct Posted Sales Invoice";
         PageManagement: Codeunit "Page Management";
         IsHandled: Boolean;
+        ConfirmQuestion: Text;
     begin
         IsHandled := false;
         OnCancelInvoiceOnBeforeTestCorrectInvoiceIsAllowed(SalesInvoiceHeader, IsHandled);
         if not IsHandled then
             CorrectPostedSalesInvoice.TestCorrectInvoiceIsAllowed(SalesInvoiceHeader, true);
-        if Confirm(GetCancelPostedInvoiceQst(SalesInvoiceHeader)) then
+        ConfirmQuestion := GetCancelPostedInvoiceQst(SalesInvoiceHeader);
+        OnCancelInvoiceOnBeforeConfirm(SalesInvoiceHeader, ConfirmQuestion);
+        if Confirm(ConfirmQuestion) then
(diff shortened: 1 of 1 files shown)


src/Layers/W1/BaseApp/Sales/History/CancelPstdSalesInvYesNo.Codeunit.al
```diff
@@ -37,12 +37,15 @@ codeunit 1323 "Cancel PstdSalesInv (Yes/No)"
         CorrectPostedSalesInvoice: Codeunit "Correct Posted Sales Invoice";
         PageManagement: Codeunit "Page Management";
         IsHandled: Boolean;
+        ConfirmQuestion: Text;
```
6CRIPT, line 40: What if the susbscriber uses a blank ConfirmQuestion.?The result would be weird for the user. Should we enforce  the text to not be empty? At least we can document that subscribers should not leave it empty.

6CRIPT (approved):
consider adding a small test that verifies a subscriber can override the confirm message, even another one double-checking that standar message shows up when no subscriber.

src/Layers/W1/BaseApp/Sales/History/CancelPstdSalesInvYesNo.Codeunit.al
AleksandricMarko: Partner using the publisher is responsible for how system works with it. There is a lot of places they can brake system but that doesn't benefit them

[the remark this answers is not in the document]
AleksandricMarko: @copilot please fix it

qasimikram (approved):
The new publisher exposes the resolved confirmation question at the correct point immediately before the dialog. Subscribers can adjust the text while default cancellation behavior remains unchanged.

PredragMaricic (approved):

PR #9441, opened 2026-07-14
[Extensibility Request] issue 30337: add OnCreateProdOrderLineOnBeforeCreateProdOrderLineFromItem event

AleksandricMarko:
## Summary
The issue author needs to completely replace the standard `Source Type = Item` processing in `CreateProdOrderLine` under specific business conditions, executing their own production order line creation and controlling the resulting `ErrorOccured` outcome. None of the existing events (`OnCreateProdOrderLineOnBeforeInitProdOrderLine`, `OnBeforeProdOrderLineInsert`, `OnAfterProdOrderLineInsert`, `OnAfterCreateProdOrderLine`) can bypass the standard flow. This PR adds an `IsHandled`-based integration event that lets an extension take full ownership of the item-line creation and set `ErrorOccured`.

Source issue repository: microsoft/AlAppExtensions; issue number: 30337

## Changes Made
- `CreateProdOrderLine` (codeunit 99000787 "Create Prod. Order Lines") - extracted the standard `Source Type = Item` branch into a new local procedure `CreateProdOrderLineFromItem` and raised a new `OnCreateProdOrderLineOnBeforeCreateProdOrderLineFromItem` integration event with `var IsHandled` and `var ErrorOccured` before calling it, so subscribers can replace the standard item-line creation and report the outcome; when `IsHandled` stays false the original behavior and all existing events are preserved.
- `OnCreateProdOrderLineOnBeforeCreateProdOrderLineFromItem` - new `[IntegrationEvent(false, false)]` publisher exposing `IsHandled`, `ErrorOccured`, the `Production Order` record, and the variant code.

Fixes [AB#641918](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/641918)

### DIFF
--- a/src/Layers/W1/BaseApp/Manufacturing/Document/CreateProdOrderLines.Codeunit.al
+++ b/src/Layers/W1/BaseApp/Manufacturing/Document/CreateProdOrderLines.Codeunit.al
@@ -245,22 +245,22 @@ codeunit 99000787 "Create Prod. Order Lines"
         OnAfterCreateProdOrderLine(ProdOrder, VariantCode, ErrorOccured);
     end;
 
-    local procedure CreateProdOrderLineFromItem(ProdOrder: Record "Production Order"; VariantCode: Code[10]; var ErrorOccured: Boolean)
+    local procedure CreateProdOrderLineFromItem(var ProductionOrder: Record "Production Order"; VariantCode: Code[10]; var ErrorOccured: Boolean)
     begin
         OnCreateProdOrderLineOnBeforeInitProdOrderLine(InsertNew);
-        InitProdOrderLine(ProdOrder."Source No.", VariantCode, ProdOrder."Location Code");
-        ProdOrderLine.Description := ProdOrder.Description;
-        ProdOrderLine."Description 2" := ProdOrder."Description 2";
-        ProdOrderLine.Validate(Quantity, ProdOrder.Quantity);
+        InitProdOrderLine(ProductionOrder."Source No.", VariantCode, ProductionOrder."Location Code");
+        ProdOrderLine.Description := ProductionOrder.Description;
+        ProdOrderLine."Description 2" := ProductionOrder."Description 2";
+        ProdOrderLine.Validate(Quantity, ProductionOrder.Quantity);
         ProdOrderLine.UpdateDatetime();
         if SalesLineIsSet then
             CopyDimFromSalesLine(SalesLine, ProdOrderLine);
-        OnBeforeProdOrderLineInsert(ProdOrderLine, ProdOrder, SalesLineIsSet, SalesLine);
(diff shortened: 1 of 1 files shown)


src/Layers/W1/BaseApp/Manufacturing/Document/CreateProdOrderLines.Codeunit.al
```diff
@@ -254,6 +245,24 @@ codeunit 99000787 "Create Prod. Order Lines"
         OnAfterCreateProdOrderLine(ProdOrder, VariantCode, ErrorOccured);
     end;
 
+    local procedure CreateProdOrderLineFromItem(ProdOrder: Record "Production Order"; VariantCode: Code[10]; var ErrorOccured: Boolean)
```
AleksandricMarko: @copilot ProdOrder: Record "Production Order" should be with var and renamed to ProductionOrder

src/Layers/W1/BaseApp/Manufacturing/Document/CreateProdOrderLines.Codeunit.al
```diff
@@ -835,6 +844,11 @@ codeunit 99000787 "Create Prod. Order Lines"
     begin
     end;
 
+    [IntegrationEvent(false, false)]
```
JakovljevicDusan, line 847: I suggest to add IsHandled and other parameters into OnCreateProdOrderLineOnBeforeInitProdOrderLine(), so we don't have to add OnCreateProdOrderLineOnBeforeCreateProdOrderLineFromItem. And to keep new procedure CreateProdOrderLineFromItem(), so OnCreateProdOrderLineOnBeforeInitProdOrderLine behaves as OnBeforeCreateProdOrderLineFromItem()....

AleksandricMarko: we introduced the rule that it is not allowed to convert regular event to IsHandled as purpose will be changed
That is why there is additional event out of procedure

PredragMaricic (approved):

qasimikram (changes requested):
S1: Remove `var` from the `ProductionOrder` parameter of `OnCreateProdOrderLineOnBeforeCreateProdOrderLineFromItem`. The approved event contract exposes the order as read-only. With the current signature, a subscriber can mutate it while leaving `IsHandled = false`, causing the standard helper and the final `OnAfterCreateProdOrderLine` event to run against altered input.

The extracted helper can remain `var` to preserve the behavior of existing events inside the original item branch. Please also add regression coverage for the unhandled refactor and the fully handled replacement path.

qasimikram (approved):

PR #9443, opened 2026-07-14
[Extensibility Request] issue 30222: add handled event before RecreatePurchLines in Pay-to Vendor No. validation

AleksandricMarko:
## Summary

The issue author needs a way to conditionally replace or bypass the standard `RecreatePurchLines` call that runs when the "Pay-to Vendor No." field is validated on a Purchase Header. Some custom purchase lines carry additional information that the standard line recreation would discard, so the extension must be able to run its own recreation logic instead. The existing extensibility points around this flow do not offer an `IsHandled` mechanism to skip the standard call, forcing partners to duplicate the entire validation trigger. This PR adds a handled integration event immediately before the `RecreatePurchLines` call so subscribers can take ownership of the recreation when needed.

Source issue repository: microsoft/AlAppExtensions; issue number: 30222

## Changes Made
- `OnValidatePayToVendorNoOnBeforeRecreatePurchLines` - new `IntegrationEvent` raised in the `OnValidate` trigger of field "Pay-to Vendor No." on table 38 "Purchase Header", immediately before `RecreatePurchLines(PayToVendorTxt)`. The event exposes `Rec`, `xRec`, and a `var IsHandled` flag (initialized to `false`); the standard `RecreatePurchLines` call now runs only when `IsHandled` remains `false`. Applied in the W1 base layer and propagated to the same-named table files in the APAC, BE, CH, DACH, ES, FI, GB, IT, NA, NL, NO, RU, and SE layers.

Fixes [AB#641640](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/641640)

### DIFF
--- a/src/Layers/W1/BaseApp/Purchases/Document/PurchaseHeader.Table.al
+++ b/src/Layers/W1/BaseApp/Purchases/Document/PurchaseHeader.Table.al
@@ -304,8 +304,12 @@ table 38 "Purchase Header"
 
                 if (xRec."Buy-from Vendor No." = "Buy-from Vendor No.") and
                    (xRec."Pay-to Vendor No." <> "Pay-to Vendor No.")
-                then
-                    RecreatePurchLines(PayToVendorTxt);
+                then begin
+                    IsHandled := false;
+                    OnValidatePayToVendorNoOnBeforeRecreatePurchLines(Rec, xRec, IsHandled);
+                    if not IsHandled then
+                        RecreatePurchLines(PayToVendorTxt);
+                end;
 
                 if not SkipPayToContact then
                     UpdatePayToCont("Pay-to Vendor No.");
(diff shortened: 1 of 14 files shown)


tomasevicst (approved):

PredragMaricic (approved):

src/Layers/W1/BaseApp/Purchases/Document/PurchaseHeader.Table.al
```diff
@@ -304,8 +304,12 @@ table 38 "Purchase Header"
 
                 if (xRec."Buy-from Vendor No." = "Buy-from Vendor No.") and
                    (xRec."Pay-to Vendor No." <> "Pay-to Vendor No.")
-                then
-                    RecreatePurchLines(PayToVendorTxt);
+                then begin
+                    IsHandled := false;
+                    OnValidatePayToVendorNoOnBeforeRecreatePurchLines(Rec, xRec, IsHandled);
+                    if not IsHandled then
```
qasimikram, line 310: S1: Please add regression coverage proving that the default purchase-line recreation remains unchanged and that a handled subscriber can recreate the lines exactly once while preserving required custom values and leaving subsequent validation logic with valid line state.

qasimikram (approved):
The event is correctly scoped to the existing vendor-change condition and is propagated consistently across the country-layer copies.

S1: Please add regression coverage proving that the default purchase-line recreation remains unchanged and that a handled subscriber can recreate the lines exactly once while preserving required custom values and leaving subsequent validation logic with valid line state.

PR #9446, opened 2026-07-14
[Extensibility Request] issue 30216: add OnPostItemJnlLineJobConsumptionOnBeforePrepareJobLine event in codeunit 90 Purch.-Post

AleksandricMarko:
## Summary
The reporter needs to intervene during purchase posting immediately before `InvoicePostingInterface.PrepareJobLine(...)` runs for job-linked purchase lines. Under specific custom conditions (based on extension fields) they must adjust `PurchLine."Qty. to Invoice"` and optionally replace the standard job-line preparation with their own implementation. Existing events fire too early or after preparation has already consumed the value, so none allow this. This PR adds an `IsHandled` integration event at that exact point so subscribers can modify the quantity and/or skip the standard call.

Source issue repository: microsoft/AlAppExtensions; issue number: 30216

## Changes Made
- `OnPostItemJnlLineJobConsumptionOnBeforePrepareJobLine` - new `IsHandled` integration event raised in `PostItemJnlLineJobConsumption` right after `Qty. to Invoice` is set and before `InvoicePostingInterface.PrepareJobLine` is called, letting subscribers adjust the quantity or bypass the standard preparation. `IsHandled` is initialized to `false` before the call and the standard call is guarded by `if not IsHandled`. Applied in the W1 base layer and propagated to the country layers that carry their own copy of the codeunit.

Fixes [AB#641618](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/641618)


PredragMaricic (dismissed):

src/Layers/W1/BaseApp/Purchases/Posting/PurchPost.Codeunit.al
```diff
@@ -6001,7 +6001,10 @@ codeunit 90 "Purch.-Post"
 
             if QtyToBeInvoiced <> 0 then begin
                 PurchLine."Qty. to Invoice" := QtyToBeInvoiced;
-                InvoicePostingInterface.PrepareJobLine(PurchHeader, PurchLine, PurchLineACY);
+                IsHandled := false;
+                OnPostItemJnlLineJobConsumptionOnBeforePrepareJobLine(PurchLine, QtyToBeInvoiced, PurchHeader, PurchLineACY, IsHandled);
+                if not IsHandled then
```
qasimikram, line 6006: S1: Please add regression coverage for the default and handled posting paths, including a subscriber that changes `Qty. to Invoice`, performs replacement preparation, and completes without duplicate or missing job-posting data.

src/Layers/IT/BaseApp/Purchases/Posting/PurchPost.Codeunit.al
```diff
@@ -1,4 +1,4 @@
-﻿// ------------------------------------------------------------------------------------------------
+// ------------------------------------------------------------------------------------------------
```
qasimikram: S2: Please revert this unrelated UTF-8 BOM removal so the IT file changes only for the requested event.

qasimikram (dismissed):
The event is placed at the requested preparation boundary and is propagated consistently across the localized code copies.

S1: Please add regression coverage for the default and handled posting paths, including a subscriber that changes `Qty. to Invoice`, performs replacement preparation, and completes without duplicate or missing job-posting data.

S2: Revert the unrelated UTF-8 BOM removal in the IT file so this PR remains limited to the event change.

AleksandricMarko:
@copilot Please revert this unrelated UTF-8 BOM removal so the IT file changes only for the requested event

qasimikram (approved):
The event placement and signature support the linked scenario while preserving standard behavior when unhandled. The IsHandled scope is limited to PrepareJobLine, and the layer copies are consistent.

tomasevicst (approved):

PR #9449, opened 2026-07-14
[Extensibility Request] issue 30334: fix OnAfterCarryOutToReqWksh record order

AleksandricMarko:
## Summary
A subscriber to the `OnAfterCarryOutToReqWksh` integration event in codeunit 99000813 "Carry Out Action" needs to update the newly created target requisition line based on the source line it was copied from. The event's first parameter is passed by `var` precisely so subscribers can modify that record, but in version 28.1 the call was changed to pass the source line as the first argument and the target line by value, removing any way to persist changes to the target line. This regression blocks the reporter's process. This PR restores the previous argument order so the modifiable `var` parameter is once again the target requisition line.

Source issue repository: microsoft/ALAppExtensions; issue number: 30334

## Changes Made
- `Carry Out Action.CarryOutToReqWksh` - updated the `OnAfterCarryOutToReqWksh` call to pass the target requisition line as the (`var`) argument, so subscribers can modify and persist the target line; applied in both the W1 and NA layers.

Fixes [AB#641682](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/641682)

### DIFF
--- a/src/Layers/NA/BaseApp/Inventory/Requisition/CarryOutAction.Codeunit.al
+++ b/src/Layers/NA/BaseApp/Inventory/Requisition/CarryOutAction.Codeunit.al
@@ -217,7 +217,7 @@ codeunit 99000813 "Carry Out Action"
                 OnCarryOutToReqWkshOnAfterPlanningCompInsert(PlanningComponent2, PlanningComponent);
             until PlanningComponent.Next() = 0;
 
-        OnAfterCarryOutToReqWksh(RequisitionLine, RequisitionLine2, ReqWkshTempName, ReqJournalName, LineNo);
+        OnAfterCarryOutToReqWksh(RequisitionLine2, RequisitionLine, ReqWkshTempName, ReqJournalName, LineNo);
     end;
 
     procedure GetTransferOrdersToPrint(var TransferHeader: Record "Transfer Header")
--- a/src/Layers/NA/BaseApp/Inventory/Requisition/CarryOutAction.Codeunit.al
+++ b/src/Layers/NA/BaseApp/Inventory/Requisition/CarryOutAction.Codeunit.al
@@ -217,7 +217,7 @@ codeunit 99000813 "Carry Out Action"
                 OnCarryOutToReqWkshOnAfterPlanningCompInsert(PlanningComponent2, PlanningComponent);
             until PlanningComponent.Next() = 0;
 
-        OnAfterCarryOutToReqWksh(RequisitionLine2, RequisitionLine, ReqWkshTempName, ReqJournalName, LineNo);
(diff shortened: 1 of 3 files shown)


AleksandricMarko:
@copilot src/Layers/W1/BaseApp/Manufacturing/Inventory/Requisition/MfgCarryOutAction.Codeunit.al should not be affected with change

AleksandricMarko:
@copilot seems like current change is wrong and instead the only thing which need to be done is to set in event signature RequisitionLine2: Record "Requisition Line" as VAR parameter

alexei-dobriansky:

# **Recommendation: Accept**

<details>

This PR fixes a regression in `OnAfterCarryOutToReqWksh` (codeunit 99000813 "Carry Out Action") where the second parameter `RequisitionLine2` — the newly created target requisition line — was missing `var`, making it impossible for subscribers to persist changes to it. The fix adds `var` to `RequisitionLine2` in the event publisher declaration in both the W1 and NA layers.

The root cause is correctly addressed: adding `var` to the publisher parameter enables subscribers to modify and save the target line without altering argument order or breaking existing callers. The call site at line 220 passes a local variable, so no call-site change is needed. The existing internal subscriber in `MfgCarryOutAction.Codeunit.al` continues to work without modification because it does not need to modify `RequisitionLine2` and AL allows subscribers to receive a parameter by value even when the publisher declares it as `var`.

</details>

<details>

No suggestions.

</details>

<details>

**Risk:** Minimal. This is a one-line change per file, scoped to the event publisher signature. Adding `var` to an existing event parameter is backward-compatible: existing subscribers that declare the parameter without `var` continue to compile and receive a copy. No call-site changes are required. The NA layer change is identical to W1, which is consistent with how this area is structured. The only affected code paths are those that carry out requisition worksheet actions, and the behavior for existing subscribers does not change.

**Necessity:** The change is necessary. Without it, any extension subscriber that needs to update the target requisition line — the documented purpose of the `var` parameter — cannot do so. The scenario is clear (reported via ALAppExtensions issue 30334), the regression is confirmed, and the fix is correctly scoped to the event signature alone.

</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=1 system=github pr=9449 round=1 by=alexei-dobriansky at=2026-08-11T10:05:17Z lastSha=140565c17a3bddc7c6357e95ac757161e79578db reviewKey=72965c0cc3892486e2cff52c226063462969d5a0fd6a656dac6da5f6e5c85dc8 suggestions=

alexei-dobriansky (approved):

tomasevicst (approved):

PR #9450, opened 2026-07-14
[Extensibility Request] issue 30290: Add OnBeforeUpdateSetupOnBillToCustomerChangeInSalesHeader event

AleksandricMarko:
## Summary

The issue author has a separate extension that already handles assignment of the VAT Registration No. and related posting groups when the bill-to customer changes on a sales header. Because `SetBillToCustomerAddressFieldsFromCustomer` always calls `AltCustVATRegFacade.UpdateSetupOnBillToCustomerChangeInSalesHeader`, the standard logic can overwrite or conflict with the values set by their feature. This PR adds an `IsHandled` integration event so subscribers can skip that specific call while leaving the rest of the standard logic intact.

Source issue repository: microsoft/AlAppExtensions; issue number: 30290

## Changes Made
- `OnBeforeUpdateSetupOnBillToCustomerChangeInSalesHeader` - New `IntegrationEvent` raised in the `SetBillToCustomerAddressFieldsFromCustomer` procedure of table `Sales Header`, right before the `AltCustVATRegFacade.UpdateSetupOnBillToCustomerChangeInSalesHeader` call. `IsHandled` is initialized to `false` and, when a subscriber sets it to `true`, the standard call is skipped. The same change was propagated to every layer that has its own copy of the file.

Fixes [AB#641535](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/641535)

### DIFF
--- a/src/Layers/W1/BaseApp/Sales/Document/SalesHeader.Table.al
+++ b/src/Layers/W1/BaseApp/Sales/Document/SalesHeader.Table.al
@@ -7971,7 +7971,10 @@ table 36 "Sales Header"
         end else
             "Payment Method Code" := BillToCustomer."Payment Method Code";
 
-        AltCustVATRegFacade.UpdateSetupOnBillToCustomerChangeInSalesHeader(Rec, xRec, BillToCustomer);
+        IsHandled := false;
+        OnBeforeUpdateSetupOnBillToCustomerChangeInSalesHeader(Rec, BillToCustomer, IsHandled);
+        if not IsHandled then
+            AltCustVATRegFacade.UpdateSetupOnBillToCustomerChangeInSalesHeader(Rec, xRec, BillToCustomer);
 
         "Customer Posting Group" := BillToCustomer."Customer Posting Group";
         "Currency Code" := BillToCustomer."Currency Code";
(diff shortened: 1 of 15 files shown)


PredragMaricic (approved):

src/Layers/W1/BaseApp/Sales/Document/SalesHeader.Table.al
```diff
@@ -7971,7 +7971,10 @@ table 36 "Sales Header"
         end else
             "Payment Method Code" := BillToCustomer."Payment Method Code";
 
-        AltCustVATRegFacade.UpdateSetupOnBillToCustomerChangeInSalesHeader(Rec, xRec, BillToCustomer);
+        IsHandled := false;
+        OnBeforeUpdateSetupOnBillToCustomerChangeInSalesHeader(Rec, BillToCustomer, IsHandled);
+        if not IsHandled then
```
qasimikram, line 7976: S1: Please add regression coverage proving that the standard facade update remains effective when unhandled and that a handled subscriber's custom VAT registration and posting setup is not overwritten by the standard call.

qasimikram (approved):
The event is narrowly scoped around the requested VAT registration setup update and is propagated consistently across the country-layer copies.

S1: Please add regression coverage for both branches. Verify that the standard facade update remains effective when unhandled and that a handled subscriber's custom VAT registration and posting setup is not overwritten by the standard call.

PR #9460, opened 2026-07-15
AI Test Toolkit: language-first data-driven eval support

thloke:
## What & why  Add language-first, data-driven AI test support using `TestDataSource`, an `AIT Test Case Context` interface, and a shared platform test handler. This preserves the existing AI Test Toolkit logging and suite behavior while allowing evals to migrate away from the classic runner with minimal test-body changes.  The No. Series Copilot accuracy test is migrated as the proof of concept, and the new context interface now has parity with the classic `AIT Test Context` APIs used by existing evals.  ## Linked work  Issue/work item link to be added before marking ready for review.  ## How I validated this  - [x] I read the full diff and it contains only changes I intended. - [x] I built the affected app(s) locally with no new analyzer warnings. - [x] I ran the change in Business Central and confirmed it behaves as expected. - [x] I added or updated tests for the new behavior, or explained below why none are needed.  **What I tested and the outcome**  - Compiled and published AI Test Toolkit and the aggregate Business Foundation Tests app to local `Navision_BCPlatform`. - Ran codeunit 133689 through ALTool and confirmed `NOSERIES-GOOD.YAML` fans out to all 50 named cases (`ACCURACY_01` through `ACCURACY_50`). - Confirmed the shared before-case and after-case test-handler hooks are invoked. - Confirmed a model-independent diagnostic body completes all 50 cases successfully. - With the platform handler-isolation fix, confirmed the normally isolated codeunit persists exactly 50 rows with   50 distinct input codes, one run ID, and the range `ACCURACY_01` through `ACCURACY_50`. - Ran Eval Suite `NOS-ACCUR` directly and confirmed exactly 50 suite-linked rows, 50 distinct cases, one run ID,   and no duplicate fan-out. - The original No. Series model call reaches the test body but cannot complete locally because the environment lacks the Azure Key Vault secret `NoSeriesCopilotToolsSelectionPromptV2`.  ## Risk & compatibility  - Existing classic suite execution remains supported; the handler defers logging while running under an AIT suite to avoid duplicate entries. - Standalone handler persistence depends on the platform fix in   `private/thloke/test-handlers-outside-isolation` (`5d4ac0420fd`), which runs hooks on a non-test SQL connection. - The PR remains draft pending the linked work item and review.    

[AB#634642](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/634642)


src/Business Foundation/Test/NoSeriesCopilot/src/NoSeriesCopilotAccuTests.Codeunit.al
```diff
@@ -15,14 +15,14 @@ codeunit 133689 "No. Series Copilot Accu. Tests"
...
+    TestHandlers = "AIT Test Handler";
 
     var
         Assert: codeunit "Library Assert";
-        AITTestContext: codeunit "AIT Test Context";
         NoSeriesCopilotTestLib: codeunit "Library - No. Series Copilot";
 
-    [Test]
-    procedure NoSeriesPositiveTests();
+    [TestDataSource(Codeunit::"AIT Test Data Source", 'NOSERIES-GOOD.YAML')]
```
nikolakukrika, line 24: Can we keep the test attribute?
[Test]
[TestDataSource(Codeunit::"AIT Test Data Source", 'NOSERIES-GOOD.YAML')]

I would really like to have all tests keep the test attribute :). It is easier to find them, count them and etc...

nikolakukrika, line 24: So will I specify all the yaml files as a context here? Is it comma separated? How does this work?

src/Business Foundation/Test/NoSeriesCopilot/src/NoSeriesCopilotAccuTests.Codeunit.al
```diff
@@ -15,14 +15,14 @@ codeunit 133689 "No. Series Copilot Accu. Tests"
...
+    procedure NoSeriesPositiveTests(context: interface "AIT Test Case Context")
```
nikolakukrika, line 25: I liked the static implementation better than passing in by parameter. Static gives more freedom. So if we would have a static interface that is called AITTestCaseContext this would be the best IMO. Then we do not need to pass it around that would require more refactoring, but use the static method.

nikolakukrika: If this is difficult to do in the platform, we can do it in the app test library, given we have a place to hook in.

src/Tools/AI Test Toolkit/src/DataDriven/AITDDCurrentCase.Codeunit.al
```diff
@@ -0,0 +1,86 @@
...
+/// Bridges the dataset lineage of the currently executing language-first (<c>[TestDataSource]</c>) data-driven case
+/// to the AIT logging pipeline, and holds the per-case start baseline used when computing the log entry.
+///
+/// Under the classic app-based path the current row is carried on the platform Test Method Line's Data Input fields;
+/// under <c>[TestDataSource]</c> the platform drives the fan-out and those fields are empty, so the per-case context
+/// records its row here (single-instance, in-memory). Because the platform runs ITestHandler hooks outside the
+/// per-function isolation scope, <see cref="AIT Test Handler"/> writes the log entry directly from
+/// <c>OnAfterTestCaseRun</c> — no cross-scope buffering is required.
+/// </summary>
+codeunit 149033 "AIT DD Current Case"
```
nikolakukrika: Minor AIT Current DD Test Case

nikolakukrika: This code should go to the Test Toolkit solution as data driven tests reside there.

src/Tools/AI Test Toolkit/src/DataDriven/AITDDCurrentCase.Codeunit.al
```diff
@@ -0,0 +1,86 @@
...
+    /// <summary>Session-scoped run identifier used to correlate per-case log entries when a data-driven test
+    /// runs directly on the platform test runner (no AIT test suite context).</summary>
+    procedure GetRunId(): Guid
+    begin
+        if IsNullGuid(RunId) then
+            RunId := CreateGuid();
+        exit(RunId);
+    end;
+
+    [EventSubscriber(ObjectType::Codeunit, Codeunit::"Test Runner - Mgt", 'OnAfterRunTestSuite', '', false, false)]
```
nikolakukrika: Is this on the Suite or on the method?
If we clear the input code that is per method it should be OnAfterRunTestMethod

src/Tools/AI Test Toolkit/src/DataDriven/AITDDCurrentCase.Codeunit.al
```diff
@@ -0,0 +1,86 @@
...
+        Clear(CurrentInputCode);
+        HasCase := false;
+    end;
+
+    /// <summary>Session-scoped run identifier used to correlate per-case log entries when a data-driven test
+    /// runs directly on the platform test runner (no AIT test suite context).</summary>
+    procedure GetRunId(): Guid
+    begin
+        if IsNullGuid(RunId) then
+            RunId := CreateGuid();
```
nikolakukrika: We should refactor the AIT Test Suite context to read from this one.

src/Tools/AI Test Toolkit/src/DataDriven/AITTestDataSource.Codeunit.al
```diff
@@ -0,0 +1,109 @@
...
+using System.Test;
+using System.TestTools.TestRunner;
+
+/// <summary>
+/// Shared data source for language-first data-driven AI evals. A single instance of this codeunit is referenced
+/// by every consuming app via <c>[TestDataSource(Codeunit::"AIT Test Data Source", '&lt;dataset&gt;')]</c>; there is
+/// no need for a per-app provider. It resolves the dataset from the shared <c>Test Input Group</c>/<c>Test Input</c>
+/// tables (imported from each app's dataset resource) and returns one <see cref="ITestContext"/> per data row.
+/// </summary>
+codeunit 149038 "AIT Test Data Source" implements ITestDataSource
```
nikolakukrika, line 17: This one should be moved to the Test Runner solution and renamed.

src/Tools/AI Test Toolkit/src/TestSuite/AITALTestSuiteMgt.Codeunit.al
```diff
@@ -71,6 +79,37 @@ codeunit 149037 "AIT AL Test Suite Mgt"
...
+            exit(CodeunitMetadata."Has Test Data Source");
+        exit(false);
+    end;
+
+    /// <summary>
+    /// Adds the codeunit's test methods to the AL test suite once, without expanding the dataset into per-row
+    /// Data Input lines. Used for language-first (<c>[TestDataSource]</c>) codeunits, where the platform performs
+    /// the per-case fan-out.
+    /// </summary>
+    local procedure AddCodeunitWithoutDataExpansion(var AITTestMethodLine: Record "AIT Test Method Line")
```
nikolakukrika, line 97: If we do not add the lines, how will this work in UI?
Will we be able to see the results in the UI? 

If not we need to react on the end of the test case and udpate the individual lines. The solution should go to the Test Runner solution.

nikolakukrika: Keep in mind that data driven tests can be outside of the AIT tool. They can also run for regular tests in the Test Tool page. There is no log entries there.

src/Tools/AI Test Toolkit/src/TestSuite/AITTestSuiteMgt.Codeunit.al
```diff
@@ -444,6 +445,119 @@ codeunit 149034 "AIT Test Suite Mgt."
...
+            AITLogEntry.Status := AITLogEntry.Status::Error;
+            AITLogEntry."Original Status" := AITLogEntry.Status::Error;
+        end;
+
+        AITLogEntry."Start Time" := StartTime;
+        AITLogEntry."End Time" := CurrentDateTime();
+        if AITLogEntry."Start Time" <> 0DT then
+            AITLogEntry."Duration (ms)" := AITLogEntry."End Time" - AITLogEntry."Start Time";
+
+        AITLogEntry."Tokens Consumed" := AOAIToken.GetTotalServerSessionTokensConsumed() - StartTokens;
```
nikolakukrika, line 501: This can work for now, but please leave a TODO to update when we extend for agent tests. We will not be able to collect tokens in this way.

nikolakukrika (changes requested):

src/Business Foundation/Test/NoSeriesCopilot/src/NoSeriesCopilotAccuTests.Codeunit.al
thloke: You will add one [TestDataSource] attribute per yaml file. So something like:

```
[TestDataSource(Codeunit::"AIT Test Data Source", 'NOSERIES-GOOD-1.YAML')]
[TestDataSource(Codeunit::"AIT Test Data Source", 'NOSERIES-GOOD-2.YAML')]
```

thloke: The compiler does not allow both attributes. I verified this directly: combining `[Test]` and `[TestDataSource]` produces `AL0189 Attribute 'Test' cannot be specified, because 'TestDataSource' is already specified`. A `[TestDataSource]` method is still a test method and is reported once per generated case, so the PR keeps only `[TestDataSource]` and documents the constraint.

thloke: Verified with the compiler: stacked `[TestDataSource]` attributes compile. The README now documents one attribute per dataset and notes that case identifiers must be unique across datasets on the same method because the platform result name is `Method[caseName]`.

src/Business Foundation/Test/NoSeriesCopilot/src/NoSeriesCopilotAccuTests.Codeunit.al
thloke: Addressed at the app level without changing the platform contract in 74cd4ff39. The platform-required context parameter remains on the method, but `AIT Test Data Source.GetTestCase` now eagerly preloads the selected row into the existing single-instance `Test Input` / `AIT Test Context` path before the body runs. Existing test bodies and helper codeunits can therefore keep calling `Codeunit "AIT Test Context"` unchanged; the No. Series migration now demonstrates that compatibility pattern.

src/Tools/AI Test Toolkit/src/DataDriven/AITDDCurrentCase.Codeunit.al
thloke: The original codeunit was removed rather than only renamed. Generic current-case/run identity is now `Test Data Source Context` in Test Runner; the remaining AIT-only timing/token baselines are in `AIT Test Case State`.

thloke: Addressed in 140d57524. Generic data-driven case/run state moved to the Test Runner app as `Test Data Source Context`. AI Test Toolkit retains only AIT-specific timing/token state in `AIT Test Case State`.

src/Tools/AI Test Toolkit/src/DataDriven/AITDDCurrentCase.Codeunit.al
thloke: Changed to method/case ownership. Test Runner clears the shared current-case identity after each decorated case result, and AIT clears its method-scoped timing/token state on `OnAfterRunTestMethod`; suite cleanup remains only as a defensive boundary.

src/Tools/AI Test Toolkit/src/DataDriven/AITDDCurrentCase.Codeunit.al
thloke: Addressed. AIT logging and suite integration now read dataset/case/run identity from the shared `Test Data Source Context`; `AIT Test Run Iteration` remains only the thin suite-line adapter needed for Eval Suite dataset overrides.

src/Tools/AI Test Toolkit/src/DataDriven/AITTestDataSource.Codeunit.al
thloke: The generic parts were moved to Test Runner in 140d57524: `Test Input Data Source Mgt.` owns group resolution/enumeration, and `Test Data Source Context` owns generic case/run identity. The small `AIT Test Data Source` adapter must remain in AI Test Toolkit because it materializes `AIT Test Case Context` and applies the AIT Eval Suite line's dataset override; moving that adapter wholesale would make Test Runner depend on AI Test Toolkit.

src/Tools/AI Test Toolkit/src/TestSuite/AITALTestSuiteMgt.Codeunit.al
thloke: Addressed in Test Runner. When the platform reports `Method[caseName]`, `Test Runner - Mgt` resolves the base method and materializes a per-case `Test Method Line`, then updates that row with the outcome. This works for the regular Test Tool without AIT log entries. The AIT subscribers enrich the same runtime case with AIT method-line/log data, so individual cases are visible in both UIs without eager YAML expansion.

src/Tools/AI Test Toolkit/src/TestSuite/AITTestSuiteMgt.Codeunit.al
thloke: Added the requested TODO beside the token-delta calculation: agent tests will require operation-scoped accounting rather than the current session-wide delta.

alexei-dobriansky:

<details>

This adds language-first data-driven AI eval support through a shared data source, injected case context, platform test handler, and per-case result materialization. The direction matches the requested migration path, but the current implementation still has core correctness gaps in dataset selection, generated test lines, and standalone log metrics. No BaseApp event publisher dependency is involved in this diff.
</details>

<details>

**Fit:** Partial

The change targets the right feature area and adds most of the needed infrastructure. It does not fully satisfy the requested behavior yet because common standalone and multi-case runs can select wrong data, fail after the first generated case, or store incomplete eval results.
</details>

<details>

**S1** (🔴 High): *Suite state can choose stale datasets*
This branch uses the last suite line whenever it has a test suite code. After a suite run, that line remains in the single-instance runner while only the active-suite marker is cleared. A later standalone `[TestDataSource]` run can use the old suite dataset instead of the attribute dataset.

**S2** (🔴 High): *Standalone logs miss AI metrics*
The standalone log path records only the AOAI server-session delta and only writes accuracy when `SetAccuracy` was called. It ignores `SetTokenConsumption`, turn counts, and the fallback accuracy that the suite path records. This can store wrong tokens and 0 accuracy for successful cases.

**S3** (🔴 High): *Generated case lines reuse same line number*
`BaseTestMethodLine` still carries the base `Function` filter when `GetNextMethodNumber` is called. That helper does not clear `Function`, so every `Method[case]` uses the same next line number and the second case can hit a duplicate `Test Method Line` key. Clear the filters or compute the next line from an unfiltered codeunit range.

**S4** (🟠 Moderate): *Context interface is missing classic APIs*
The new interface does not include `GetTurnSetup(var TurnSetup)` or the suite setup helpers that existing eval code uses. Those tests cannot move to the injected context with only signature changes. Add the missing non-obsolete members or narrow the parity claim.
</details>

<details>

**Risk:** This touches shared Test Runner and AI Test Toolkit infrastructure. If wrong, data-driven evals can run against stale input rows, fail when more than one case is reported, or write misleading token and accuracy results.

**Necessity:** The feature is needed for language-first data-driven evals and the scope is reasonable for a Slice. The blockers are in the core fan-out and logging paths, so they should be fixed before merge.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=9460 round=1 by=alexei-dobriansky at=2026-09-04T08:38:25Z lastSha=74cd4ff39680a86224292db76b6a3c2d8022f504 reviewKey=54f4a30875bdc5c03a9613a2dde2fc3ea16f35d5451b98927a00d546a3aa1c12 suggestions=S1@95e69e31,S2@64ecb527,S3@b4cee31e,S4@f5f6ee36

PR #9461, opened 2026-07-15
[Extensibility Request] issue 30335: add OnBeforeRecreateSalesLine IsHandled event

AleksandricMarko:
## Summary

The issue author needs to conditionally replace the standard sales line recreation logic inside `RecreateSalesLinesHandleSupplementTypes` in table 36 "Sales Header". Under specific business conditions based on extension fields, their extension must fully take over sales line creation, which requires bypassing the standard `CreateSalesLine` call at the caller level. No existing extensibility point sits immediately before that call, so this PR adds a new `IsHandled` integration event that lets subscribers skip the standard creation and run an equivalent custom implementation.

Source issue repository: microsoft/AlAppExtensions; issue number: 30335

## Changes Made
- `OnBeforeRecreateSalesLine` - new `IsHandled` integration event published in `RecreateSalesLinesHandleSupplementTypes` immediately before `CreateSalesLine`, allowing subscribers to skip the standard sales line recreation; the same change is propagated to all country layers that have their own copy of the table.

Fixes [AB#641988](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/641988)

### DIFF
--- a/src/Layers/W1/BaseApp/Sales/Document/SalesHeader.Table.al
+++ b/src/Layers/W1/BaseApp/Sales/Document/SalesHeader.Table.al
@@ -4602,7 +4602,10 @@ table 36 "Sales Header"
         ShouldCreateSalsesLine := not TempSalesLine.IsExtendedText();
         OnRecreateSalesLinesHandleSupplementTypesOnAfterCalcShouldCreateSalsesLine(TempSalesLine, ShouldCreateSalsesLine, SalesLine);
         if ShouldCreateSalsesLine then begin
-            CreateSalesLine(TempSalesLine);
+            IsHandled := false;
+            OnBeforeRecreateSalesLine(IsHandled, SalesLine, TempSalesLine, Rec);
+            if not IsHandled then
+                CreateSalesLine(TempSalesLine);
             ExtendedTextAdded := false;
             OnAfterRecreateSalesLine(SalesLine, TempSalesLine, Rec);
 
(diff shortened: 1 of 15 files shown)


PredragMaricic (approved):

src/Layers/W1/BaseApp/Sales/Document/SalesHeader.Table.al
```diff
@@ -4602,7 +4602,10 @@ table 36 "Sales Header"
         ShouldCreateSalsesLine := not TempSalesLine.IsExtendedText();
         OnRecreateSalesLinesHandleSupplementTypesOnAfterCalcShouldCreateSalsesLine(TempSalesLine, ShouldCreateSalsesLine, SalesLine);
         if ShouldCreateSalsesLine then begin
-            CreateSalesLine(TempSalesLine);
+            IsHandled := false;
+            OnBeforeRecreateSalesLine(IsHandled, SalesLine, TempSalesLine, Rec);
+            if not IsHandled then
```
qasimikram, line 4607: S1: Please add regression coverage proving that the unhandled path remains unchanged and that a subscriber can create exactly one replacement line without duplicate creation while the existing post-creation processing still runs correctly.

qasimikram (approved):
The event is narrowly placed around the requested `CreateSalesLine` call and is implemented consistently across the country-layer copies.

S1: Please add regression coverage for this new handled path. Verify that the unhandled path remains unchanged and that a subscriber can create exactly one replacement line without duplicate creation while the existing post-creation processing continues to run correctly.

PR #9463, opened 2026-07-15
[Extensibility Request] issue 30336: add OnValidateNoOnBeforeModifyCapNeedEntries event in Prod. Order Routing Line

AleksandricMarko:
## Summary
The issue author is building a customization on table 5409 "Prod. Order Routing Line" that must run after the standard Work Center / Machine Center transfer logic populates the routing line, but before capacity need entries are updated. They need access to both `Rec` and `xRec` to compare the previous and current routing line state at that exact point, which no existing event provides. This PR adds an integration event at that precise location so extensions can subscribe without duplicating validation logic.

Source issue repository: microsoft/AlAppExtensions; issue number: 30336

## Changes Made
- `Prod. Order Routing Line` (table 5409) - Added integration event `OnValidateNoOnBeforeModifyCapNeedEntries(Rec, xRec)` in the `OnValidate` trigger of field `No.`, raised after `WorkCenterTransferFields()`/`MachineCtrTransferFields()` and immediately before `ModifyCapNeedEntries()`; new publisher declared alongside the other integration events. The same change is mirrored in the IT layer counterpart.

Fixes [AB#641735](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/641735)

### DIFF
--- a/src/Layers/IT/BaseApp/Manufacturing/Document/ProdOrderRoutingLine.Table.al
+++ b/src/Layers/IT/BaseApp/Manufacturing/Document/ProdOrderRoutingLine.Table.al
@@ -2176,7 +2176,7 @@ table 5409 "Prod. Order Routing Line"
     end;
 
     [IntegrationEvent(false, false)]
-    local procedure OnValidateNoOnBeforeModifyCapNeedEntries(var ProdOrderRoutingLine: Record "Prod. Order Routing Line"; var xProdOrderRoutingLine: Record "Prod. Order Routing Line")
+    local procedure OnValidateNoOnBeforeModifyCapNeedEntries(var ProdOrderRoutingLine: Record "Prod. Order Routing Line"; xProdOrderRoutingLine: Record "Prod. Order Routing Line")
     begin
     end;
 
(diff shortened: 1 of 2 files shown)


6CRIPT (dismissed):
Should we keep xRex without "var" instead? The OP suggested so in https://github.com/microsoft/ALAppExtensions/issues/30336

PredragMaricic (dismissed):

qasimikram (changes requested):
S1: Remove `var` from `xProdOrderRoutingLine` in both publisher declarations. The linked request needs the previous value for comparison, not mutation. With the current signature, a subscriber can change `xRec` before the subsequent `(xRec."No." <> '')` condition and unintentionally alter whether `CalcStartingEndingDates` runs.

AleksandricMarko:
@copilot please remove var for xProdOrderRoutingLine parameter in both files

qasimikram (approved):
The new publisher is correctly placed before capacity-need updates in both layers. Its name, parameters, and IntegrationEvent(false, false) settings match the requested additive scenario without bypassing standard behavior.

tomasevicst (approved):

PR #9464, opened 2026-07-15
[Extensibility Request] issue 30349: add OnBeforeCheckTrackingIfRequired event to Item Journal Line

AleksandricMarko:
## Summary
The reporter needs clearer, user-facing error messages when item tracking (Serial/Lot No.) is missing during document posting. The standard `CheckTrackingIfRequired` validation reports errors against the temporary "Item Journal Line" record, so users cannot tell which document line or item is affected. This PR adds an integration event at the start of `CheckTrackingIfRequired` so extensions can run their own validation and raise a more meaningful error before the standard checks.

Source issue repository: microsoft/AlAppExtensions; issue number: 30349

## Changes Made
- `CheckTrackingIfRequired` (table 83 "Item Journal Line") - raise a new `OnBeforeCheckTrackingIfRequired` integration event at the start of the procedure so subscribers can perform custom item-tracking validation before the standard `TestField` checks; the same change is propagated to the country layer counterparts (APAC, CH, ES, FR, GB, IT, RU).

Fixes [AB#642241](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/642241)

### DIFF
--- a/src/Layers/APAC/BaseApp/Inventory/Journal/ItemJournalLine.Table.al
+++ b/src/Layers/APAC/BaseApp/Inventory/Journal/ItemJournalLine.Table.al
@@ -3665,6 +3665,7 @@ table 83 "Item Journal Line"
     /// <param name="ItemTrackingSetup">Item tracking setup to use.</param>
     procedure CheckTrackingIfRequired(ItemTrackingSetup: Record "Item Tracking Setup")
     begin
+        OnBeforeCheckTrackingIfRequired(Rec, ItemTrackingSetup);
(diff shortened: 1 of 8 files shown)


PredragMaricic (approved):

qasimikram (approved):
The new publisher is raised before both standard tracking validations and exposes the requested item journal line and tracking setup context. The implementation is consistent across all localized table copies and does not alter default behavior.

PR #9465, opened 2026-07-15
[Extensibility Request] issue 30350: add OnBeforeTestFirstApplyItemLedgerEntryTracking event

AleksandricMarko:
## Summary
The reporter needs to raise clearer, document-oriented errors before the standard item tracking validation runs, because the built-in `TestField` errors reference the temporary `Item Journal Line` record and do not indicate which document line or item is missing Serial/Lot tracking. They also need to conditionally skip lot checking in a specific scenario. This PR adds an integration event at the very start of `TestFirstApplyItemLedgerEntryTracking` so an extension can validate up front and adjust the `Item Tracking Code` (passed by `var`) before the standard checks.

Source issue repository: microsoft/AlAppExtensions; issue number: 30350

## Changes Made
- `OnBeforeTestFirstApplyItemLedgerEntryTracking` - new integration event published at the start of the `TestFirstApplyItemLedgerEntryTracking` procedure in codeunit 22 "Item Jnl.-Post Line"; the `Item Tracking Code` parameter is `var` so subscribers can amend it (e.g. skip lot checking) before the standard tracking validation. Added in the W1 base layer and mirrored into the CH, ES, IT, RU and APAC layer counterparts.

Fixes [AB#642242](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/642242)


src/Layers/APAC/BaseApp/Inventory/Posting/ItemJnlPostLine.Codeunit.al
```diff
@@ -7449,6 +7451,11 @@ codeunit 22 "Item Jnl.-Post Line"
     begin
     end;
 
+    [IntegrationEvent(false, false)]
+    local procedure OnBeforeTestFirstApplyItemLedgerEntryTracking(ItemLedgEntry: Record "Item Ledger Entry"; OldItemLedgEntry: Record "Item Ledger Entry"; var ItemTrackingCode: Record "Item Tracking Code")
```
AleksandricMarko: @copilot in  local procedure OnBeforeTestFirstApplyItemLedgerEntryTracking all parameters with Ledg in name should be with Ledger, so not ItemLedgEntry but ItemLedgerEntry
Make changes in all files

PredragMaricic (approved):

qasimikram (approved):
The event is raised before the standard item-tracking validation and provides the requested ledger-entry context. Passing the local item-tracking code by reference allows subscribers to refine the checks while leaving default behavior unchanged.

qasimikram (approved):
The publisher placement and signature match the requested scenario. The scoped var Item Tracking Code permits customized serial/lot validation while leaving default behavior unchanged without subscribers. Layer copies are consistent. No actionable findings.

JakovljevicDusan (approved):

alexei-dobriansky:

<details>

This PR adds a new integration event `OnBeforeTestFirstApplyItemLedgerEntryTracking`, raised at the very start of `TestFirstApplyItemLedgerEntryTracking` in codeunit 22 "Item Jnl.-Post Line", with the "Item Tracking Code" parameter passed by `var`. It lets an extension raise a clearer, document-oriented error before the standard `TestField` checks run, and adjust the tracking code to skip a check. It answers AB#642242 (ALAppExtensions #30350).

The core design holds up. `TestFirstApplyItemLedgerEntryTracking` receives `ItemTrackingCode` **by value** from its only caller (line 1784, which passes `GlobalItemTrackingCode`), so a `var` change made by a subscriber affects only this procedure's local copy and cannot leak back into the global tracking code used elsewhere in posting. That is the right blast radius for what the request needs. The country-layer work is also complete: `ItemJnlPostLine.Codeunit.al` exists in exactly six layers (W1, APAC, CH, ES, IT, RU) and all six get the identical call and publisher. Two things are worth a second look. The `"Drop Shipment"` serial and lot checks on the lines below are *not* gated by `ItemTrackingCode`, so changing that record cannot skip them - if the reporter's scenario involves drop shipment, this event will not actually solve their problem. And the new publisher uses different parameter names from its sibling `OnAfterTestFirstApplyItemLedgerEntryTracking` five lines below.
</details>

<details>

**S1** - *Parameter names differ from the sibling event*
The new publisher uses `ItemLedgerEntry` and `OldItemLedgerEntry`, while `OnAfterTestFirstApplyItemLedgerEntryTracking` just below and the procedure itself use `ItemLedgEntry` and `OldItemLedgEntry`. Publisher parameter names appear in generated subscriber stubs and cannot be changed later without churn for partners. Please rename them to match.

**S2** - *Changing the tracking code cannot skip the drop shipment checks*
The two `"Drop Shipment"` checks do not read `ItemTrackingCode`, so a subscriber cannot skip them by changing that record. The linked request mentions conditionally skipping lot checking; please confirm the reporter's scenario is covered, and add `var IsHandled: Boolean` if they also need to bypass the drop shipment checks.

**S3** - *Consider an IsHandled flag for consistency*
Most `OnBefore` events in this codeunit carry `var IsHandled: Boolean` so a subscriber can replace the standard validation. Here the only way to stop the standard checks is to raise an error or to blank the tracking code, which is indirect. Please consider adding the flag while the signature is still open.

**S4** - *Unrelated removal of the trailing newline*
All six files lose the blank line at the end of the file. This is unrelated to the change and adds noise to the diff. Please restore it.

**S5** - *Confirm the failing unit test jobs are unrelated*
Four unit test jobs are failing: AT IntegrationTests, CH LegacyTestsBucket2, IN UncategorizedTests and NL UncategorizedTests. An empty publisher should not change behavior, and the scatter across unrelated buckets looks like flakiness, but please confirm rather than assume.
</details>

<details>

**Risk:** The diff adds no logic, so with no subscriber attached nothing changes, and all `Build Apps` jobs are green in every country. The forward-looking risk is real but contained. `TestFirstApplyItemLedgerEntryTracking` enforces serial and lot number integrity when applying item ledger entries, and passing "Item Tracking Code" by `var` deliberately lets a subscriber turn `"SN Specific Tracking"` or `"Lot Specific Tracking"` off before those checks run - that is item tracking integrity during posting, so a careless subscriber can let mismatched serial or lot numbers through and corrupt traceability. Two things keep this acceptable: the parameter is a by-value copy in the caller, so the change cannot escape this procedure, and the drop shipment checks stay unconditional. Country-layer coverage is complete, so no layer diverges. There is no table, API or upgrade impact. A test is not warranted for an empty publisher and its absence should not block; the four red test jobs in S5 are the only open question.

**Necessity:** The scenario is concrete and the justification is good. The built-in `TestField` errors here fire against the temporary "Item Journal Line" record and tell the user nothing about which document line or item is missing tracking, which is a genuine, recurring usability problem in posting, and an extension cannot improve it without a hook before the checks. The change is proportionate - one event call plus one publisher, mirrored across the layers that need it - and it is the right channel, with an approved work item behind it. My suggestions are about the shape of the contract rather than whether to add it; S1 and S2 are worth closing before merge because the signature is frozen once released.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9465 round=1 by=alexei-dobriansky at=2026-08-03T09:55:00Z lastSha=b9e342966d2537ea1d9312998593129e88e54541 suggestions=S1,S2,S3,S4,S5

AleksandricMarko:
@copilot pull origin main into this branch and resolve conflicts if exist

PR #9466, opened 2026-07-15
[Extensibility Request] issue 30351: extend OnCheckExpirationDateOnBeforeAssignExpirationDate event

AleksandricMarko:
## Summary
The issue author extends the standard item journal posting flow and needs to replicate the expiration-date assignment logic in a modified form. To do that reliably from a subscriber, they need access to the item tracking code, the item journal line, and the sign factor at the point where the expiration date is assigned. This PR appends those three values to the existing `OnCheckExpirationDateOnBeforeAssignExpirationDate` integration event so subscribers can drive the same logic through the existing `IsHandled` pattern.

Source issue repository: microsoft/AlAppExtensions; issue number: 30351

## Changes Made
- `OnCheckExpirationDateOnBeforeAssignExpirationDate` - appended `GlobalItemTrackingCode`, `ItemJnlLine2` and `SignFactor` parameters (after the existing `var IsHandled` parameter) to the integration event and passed them from `CheckExpirationDate` in codeunit 22 "Item Jnl.-Post Line"; the same change was propagated to the same-named files in the country layers that carry their own copy.

Fixes [AB#642240](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/642240)

### DIFF
--- a/src/Layers/APAC/BaseApp/Inventory/Posting/ItemJnlPostLine.Codeunit.al
+++ b/src/Layers/APAC/BaseApp/Inventory/Posting/ItemJnlPostLine.Codeunit.al
@@ -4245,7 +4245,7 @@ codeunit 22 "Item Jnl.-Post Line"
             ExpirationDateChecked := true;
         end;
         IsHandled := false;
-        OnCheckExpirationDateOnBeforeAssignExpirationDate(TempTrackingSpecification, ExistingExpirationDate, IsHandled);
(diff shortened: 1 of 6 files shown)


6CRIPT (approved):

PredragMaricic (approved):

qasimikram (approved):
The additional event context matches the linked request and is appended after the existing handled parameter, preserving compatibility for current subscribers. The change is propagated consistently across all localized copies and leaves default behavior unchanged.

JakovljevicDusan (approved):

PR #9467, opened 2026-07-15
[Extensibility Request] issue 30346: make InsertPostValueEntryToGL public in Item Jnl.-Post Line

AleksandricMarko:
## Summary
An extension author wants to post additional variances and reuse the standard General Ledger posting logic that already lives in codeunit 22 `Item Jnl.-Post Line`. They already call the public `PostInventoryToGL` procedure, but `InsertPostValueEntryToGL` was declared `local` and therefore unreachable from extensions. This PR exposes that procedure so extensions can invoke the standard posting flow.

Source issue repository: microsoft/AlAppExtensions; issue number: 30346

## Changes Made
- `InsertPostValueEntryToGL` (codeunit 22 `Item Jnl.-Post Line`) - changed from a `local procedure` to a public `procedure` so extensions can call the standard value-entry-to-G/L posting logic. Applied in the W1 base layer and mirrored to the same-named counterparts in the APAC, CH, ES, IT, and RU layers.

Fixes [AB#642120](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/642120)

### DIFF
--- a/src/Layers/APAC/BaseApp/Inventory/Posting/ItemJnlPostLine.Codeunit.al
+++ b/src/Layers/APAC/BaseApp/Inventory/Posting/ItemJnlPostLine.Codeunit.al
@@ -5312,7 +5312,7 @@ codeunit 22 "Item Jnl.-Post Line"
         exit((ItemJnlLine."Value Entry Type" = ItemJnlLine."Value Entry Type"::Revaluation) and (ItemJnlLine.Quantity <> 0));
     end;
 
-    local procedure InsertPostValueEntryToGL(ValueEntry: Record "Value Entry")
+    procedure InsertPostValueEntryToGL(ValueEntry: Record "Value Entry")
     var
         PostValueEntryToGL: Record "Post Value Entry to G/L";
     begin
--- a/src/Layers/CH/BaseApp/Inventory/Posting/ItemJnlPostLine.Codeunit.al
+++ b/src/Layers/CH/BaseApp/Inventory/Posting/ItemJnlPostLine.Codeunit.al
@@ -5284,7 +5284,7 @@ codeunit 22 "Item Jnl.-Post Line"
         exit((ItemJnlLine."Value Entry Type" = ItemJnlLine."Value Entry Type"::Revaluation) and (ItemJnlLine.Quantity <> 0));
     end;
 
-    local procedure InsertPostValueEntryToGL(ValueEntry: Record "Value Entry")
+    procedure InsertPostValueEntryToGL(ValueEntry: Record "Value Entry")
     var
         PostValueEntryToGL: Record "Post Value Entry to G/L";
     begin
--- a/src/Layers/ES/BaseApp/Inventory/Posting/ItemJnlPostLine.Codeunit.al
+++ b/src/Layers/ES/BaseApp/Inventory/Posting/ItemJnlPostLine.Codeunit.al
@@ -5293,7 +5293,7 @@ codeunit 22 "Item Jnl.-Post Line"
         exit((ItemJnlLine."Value Entry Type" = ItemJnlLine."Value Entry Type"::Revaluation) and (ItemJnlLine.Quantity <> 0));
     end;
 
-    local procedure InsertPostValueEntryToGL(ValueEntry: Record "Value Entry")
+    procedure InsertPostValueEntryToGL(ValueEntry: Record "Value Entry")
     var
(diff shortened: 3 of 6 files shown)


PredragMaricic (approved):
This precisely exposes the requested procedure across all affected localized implementations while leaving its body and runtime behavior unchanged. The public API commitment is intentional and consistent. No blocking findings.

JakovljevicDusan (approved):

alexei-dobriansky:

<details>

This PR changes `InsertPostValueEntryToGL` in codeunit 22 "Item Jnl.-Post Line" from `local procedure` to a public `procedure`, so an extension that posts its own variances can reuse the standard logic that queues a value entry for G/L posting. It answers AB#642120 (ALAppExtensions #30346).

The change is minimal and correct for what it claims. It is a visibility modifier only - no logic, no signature change, no new branch - so no existing code path behaves differently, and every `Build Apps` job is green. The country-layer handling is right and complete: `ItemJnlPostLine.Codeunit.al` exists in exactly six layers (W1, APAC, CH, ES, IT, RU) and all six are changed identically, so no layer is left behind. The one thing worth thinking about before merge is that the procedure was written under assumptions that only held while it was local. It is called in exactly three places, always immediately after the value entry has been inserted, so `ValueEntry."Entry No."` is always a real key and `PostValueEntryToGL.Insert()` can never hit a duplicate. Neither assumption is checked in code, and neither survives being callable from outside.
</details>

<details>

**S1** - *Insert can fail with a duplicate key when called twice*
`PostValueEntryToGL.Insert()` is called without a conflict check. Inside this codeunit that is safe, because each value entry is queued once. An extension can now call the procedure for a value entry the standard flow already queued, which gives a runtime error. Please guard with `if not PostValueEntryToGL.Get(ValueEntry."Entry No.") then` or use the return value of `Insert()`.

**S2** - *No check that the value entry is a saved record*
The procedure copies `ValueEntry."Entry No."` straight into the key of "Post Value Entry to G/L". If an extension passes an unsaved record, the entry number is 0 and a bad row is queued for the "Post Inventory Cost to G/L" run. Please add `ValueEntry.TestField("Entry No.")`, or make the precondition explicit in the code.

**S3** - *Document the precondition for external callers*
The three internal call sites all run right after the value entry is inserted and inside the posting transaction. That context is now invisible to an extension author. Please add a short comment stating that the value entry must already be inserted and that the call belongs inside the posting transaction.

**S4** - *Making a procedure public is a one-way door*
Once released as public, this procedure cannot be made local again without a breaking change, and its body then becomes part of the contract. Please confirm the requester genuinely needs to *invoke* the logic, since the existing `OnInsertPostValueEntryToGLOnAfterTransferFields` and `OnAfterInsertPostValueEntryToGL` events already cover the observe-and-adjust case.
</details>

<details>

**Risk:** Regression risk for existing behavior is very low and easy to argue: the diff touches only the `local` keyword on one procedure in six files, adds no code, and removes none, so every current caller executes exactly the same instructions. All country `Build Apps` jobs pass and no unit test fails. The real risk is the widened API surface in a financially sensitive area. "Post Value Entry to G/L" drives the "Post Inventory Cost to G/L" batch, so a badly formed row queued by an extension shows up as an inventory-to-G/L reconciliation difference rather than as an immediate error - exactly the kind of problem that is expensive to trace back. S1 and S2 are what keep that risk bounded. There is no upgrade or data impact, and no event or table signature changes. A test is not warranted here: there is no behavior to assert, and BCApps does not test visibility modifiers.

**Necessity:** The request is legitimate and well targeted. The reporter already uses the public `PostInventoryToGL` and needs the matching queueing step to post additional variances through the standard flow; reimplementing the `IsPostToGL` rules in an extension would duplicate cost-posting logic that must stay in sync with the platform, which is worse. An event would not solve it, because the requester needs to *call* the logic rather than observe it, so exposing the procedure is the right mechanism. The scope is also right for a bug-channel change: six one-word edits, no behavior change, and a linked, tagged work item. I would merge this once S1 and S2 are addressed, since they are small and they are what make the procedure safe to call from outside.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9467 round=1 by=alexei-dobriansky at=2026-08-03T09:55:00Z lastSha=14080f4ac00ada84bff060a7221c6cbf34d72023 suggestions=S1,S2,S3,S4

alexei-dobriansky (approved):

PR #9538, opened 2026-07-16
[main]- Requests to Approve: Open Record shows wrong Requisition Worksheet batch after viewing a different batch

v-ankitgoyal:
[AB#642418](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/642418)

Issue: - Requests to Approve: Open Record shows wrong Requisition Worksheet batch after viewing a different batch

Root cause: -
When the worksheet is opened positioned on a specific batch record (for example, via Page Management from an approval entry's Open Record), the Requisition Line carries both Journal Batch Name and Worksheet Template Name. Page 291's OnOpenPage only short-circuited when Worksheet Template Name = '' (OpenedFromBatch), so with the template populated it fell through to the template-selection path. Because the page uses SaveValues = true, CurrentJnlBatchName was restored from the previously viewed batch, and OpenJnl filtered on that stale batch instead of the record's actual batch.

Solutions: -
Added a dedicated GetCurrentJnlBatchName() helper, called from OnOpenPage, that detects when the page is opened on a specific batch record and uses the batch from the record (overriding the SaveValues-restored value), scoping the page to the record's template:

// OnOpenPage
if OpenedFromBatch then begin
    CurrentJnlBatchName := Rec."Journal Batch Name";
    ReqJnlManagement.OpenJnl(CurrentJnlBatchName, Rec);
    SetControlAppearanceFromWkshBatch();
    exit;
end;

if OpenedFromWkshBatchRecord() then
    exit;

OnBeforeTemplateSelection(Rec, CurrentJnlBatchName);

// helper
local procedure GetCurrentJnlBatchName(): Boolean
begin
    if (Rec."Journal Batch Name" = '') or (Rec."Worksheet Template Name" = '') then
        exit(false);

    CurrentJnlBatchName := Rec."Journal Batch Name";
    Rec.FilterGroup := 2;
    Rec.SetRange("Worksheet Template Name", Rec."Worksheet Template Name");
    Rec.FilterGroup := 0;
    ReqJnlManagement.OpenJnl(CurrentJnlBatchName, Rec);
    SetControlAppearanceFromWkshBatch();
    exit(true);
end;


v-ankitgoyal:

PredragMaricic (approved):

v-praghav (approved):

PR #9539, opened 2026-07-16
Bugs/bug 642418 master requests to approve open record 2

v-ankitgoyal:
Fixes : [AB#642418](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/642418)

### DIFF
--- a/src/Layers/NA/BaseApp/Inventory/Requisition/ReqWorksheet.Page.al
+++ b/src/Layers/NA/BaseApp/Inventory/Requisition/ReqWorksheet.Page.al
@@ -1172,6 +1172,18 @@ page 291 "Req. Worksheet"
             SetControlAppearanceFromWkshBatch();
             exit;
         end;
+        // When the page is opened positioned on a specific worksheet batch (for example, from the
+        // Open Record action on an approval entry), use the batch from the record instead of the
+        // previously saved batch name that SaveValues would otherwise restore.
+        if (Rec."Journal Batch Name" <> '') and (Rec."Worksheet Template Name" <> '') then begin
+            CurrentJnlBatchName := Rec."Journal Batch Name";
+            Rec.FilterGroup := 2;
+            Rec.SetRange("Worksheet Template Name", Rec."Worksheet Template Name");
+            Rec.FilterGroup := 0;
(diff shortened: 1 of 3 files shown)


alexei-dobriansky:

<details>

This PR changes the Req. Worksheet page in W1 and NA so an Open Record action that opens a specific requisition worksheet batch uses the batch from `Rec` instead of the saved batch name restored by `SaveValues`.

The new condition matches the failing launch shape: both `"Worksheet Template Name"` and `"Journal Batch Name"` are already present. It then applies the template filter and calls `ReqJnlManagement.OpenJnl`, which validates and filters the batch. The W1 test covers the main regression by opening two different batches in sequence and checking that each page shows its own batch.
</details>

<details>

**S1** - *Keep the open worksheet event*  
The new record-open path skips `OnBeforeOpenReqWorksheet`, which was raised before `OpenJnl` in this flow before the change. Call the event before `ReqJnlManagement.OpenJnl` in both W1 and NA so extensions still get the same hook.

**S2** - *Use the linked bug number in the test*  
The new test comment uses `[SCENARIO 639932]`, but this PR links AB#642418. Use the same scenario number as the bug so the test is easy to trace later.
</details>

<details>

**Risk:** The change is in page-open control flow for Req. Worksheet, not in posting or amount calculation. The main regression surface is partner extensions that subscribe to the page open events, plus W1/NA behavior parity. The PR has test coverage for W1, but the current PR build result is not present because the old build check was deleted as stale.

**Necessity:** The change is needed because Requests to Approve must open the batch from the selected approval entry, even after the user previously viewed another batch. The scope is narrow and matches the bug scenario.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=9539 round=1 by=alexei-dobriansky at=2026-08-23T00:07:55.713Z lastSha=0b54287cb8c44798bce22d56b8ac2051ab6a73cc reviewKey=ee2f4593fd593c2f2581e6715e5bc2fbcfdba8c07bcf58c84dfa9ad6260602ab suggestions=S1@037f84a0,S2@171db326

PR #9571, opened 2026-07-17

neeleshsinghal:
[AB#641061](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/641061)

### DIFF
--- a/src/Layers/W1/BaseApp/Pricing/PriceList/PriceListManagement.Codeunit.al
+++ b/src/Layers/W1/BaseApp/Pricing/PriceList/PriceListManagement.Codeunit.al
@@ -789,10 +789,17 @@ codeunit 7017 "Price List Management"
                 end;
             until not PriceSourceList.Next(PriceSource);
 
-        if NonEmptyCount = 1 then
-            PriceListLine.SetView(FirstMatchView)
-        else
-            ClearSourceFilters(PriceListLine);
+        case NonEmptyCount of
+            0:
+                begin
+                    MarkingIsUsed := true;
+                    ClearSourceFilters(PriceListLine);
+                end;
+            1:
+                PriceListLine.SetView(FirstMatchView);
+            else
+                ClearSourceFilters(PriceListLine);
+        end;
     end;
 
     local procedure MarkPriceListLines(var PriceListLine: Record "Price List Line")
--- a/src/Layers/W1/BaseApp/Pricing/PriceList/PriceListManagement.Codeunit.al
+++ b/src/Layers/W1/BaseApp/Pricing/PriceList/PriceListManagement.Codeunit.al
@@ -631,6 +631,8 @@ codeunit 7017 "Price List Management"
     end;
 
     procedure SetPriceListLineFilters(var PriceListLine: Record "Price List Line"; PriceSourceList: Codeunit "Price Source List"; AmountType: Enum "Price Amount Type")
+    var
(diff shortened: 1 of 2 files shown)


qasimikram (changes requested):
S1 (blocking): Please preserve an empty result when none of the price sources match. In `BuildSourceFilters`, `NonEmptyCount = 0` currently leaves `MarkingIsUsed` false and then calls `ClearSourceFilters`. The caller consequently skips `MarkedOnly(true)`, leaving only the status, price type, and amount type filters and potentially showing unrelated price lines. Please handle the zero-match case explicitly and add a regression test where every source is empty.

Question on performance coverage: The new fast path avoids marking only when exactly one source has matching rows. When both `All Customers` and a customer-specific or group source contain rows, `NonEmptyCount` reaches 2 and the implementation returns to marking every matching row, including a potentially very large `All Customers` set. Could you confirm one of the following?

1. The production reproduction has only one non-empty source, with before/after timing showing that this change meets the <5-second target.
2. The multi-source scenario has also been tested at scale and remains acceptably fast when both `All Customers` and customer/group rows exist.

qasimikram (commented):
The zero-match regression is addressed: `NonEmptyCount = 0` now forces `MarkedOnly(true)`, and T222 covers the case where unrelated price lines exist but none match the current customer's sources.

One performance-coverage question remains. The fast path avoids marking only when exactly one source has matching rows. If both `All Customers` and a customer-specific or group source contain rows, `NonEmptyCount` reaches 2 and the implementation still marks every matching row, including a potentially large `All Customers` set. Could you confirm one of the following?

1. The production reproduction has only one non-empty source, with before/after timing showing that this change meets the <5-second target.
2. The multi-source scenario has also been tested at scale and remains acceptably fast when both `All Customers` and customer/group rows exist.

Non-blocking test suggestion: use `Validate` for the customer group fields and `Modify(true)` in T221/T222 so the fixtures exercise the same validation path as production setup.

src/Layers/W1/BaseApp/Pricing/PriceList/PriceListManagement.Codeunit.al
```diff
@@ -745,28 +748,66 @@ codeunit 7017 "Price List Management"
...
+                    NonEmptyCount += 1;
+                    case NonEmptyCount of
+                        1:
+                            FirstMatchView := PriceListLine.GetView(false);
+                        2:
+                            begin
+                                MarkingIsUsed := true;
+                                CurrentMatchView := PriceListLine.GetView(false);
+                                PriceListLine.SetView(FirstMatchView);
+                                MarkPriceListLines(PriceListLine);
```
qasimikram, line 782: Could you confirm the performance coverage for this fallback path? When both `All Customers` and a customer-specific or group source contain rows, this branch still marks every matching row, including a potentially large `All Customers` set.

Please confirm either that the production reproduction has only one non-empty source with before/after timing meeting the <5-second target, or that this multi-source path was tested at scale and remains acceptably fast.

qasimikram (approved):
The blocking zero-match regression is fixed and covered by T222. I found no remaining correctness issue in the updated diff. Approving with the open performance-coverage thread left for confirmation before merge.

v-ankitgoyal (approved):

PR #9585, opened 2026-07-17
[master] Description Field value gets automatically reverted the original value after changing it.

neeleshsinghal:
**Issue:** After validating Quantity on a sales line, editing the Description reverts it to the original item value when "Calc. Inv. Discount" is enabled (Bug 638838).

**Cause:** In CalculateSalesSubPageTotals, NeedRefreshSalesLine := false; was reset after the if TotalsUpToDate then exit; early-exit, so the flag set true by the Quantity validation stayed stuck, and RefreshSalesLine's Rec.Find() reloaded the line from the DB, discarding the uncommitted Description edit.

**Solution:** Moved NeedRefreshSalesLine := false; before the early-exit, so when totals are already up-to-date (no recalculation), the flag is cleared and RefreshSalesLine no longer overwrites the user's edit.

Workitem: [Bug 638838](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/638838): [master] [all-e] RFH: [SaaS] Description Field value gets automatically reverted the original value after changing it.

Fixes [AB#638838](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/638838)

### DIFF
--- a/src/Layers/BE/BaseApp/Utilities/DocumentTotals.Codeunit.al
+++ b/src/Layers/BE/BaseApp/Utilities/DocumentTotals.Codeunit.al
@@ -80,10 +80,11 @@ codeunit 57 "Document Totals"
         if IsHandled then
             exit;
 
+        NeedRefreshSalesLine := false;
+
         if TotalsUpToDate then
             exit;
         TotalsUpToDate := true;
-        NeedRefreshSalesLine := false;
 
         SalesSetup.GetRecordOnce();
         TotalSalesLine2.Copy(TotalSalesLine);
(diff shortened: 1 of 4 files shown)


v-ankitgoyal (approved):

qasimikram (approved):
The stale `NeedRefreshSalesLine` state is cleared before the `TotalsUpToDate` early exit across W1, NA, and BE, which prevents `RefreshSalesLine` from reloading the database record and discarding the uncommitted Description edit. The added UI regression test covers the Quantity validation followed by Description editing with automatic invoice discount calculation enabled.

PR #9612, opened 2026-07-20
Fixing correct value for Legacy Subcontracting flag in Italy during u…

PredragMaricic:
## What & why

After upgrading from 27.x to 28.3, customers using Italian legacy subcontracting lose access to the legacy Subcontracting pages because the Legacy Subcontracting flag in Manufacturing Setup is incorrectly initialized to False.

## Linked work

Fixes [AB#642627](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/642627)


AleksandricMarko (dismissed):

JakovljevicDusan (dismissed):

AndreiPanko (approved):

AleksandricMarko (approved):

src/Layers/IT/BaseApp/Local/Manufacturing/Document/LegacySubcUpgrade.Codeunit.al
```diff
@@ -33,25 +34,31 @@ codeunit 99008502 "Legacy Subc. Upgrade"
...
-        if ManufacturingSetup.Get() then begin
-            HasLegacySubcontractingData := LegacySubcFeatureHandler.DatabaseHasLegacySubcontractingData();
-            if ManufacturingSetup."Legacy Subcontracting" <> HasLegacySubcontractingData then begin
-                ManufacturingSetup."Legacy Subcontracting" := HasLegacySubcontractingData;
-                ManufacturingSetup.Modify();
-            end;
-        end;
+        if ManufacturingSetup.Get() then
+            if not ManufacturingSetup."Legacy Subcontracting" then
+                if LegacySubcFeatureHandler.DatabaseHasLegacySubcontractingData() or SubcontractorWorkCentersExist() then begin
```
6CRIPT, line 44: The new Work Center detection looks correct, but this procedure exits when the existing legacy subcontracting upgrade tag is already registered.

Are the affected customers expected to have already executed the previous version of this upgrade? If so, they will already have this tag and this new detection will never run for them.

6CRIPT (changes requested):

src/Layers/IT/BaseApp/Local/Manufacturing/Document/LegacySubcUpgrade.Codeunit.al
PredragMaricic: It is a good point. Discussed already with Andrei, that is ok as we have a simple workaround provided if needed.

6CRIPT (approved):

PR #9630, opened 2026-07-21
[Master]-Act Consumption Qty field in Component of a Firm Planned Order has amounts that point to a Released Production Order that has the same Document No if the component is the same

neeleshsinghal:


neeleshsinghal:
Superseded by #9660 (recreated as a single clean commit with only my changes).

PR #9638, opened 2026-07-21
Fixed Filtering the Sales Order List using the 'Delayed' view the list shows sales orders that are out of scope

6CRIPT:
Fixes [AB#642628](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/642628)

Needs a backport to 28.x

### DIFF
--- a/src/Layers/W1/BaseApp/Sales/Document/SalesOrderList.Page.al
+++ b/src/Layers/W1/BaseApp/Sales/Document/SalesOrderList.Page.al
@@ -1163,7 +1163,7 @@ page 9305 "Sales Order List"
         view(Delayed)
         {
             Caption = 'Delayed';
-            Filters = where("Document Type" = const(Order), Status = const(Released), "Completely Shipped" = const(false), "Late Order Shipping" = const(true));
+            Filters = where("Document Type" = const(Order), Status = const(Released), "Completely Shipped" = const(false), "Date Filter" = filter(<= '%workdate'), "Late Order Shipping" = const(true));
         }
         view(Released)
         {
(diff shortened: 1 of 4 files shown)


PredragMaricic (approved):

src/Layers/W1/BaseApp/Sales/Document/SalesOrderList.Page.al
```diff
@@ -1163,7 +1163,7 @@ page 9305 "Sales Order List"
         view(Delayed)
         {
             Caption = 'Delayed';
-            Filters = where("Document Type" = const(Order), Status = const(Released), "Completely Shipped" = const(false), "Late Order Shipping" = const(true));
+            Filters = where("Document Type" = const(Order), Status = const(Released), "Completely Shipped" = const(false), "Date Filter" = filter(<= '%workdate'), "Late Order Shipping" = const(true));
```
qasimikram, line 1166: S1: Please add a regression test for the bookmarked `Delayed` view. Create one released, incompletely shipped order with an outstanding line dated on/before WorkDate and another dated after WorkDate, open the Sales Order List through the bookmarked `Delayed` view, and verify only the overdue order appears. A test that only exercises the FlowField or normal page opening could pass because `OnOpenPage` already sets `Date Filter`, while missing the initialization-order regression described by AB#642628.

qasimikram (approved):
The filter change correctly supplies the `Date Filter` used by the `Late Order Shipping` FlowField when the predefined `Delayed` view is opened directly.

S1: Please add regression coverage for the bookmarked-view path, including an overdue order and a future-dated order. The actionable details are in the unresolved inline thread.

src/Layers/W1/BaseApp/Sales/Document/SalesOrderList.Page.al
6CRIPT: The predefined/bookmarked page view cannot be selected through the AL TestPage API. OpenView() only opens the page in view mode, and testing the FlowField, the cue drilldown, or applying the filters manually would pass even if this view change were reverted. Therefore, such an AL test would not provide regression coverage for the reported initialization-order issue.

PR #9640, opened 2026-07-21
[Master]-Fixed Asset: Declining-Balance method - wrong calculation when running depreciation

neeleshsinghal:


neeleshsinghal (commented):
I dont think this makes any difference as we may not even need this

neeleshsinghal:
Superseded by #9662 (recreated as a single clean commit with only my changes).

PR #9647, opened 2026-07-22
[Payables Agent] Write transaction in Payable Agent code triggered by Confirm dialog

Explorer986:
## What & why

GetAgent unconditionally called PayablesAgentSetup.Modify(), opening a write transaction even when nothing changed. When invoked from agent action visibility checks (ShowCanCreateAgent) around a Confirm dialog, this left an open write transaction that blocked subsequent Page.RunModal calls. Now only persist when the setup record actually changed.

Fixes [AB#641849](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/641849)

## How I validated this

### DIFF
--- a/src/Apps/W1/PayablesAgent/app/Setup/PayablesAgentSetup.Codeunit.al
+++ b/src/Apps/W1/PayablesAgent/app/Setup/PayablesAgentSetup.Codeunit.al
@@ -201,19 +201,26 @@ codeunit 3307 "Payables Agent Setup"
     procedure GetAgent(var Agent: Record Agent): Boolean
     var
         PayablesAgentSetup: Record "Payables Agent Setup";
+        SetupChanged: Boolean;
     begin
         PayablesAgentSetup.GetSetup();
         // We attempt to find the agent by the security id stored in the setup record.
         if Agent.Get(PayablesAgentSetup."User Security Id") then
             exit(true);
         // If the agent could not be found, and there was a user security id configured, we need to clear it, since it is not valid anymore.
-        if not IsNullGuid(PayablesAgentSetup."User Security Id") then
+        if not IsNullGuid(PayablesAgentSetup."User Security Id") then begin
             Clear(PayablesAgentSetup."User Security Id");
+            SetupChanged := true;
+        end;
         // If the agent could not be found from the configured security id, we attempt to find it by the user name.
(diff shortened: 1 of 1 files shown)


alexei-dobriansky:

<details>

This PR makes `GetAgent` call `PayablesAgentSetup.Modify()` only when the setup record is changed. That removes the repeated write when the setup already points to an existing agent.

The fix does not fully remove writes from the action-visibility path that can run around a `Confirm` dialog. `ShowCanCreateAgent` still reaches `AllowCreateNewAgent`, which calls `GetAgent`; `GetAgent` can still insert the setup record through `GetSetup()`, clear a stale security id, or cache an agent found by name. The analogous SOA and Custom Agent create checks are read-only, so this path should also avoid repair writes during visibility checks.
</details>

<details>

**S1** - *Read-only visibility still writes setup*
Change the visibility path so it never calls a procedure that can insert or modify `Payables Agent Setup`. `GetAgent` still writes on missing setup, stale security id, or agent-found-by-name paths. The `Confirm` and `Page.RunModal` bug can still happen in those cases.

**S2** - *Regression test covers write-free visibility*
Add a regression test for the `Confirm`/`Page.RunModal` scenario, or for `ShowCanCreateAgent` proving that visibility checks do not write setup data. Cover missing setup, stale security id, and agent-found-by-name paths. This bug is easy to reintroduce because the write happens indirectly through `GetAgent`.
</details>

<details>

**Risk:** The changed code runs from Payables Agent action visibility and idle agent checks, including UI flows around modal dialogs. A remaining write transaction can still block `Page.RunModal` after the user confirms. The PR does not change public signatures and does not introduce a `Commit()` before `Error()`, but the transaction side effect remains in some setup states.

**Necessity:** The linked Bug is valid and important: a background visibility check can break a normal user flow after a committed `User` change. The fix is needed, but the scope should be a read-only lookup path for visibility checks, with setup repair moved to an explicit setup or activation flow.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9647 round=1 by=alexei-dobriansky at=2026-08-03T10:19:26Z lastSha=2985c683b1717d299f7ca1087cccdfbe415bf050 suggestions=S1,S2

Explorer986:
This is already fixed

PR #9648, opened 2026-07-22
[master]-[BE] [PEPPOL] There is a problem between the totals on the invoice printout and the XML PEPPOL with Payment discount

MSNehaNawal:
[Bug 643204](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/643204): [ALL-E] [BE] [PEPPOL] There is a problem between the totals on the invoice printout and the XML PEPPOL file that is generated when using a discount % in the payment condtions. So in Belgium we calc the taxable amount on the total amount - discount%

Fixes [AB#643204](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/643204)

### DIFF
--- a/src/Apps/BE/PeppolBE/App/src/PEPPOL30BESubscribers.Codeunit.al
+++ b/src/Apps/BE/PeppolBE/App/src/PEPPOL30BESubscribers.Codeunit.al
@@ -4,7 +4,6 @@
 // ------------------------------------------------------------------------------------------------
 namespace Microsoft.Peppol.BE;
 
-using Microsoft.Finance.VAT.Calculation;
 using Microsoft.Peppol;
 
 codeunit 37314 "PEPPOL30 BE Subscribers"
@@ -22,24 +21,5 @@ codeunit 37314 "PEPPOL30 BE Subscribers"
         Rec."PEPPOL 3.0 Sales Format" := Rec."PEPPOL 3.0 Sales Format"::"PEPPOL 3.0 - BE Sales";
         Rec."PEPPOL 3.0 Service Format" := Rec."PEPPOL 3.0 Service Format"::"PEPPOL 3.0 - BE Service";
         Rec.Modify();
-    end;
-
-    [EventSubscriber(ObjectType::Codeunit, Codeunit::"PEPPOL30 Common", 'OnAfterGetSalesTotals', '', false, false)]
-    local procedure OnAfterGetSalesTotalsRemovePaymentDiscount(PostedDocHeaderRecRef: RecordRef; var TempVATAmtLine: Record "VAT Amount Line" temporary; PEPPOLFormat: Enum "PEPPOL 3.0 Format")
-    begin
-        // In Belgium the payment discount must not reduce the PEPPOL document totals, so the XML matches
-        // the invoice printout and VAT is reported on the full amount. Clearing the payment discount on the
-        // VAT amount line buffer keeps TaxableAmount, TaxExclusiveAmount, TaxInclusiveAmount, PayableAmount and
-        // the payment discount AllowanceCharge aligned with the printout.
-        if not (PEPPOLFormat in [PEPPOLFormat::"PEPPOL 3.0 - BE Sales", PEPPOLFormat::"PEPPOL 3.0 - BE Service"]) then
-            exit;
-
-        if TempVATAmtLine.FindSet() then
-            repeat
-                if TempVATAmtLine."Pmt. Discount Amount" <> 0 then begin
-                    TempVATAmtLine."Pmt. Discount Amount" := 0;
-                    TempVATAmtLine.Modify();
-                end;
-            until TempVATAmtLine.Next() = 0;
     end;
 }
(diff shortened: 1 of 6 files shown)


src/Apps/W1/PEPPOL/App/src/Common/PEPPOL30Impl.Codeunit.al
```diff
@@ -1563,6 +1565,28 @@ codeunit 37201 "PEPPOL30 Impl."
         exit(CountryRegion."ISO Code" = 'DK');
     end;
 
+    local procedure IsBelgianCompany(): Boolean
```
Groenbech96: Dont we have a BE localization of PEPPOL?

Groenbech96 (changes requested):

src/Apps/W1/PEPPOL/App/src/Common/PEPPOL30Impl.Codeunit.al
MSNehaNawal: @Groenbech96 Peppol BE, do not have this codeunit.
Below is the call stack-
image

djukicmilica: You need to use different service code in BE env. ("PEPPOL 3.0 - BE Sales")

and do changes in
https://github.com/microsoft/BCApps/tree/main/src/Apps/BE/PeppolBE/App

MSNehaNawal: @Groenbech96 @djukicmilica
I have made the suggested changes, can you please have a look.

src/Apps/BE/PeppolBE/App/src/PEPPOL30BESubscribers.Codeunit.al
```diff
@@ -22,4 +23,23 @@ codeunit 37314 "PEPPOL30 BE Subscribers"
...
+        // In Belgium the payment discount must not reduce the PEPPOL document totals, so the XML matches
+        // the invoice printout and VAT is reported on the full amount. Clearing the payment discount on the
+        // VAT amount line buffer keeps TaxableAmount, TaxExclusiveAmount, TaxInclusiveAmount, PayableAmount and
+        // the payment discount AllowanceCharge aligned with the printout.
+        if not (PEPPOLFormat in [PEPPOLFormat::"PEPPOL 3.0 - BE Sales", PEPPOLFormat::"PEPPOL 3.0 - BE Service"]) then
+            exit;
+
+        if TempVATAmtLine.FindSet() then
+            repeat
+                if TempVATAmtLine."Pmt. Discount Amount" <> 0 then begin
```
djukicmilica: Can we use ModifyAll instead of the loop?

Groenbech96:
**Recommendation: implement this through the `PEPPOL Tax Info Provider` interface instead of a new integration event.**

This module currently has **no** integration events — all format-specific behavior goes through the enum-implements-interface pattern (the BE format values in `PEPPOL30FormatBE.EnumExt.al` already override `PEPPOL30 Validation` and the posted-document iterator). Adding the module''s first `[IntegrationEvent]` that mutates the shared `TempVATAmtLine` buffer introduces a second, less discoverable extensibility mechanism.

`Pmt. Discount Amount` is written to the buffer in exactly one place — `PEPPOL30 Impl.GetTaxTotals` (`VATAmtLine."Pmt. Discount Amount" += SalesLine."Pmt. Discount Amount";`) — and `GetTaxTotals` is part of the `PEPPOL Tax Info Provider` interface. So the same result can be achieved on-pattern:

**1.** Add a BE codeunit implementing `PEPPOL Tax Info Provider` that delegates every method to codeunit `PEPPOL30` (it''s `Access = Public`), and in `GetTaxTotals` zeroes the payment discount before delegating:

```al
procedure GetTaxTotals(SalesLine: Record "Sales Line"; var VATAmtLine: Record "VAT Amount Line")
begin
    SalesLine."Pmt. Discount Amount" := 0; // BE: exclude payment discount from PEPPOL totals
    PEPPOL30.GetTaxTotals(SalesLine, VATAmtLine);
end;
```

**2.** Point the BE format values at it in `PEPPOL30FormatBE.EnumExt.al` (on both `PEPPOL 3.0 - BE Sales` and `PEPPOL 3.0 - BE Service`):

```al
Implementation = ... , "PEPPOL Tax Info Provider" = "PEPPOL30 BE Tax Info";
```

**3.** Remove the `OnAfterGetSalesTotals` event from `PEPPOL30 Common` and the subscriber from `PEPPOL30 BE Subscribers`.

Because `GetTaxTotals` is the **sole writer** of that field, zeroing the by-value `SalesLine` copy before accumulation produces the identical buffer as the post-loop in the subscriber (every line ends with `Pmt. Discount Amount = 0`), so all downstream totals and the payment-discount `AllowanceCharge` (auto-skipped via its `= 0` guard) stay consistent. The conditional handling is fine as-is — it only takes effect when a payment discount is actually present.

This keeps all format-specific logic behind the interface, avoids mutating shared state through an event, and needs no new publisher.

djukicmilica:
Also, add some tests.

v-ankitgoyal (approved):

DavidHolusa (approved):

Groenbech96 (approved):
Great.

Groenbech96:
Closed, reopen to retrigger build

djukicmilica (approved):

ventselartur (approved):

PR #9651, opened 2026-07-22
[Extensibility Request] issue 30352: add SkipNewExpirationDateCheck to OnCheckExpirationDateOnAfterCalcSumLot

AleksandricMarko:
## Summary
The reporter needs a way to bypass the mandatory `New Expiration Date` validation performed during item journal posting in codeunit 22 "Item Jnl.-Post Line". The existing `OnBeforeCheckExpirationDate` (IsHandled) event would force them to replicate the whole procedure and skip several downstream events, risking conflicts with other extensions. This PR extends the existing `OnCheckExpirationDateOnAfterCalcSumLot` integration event with a new `var SkipNewExpirationDateCheck: Boolean` parameter so a subscriber can skip only that single `TestField` check without affecting any other logic.

Source issue repository: microsoft/ALAppExtensions; issue number: 30352

## Changes Made
- `OnCheckExpirationDateOnAfterCalcSumLot` - appended a new `var SkipNewExpirationDateCheck: Boolean` parameter to the integration event; `CheckExpirationDate` now initializes it to `false`, passes it to the event, and only performs `TempTrackingSpecification.TestField("New Expiration Date", ...)` when the subscriber has not set it to `true`. Adding a parameter to an existing integration event is non-breaking for current subscribers. The same change was mirrored across the W1 base layer and its country-layer counterparts (ES, APAC, CH, RU, IT).

Fixes [AB#642607](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/642607)

### DIFF
--- a/src/Layers/W1/BaseApp/Inventory/Posting/ItemJnlPostLine.Codeunit.al
+++ b/src/Layers/W1/BaseApp/Inventory/Posting/ItemJnlPostLine.Codeunit.al
@@ -4198,6 +4198,7 @@ codeunit 22 "Item Jnl.-Post Line"
         SumOfEntries: Decimal;
         SumLot: Decimal;
         IsHandled: Boolean;
+        SkipNewExpirationDateCheck: Boolean;
     begin
         IsHandled := false;
         OnBeforeCheckExpirationDate(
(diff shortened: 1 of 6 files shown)


src/Layers/W1/BaseApp/Inventory/Posting/ItemJnlPostLine.Codeunit.al
```diff
@@ -4278,8 +4279,9 @@ codeunit 22 "Item Jnl.-Post Line"
                         SumLot := SignFactor * ItemTrackingMgt.SumNewLotOnTrackingSpec(TempTrackingSpecification)
                     else
                         SumLot := SignFactor * TempTrackingSpecification."Quantity (Base)";
-                    OnCheckExpirationDateOnAfterCalcSumLot(SumLot, SignFactor, TempTrackingSpecification);
-                    if (SumOfEntries > 0) and
+                    SkipNewExpirationDateCheck := false;
+                    OnCheckExpirationDateOnAfterCalcSumLot(SumLot, SignFactor, TempTrackingSpecification, SkipNewExpirationDateCheck);
+                    if (not SkipNewExpirationDateCheck) and (SumOfEntries > 0) and
```
qasimikram, line 4284: S1: Please add regression coverage proving that the default path still raises the existing expiration-date validation and that setting `SkipNewExpirationDateCheck := true` skips only this `TestField` without bypassing the rest of `CheckExpirationDate`.

qasimikram (approved):
The implementation is consistent across the W1 and localized code copies and scopes the bypass to the requested expiration-date validation.

S1: Please add regression coverage for the new subscriber-controlled path. The test should prove that the default path still raises the existing validation error and that setting `SkipNewExpirationDateCheck := true` skips only this `TestField` without bypassing the remainder of `CheckExpirationDate`.

PredragMaricic (approved):

PR #9653, opened 2026-07-22
[Master][ALL-E] No Reservation created when the item is selected by Description in Sales Lines

sanjmaurya:
[Bug 643213](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/643213): [Master][ALL-E] No Reservation created when the item is selected by Description in Sales Lines

Fixes [AB#643213](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/643213)

**Issue :-**  No Reservation created when the item is selected by Description in Sales Lines  

**Cause :-**  Reserve = Always items, the automatic reservation in NoOnAfterValidate (SalesOrderSubform.Page.al) is gated by (Rec."Outstanding Qty. (Base)" <> 0) and (Rec."No." <> xRec."No."). That guard is evaluated after a CurrPage.SaveRecord() call.

Manual description entry: "No." is resolved inside the field's OnValidate, xRec."No." stays blank, so the guard holds → reservation created. (Already working after the CurrFieldNo edits.)
Description lookup (dropdown): "No." is set earlier in OnAfterLookup → SaveLookupSelection. By the time the inner CurrPage.SaveRecord() runs, xRec re-syncs to the saved record, so Rec."No." <> xRec."No." flips to false and AutoReserve() is skipped.
 
**Solutions :-**   n NoOnAfterValidate, capture whether "No." changed into a local NoHasChanged before any CurrPage.SaveRecord() runs, and use that captured value in the reservation guard. This preserves the working manual path (value stays true) and fixes the lookup path.

Two files now carry the fix:

SalesLine.Table.al — CurrFieldNo := FieldNo("No.") set at the three points where Description resolves to an item No. (exact/similar match branches, SaveLookupSelection, RestoreLookupSelection), matching the Purchase Line pattern.
SalesOrderSubform.Page.al — capture NoHasChanged before SaveRecord and use it in the AutoReserve guard.


v-ankitgoyal (dismissed):

PredragMaricic (changes requested):
S1 (High): The lookup-reservation flag can survive an error.

`AutoReserveItemFromLookup` is page-global and is cleared only after `NoOnAfterValidate()` returns. If that validation errors, the reset is skipped. A later Type, No., or Item Reference validation can then incorrectly treat the item as newly selected through Description lookup and invoke `AutoReserve()`.

Please pass the lookup-restored state directly into the validation flow, or otherwise guarantee cleanup across errors, and add an error/retry regression test.

alexei-dobriansky:

<details>

Fixes missing auto-reservation when a "Reserve = Always" item is selected through the Description field's lookup dropdown on Sales Order lines. The root cause: by the time `NoOnAfterValidateInternal` calls `CurrPage.SaveRecord()`, `xRec."No."` has already been synced to the saved record, so `Rec."No." <> xRec."No."` evaluates to `false` and `AutoReserve()` is skipped.

The fix introduces `RestoreLookupSelectionWithResult()` returning a boolean, sets `CurrFieldNo` to simulate the "No." field context, captures `NoHasChanged` before `SaveRecord`, and uses it in the reservation guard. The approach is correct: it evaluates the change flag early (before SaveRecord resyncs xRec) and threads it through a dedicated parameter, avoiding pollution of other callers. The old `RestoreLookupSelection()` is preserved as a wrapper for backward compatibility. The fix is applied to all 7 layer-specific SalesOrderSubform pages and all layer-specific SalesLine.Table.al files. Purchase/Service lines do not have this auto-reserve pattern so the scope is complete. Tests cover both the happy path (reservation created) and a negative case (leaked state after error does not cause spurious reservation).
</details>

<details>

**S1** - *Fix malformed AB# link in PR body*
The PR body says `Fixes AB#6493213` but the actual work item is AB#643213 (Bug 643213). The extra digit prevents ADO from linking the PR to the bug automatically.

**S2** - *Verify CurrFieldNo is safe to set unconditionally in RestoreLookupSelectionWithResult*
`CurrFieldNo := FieldNo("No.")` is set inside the `if LookupStateManager.IsRecordSaved()` block for all Type values (G/L Account, Resource, Fixed Asset, Item Charge), not just Item. If the "No." OnValidate trigger for non-Item types has logic gated by `CurrFieldNo`, this could trigger unintended side effects for those types. Confirm those code paths are inert when CurrFieldNo equals "No.".

**S3** - *Second test may not catch the leak it claims to guard against*
`LookupAutoReserveStateDoesNotSurviveErrorInSalesOrder` uses `asserterror` which rolls back database changes, but the page-level `LookupSelectionRestored` local variable lives on the stack of the OnValidate trigger — it is inherently per-invocation and cannot leak across separate `SetValue` calls. The test therefore passes trivially regardless of the fix. Consider whether this test actually validates a real regression scenario or is a no-op.
</details>

<details>

**Risk:** Low-medium. The fix narrows the condition for `AutoReserve()` by adding an OR branch (`NoHasChanged`), meaning it can only INCREASE reservation calls, never decrease them for normal paths. The `NoOnAfterValidate()` public wrapper is preserved, so partner extensions subscribing to `OnBeforeNoOnAfterValidate` are unaffected. The `RestoreLookupSelection()` overload is kept for binary compatibility. The `CurrFieldNo` assignment may fire existing field-level subscribers for "No." on non-Item types — low probability but worth verifying.

**Necessity:** High. Items with Reserve=Always not being reserved when selected by description is a functional regression that corrupts availability planning. The scope (Sales Order only) is correct — Purchase Order Subform has no auto-reserve logic.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9653 round=1 by=alexei-dobriansky at=2026-08-03T09:30:00Z lastSha=ef0412fe2906a8167dae05677df7e06f627c8d95 suggestions=S1,S2,S3

sanjmaurya:
> ## Agentic PR Review - Round 1
> # **Recommendation: Accept with Suggestions**
> **What this PR does**
> Fixes missing auto-reservation when a "Reserve = Always" item is selected through the Description field's lookup dropdown on Sales Order lines. The root cause: by the time `NoOnAfterValidateInternal` calls `CurrPage.SaveRecord()`, `xRec."No."` has already been synced to the saved record, so `Rec."No." <> xRec."No."` evaluates to `false` and `AutoReserve()` is skipped.
> 
> The fix introduces `RestoreLookupSelectionWithResult()` returning a boolean, sets `CurrFieldNo` to simulate the "No." field context, captures `NoHasChanged` before `SaveRecord`, and uses it in the reservation guard. The approach is correct: it evaluates the change flag early (before SaveRecord resyncs xRec) and threads it through a dedicated parameter, avoiding pollution of other callers. The old `RestoreLookupSelection()` is preserved as a wrapper for backward compatibility. The fix is applied to all 7 layer-specific SalesOrderSubform pages and all layer-specific SalesLine.Table.al files. Purchase/Service lines do not have this auto-reserve pattern so the scope is complete. Tests cover both the happy path (reservation created) and a negative case (leaked state after error does not cause spurious reservation).
> 
> **Suggestions**
> **S1** - _Fix malformed AB# link in PR body_ The PR body says `Fixes AB#6493213` but the actual work item is AB#643213 (Bug 643213). The extra digit prevents ADO from linking the PR to the bug automatically.
> 
> **S2** - _Verify CurrFieldNo is safe to set unconditionally in RestoreLookupSelectionWithResult_ `CurrFieldNo := FieldNo("No.")` is set inside the `if LookupStateManager.IsRecordSaved()` block for all Type values (G/L Account, Resource, Fixed Asset, Item Charge), not just Item. If the "No." OnValidate trigger for non-Item types has logic gated by `CurrFieldNo`, this could trigger unintended side effects for those types. Confirm those code paths are inert when CurrFieldNo equals "No.".
> 
> **S3** - _Second test may not catch the leak it claims to guard against_ `LookupAutoReserveStateDoesNotSurviveErrorInSalesOrder` uses `asserterror` which rolls back database changes, but the page-level `LookupSelectionRestored` local variable lives on the stack of the OnValidate trigger — it is inherently per-invocation and cannot leak across separate `SetValue` calls. The test therefore passes trivially regardless of the fix. Consider whether this test actually validates a real regression scenario or is a no-op.
> 
> **Risk assessment and necessity**
> **Risk:** Low-medium. The fix narrows the condition for `AutoReserve()` by adding an OR branch (`NoHasChanged`), meaning it can only INCREASE reservation calls, never decrease them for normal paths. The `NoOnAfterValidate()` public wrapper is preserved, so partner extensions subscribing to `OnBeforeNoOnAfterValidate` are unaffected. The `RestoreLookupSelection()` overload is kept for binary compatibility. The `CurrFieldNo` assignment may fire existing field-level subscribers for "No." on non-Item types — low probability but worth verifying.
> 
> **Necessity:** High. Items with Reserve=Always not being reserved when selected by description is a functional regression that corrupts availability planning. The scope (Sales Order only) is correct — Purchase Order Subform has no auto-reserve logic.
> 
> > [AI-PR-REVIEW] version=1 system=github pr=9653 round=1 by=alexei-dobriansky at=2026-08-03T09:30:00Z lastSha=ef0412fe2906a8167dae05677df7e06f627c8d95 suggestions=S1,S2,S3

Hi [alexei-dobriansky](https://github.com/alexei-dobriansky)
S1 is resolved

S2. — CurrFieldNo := FieldNo("No.") in RestoreLookupSelectionWithResult
Confirmed safe — no code change needed. I traced every CurrFieldNo-gated path reachable during Validate("No.").
Setting CurrFieldNo := FieldNo("No.") only alters behavior for logic gated specifically on CurrFieldNo = FieldNo("No."). Before the fix, CurrFieldNo held FieldNo(Description) (non-zero) during the restore call, so any CurrFieldNo <> 0 gate already fired — the change introduces nothing new there.

There are three = FieldNo("No.") gates reachable during Validate("No."), and all are inert for non-Item types (G/L Account, Resource, Fixed Asset, Charge):

1. CheckWarehouse — if CurrFieldNo = FieldNo("No.") then CheckWarehouse(false); is nested inside if (xRec."No." <> "No.") and (Quantity <> 0). On the restore path the No. was already assigned in SaveLookupSelection to the same value, so xRec."No." = "No." and the block is skipped for all types — not just non-Item.
2. CheckItemAvailable — its body is guarded by (Type = Type::Item), so it's inert for non-Item types.
3. CheckCreditLimit — the CurrFieldNo = FieldNo("No.") term appears only as part of an Item-specific exclusion; it has no effect for non-Item types.

S3:- resolved

alexei-dobriansky:

# **Recommendation: Accept**

<details>

The core fix is unchanged from Round 1: it captures whether `"No."` changed into a local `NoHasChanged` before `CurrPage.SaveRecord()` resyncs `xRec`, threads a `SelectionRestored` flag out of `RestoreLookupSelectionWithResult()`, and uses it in the `AutoReserve()` guard so that selecting a `Reserve = Always` item through the Description lookup still creates the reservation.

Since Round 1 the only author change is a rework of the negative test `LookupAutoReserveStateDoesNotSurviveErrorInSalesOrder` (W1 and RU test codeunit 137049). The test now asserts directly against the single-instance `Lookup State Manager`: it saves a lookup selection, asserts `IsRecordSaved()` is true, forces an error during the `No.` validation flow, and asserts `IsRecordSaved()` is false afterward. This is a meaningful regression guard - the state manager is in-memory and non-transactional, so it survives the `asserterror` rollback, and the assertion verifies the production code clears it (`RestoreLookupSelectionWithResult` -> `ClearSavedRecord`) *before* `OnBeforeNoOnAfterValidate` fires. The unused `Library - Item Reference` codeunit and `Item Reference` record were removed with no remaining references.
</details>

<details>

| ID  | Title                                                        | Status    | Author response |
|-----|--------------------------------------------------------------|-----------|-----------------|
| S1  | Fix malformed AB# link in PR body                            | Addressed | PR body now links `AB#643213` (both the title and the `Fixes` line); ADO auto-link restored. |
| S2  | Verify CurrFieldNo is safe to set unconditionally            | Disputed  | Confirmed safe, no code change. Traced all three `CurrFieldNo = FieldNo("No.")` gates: `CheckWarehouse` is skipped because `No.` is unchanged on the restore path; `CheckItemAvailable` and the `CheckCreditLimit` term are `Type = Item`-guarded, so they are inert for G/L Account, Resource, Fixed Asset and Item Charge. Verified against `SalesLine.Table.al` and accepted. |
| S3  | Second test may not catch the leak it claims to guard against | Addressed | Test reworked to assert `Lookup State Manager.IsRecordSaved()` transitions true -> false across the forced error. Verified: the single-instance state is cleared in `RestoreLookupSelectionWithResult` before `OnBeforeNoOnAfterValidate`, and survives rollback, so the assertion is real - it would fail if the clear regressed. |
</details>

<details>

*None.* The only change since Round 1 is the test rework, which addresses S3. `LibraryItemReference` removal leaves no dangling references (W1 and RU), the RU codeunit mirrors W1 exactly, and each test that saves a lookup selection also clears it (or overwrites it via `SaveItemLookupSelection`), so no single-instance state leaks between tests.
</details>

<details>

**Risk:** Low. The Round-2 delta is test-only - no production behavior changed since Round 1. The reworked negative test is now a genuine assertion of the clear-before-error contract rather than a trivially-passing check. Production risk remains as assessed in Round 1: the guard can only ever *increase* reservation calls, the `NoOnAfterValidate()` wrapper and `RestoreLookupSelection()` overload are preserved for compatibility, and the `CurrFieldNo` assignment was confirmed inert for non-Item types (S2).

**Necessity:** High, unchanged from Round 1. `Reserve = Always` items not being reserved when selected by Description corrupts availability planning. Scope (Sales Order) remains correct.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9653 round=2 by=alexei-dobriansky at=2026-08-03T13:13:00Z lastSha=b8c6037c0fb3e0aa61348bfaa0aeb79a547404b6 suggestions=S1:addressed,S2:disputed,S3:addressed parentRound=1

alexei-dobriansky (approved):

v-praghav (approved):

PredragMaricic (approved):

PR #9654, opened 2026-07-22
Default new E-Document services to import v2 and expose the version selector

Groenbech96:
## What & why

Step 1 of moving the E-Document framework's inbound import onto the **v2 draft pipeline**: make Version 2.0 the default for **newly created** services and expose the version selector in the UI. Concretely:

- **Default new services to v2** — add `InitValue = "Version 2.0"` to `field(31) "Import Process"` on table `E-Document Service`. Any service initialized via `Rec.Init()` (the service card's "New", programmatic creation) now starts on the v2 draft pipeline. Existing/stored services are untouched.
- **Expose the version on the service card** — remove `Visible = false` from the `"Import Process"` field on page `E-Document Service` so users can see and switch the version. The value-driven `PurchaseDraft` (v2) and `ImportParamenters` (v1) groups already toggle off this field.

**Deliberately NOT in scope:** no obsoletion/deprecation of v1. The local format apps (PEPPOL, XRechnung, ZUGFeRD, OIOUBL, Factura-E, Verifactu, FacturX, PINT ANZ, EReporting) are **not yet migrated** off the v1 inbound path, so v1 stays fully functional and selectable. Deprecation is a later step once formats are migrated.

## Linked work

Fixes #

## How I validated this

- Added unit test `NewServiceDefaultsToImportProcessV2` in `EDocProcessTest.Codeunit.al`: a freshly `Init()`-ed `E-Document Service` asserts `"Import Process" = "Version 2.0"`.
- **Test-preservation:** the existing receive/flow/E2E suites rely on service factories producing v1 (so the v1 mock subscribers run). All nine `Init()`-only factories in `LibraryEDocument.Codeunit.al` are now pinned to `Version 1.0`, so those suites exercise v1 exactly as before — the only observers of the new v2 default are production `Init()` and the new unit test. Tests that need v2 already opt in explicitly (`"Import Process" := "Version 2.0"`).
- Verified statically that the `InitValue` default only reaches records via `Rec.Init()`: the sole non-factory inline service creators are `EDocAPITest` (version-agnostic), the Microsoft365 Outlook test (no `Init()` → unaffected), and demo-data codeunits (not consumed by any test; the inbound demo service is already explicitly v2).
- **Not built/run locally**: this environment has no AL symbols/compiler, so compilation and in-product verification are left to CI and the PR author. The change is a strict subset of the previously reviewed #9649 (which this supersedes), minus all obsoletion.

## Risk & compatibility

- **New services only.** `InitValue` changes the default for records created via `Init()`; existing service records keep their stored value. No upgrade/data migration.
- **No breaking changes / no obsoletes.** v1 remains selectable and functional; nothing is marked `Obsolete` and no `#if`/pragma guards were added.
- Demo data: the outbound-purposed `E-DOCUMENTS` demo service inherits the new v2 default (Import Process is irrelevant to its send workflow); the inbound demo service was already v2.
- Follow-up: migrate the local format apps onto v2, then deprecate the v1 inbound path.


Groenbech96:
Combined into #9702 per request -- that PR now contains these commits plus the draft-format selector so v2 is usable end-to-end. Closing this one in favor of #9702.

PR #9657, opened 2026-07-22
[Extensibility Request] issue 30359: add Direction parameter to OnBeforeScheduleRoutingLine

AleksandricMarko:
## Summary
The reporter wants to influence the starting- and ending-datetime calculation in production order routing. To do that, subscribers need access to the scheduling `Direction` so they can adjust it before the routing line is scheduled. This PR exposes `Direction` as a `var` parameter on the existing publisher event so extensions can modify routing behaviour without breaking the base logic.

Source issue repository: microsoft/ALAppExtensions; issue number: 30359

## Changes Made
- `OnBeforeScheduleRoutingLine` (codeunit 99000774 "Calculate Routing Line") - appended `var Direction: Option Forward,Backward` to the publisher signature and passed the in-scope `Direction` from `CalculateRoutingLine` when raising the event, allowing subscribers to read and change the scheduling direction before the line is scheduled.

Fixes [AB#642790](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/642790)

### DIFF
--- a/src/Layers/W1/BaseApp/Manufacturing/Routing/CalculateRoutingLine.Codeunit.al
+++ b/src/Layers/W1/BaseApp/Manufacturing/Routing/CalculateRoutingLine.Codeunit.al
@@ -1364,7 +1364,7 @@ codeunit 99000774 "Calculate Routing Line"
         CalcExpectedCost(ProdOrderRoutingLine, TotalQtyPerOperation, TotalCapacityPerOperation);
 
         IsHandled := false;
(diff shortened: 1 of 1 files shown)


qasimikram (approved):
The change matches the linked extensibility request and exposes the existing scheduling direction at the correct pre-scheduling event. The potentially modified value is consumed immediately by the existing forward/backward branch, while default behavior remains unchanged when there are no subscribers.

JakovljevicDusan (approved):

PR #9660, opened 2026-07-22
[Master]-Act Consumption Qty field in Component of a Firm Planned Order has amounts that point to a Released Production Order that has the same Document No if the component is the same

neeleshsinghal:
[AB#643947](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/643947)

Recreated as a single, clean commit containing only the author's changes (Copilot build/test/CI commits from the previous branch have been dropped).

Supersedes #9630.

### DIFF
--- a/src/Layers/W1/BaseApp/Manufacturing/Document/ProdOrderCompLineList.Page.al
+++ b/src/Layers/W1/BaseApp/Manufacturing/Document/ProdOrderCompLineList.Page.al
@@ -5,6 +5,7 @@
 namespace Microsoft.Manufacturing.Document;
 
 using Microsoft.Finance.Dimension;
+using Microsoft.Inventory.Ledger;
 
 page 5407 "Prod. Order Comp. Line List"
 {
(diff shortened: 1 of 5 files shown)


v-ankitgoyal (approved):

src/Layers/W1/BaseApp/Manufacturing/Document/ProdOrderCompLineList.Page.al
```diff
@@ -138,10 +139,28 @@ page 5407 "Prod. Order Comp. Line List"
                     AutoFormatType = 0;
                     ApplicationArea = Manufacturing;
                 }
-                field("Act. Consumption (Qty)"; Rec."Act. Consumption (Qty)")
+                field("Act. Consumption (Qty)"; ActConsumptionQtyValue)
```
qasimikram, line 142: S1: Rebinding this control from `Rec."Act. Consumption (Qty)"` to a page variable drops the table field's `AccessByPermission = TableData "Production Order" = R;` behavior. Please add the same property to this replacement control and its W1/IT counterparts so users without Production Order read permission do not gain visibility to the field.

qasimikram (approved):
Approve with suggestions.

S1 - Preserve the field permission gate: The replacement page-variable controls should retain `AccessByPermission = TableData "Production Order" = R;` from the original `Prod. Order Component` FlowField. Please apply this to both page controls in the W1 and IT variants so UI element removal behavior remains unchanged.

S2 - Cover the preserved Released/Finished path: The regression test covers the Firm Planned zero-value branch, but not the new `CalcFields` assignment and custom drilldown used for Released/Finished components. Please add a positive test asserting that posted consumption is still displayed, ideally also validating the drilldown filters.

src/Layers/W1/BaseApp/Manufacturing/Document/ProdOrderCompLineList.Page.al
neeleshsinghal: Its not visible until you personilize

PR #9695, opened 2026-07-23
621963 Deprecate table 9650 "Custom Report Layout" (Pending, tag 29.0) + country forks

V-CWurtzen:
## What & why

Deprecates table **9650 "Custom Report Layout"** (`ObsoleteState = Pending`, tag `29.0`), mirroring the
already-Pending page 9650. The legacy custom-layout table is superseded by the system report-layout
tables — **"Tenant Report Layout"** and **"Report Layout Selection"**, surfaced by the system *Report
Layouts* page. Marking it Pending starts the retirement clock and points partners at the modern surface.

No schema is removed and no data-migration behaviour changes: the table stays defined and present, and
remains in the cloud-migration replication set. Physical removal is a later `Removed` / `CLEANSCHEMA` step.

## Design: the table stays, its legacy surface does not

Table 9650 is marked Pending but deliberately **not** wrapped in `#if not CLEAN29`. It has live schema
dependents — `TableRelation`, `CalcFormula` and `tabledata` permissions — that must keep resolving in the
Clean build; wrapping the table would turn those into errors no pragma can suppress.

The legacy surface that *can* go is obsoleted and compiled out of the Clean build: the layout-description
FlowFields that look up the table, page 9653 *"Report Layouts Part"* and its part on *Report Layout
Selection*, and the consumer controls, triggers and `CalcFields` calls that fed them.

Where a construct leaves the Clean build it is marked obsolete rather than having the diagnostic
suppressed. The suppressions that remain are scoped to the referencing statement and sit inside the same
`#if not CLEAN29` guard, so they retire together with the code they cover.

## Scope

W1 BaseApp, Tests and DemoTool; the two CZ localization report-selection pages bound to a wrapped field;
and fork parity for every existing fork of each changed W1 file. Test layers are included because they
compile in the Clean build.

Metadata and reference-guarding only — obsoletion markers, `#if not CLEAN29` guards and balanced pragmas.
No functional or behavioural change.

## For reviewers

User-visible in the Clean build: the layout *description* columns on the report-selection setup pages, the
custom-layout column on *Interact. Tmpl. Languages*, and the **Custom Layouts** part on *Report Layout
Selection* disappear. The functionality is superseded by the system tables, but it is a visible change and
worth acknowledging as such.

Fixes [AB#621963](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/621963)


src/Layers/W1/BaseApp/Foundation/Reporting/CustomReportLayout.Table.al
```diff
@@ -11,10 +11,14 @@ using System.Reflection;
 using System.Security.AccessControl;
 using System.Utilities;
 
+#pragma warning disable AS0105
 table 9650 "Custom Report Layout"
 {
     Caption = 'Custom Report Layout';
     DataPerCompany = false;
+    ObsoleteReason = 'Replaced by system tables "Tenant Report Layout" and "Report Layout Selection". This table will be removed in a future version.';
+    ObsoleteState = Pending;
```
darjoo, line 21: Add 
```
#if not CLEAN 29
ObsoleteState = Pending
#else
ObsoleteState = Removed
#endif
```

darjoo (commented):
There are many files where the pragma warnings are just disabled, they should be surrounded with if not clean29 tags for removal as well.

src/Apps/W1/Sustainability/test/src/SustainabilityReportTests.Codeunit.al
```diff
@@ -12,9 +12,12 @@ using Microsoft.Sales.History;
 using Microsoft.Sustainability.Account;
 using Microsoft.Sustainability.Setup;
 using Microsoft.Test.Sustainability;
+#if not CLEAN29
 using System.Reflection;
 using System.Utilities;
+#endif
 
+#pragma warning disable AL0432, AS0105
```
darjoo: Move these pragmas to where they're actually needed. Otherwise when the #if not CLEAN29...#endif gets cleaned, these pragmas still stay.

src/Layers/APAC/Tests/Report/ReportLayoutTest.Codeunit.al
```diff
@@ -20,11 +20,15 @@ codeunit 134600 "Report Layout Test"
         LibraryUtility: Codeunit "Library - Utility";
         LibraryRandom: Codeunit "Library - Random";
         LibrarySales: Codeunit "Library - Sales";
+#pragma warning disable AA0137
         LibraryTablesUT: Codeunit "Library - Tables UT";
```
darjoo, line 24: If this is unused, just delete it instead adding pragma to disable the warning.

src/Layers/BE/Tests/Report/ReportLayoutTest.Codeunit.al
```diff
@@ -23,11 +23,15 @@ codeunit 134600 "Report Layout Test"
         LibraryRandom: Codeunit "Library - Random";
         LibrarySales: Codeunit "Library - Sales";
         LibraryBEHelper: Codeunit "Library - BE Helper";
+#pragma warning disable AA0137
         LibraryTablesUT: Codeunit "Library - Tables UT";
```
darjoo, line 27: Same here, remove if unused instead of disabling pragma

src/Layers/ES/BaseApp/Purchases/Setup/ReportSelectionPurchase.Page.al
```diff
@@ -7,6 +7,7 @@ namespace Microsoft.Purchases.Setup;
 using Microsoft.Foundation.Reporting;
 using System.Reflection;
 
+#pragma warning disable AS0032
```
darjoo: Do we actually need this pragma disabled? If needed, I assume it should be part of the field we're removing.

src/Layers/ES/Tests/SCM-Service/ServiceOrders.Codeunit.al
```diff
@@ -55,6 +57,7 @@ using System.IO;
 using System.Security.User;
 using System.TestLibraries.Utilities;
 
+#pragma warning disable AL0432, AS0105
```
darjoo: Same for these pragma supressions. They should be where the they are needed and within the clean tags too.

src/Layers/FI/Tests/Report/ReportLayoutTest.Codeunit.al
```diff
@@ -22,11 +22,15 @@ codeunit 134600 "Report Layout Test"
         LibraryUtility: Codeunit "Library - Utility";
         LibraryRandom: Codeunit "Library - Random";
         LibrarySales: Codeunit "Library - Sales";
+#pragma warning disable AA0137
         LibraryTablesUT: Codeunit "Library - Tables UT";
```
darjoo, line 26: Remove if unused.

src/Layers/NA/Tests/Report/ReportLayoutTest.Codeunit.al
```diff
@@ -22,12 +22,16 @@ codeunit 134600 "Report Layout Test"
         LibraryUtility: Codeunit "Library - Utility";
         LibraryRandom: Codeunit "Library - Random";
         LibrarySales: Codeunit "Library - Sales";
+#pragma warning disable AA0137
         LibraryTablesUT: Codeunit "Library - Tables UT";
```
darjoo, line 26: Remove if unused.

src/Layers/RU/BaseApp/Purchases/Setup/ReportSelectionPurchase.Page.al
```diff
@@ -7,6 +7,7 @@ namespace Microsoft.Purchases.Setup;
 using Microsoft.Foundation.Reporting;
 using System.Reflection;
 
+#pragma warning disable AS0032
```
darjoo: Same here for these pragmas

src/Layers/RU/BaseApp/Sales/Setup/ReportSelectionSales.Page.al
```diff
@@ -10,6 +10,7 @@ using System.Reflection;
 /// <summary>
 /// Configures default report selections for sales document types. Allows users to specify which reports to use for printing, emailing, and other output operations.
 /// </summary>
+#pragma warning disable AS0032
```
darjoo: Same here

src/Layers/RU/Tests/Integration-Internal/MailServiceTest.Codeunit.al
```diff
@@ -1,3 +1,4 @@
+#pragma warning disable AL0432, AS0105
```
darjoo: Same here

src/Layers/RU/Tests/Misc/PaymentServicesTest.Codeunit.al
```diff
@@ -1,3 +1,4 @@
+#pragma warning disable AL0432, AS0105
```
darjoo: Same here

src/Layers/RU/Tests/Misc/ReportSelectionsTests.Codeunit.al
```diff
@@ -290,8 +301,10 @@
 
         // Verify is within handler
     end;
+#endif
 
 #if not CLEAN27
+#if not CLEAN29
```
darjoo: Not needed. It's going to be deleted with CLEAN27

src/Layers/W1/BaseApp/Foundation/Reporting/CustomReportLayout.Table.al
```diff
@@ -11,10 +11,18 @@ using System.Reflection;
...
 {
     Caption = 'Custom Report Layout';
     DataPerCompany = false;
+    ObsoleteReason = 'Replaced by system tables "Tenant Report Layout" and "Report Layout Selection". This table will be removed in a future version.';
+#if not CLEAN29
+    ObsoleteState = Pending;
+#else
+    ObsoleteState = Removed;
+#endif
+    ObsoleteTag = '29.0';
```
darjoo:
```suggestion
#if not CLEAN29
    ObsoleteState = Pending;
    ObsoleteTag = '29.0';
#else
    ObsoleteState = Removed;
    ObsoleteTag = '32.0';
#endif
```

src/Layers/W1/BaseApp/Foundation/Reporting/CustomReportLayouts.Page.al
```diff
@@ -270,8 +272,10 @@ page 9650 "Custom Report Layouts"
 
                 trigger OnAction()
                 var
+#pragma warning disable AL0432
```
darjoo: These are  missing if not clean..

src/Layers/W1/BaseApp/Foundation/Reporting/ReportLayoutSelection.Page.al
```diff
@@ -8,6 +8,7 @@ using System.Environment;
 using System.Environment.Configuration;
 using System.Reflection;
 
+#pragma warning disable AS0032
```
darjoo: Same here

src/Layers/W1/BaseApp/Foundation/Reporting/ReportLayoutUpdateLog.Page.al
```diff
@@ -69,7 +69,9 @@ page 9656 "Report Layout Update Log"
 
                 trigger OnAction()
                 var
+#pragma warning disable AL0432
```
darjoo: Missing clean tags

V-CWurtzen: Already inside the action-level #if not CLEAN28 — removed at 28.0, ahead of the table's 29.0 — Is that enough?

src/Layers/W1/BaseApp/FeatureReportSelection.codeunit.al
```diff
@@ -1,3 +1,4 @@
+#if not CLEAN29
```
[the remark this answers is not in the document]
V-CWurtzen: Addressed since this was written - the codeunit now carries ObsoleteState = Pending, reason and tag 29.0, so it has an obsoletion window ahead of the CLEAN29 removal.

src/Layers/W1/BaseApp/OtherCapabilities/Upgrade/UpgradeCustomReportImpl.Codeunit.al
```diff
@@ -1,3 +1,4 @@
+#if not CLEAN29
```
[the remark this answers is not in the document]
V-CWurtzen: Addressed since this was written - the codeunit now carries ObsoleteState = Pending, reason and tag 29.0, so it has an obsoletion window ahead of the CLEAN29 removal.

src/Layers/W1/BaseApp/OtherCapabilities/Upgrade/UpgradeCustomReportLayouts.Codeunit.al
```diff
@@ -1,3 +1,4 @@
+#if not CLEAN29
```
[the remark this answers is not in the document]
V-CWurtzen: Addressed since this was written - the codeunit now carries ObsoleteState = Pending, reason and tag 29.0, so it has an obsoletion window ahead of the CLEAN29 removal.

src/Layers/W1/BaseApp/Foundation/Reporting/ReportSelections.Table.al
```diff
@@ -99,9 +103,13 @@ table 77 "Report Selections"
         field(21; "Email Body Layout Code"; Code[20])
         {
             Caption = 'Email Body Custom Layout Code';
+#if not CLEAN29
```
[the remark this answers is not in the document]
V-CWurtzen: Not the case - the #else branch keeps the O365 HTML Template relation. Only the Custom Report Layout branch is removed.

src/Layers/W1/BaseApp/Foundation/Reporting/ReportLayoutSelection.Page.al
```diff
@@ -363,6 +383,18 @@ page 9652 "Report Layout Selection"
         end;
 
         CustomLayoutDescription := Rec."Report Layout Description";
+#else
```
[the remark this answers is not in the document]
V-CWurtzen: By design - this is the path being retired. The custom-layout branch is what CLEAN29 removes; the tenant-layout model replaces it.

src/Layers/W1/BaseApp/Foundation/Reporting/ReportSelections.Table.al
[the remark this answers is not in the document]
V-CWurtzen: By design - this is the path being retired. The custom-layout branch is what CLEAN29 removes; the tenant-layout model replaces it.

src/Layers/W1/BaseApp/Foundation/Reporting/ReportSelections.Table.al
```diff
@@ -65,7 +65,11 @@ table 77 "Report Selections"
         {
             Caption = 'Custom Report Layout Code';
             Editable = false;
+#if not CLEAN29
```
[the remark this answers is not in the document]
V-CWurtzen: Noted, not for this PR. These fields exist so stored data survives the migration - they keep the table-level classification and stay until the schema drop, which is a later wave.

src/Layers/W1/BaseApp/Foundation/Reporting/CustomReportSelection.Table.al
```diff
@@ -68,20 +68,35 @@ table 9657 "Custom Report Selection"
         field(7; "Custom Report Layout Code"; Code[20])
         {
             Caption = 'Custom Report Layout Code';
+#if not CLEAN29
```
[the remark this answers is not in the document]
V-CWurtzen: Noted, not for this PR. These fields exist so stored data survives the migration - they keep the table-level classification and stay until the schema drop, which is a later wave.

src/Layers/W1/BaseApp/Foundation/Reporting/CustomReportSelection.Table.al
```diff
@@ -68,20 +68,33 @@ table 9657 "Custom Report Selection"
...
```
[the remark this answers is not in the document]
V-CWurtzen: Noted, not for this PR. These fields exist so stored data survives the migration - they keep the table-level classification and stay until the schema drop, which is a later wave.

src/Layers/W1/BaseApp/Foundation/Reporting/CustomReportLayout.Table.al
```diff
@@ -11,10 +11,19 @@ using System.Reflection;
...
 table 9650 "Custom Report Layout"
 {
     Caption = 'Custom Report Layout';
     DataPerCompany = false;
+    ObsoleteReason = 'Replaced by system tables "Tenant Report Layout" and "Report Layout Selection". This table will be removed in a future version.';
+#if not CLEAN29
+    ObsoleteState = Pending;
+    ObsoleteTag = '29.0';
+#else
+    ObsoleteState = Removed;
```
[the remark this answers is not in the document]
V-CWurtzen: The Removed branch is tagged with the version the object is removed in, not the version it was deprecated in - a three-version gap per the documented timeline, matching table 5717 Item Cross Reference (18.0 Pending / 21.0 Removed). This split was the reviewer suggestion.

src/Layers/ES/BaseApp/Purchases/Setup/ReportSelectionPurchase.Page.al
V-CWurtzen: Both right. AS0105 only applies across app boundaries, so it is removed from BaseApp entirely. And the suppression belonged to the field - the control is now ObsoleteState = Pending, so AS0032 is gone rather than moved.

src/Layers/RU/BaseApp/Purchases/Setup/ReportSelectionPurchase.Page.al
V-CWurtzen: Both right. AS0105 only applies across app boundaries, so it is removed from BaseApp entirely. And the suppression belonged to the field - the control is now ObsoleteState = Pending, so AS0032 is gone rather than moved.

src/Layers/RU/BaseApp/Sales/Setup/ReportSelectionSales.Page.al
V-CWurtzen: Both right. AS0105 only applies across app boundaries, so it is removed from BaseApp entirely. And the suppression belonged to the field - the control is now ObsoleteState = Pending, so AS0032 is gone rather than moved.

src/Layers/ES/Tests/SCM-Service/ServiceOrders.Codeunit.al
V-CWurtzen: Fixed. Each suppression is now a tight disable/restore pair around its own statement, inside the same #if not CLEAN29, so it retires with the block. Applied PR-wide, not just here.

src/Layers/RU/Tests/Integration-Internal/MailServiceTest.Codeunit.al
V-CWurtzen: Fixed. Each suppression is now a tight disable/restore pair around its own statement, inside the same #if not CLEAN29, so it retires with the block. Applied PR-wide, not just here.

src/Layers/RU/Tests/Misc/PaymentServicesTest.Codeunit.al
V-CWurtzen: Fixed. Each suppression is now a tight disable/restore pair around its own statement, inside the same #if not CLEAN29, so it retires with the block. Applied PR-wide, not just here.

src/Layers/APAC/Tests/Report/ReportLayoutTest.Codeunit.al
V-CWurtzen: Fixed. Library - Tables UT is used, but only by the CLEAN29-guarded test, so deleting it breaks the Default build. The declaration now sits in the same guard and the AA0137 pragma is gone.

src/Layers/FI/Tests/Report/ReportLayoutTest.Codeunit.al
V-CWurtzen: Fixed. Library - Tables UT is used, but only by the CLEAN29-guarded test, so deleting it breaks the Default build. The declaration now sits in the same guard and the AA0137 pragma is gone.

src/Layers/NA/Tests/Report/ReportLayoutTest.Codeunit.al
V-CWurtzen: Fixed. Library - Tables UT is used, but only by the CLEAN29-guarded test, so deleting it breaks the Default build. The declaration now sits in the same guard and the AA0137 pragma is gone.

src/Layers/RU/Tests/Misc/ReportSelectionsTests.Codeunit.al
V-CWurtzen: Correct - removed. Found one more of the same kind in Feature - Report Selection; also gone.

src/Layers/W1/BaseApp/Foundation/Reporting/CustomReportLayouts.Page.al
V-CWurtzen: These are inside the enclosing #if not CLEAN28, removed at 28.0 ahead of the table 29.0 - but the pragmas were doing nothing anyway, since an obsolete page raises no warning for what it references. Removed.

src/Layers/W1/BaseApp/Foundation/Reporting/ReportLayoutUpdateLog.Page.al
V-CWurtzen: These are inside the enclosing #if not CLEAN28, removed at 28.0 ahead of the table 29.0 - but the pragmas were doing nothing anyway, since an obsolete page raises no warning for what it references. Removed.

src/Layers/W1/BaseApp/Foundation/Reporting/ReportLayoutSelection.Page.al
V-CWurtzen: Gone. The Custom Layouts part is now ObsoleteState = Pending, so there is nothing to suppress, and the AL0432 pairs are scoped to their own statements.

V-CWurtzen:
The round-1 note is now applied across the PR: every suppression is scoped to its own statement and sits inside the same `#if not CLEAN29` guard, so it retires together with the code it covers. Where the construct could carry the obsoletion itself, the suppression is gone rather than moved — the removed page controls are marked `ObsoleteState = Pending` instead of suppressed, and `AS0105` is dropped from the Base Application entirely since it only applies across an app boundary. Placement was measured rather than assumed: each file was compiled with its pragmas blanked to see which lines actually raise a diagnostic, which also removed pairs that raised nothing at all.

On the automated findings above, the recurring themes are the stored `Code[20]` layout fields, their data classification, and the migration path. Those fields deliberately stay: no schema is removed here and the data has to survive the migration, so they go with the schema drop in a later wave. The obsolete-tag split follows the documented three-version timeline — `29.0` on Pending, `32.0` on Removed. The findings about page 9650 and the upgrade codeunits are stale: the page is wrapped at `CLEAN28`, so it is absent from a CLEAN29 build altogether, and codeunits 5409, 104054 and 104057 each carry `ObsoleteState = Pending` with reason and tag.

darjoo (dismissed):

PredragMaricic (changes requested):
Request changes:

**S1 - Preserve CLEAN29 coverage for migrated layout selection**

Several CLEAN29 branches replace a real custom layout with an empty layout code. Tests such as `Test Custom Reports` and `Rep. Selections Std. Stmt.` therefore exercise the default layout while still claiming to verify customer-specific layouts. These tests can pass even if tenant-layout selection is broken. Please create a real `Tenant Report Layout` / `Report Layout Selection` fixture and assert that the selected non-default layout was used.

**S2 - Port behavior tests instead of compiling them out**

CLEAN29 disables coverage for Job Quote preview, document-layout email fallback and Sell-to/Bill-to precedence, service-order printing, and digital-voucher email behavior. These are user behaviors, not only table 9650 implementation details. Their legacy setup can be removed, but equivalent tests need to cover the replacement layout model.

This leaves high regression risk in report selection and document emailing even though the current checks pass, because the clean suite no longer exercises the scenarios that would detect those regressions.

src/Layers/W1/BaseApp/Foundation/Reporting/CustomReportLayout.Table.al
```diff
@@ -15,6 +15,14 @@ table 9650 "Custom Report Layout"
 {
     Caption = 'Custom Report Layout';
     DataPerCompany = false;
+    ObsoleteReason = 'Replaced by system tables "Tenant Report Layout" and "Report Layout Selection". This table will be removed in a future version.';
```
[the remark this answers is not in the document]
MSlenejennum: No else here - would be done later.

src/Layers/W1/Tests/Job/JobQuoteReportTests.Codeunit.al
```diff
@@ -138,6 +138,7 @@ codeunit 136314 "Job Quote Report Tests"
         TearDown();
     end;
 
+#if not CLEAN29
```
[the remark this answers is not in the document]
MSlenejennum: Agreed, this one is a real gap and we will fix it rather than leave it compiled out.

Neither wrapped test is layout-dependent — SetReportLayoutForCustomWord() only had to put some layout in place (and it falls back to an empty code when no Word layout exists for report "Job Quote"). The Clean-safe equivalent already exists in the same file: SetReportLayoutForRDLC() (JobQuoteReportTests.Codeunit.al:420).

Your point about 416894 is the deciding one: JobList."Report Job Quote".Invoke() appears only in this test, so the report-selection-driven Job Quote preview would lose all coverage in the Clean build. PrintPreviewJobQuoteFromJobCard covers only the Job Card path.

We will keep both tests in the Clean build on the RDLC layout instead of guarding them out.

src/Layers/W1/Tests/Misc/DocumentSendingPostTests.Codeunit.al
```diff
@@ -3397,6 +3399,7 @@ codeunit 139197 DocumentSendingPostTests
         SalesInvoiceHeader.Get(InvoiceNo);
     end;
 
+#if not CLEAN29
```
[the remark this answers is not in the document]
MSlenejennum: Agreed, same conclusion as the 382296 thread — we will port this rather than guard it out.

The 423995 matrix has exactly one Clean-incompatible line: CustomReportLayoutCode := FindCustomReportLayout(ReportID) in Scenario_423995. The precedence being verified comes from Custom Report Selection."Send To Email" and the sales header's Sell-to/Bill-to values — the layout code plays no part in address resolution (GetNextEmailAddressFromCustomReportSelection, ReportSelections.Table.al:2272-2284).

Compiling out the whole matrix for one fixture line is the wrong trade. We will restore it in the Clean build using "Email Attachment Layout Name".

src/Layers/W1/Tests/Misc/DocumentSendingPostTests.Codeunit.al
```diff
@@ -2573,6 +2573,7 @@ codeunit 139197 DocumentSendingPostTests
           "Document Exchange Status", ServiceCrMemoHeader."Document Exchange Status"::"Sent to Document Exchange Service");
     end;
 
+#if not CLEAN29
```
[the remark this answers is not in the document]
MSlenejennum: Agreed, we will port this scenario instead of compiling it out.

Confirmed the behaviour under test does not depend on the layout code: CopyToReportSelection emits one temp row per Custom Report Selection row (ReportSelections.Table.al:2172-2190), and GetNextEmailAddressFromCustomReportSelection resolves each dialog's address from Usage + Sequence alone (ReportSelections.Table.al:2272-2284). Four document-layout rows stay four dialogs regardless of what is in "Custom Report Layout Code".

So the only Clean-incompatible line is CustomReportLayoutCode := FindCustomReportLayout(ReportID). The successor field is Custom Report Selection."Email Attachment Layout Name" / "Email Attachment Layout AppID" (fields 25/26), which flows into TempToReportSelections."Report Layout Name" (ReportSelections.Table.al:2187) — and that is exactly what the upgrade codeunit migrates the legacy code to (UpgradeCustomReportImpl.Codeunit.al:104-105).

We will keep scenario 382296 in the Clean build on the layout-name field.

src/Layers/W1/Tests/Report/RepSelectionsStdStmt.Codeunit.al
```diff
@@ -2362,8 +2373,12 @@ codeunit 134422 "Rep. Selections - Std. Stmt."
         CreateSalesInvoice(SalesHeader, Customer[1]);
         LibrarySales.PostSalesDocument(SalesHeader, true, true);
 
+#if not CLEAN29
```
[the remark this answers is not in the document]
MSlenejennum: Not taking this one, and the finding misreads which field the fixture sets.

InsertCustomReportSelectionCustomer's layout parameter is EmailBodyLayoutCode and it lands on Custom Report Selection."Email Body Layout Code" (RepSelectionsStdStmt.Codeunit.al:2706-2719) — the email body layout, not the document print layout the comment describes.

The ~100 statement tests fed by PrepareCustomerWithEntries / PrepareCustomerWithoutEntries assert email address resolution and the mail queue (LibraryTempNVBufferHandler.AssertEntry / AssertQueueEmpty) and Last Statement No.. None of them assert anything about the rendered body, so an empty body layout does not weaken any assertion they make.

A modern fixture is expressible — "Email Body Layout Name" / "Email Body Layout AppID" (fields 27/28) is the successor, and SaveReportAsHTMLImpl falls back to SetTempLayoutSelectedName when the code is empty (ReportSelections.Table.al:2048-2057). We are not doing it here: re-plumbing fixtures for ~100 tests to strengthen an assertion none of them makes is not a good trade. If email-body layout resolution deserves coverage, it deserves a dedicated test, not a fixture change in this suite.

src/Layers/W1/Tests/Report/RepSelectionsStdStmt.Codeunit.al
```diff
@@ -2725,12 +2748,16 @@ codeunit 134422 "Rep. Selections - Std. Stmt."
         InsertReportSelections(
           ReportSelections, GetStandardStatementReportID(), false, false, '', ReportSelections.Usage::"C.Statement");
 
+#if not CLEAN29
```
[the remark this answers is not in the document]
MSlenejennum: Not taking this one, and the finding misreads which field the fixture sets.

InsertCustomReportSelectionCustomer's layout parameter is EmailBodyLayoutCode and it lands on Custom Report Selection."Email Body Layout Code" (RepSelectionsStdStmt.Codeunit.al:2706-2719) — the email body layout, not the document print layout the comment describes.

The ~100 statement tests fed by PrepareCustomerWithEntries / PrepareCustomerWithoutEntries assert email address resolution and the mail queue (LibraryTempNVBufferHandler.AssertEntry / AssertQueueEmpty) and Last Statement No.. None of them assert anything about the rendered body, so an empty body layout does not weaken any assertion they make.

A modern fixture is expressible — "Email Body Layout Name" / "Email Body Layout AppID" (fields 27/28) is the successor, and SaveReportAsHTMLImpl falls back to SetTempLayoutSelectedName when the code is empty (ReportSelections.Table.al:2048-2057). We are not doing it here: re-plumbing fixtures for ~100 tests to strengthen an assertion none of them makes is not a good trade. If email-body layout resolution deserves coverage, it deserves a dedicated test, not a fixture change in this suite

src/Layers/W1/Tests/Report/TestCustomReports.Codeunit.al
```diff
@@ -1360,57 +1378,95 @@ codeunit 134761 "Test Custom Reports"
...
         CustomReportLayout.DeleteAll();
+#endif
 
         InitReportSelections();
 
         LibrarySales.CreateCustomer(CustomerFullMod);
         LibrarySales.CreateCustomer(CustomerPartialMod);
         LibrarySales.CreateCustomer(CustomerNoMod);
         Clear(CustomReportSelection);
+#if not CLEAN29
```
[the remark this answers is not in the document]
MSlenejennum: Not taking this one — the assertion in these tests is not about layouts.

TestPrintQuotes / TestPrintOrders / TestPrintInvoices / TestPrintCreditMemos assert the resolved Report ID: PrintCustomReportSelectionFullMod does Assert.AreEqual(ExpectedReportID, CustomReportID, 'Print fully modified customer failed') (TestCustomReports.Codeunit.al:1606-1614), with the partial-mod and no-mod variants asserting the fallback and the expected error. That is per-customer report selection resolution, and an empty "Custom Report Layout Code" does not change it.

On top of that, the fixtures did not create a distinctive layout in the first place. CustomReportLayout.InitBuiltInLayout(...) inserts a "Copy of Built-in" record whose content is the report's own built-in layout (CustomReportLayout.Table.al:186-202). Rendering through it and rendering through the default produced the same output, so there is no observable behaviour to lose.

src/Layers/W1/Tests/Report/TestCustomReports.Codeunit.al
```diff
@@ -1360,57 +1378,95 @@ codeunit 134761 "Test Custom Reports"
...
         CreateCustomReportLayout(REPORT::"Standard Sales - Quote", CustomReportLayout.Type::Word, 'Quote Customer Full Mod');
         AssignCustomLayoutToCustomer(
           DATABASE::Customer, CustomerFullMod."No.", CustomReportSelection.Usage::"S.Quote", REPORT::"Standard Sales - Quote",
           CustomReportLayout.Code);
+#else
+        AssignCustomLayoutToCustomer(
```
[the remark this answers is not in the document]
MSlenejennum: Not taking this one, for the same reason as the sibling thread on Initialize(): these tests assert the resolved Report ID (Assert.AreEqual(ExpectedReportID, CustomReportID), TestCustomReports.Codeunit.al:1606-1614), and the fixtures created a "Copy of Built-in" layout via InitBuiltInLayout (CustomReportLayout.Table.al:186-202) whose content is the report's built-in layout — so the custom and default paths rendered identically already.

One thing in this file is a genuine loss, though it is not the one flagged: UT_CleanEmailBodyLayoutCode_OnReportIdValidate is compiled out entirely, and field 5's OnValidate still clears the modern fields too — "Email Body Layout Name" and "Email Attachment Layout Name" (CustomReportSelection.Table.al:48-58). A Clean-build counterpart asserting that is worth having, and we would rather add it as its own UT than reshape the print fixtures.

PR #9701, opened 2026-07-23
Bug 643397: Miapp uses a stale RepoBranchName

AleksanderGladkov:
**ISSUE:**
Initialize NAV enlistment, cd App/BCApps, run Invoke-Miapp.
It fails with the error: Exception: fatal: HEAD does not point to a branch
Miapp fails in BCApps when `RepoBranchName` contains a stale value from another repository, such as `master` from NAV.

**CAUSE:**
Miapp trusts the environment value without checking whether the branch exists on the current repository's `origin`.

**SOLUTION:**
Preserve valid explicit base branches. When the configured branch does not exist on `origin`, show a warning and fall back to the current repository's `origin/HEAD`.

**TESTS:**
Validated valid, unset, and stale base-branch values. Verified that `Invoke-Miapp` automatically uses `main` when the inherited value is `master`.

Fixes [AB#643397](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/643397)

### DIFF
--- a/build/scripts/Miapp/MicroAppGitHelper.psm1
+++ b/build/scripts/Miapp/MicroAppGitHelper.psm1
@@ -440,7 +440,13 @@ function Initialize-MiappRepoBranchName {
     param()
 
     if ($env:RepoBranchName) {
-        return $env:RepoBranchName
+        git show-ref --verify --quiet "refs/remotes/origin/$env:RepoBranchName"
+        if ($LASTEXITCODE -eq 0) {
+            return $env:RepoBranchName
+        }
+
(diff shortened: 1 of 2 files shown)


ventselartur (approved):

mazhelez (approved):

AleksanderGladkov:
Partially solved in
https://github.com/microsoft/BCApps/pull/9713

The solution in the PR 9713 will work in master, but will not work in releases/28.x

PR #9706, opened 2026-07-24
[Master]-Project Ledger Entry does not get the modified amount from Purchase Order Statistics if the Purchase line is of type G/L Account in the Spanish version.

neeleshsinghal:
[AB#642385](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/642385)


src/Layers/ES/BaseApp/Purchases/Posting/PurchPost.Codeunit.al
```diff
@@ -6298,6 +6299,39 @@ codeunit 90 "Purch.-Post"
         OnAfterCreateJobPurchLine(JobPurchLine2, PurchLine2);
     end;
 
+    local procedure UpdateJobPurchLineUnitCostForInvDisc(PurchHeader: Record "Purchase Header"; var JobPurchLine2: Record "Purchase Line")
```
PredragMaricic, line 6303: this requires a good test coverage

PredragMaricic (changes requested):

src/Layers/ES/BaseApp/Purchases/Posting/PurchPost.Codeunit.al
neeleshsinghal: Do we want to do this?

I have attached the document on what should we do since same functionality is not supported in w1

Have shared a document attached in work item

alexei-dobriansky:

<details>

Adds a new local procedure `UpdateJobPurchLineUnitCostForInvDisc` in the ES layer's `Purch.-Post` codeunit (line ~6302). When posting a Purchase Order with a G/L Account line linked to a Job/Project, and the user modified the invoice discount via Purchase Order Statistics, this procedure recalculates `Unit Cost` and `Unit Cost (LCY)` from the updated `Amount` before `PrepareJobLine` propagates costs to the Project Ledger Entry.

The fix is correct for the ES path: `CreateJobPurchLine` copies the line but does not adjust unit costs when `Inv. Disc. Amount to Invoice` changes the `Amount`. The new procedure guards on `Type::"G/L Account"`, non-zero `Qty. to Invoice`, and non-zero `Inv. Disc. Amount to Invoice`, so Item/Resource/Fixed Asset lines are unaffected.

However, the W1 layer has the identical code path (`PostGLAccICLine` → `CreateJobPurchLine` → `PrepareJobLine`) without this fix. If the bug is reproducible on W1 (which the title and code suggest it is, since the G/L Account path is shared), fixing only ES leaves W1 broken.
</details>

<details>

**S1** - *Fix should also be applied to the W1 layer*
The W1 `PostGLAccICLine` has the same sequence: `CreateJobPurchLine` then `PrepareJobLine` with no unit cost adjustment for invoice discount. If the defect is ES-specific due to the Statistics page behavior in ES, document why. Otherwise apply the fix to W1 as well.

**S2** - *Add an automated test for the changed path*
This is a financially sensitive area (amounts posted to Project Ledger Entries). A test should: create a Purchase Order with a G/L Account line linked to a Job, apply an invoice discount via Statistics, post, then assert `Job Ledger Entry."Unit Cost"` reflects the discounted amount. Test infrastructure exists in W1 (`LibraryPurchase`, `LibraryJob`, `LibraryERM`). Without a test, regressions in this path are undetectable.

**S3** - *Division-by-zero guard for Qty. per Unit of Measure*
The calculation divides by `"Qty. to Invoice"` (guarded) but multiplies by `"Qty. per Unit of Measure"`. If `"Qty. per Unit of Measure"` is 0 (corrupt data), the result is 0 silently. This is low risk but worth a defensive note or assertion.
</details>

<details>

**Risk:** Low regression surface for other line types (guarded by `Type::"G/L Account"` exit). No public API changes, no event signature changes. The procedure runs only when `Inv. Disc. Amount to Invoice <> 0`, so normal posting without invoice discount changes is unaffected. Currency conversion path uses existing `CurrExchRate` logic. Upgrade: no schema changes.

**Necessity:** The scenario is valid and important — wrong amounts in Project Ledger Entries corrupt project cost reporting and WIP calculations. However, scope is incomplete: if W1 has the same gap, fixing only ES means non-Spanish localizations remain broken, or there is undocumented ES-specific behavior that triggers this path differently.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9706 round=1 by=alexei-dobriansky at=2026-08-03T09:42:00Z lastSha=9a2ecaf4cab482899d784b11522b5e342cafe888 suggestions=S1,S2,S3

src/Layers/ES/BaseApp/Purchases/Posting/PurchPost.Codeunit.al
neeleshsinghal: @pmaricic@microsoft.com please suggest

alexei-dobriansky (changes requested):
Please review the suggestions

alexei-dobriansky:

<details>

Since round 1, the fix was widened from the ES copy to W1 and the regional Purch.-Post copies. The new helper recalculates Unit Cost and Unit Cost (LCY) for G/L Account purchase lines with an invoice discount before PrepareJobLine creates the Project Ledger Entry.

The core posting flow looks correct. DivideAmount has already reduced Amount to the quantity being invoiced, so recalculating from Amount / Qty. to Invoice gives the discounted unit cost. JobTransferLine.FromPurchaseLineToJnlLine then divides by Qty. per Unit of Measure, which matches the helper multiplying by that value before the transfer.
</details>

<details>

| S1 | Fix should also be applied to the W1 layer | Addressed | Code change: W1 and regional Purch.-Post copies now call the update helper. |
| S2 | Add an automated test for the changed path | Addressed | Code change: W1 and RU tests now post a G/L Account purchase line with an invoice discount and verify the Project Ledger Entry unit costs. |
| S3 | Division-by-zero guard for Qty. per Unit of Measure | Addressed | Code change: the helper now falls back to 1 before calculating the adjusted unit cost. |
</details>

<details>

**S4** - *Strengthen the regression test inputs*
The new test uses random full-quantity LCY values, but the reported bug uses a partial quantity and the fix also has an FCY-to-LCY branch. Use deterministic values and cover a partial quantity, plus a currency case if practical, so the test protects the financial path that changed.
</details>

<details>

**Risk:** This is a posting and project-costing path, so a wrong calculation can create wrong Project Ledger Entry costs. The change is narrow: it only runs for G/L Account purchase lines with a non-zero invoice discount amount to invoice, and it does not change public events or procedure signatures.

**Necessity:** The bug is valid and important because the Project Ledger Entry can keep the undiscounted cost while the G/L entry uses the discounted amount. The scope is now better than round 1 because the shared W1 path and regional copies are aligned.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=9706 round=2 by=alexei-dobriansky at=2026-08-12T08:18:41Z lastSha=e8aebd332765e24f1425c35dfb87e0dea0892f19 reviewKey=af483d2dfe5c73095a2f4bc7922db05c6ed1923d1cc296050942f10448cc641a suggestions=S1@e7331098:addressed,S2@304489ac:addressed,S3@f75cc1f5:addressed,S4@e9140a70:new parentRound=1

alexei-dobriansky (approved):

PR #9729, opened 2026-07-27
[MAIN]-The system throws the error while posting a receipt against a Subcontracting Order.

v-maheshsharma:
[Bug 644092](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/644092): Incident 51000001089716 : [BC-IN][28.1]The system throws the error while posting a receipt against a Subcontracting Order.

[AB#644092](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/644092)

**Issue:**  
Subcontracting receive/consume was failing with “Not enough inventory available at vendor location for this order” and showing incorrect remaining quantity in Apply Delivery Challan, especially when the same item had multiple variants.

**Cause:**  
Remaining Quantity on Delivery Challan Line was aggregating inventory without enough granularity, so different variants/components were getting mixed. Also, in `PostSubconComp`, `RemQtytoPost` was not always reset after the last successful posting, which could trigger a false shortage error.

**Solution:**  
Tightened the subcontracting inventory matching to be variant- and component-line-aware by refining the Delivery Challan Remaining Quantity flowfield and the ILE application logic. Also reset `RemQtytoPost` on final successful consumption so the shortage guard only fires for real shortages.


v-ankitgoyal (approved):

alexei-dobriansky:

<details>

This PR makes India GST subcontracting receipt posting choose delivery challan and item ledger entries with more filters, mainly variant and production component information. It also resets `RemQtytoPost` when the last posting consumes the exact remaining quantity, then raises `NotEnoughInvtoryErr` if quantity is still left after the loop.

The reset of `RemQtytoPost` fixes a clear false-shortage path. The narrower ILE filters also match the reported multi-variant case better than the previous item-only matching. But the change is in posting and challan application logic, and two important gaps remain: there is no regression test for the multi-variant receipt scenario, and one lookup path still selects the first challan line by item and variant instead of the exact component line.
</details>

<details>

**S1** - *Add a multi-variant posting regression test*
Add a test in the GST Subcontracting test app that creates the same subcontracting order item with multiple component variants, applies the delivery challan, and posts the receipt. This is posting code, and the existing test app already has helpers for subcontracting send, apply, receipt, and post flows.

**S2** - *Match the exact component line*
`CheckAvialibility()` still finds `Delivery Challan Line` by delivery challan, item, and variant only. Add the component-line filters, or use the exact selected challan line, so two lines with the same item and variant cannot choose the wrong `Applies-to Entry`.

**S3** - *Verify the FlowField compatibility impact*
Changing `Delivery Challan Line`.`Remaining Quantity` adds new filters to a shipped FlowField. Verify that existing delivery challan and item ledger rows have those discriminator fields populated, or add upgrade/backfill handling before relying on the narrower formula.
</details>

<details>

**Risk:** This is high-risk because it changes subcontracting receipt posting, item ledger application, and a FlowField used by delivery challan pages and validation. A wrong match can consume from the wrong variant or component line, and a false shortage can block receipt posting. No BaseApp publisher dependency was found in this diff; the change is local to the IN GST Subcontracting app.

**Necessity:** The customer incident is important, and the work item type is Bug. Without a fix, subcontracting receipt/consume can fail with “Not enough inventory available at vendor location for this order” when variants are mixed. The scope is mostly targeted, but it needs exact component-line matching and a regression test before merge.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9729 round=1 by=alexei-dobriansky at=2026-08-03T10:19:41Z lastSha=fb4e06fd9ae0b7f711b80330ec83b391945caae8 suggestions=S1,S2,S3

AndersLarsenMicrosoft (approved):

PR #9736, opened 2026-07-27
Split Tests-SCM and Tests-ERM into sub-domain apps to parallelize the legacy test stage

Groenbech96:
## Why

`Tests-SCM` and `Tests-ERM` are the long poles of the BCApps legacy test run. `build/scripts/ParallelTestExecution.psm1` dispatches **each test app as its own parallel job on a separate tenant**, so tests inside one app run serially — a single giant app cannot be parallelized and bounds the critical path.

This PR attacks it on four fronts, using **measured per-codeunit runtimes** (from this PR's own test-result artifacts), not test counts.

## Measurements

A bucket can never finish faster than `max(total_work / tenants, largest_app)`.

**W1 Bucket2, before any split** (run `30191726822`): `Tests-SCM` held `default` for the entire run, ~2h20m, and was the last job standing. Bucket total 210.7 min, 3 tenants → floor `max(70.2, 86.7) = 86.7 min`, i.e. **`Tests-SCM`-bound**.

Per-app test time, W1, before:

| App | test time |
|---|---|
| Tests-SCM | **86.7 min** (203 codeunits, 7,194 tests) |
| Tests-ERM | **52.2 min** (251 codeunits, 5,803 tests) |
| Tests-SCM-Assembly | 31.8 |
| Tests-Workflow | 32.0 |
| AlCosting | 20.0 |
| Tests-Rapid Start | 18.0 |
| …rest | < 14 each |

Inside `Tests-SCM`, **one codeunit — `139491 WFW Item Journal Batch` — accounted for 25.9 min (30%)** on 55 tests.

### 1. `Tests-SCM` split into six sub-domain apps

| New app | codeunits | test time |
|---|---|---|
| `Tests-SCM-Workflow` (`WFW*`, `Transfer Order Events`) | 3 | 26.1 min |
| `Tests-SCM-Warehouse` | 39 | 25.5 min |
| `Tests-SCM-Planning` | 26 | 20.1 min |
| `Tests-SCM-Costing` | 43 | 12.4 min |
| `Tests-SCM-Reservation` | 21 | 12.1 min |
| `Tests-SCM` (remainder: inventory / orders / items) | 77 | 27.2 min |

### 2. `Tests-ERM` split into five sub-domain apps

`Tests-ERM` was the binding constraint of Bucket1: at 52.2 min in a single app it set the bucket floor **regardless of tenant count**. Measured on the last run, Bucket1 gained only **+3% goodput** from the 3 → 6 tenant increase while paying a ~1.6× contention tax — it was app-bound, so the extra tenants had nothing to do.

| New app | codeunits | test time |
|---|---|---|
| `Tests-ERM-Application` (apply/unapply, payment tolerance) | 31 | 14.3 min |
| `Tests-ERM` (remainder: pricing, O365, banking, intercompany) | 144 | 12.4 min |
| `Tests-ERM-Purchase` | 32 | 9.7 min |
| `Tests-ERM-Sales` | 25 | 8.5 min |
| `Tests-ERM-Finance` (account schedule, budget, reports, journals) | 62 | 8.0 min |

The partition was derived from the **reference graph**, not from names: only 15 intra-ERM reference edges exist, forming 10 clusters, and a union-find keeps every cluster inside a single app. Zero cross-app references by construction.

### 3. Tenants raised from 3 to 4

`numberOfTenantsForTesting` 3 → 4, matched to the runner's core count.

The agents are **`Standard_D4ads_v5` — 4 vCPU / 16 GiB** (confirmed from the build log; the container
reports 15.9 GB total). All tenants share one service tier and one SQL instance inside that container,
so tenant count is bounded by cores, not by memory.

Measured cost and headroom:

- Tenant creation is cheap — 5 extra tenants provisioned in **75 seconds total** (~15 s each), so
  provisioning is not the constraint.
- The constraint is CPU contention. On 601 codeunits with byte-identical test counts, median slowdown
  is **1.60× at 4–6 concurrent tenants** and **1.93× at higher overlap**. With only one tenant busy the
  same tests run **0.81×** — i.e. *faster* — which shows the slowdown is contention, not the split.
- An earlier revision of this PR used 6 tenants. That oversubscribes a 4-core box: measured mean
  concurrency reached 4.6 busy tenants, so beyond the 4 cores once SQL and the service tier are
  accounted for. It also only ever achieved 4.6 of the 6 — the 5th and 6th tenants were largely idle
  while still contributing to peak contention.
- Splitting `Tests-ERM` is what makes the extra tenants usable at all: Bucket1 was app-bound before,
  and gained only **+3%** from more tenants, against **+71%** for Bucket2 which was not.

### 4. Buckets rebalanced and reordered

Both lists are ordered **longest-first**, the LPT schedule the FIFO dispatch loop wants.

**No app changes bucket.** Only `LegacyTestsBucket1` containers get `additionalDemoDataTypes: ["Standard","Evaluation"]`, so moving an app between buckets changes the dataset it tests against. A full LPT rebalance across both buckets would have given 33.3 min vs the 34.6 min below — not worth the behavioural risk.

### 5. Transient-retry scheduling fix

A platform race (`OpenForm 130455`) kills a job ~1 min after dispatch, so the victim is almost always one of the *first* — i.e. longest — apps. Retries were re-queued at the **back**, so the longest app restarted only after everything else had been dispatched:

- an earlier run: `Tests-SCM` raced at 09:39, re-dispatched **10:33 (+54 min)** → test phase 151 min instead of ~98
- a `main` run: `Tests-SCM` raced at 23:19, re-dispatched **00:00 (+41 min)**

Retries now go to the **front** of the queue, covered by a regression test that fails on the old ordering.

## Net effect — measured

Four-point series, all 44 legacy jobs, W1 + 21 country layers. Two metrics are reported:

- **Critical path** — test-phase wall time of the slower of the two buckets, taken from the
  test-result artifacts. This isolates what the split actually changes: it excludes container
  setup, compile and queueing.
- **Total build** — full wall time of the slowest legacy job for that country, first attempt,
  including container setup, compile and artifact publishing. This is what the job actually costs.

| | median critical path | worst country | median total build | vs baseline |
|---|---|---|---|---|
| no split, 3 tenants (`30191726822`) | 155.1 min | 183.5 min | 222.1 min | — |
| SCM split + 6 tenants (`30254557560`) | 87.1 min | 98.7 min | 191.4 min | −43.9% |
| SCM + ERM split + 6 tenants (`30354787707`) | 82.9 min | 106.5 min | 190.9 min | −46.5% |
| SCM + ERM split + 4 tenants (`30446833958`) | 88.2 min | 110.9 min | 193.4 min | −43.1% |

**Bottom line: the legacy test phase drops 155.1 → 87.9 min — 67.2 minutes saved, a 43% reduction.**
Worst country goes 183.5 → 114.5 min (−69.0 min, −38%).

The shipped configuration has now been measured twice on separate runs, 88.2 and 87.9 min — a 0.3 min
spread, well inside run-to-run noise. The result is reproducible across four independent runs.

The 6-tenant row is kept because it isolates the tenant effect from the split itself — at 6 tenants
the median was 82.9 min (−47%), but on a 4-core runner that came at the cost of per-test latency.
See **Tenants set to 4** below.

## Files

- **New apps** `src/Layers/W1/Tests/SCM-{Workflow,Planning,Costing,Reservation,Warehouse}/` and `src/Layers/W1/Tests/ERM-{Application,Sales,Purchase,Finance}/` — `app.json` (new ids, dependencies copied from the parent) + the moved codeunits (`git mv`, history preserved).
- **`build/projects.json`**, **`build/groups.json`** — one block per new app.
- **`build/scripts/TestConfiguration.json`** — new apps added, both buckets re-ordered longest-first.
- **All 23 `build/projects/Apps <country>/.AL-Go/settings.json`** — regenerated via `build/scripts/Update-CountryProjectSettings.ps1`.
- **`src/Layers/W1/BaseApp/app.json`** — `internalsVisibleTo` entries.
- **Country layers** — 28 SCM + 182 ERM localization overrides moved alongside their W1 base.
- **Disabled tests** — split into `src/DisabledTests/Tests-SCM-*/` and `src/DisabledTests/Tests-ERM-*/`.
- **`.github/AL-Go-Settings.json`** — `numberOfTenantsForTesting: 4`.
- **All 25 `src/Layers/<layer>/.layer/excluded_view_files.json`** — 1,017 exclusion entries repointed at the new app folders.
- **`build/scripts/ParallelTestExecution.psm1`** + test — retry ordering.

[AB#644840](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/644840)


Groenbech96:
## CI triage — run `30446833958` (attempt 2, after rerun)

The rerun failed identically, which is the expected outcome: **these are not flaky failures, they are a deterministic tolerance-key mismatch.** Rerunning cannot clear them.

**151 residual failures out of 593,548 tests (0.025%).**

| Group | Count | Cause |
|---|---|---|
| **A** | **97** | Known-unstable test whose codeunit **moved to a different app** in this PR → tolerance key misses |
| **B** | **38** | Known-unstable test in the same job, which never got the tolerance rewrite (cascade from A) |
| **C** | **16** | Not on the unstable list — `137264 SCM Package Tracking Sales`, GB only, pre-existing |

### Why group A happens

`build/scripts/TestTolerance/TestTolerance.psm1` keys the unstable-tests artifact on
`extensionId::codeunitId::testMethod`, where `extensionId` is the **app id**. Splitting a test app
gives its codeunits a new app id, so the entry no longer matches and an already-known-flaky failure
becomes a hard one.

Every codeunit driving group A is listed in the current `unstable-tests-main` artifact under its
**old** app:

| Codeunit | Listed under | Now lives in | Entries |
|---|---|---|---|
| `137023 SCM Reservation Worksheet` | `Tests-SCM` | `Tests-SCM-Reservation` | 14 |
| `137504 SCM Warehouse Unit Tests` | `Tests-SCM` | `Tests-SCM-Warehouse` | 7 |
| `137085 Cost Adjustment Features` | `Tests-SCM` | `Tests-SCM-Costing` | 6 |
| `137402 SCM Costing Batch` | `Tests-SCM` | `Tests-SCM-Costing` | 1 |
| `134911 ERM Create Finance Charge Memo` | `Tests-ERM` | `Tests-ERM-Finance` | 2 |
| `134992 ERM Financial Reports IV` | `Tests-ERM` | `Tests-ERM-Finance` | 1 |

These tests already fail on `main` — that is why they are on the list. The PR does not break them; it
breaks the *toleration* that normally absorbs them.

### Group B is cascade, verified

When one tenant hits a non-tolerated failure the pipeline throws, so the remaining tenants' result
files never get the tolerance rewrite applied. If that is the mechanism, every group-B failure should
sit in a job that also contains a group-A failure.

**Measured: 0 of the 25 failing jobs have a group-B failure without a group-A failure.**

### Failure pattern matches the mechanism

All **22 Bucket1** jobs failed, plus **3 Bucket2** (CA, US, GB). Bucket1 holds `Tests-SCM-Costing`
and `Tests-ERM-Finance`, whose moved codeunits run in every country, so every Bucket1 job hits at
least one. Bucket2 holds `Tests-SCM-Reservation` and `Tests-SCM-Warehouse`, whose broken entries only
manifest in CA and US.

### The 16 in group C are pre-existing

All `137264 SCM Package Tracking Sales`, GB Bucket2 only:

- baselines `30191726822` / `30301545614` (no split) — **pass**
- `30254557560` (SCM split only) — **fail in GB**
- `30354787707`, `30446833958` — **fail in GB**

They predate the `Tests-ERM` work and appeared with the earlier `Tests-SCM` split. The file is not
excluded in any layer and exists only in W1 and NL, so the only thing that changed for it is which
app it belongs to — i.e. which tests share its tenant and in what order. This looks like a test-order
or shared-state dependency and deserves separate triage; it is not a blocker for this PR.

### Options

1. **Merge on a documented exception**, using this attribution as the evidence. The artifact is
   regenerated from failures observed on `main`, so the entries re-add themselves under the new app
   ids on the next detection cycle and the noise disappears by itself.
2. **Regenerate the artifact first** — remap `extensionId` for the 12 moved codeunits so the entries
   match the new app ids, then rebuild. This also prevents the same breakage on every future test-app
   split, of which there will be more.

Relaxing the matching to `codeunitId::testMethod` was prototyped and **rejected**: 52 codeunit ids are
owned by different apps in different countries (e.g. `144013` lives in `Payment Management FR-Tests`,
`Reverse Charge VAT GB Tests` *and* `Tests-Local`), and the unstable list is a single global artifact
applied to all 22 country runs. It would be safe today only by coincidence.

Groenbech96:
## Group C root cause — GB posting-date check (correction to the mechanism)

> **Correction.** An earlier version of this comment explained the trigger as "119 codeunits ran
> before it in `Tests-SCM`, only 11 in `Tests-SCM-Reservation`". That is an oversimplification that
> does not fully hold — two codeunits in the new app (`137052`, `137059`) *do* clear the flag in
> `Initialize()` and commit it, and both run before `137264`. The verified chain is below; the last
> link is stated as unverified rather than guessed.

The 16 failures (`137264 SCM Package Tracking Sales`, GB only) are a **latent test-isolation defect
exposed by the split, not caused by it**. Fixed in `eb65dbf965`.

### Verified

1. The failure is `Unhandled UI: Confirm Posting Date %1 is different from Work Date %2`.
2. GB's `SalesHeader.TestPostingDate()` raises that Confirm only when **both** hold:
   `Sales & Receivables Setup."Posting Date Check on Posting"` is enabled **and**
   `"Posting Date" <> WorkDate()`.
3. The second condition is permanently true for these tests: the shared `CreateSalesOrder` helper
   sets the sales order's posting date to `CalcDate('<+5D>', WorkDate())`. That is pre-existing —
   it is present at the merge base (13 occurrences) and `main` has not touched this file.
4. So the only thing that can have flipped is the **flag**.
5. **GB demo data enables the flag** (`src/Layers/GB/DemoTool/CreateSalesReceivablesS.Codeunit.al:151`
   and the purchases equivalent).
6. **`137264` never clears it.** Its `Initialize()` calls `LibraryERMCountryData.UpdateLocalData()`,
   which **in the GB layer is an empty stub** (`begin exit; end;`). The procedures that actually
   clear the flag — `UpdateSalesReceivablesSetup()` / `UpdatePurchasesPayablesSetup()` — are never
   called, and the codeunit does not use `LibrarySetupStorage`, so it neither saves nor restores
   setup around itself.
7. It therefore depends entirely on **ambient setup state left by other codeunits on the same
   tenant** — state that is committed, and that the split changed by changing which codeunits share
   the tenant and in what order.

### Evidence it is not a product regression

- All **22 countries** run this codeunit with 26 tests; **only GB** fails.
- Baselines `30191726822` and `30301545614`: **26 tests, 0 failures, 0 skipped** — genuinely
  passing, not tolerated.
- First appeared in `30254557560`, the run where the `Tests-SCM` split landed.

### The fix

Clear the setup in `Initialize()`, the pattern **283 other W1 test codeunits already use**:

```al
LibraryERMCountryData.UpdateSalesReceivablesSetup();
LibraryERMCountryData.UpdatePurchasesPayablesSetup();
```

qasimikram (dismissed):

onbuyuka (dismissed):

Groenbech96:
## Both orphaned-`DisabledTest` findings: **fixed** in `0849be18ae`

Replying to the two AL-review findings from 2026-07-31:

- [#issuecomment-5144139101](https://github.com/microsoft/BCApps/pull/9736#issuecomment-5144139101) — codeunit `134911 ERM Create Finance Charge Memo`
- [#issuecomment-5144139332](https://github.com/microsoft/BCApps/pull/9736#issuecomment-5144139332) — codeunits `137402 SCM Costing Batch` / `137289 SCM Inventory Costing IV`

**Both findings were correct, and both are now fixed.** The diagnosis in each comment was exact: `Get-DisabledTests` resolves `src/DisabledTests/<AppName with _ for spaces>/*.json` strictly per test app, so entries left under the old app are dead once the codeunit moves.

All 5 orphaned entries have been moved to the owning app:

| Codeunit | Method | Moved from → to |
|---|---|---|
| 134911 | `SuggestFinanceChargeMemo` | `Tests-ERM` → `Tests-ERM-Finance` |
| 134911 | `SuggestFinanceChargeMemoLine` | `Tests-ERM` → `Tests-ERM-Finance` |
| 137402 | `PurchasePostingDoesNotPropagateACYToSourceCurrencyOnGLEntry` | `Tests-SCM` → `Tests-SCM-Costing` |
| 137289 | `AdjustCostForDifferentCurrencyExchangeRate` | `Tests-SCM` → `Tests-SCM-Costing` |
| 137289 | `PostPurchOrderWithDiffCurrencyExchangeRate` | `Tests-SCM` → `Tests-SCM-Costing` |

### Verification

**Before the fix** (run `30639053781`) these ran and failed exactly as predicted:

- `SuggestFinanceChargeMemo` / `SuggestFinanceChargeMemoLine` — failed in **CZ, DK, IN, IS, MX, NO, RU, SE** (`There is no Finance Charge Memo Header within the filter`)
- the three SCM ones — failed in **RU** (`...inconsistencies in the G/L Entry table`, and the ACY assert)

**After the fix** (run `30889678357`) all five are absent from **all 46** legacy test-result artifacts, i.e. correctly excluded again in every country.

### Completeness check

Rather than fixing only the reported five, I diffed the entire disabled-test corpus across both branches:

- `main` **21520** entries vs PR **21520** entries — **0 dropped, 0 added** (content now identical)
- Codeunit → owning-app orphan scan: the PR's orphan set is now **exactly equal** to `main`'s pre-existing set, so these were the only two clusters the split introduced.

Thanks — both findings were accurate and actionable.

Groenbech96:
## CI triage — run `30889678357` (head `3d93375`, attempt 1)

**13 legacy test jobs failed. None of the failures require a code change to this PR, and none can be cleared by a rerun.** Breakdown below.

### 1. The 5 previously-regressed tests are gone ✅

The orphaned-`DisabledTest` regression is fixed (`0849be18ae`). `SuggestFinanceChargeMemo`, `SuggestFinanceChargeMemoLine`, `AdjustCostForDifferentCurrencyExchangeRate`, `PostPurchOrderWithDiffCurrencyExchangeRate` and `PurchasePostingDoesNotPropagateACYToSourceCurrencyOnGLEntry` are absent from all 46 artifacts. Full disabled-test corpus is now byte-equivalent to `main` (21520 entries both sides, 0 dropped / 0 added).

### 2. Test failures — still the tolerance-key mismatch already documented above

7 jobs failed on test failures. Measured against the `unstable-tests-main` artifact actually used by this run (363 entries):

| Job | Failures | Tolerated now | Would be tolerated if `extensionId` were ignored | Genuinely untolerable |
|---|---:|---:|---:|---:|
| CA/B1 | 22 | 14 | 22 | 0 |
| CA/B2 | 21 | 0 | 21 | 0 |
| GB/B1 | 6 | 4 | 6 | 0 |
| IT/B1 | 1 | 0 | 1 | 0 |
| US/B1 | 8 | 0 | 8 | 0 |
| US/B2 | 21 | 0 | 21 | 0 |
| RU/B1 | 1 | 0 | 0 | **1** |

**60 of 78 known-unstable failures miss the tolerance key purely because the app id changed:**

| Results `extensionid` | List `extensionid` | Meaning |
|---|---|---|
| `4f458f6b-…` (Tests-SCM-Costing) | `6f0b528b-…` (Tests-SCM) | 137085, 137103, 137402 |
| `8f41d123-…` (Tests-SCM-Reservation) | `6f0b528b-…` (Tests-SCM) | 137023 |
| `6ed824d3-…` (Tests-ERM-Finance) | `fa3e2564-…` (Tests-ERM) | 134979, 134992 |

This is the same deterministic `extensionId::codeunitId::testMethod` mismatch triaged in [#issuecomment-5127488885](https://github.com/microsoft/BCApps/pull/9736#issuecomment-5127488885) — reruns reproduce it identically, and relaxing the key to `codeunitId::testMethod` was already prototyped and rejected there (52 codeunit ids are owned by different apps in different countries). It still needs one of the two documented decisions: **merge on a documented exception**, or **remap `extensionId` in the artifact for the moved codeunits first**. I have not touched `build/scripts/TestTolerance/TestTolerance.psm1` — it is outside this PR's diff and loosening a repo-wide failure-suppression mechanism is not a call to make autonomously.

### 3. One genuinely non-tolerated test (RU)

`133772 Remittance Purch & Sales UT::AppliedAmountOnRemittanceAdviceEntriesReportWhenApplyUapplyApply` — `Assert.Fail failed. No row found where Field 'LAmountWDiscCur' = <-257.79>`.

- Passes on `main` (68/68 across three nightlies, all countries)
- Fails in **RU only**, in both PR runs
- Not a disabled-entry problem: all 8 of this codeunit's exclusions migrated correctly to `Tests-ERM-Purchase`

The codeunit moved `Tests/ERM` → `Tests/ERM-Purchase`, so this looks like the same class as the group-C `137264` finding — a latent test-isolation/ordering dependency exposed by the split rather than caused by it. Needs its own triage; flagging rather than guessing at a fix.

### 4. Six jobs failed on BC platform instability (transient)

BE/B1, DE/B1, ES/B1, SE/B1, FI/B2, NZ/B2 recorded **0 test failures** yet failed, and ran 1000–3000 fewer tests than their peers. Each has exactly one app whose background job died:

```
Exception occurred while running tests: ClientSession State is InError (Wait time 30 seconds)
Test results file not found at '…\TestResults-<tenant>.xml'. Skipping tolerance check.
(job emitted terminating error: Test execution failed)
```

preceded by BC server `ArgumentNullException: Value cannot be null. (Parameter 'encoder')` in `System.Drawing` / `NavMediaImage.BytesAsync`. The victim app is different every time (`Tests-ERM` in DE+ES, `Tests-SCM-Assembly` in FI+NZ, `Tests-Misc` in BE, `Tests-Workflow` in SE) — a container/GDI+ session crash, not a code path.

Worth noting: `Test-TransientTestFailure` only matches the `InteractionManager` / page-130455 race, so `ClientSession State is InError` is classified `Failed` and never retried, and no results file is produced so tolerance cannot absorb it either. Adding that signature to the transient classifier would let the existing one-retry path absorb these — happy to do that as a follow-up if wanted, but it is a change to this PR's dispatcher semantics so I have not pushed it unprompted.

---

**Nothing pushed this iteration.** Items 2 and 3 need a human decision; item 4 is infrastructure noise that a rerun alone will not reliably clear while item 2 stands.

Groenbech96:
## Clarification: yes to the tolerance half, **no** to "just disable the last one"

Re-measured run `30889678357` against the current `unstable-tests-main` (380 entries, generated 2026-08-05 05:41):

| | Failure instances |
|---|---:|
| Total | **80** |
| Already matched (`extensionId` unchanged) | 18 |
| **On the list, blocked *only* by `extensionId`** | **61** |
| **Not on the list at all** | **1** |

So the first half of the reading is exactly right: **79 of 80 are failures `main` already tolerates**, and 61 of them fail purely because the split gave the codeunit a new app id.

### But the remaining one should probably not be disabled

`133772 Remittance Purch & Sales UT::AppliedAmountOnRemittanceAdviceEntriesReportWhenApplyUapplyApply`

| Where | Result |
|---|---|
| `main`, **RU**, nightly `30774412143` | **PASS** |
| `main`, **RU**, nightly `30865466451` | **PASS** |
| This PR, other 22 countries | **PASS** (22/22) |
| This PR, **RU** | **FAIL** (both PR runs) |

It is green in RU on `main` and green everywhere else on this branch — it regressed *only* on this branch, in *one* country. That is not the "RU base-app behaviour" pattern the other RU exclusions describe; disabling it would be **globally** disabling a test that genuinely passes in 22 countries, and would hide the regression rather than record it.

The mechanism looks like the group-C `137264` case, which we **fixed** (`eb65dbf965`) rather than disabled. This codeunit has the same weakness — `Initialize()` is only:

```al
local procedure Initialize()
begin
    LibraryVariableStorage.Clear();
    DeleteObjectOptionsIfNeeded();
end;
```

No `LibrarySetupStorage`, no `LibraryERMCountryData` reset — so it inherits whatever setup state the other codeunits on its tenant committed, and the split changed which codeunits those are. The assert is on an applied amount (`LAmountWDiscCur`), which RU application/rounding setup influences.

### One more thing the extensionId framing misses

**6 of the 13 failing jobs had zero test failures.** BE/DE/ES/SE Bucket1 and FI/NZ Bucket2 each lost one app to a BC platform crash:

```
Exception occurred while running tests: ClientSession State is InError (Wait time 30 seconds)
Test results file not found at '…\TestResults-<tenant>.xml'. Skipping tolerance check.
```

No results file is written, so **neither tolerating nor disabling touches these** — and `Test-TransientTestFailure` doesn't match this signature, so they're never retried either. Even a perfect answer to the `extensionId` question would not have made *this* run green.

### Suggested split of the decision

1. **`extensionId` / 61 failures** — the documented choice: merge on exception, or remap the artifact. Unchanged.
2. **The 1 RU test** — triage as an isolation defect (like `137264`), not a disable. If it must be unblocked quickly, prefer an explicit temporary exclusion with a bug annotation naming the split as the trigger, so it isn't mistaken for a genuine RU behaviour difference.
3. **The 6 platform crashes** — add `ClientSession State is InError` to `Test-TransientTestFailure` so the existing one-retry path absorbs them. Happy to push that on request.

Groenbech96:
## Bug filed for the RU test: **AB#645167**

[645167 — \[BCApps\] AppliedAmountOnRemittanceAdviceEntriesReportWhenApplyUapplyApply fails in RU after the Tests-ERM split (test isolation)](https://dev.azure.com/dynamicssmb2/Dynamics%20SMB/_workitems/edit/645167)

| Field | Value |
|---|---|
| Type | Bug — Test Defect |
| Area | `Dynamics SMB\AI Business Solutions\Finance` |
| Iteration | `Dynamics SMB\Wave 2 2026` |
| Severity / Priority | 3 / 2 |
| Parent | [644840 — Improve App Test Build in BCApps](https://dev.azure.com/dynamicssmb2/Dynamics%20SMB/_workitems/edit/644840) |

The bug records the full evidence: PASS on `main` in RU (nightlies `30774412143` / `30865466451`), PASS in the other 22 countries on this branch, FAIL in RU in both PR runs — plus the stack trace, the `Initialize()` gap (no `LibrarySetupStorage` / `LibraryERMCountryData` reset), and the parallel to `137264 SCM Package Tracking Sales` which was fixed rather than disabled.

It also explicitly records that this is **not** a DisabledTest or tolerance-key problem, so it does not get conflated with the `extensionId` discussion above.

Groenbech96:
## Test failure analysis — run `30997345842` (head `8374a11`)

**The compile fix worked: all 44 `Build Apps` jobs passed (0 failed).** 11 test jobs failed. Breakdown below — **nothing here is a new regression from this PR.**

### Overall

| | Failure instances |
|---|---:|
| Total | **135** |
| Already tolerated | 72 |
| **On the list, blocked only by `extensionId`** | **62** |
| **Not on the list at all** | **1** |

### Per job

| Job | Fails | Tolerated | extId-blocked | Not on list | Verdict |
|---|---:|---:|---:|---:|---|
| CA/B2 | 21 | 0 | 21 | 0 | extId |
| US/B2 | 21 | 0 | 21 | 0 | extId |
| CA/B1 | 10 | 2 | 8 | 0 | extId |
| US/B1 | 10 | 2 | 8 | 0 | extId |
| IT/B1 | 3 | 2 | 1 | 0 | extId |
| GB/B1 | 2 | 0 | 2 | 0 | extId |
| CZ/B1 | 1 | 0 | 1 | 0 | extId |
| RU/B1 | 1 | 0 | 0 | **1** | AB#645167 |
| **FR (Default)** | 66 | **66** | 0 | 0 | **stale list — see below** |
| **W1/B1** | 0 | – | – | – | **platform crash** |
| **NO/B2** | 0 | – | – | – | **platform crash** |

### 1. FR (Default) — a race with the unstable-test detector, not a regression

This looked alarming (66 failures in `Factur-X CII XML Tests`, `PEPPOL BIS 3.0 XML Tests`, `Export E-Reporting Tests`, all `VAT Bus. Posting Group ... cannot be found in the related table`). It is **not** caused by this PR:

- This PR touches **nothing** under E-Reporting FR — verified against the full PR file list.
- The app's `extensionId` (`9b8abc90-…`) is **unchanged** by this PR, so the tolerance key matches normally.
- The job downloaded unstable-tests artifact **`8931707777`** — **392 entries, generated 13:15**, which contained **51** of these entries and reported *"15 failure(s) are NOT in the unstable tests list"*.
- The current artifact — **408 entries, 16:41** — contains **all 66**. 51 → 66 is exactly the 15 that failed.

So the detector added them **after** this run started. **A rerun of the FR job now would tolerate all 66 and go green.** Pure timing.

### 2. W1/B1 and NO/B2 — BC platform crash, zero test failures

Both recorded **0** failures across every artifact, yet failed. Same signature as last run:

```
Exception occurred while running tests: Cannot open page 130455 ...     <- known race, retried once
Exception occurred while running tests: ClientSession State is InError  <- NOT matched, hard Failed
Test results file not found at '...\TestResults-<tenant>.xml'. Skipping tolerance check.
```

The sequence is now clear: the **first** failure is the known `page 130455` race, which `Test-TransientTestFailure` catches and re-queues. The job then dies on **`ClientSession State is InError`**, which the classifier does **not** match — so it is graded `Failed`, and because no results file is written, tolerance cannot absorb it either.

Victims differ per run (`Tests-Workflow` on W1, `Tests-SCM-Assembly` on NO), confirming infrastructure rather than a code path.

**This is a one-line fix in this PR's own dispatcher** — adding `ClientSession State is InError` to `Test-TransientTestFailure` would let the existing one-retry path absorb it. Happy to push it on request; I have not, since it changes this PR's retry semantics.

### 3. The 62 `extensionId`-blocked failures — unchanged

| Count | Codeunit | Jobs |
|---:|---|---|
| 28 | `137023 SCM Reservation Worksheet` | CA/B2, US/B2 |
| 14 | `137504 SCM Warehouse Unit Tests` | CA/B2, US/B2 |
| 12 | `137085 Cost Adjustment Features` | CA/B1, US/B1 |
| 4 | `134979 Reminder Automation Tests` | CA/B1, CZ/B1, IT/B1, US/B1 |
| 2 | `137103 Cost Adjustment Parallel Run` | CA/B1, US/B1 |
| 1 | `134897 ERM Source Currency` | GB/B1 |
| 1 | `134992 ERM Financial Reports IV` | GB/B1 |

All already known-unstable on `main`; they only miss the tolerance key because the split gave the moved codeunits new app ids. Still the outstanding decision.

### 4. The one genuine non-tolerated failure

`133772 Remittance Purch & Sales UT::AppliedAmountOnRemittanceAdviceEntriesReportWhenApplyUapplyApply` (RU) — tracked as **AB#645167**. Unchanged.

---

### Summary

Of 11 failing jobs: **7 are purely the `extensionId` issue**, **1 is FR needing only a rerun**, **2 are BC platform crashes**, and **1 is the known RU bug**. Resolving the `extensionId` question plus a rerun would leave only the RU test and the platform flakiness.

aholstrup1 (approved):

qasimikram (approved):

PR #9749, opened 2026-07-27
OIOUBL fixes after schematron update

mynjj:
# Problem

We have always exported a technically invalid OIOUBL document whenever we had a document with a header-level discount (or line level discount).

The approach used to be to report the tax of the discount as a `ReverseCharge` with 0% impact in VAT, and have the lines report the TaxableAmount after discounts.

A recent change in the OIOUBL schematron enforced a rule that now makes this kind of exported documents invalid (https://git.erst.dk/openebusiness/common/-/blob/master/released/oioubl/OIOUBLchangelist1.17.adoc?ref_type=heads)

Concretely, the validation asks for all the tax categories reported along the document to also exist on the summary section of the document and viceversa. This was not true because we were hardcoding ReverseCharge as the discount, but the lines only had others like StandardRated. Therefore the `ReverseCharge` was not appearing in the summary.

# Solution

A possible solution could be to add yet another synthetic ReverseCharge in the summary with zero impact, but this is semantically incorrect.

The correct solution is to stop exporting the discount as a ReverseCharge and instead take the value that each of the applicable VAT types of the lines get. This means several AllowanceCharge sections instead of just one, with the right applicable VAT.

By changing the discount's VAT from 0% to x% we are forced to also fix the TaxableAmount of the corresponding lines, since another validation enforces that the totals section add up. Instead of having the line report it's TaxTotal with the value after the discount, report the raw amount before the discount. This way the latter header level section cancels out appropriately and the totals match.

### Header level

| Section | Changed? | What and Why |
|---|---|---|
| `cac:AllowanceCharge` (header level discount) | Yes | Before we always emitted a single category `ReverseCharge`, now we emit several per (category, vat%) pairs. (The new enforced validation, main fix)|
| `cac:TaxTotal` and `cac:TaxSubtotal` per category| Yes| Before it only reported one TaxSubtotal and collapsed in case there were multiple rates in the invoice. Now there's one SubTotal per category|
| Totals like `cac:LegalMonetaryTotal`|No | Untouched, total-wise we are doing the same, it's just correct VAT classification of the discount's VAT share. |

### Line level

| Section | Changed? | What and Why |
|---|---|---|
| `cac:AllowanceCharge` (line level discount)| Yes | Same fix from ReverseCharge to the corresponding |
|  `cac:TaxTotal`, `cac:TaxSubtotal` (line tax) | Yes | As explained above, to consider the header level discount to be subtracted later, the lines are reported with it's full amount. Those values are f.ex for SalesInvoice `SalesInvoiceLine.Amount + SalesInvoiceLine."Inv. Discount Amount"` as Taxable Base, the tax amount is computed as the tax amount of the sales line minus the line discount contribution (taking computed values from the line to avoid rounding problems, instead of recalculating from the Taxable Base) |

# Verification

Several scenarios with header-level and line-level discounts have been tried manually and validated with https://oioubl.nemhandel.dk/validation. 

The totals section hasn't changed, so total-wise we are safe. 

Generated tests have been validated as the expected behavior change and ran before and after the changes to ensure coverage.

---

Fixes [AB#638267](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/638267)


Groenbech96 (dismissed):
Dont you want a test case or two for this change? NVM, i did not see this was draft-

src/Apps/DK/OIOUBL/app/src/Management/OIOUBLTaxGroupBuffer.Table.al
```diff
@@ -0,0 +1,43 @@
...
+        }
+        field(5; "OIOUBL-Inv. Discount Amount"; Decimal)
+        {
+            Caption = 'Inv. Discount Amount';
+        }
+    }
+
+    keys
+    {
+        key(PK; "OIOUBL-Tax Category ID", "OIOUBL-VAT %")
```
AleksanderGladkov, line 38: @mynjj For me it looks weird to have decimal field as part of the primary key. Can we just one field in the key like Entry No or Line ID etc.?

mynjj: @AleksanderGladkov , can you elaborate on your proposal? My goal is to have unique (category, VAT%) being accumulated while iterating over the lines. It works since the decimals are coming from the same source (the VAT % on the lines that comes from VAT Posting Setup). Also note this is a temp table.

I assume the concern comes from the possibility of rounding issues making two records out of one with VAT 1.1 and another one with 1.1000000001 or something like this. I could serialize it into a line id, round, or juggle with an entry number, but the end result is the same I think, and in the scenario of fixed values in VAT Posting Setup is not really a concern, there are no operations on these values.

AleksanderGladkov: @mynjj I understand your intention, I just never seen decimals in primary key before.
As about my proposal, if you have the only Entry No. / Line ID as primary key, you will still be able to apply filters to tax category and VAT % to understand if the line exists.

Also my concern is that in future the OIOUBL rules could change again and for example you will have to group amounts by some other property - like today you group by tax category and VAT %, and next year they introduce tax subcategory, and your table will not work anymore. With Line ID it will be easy to adjust the solution.

mynjj: Thanks Sasha,

I think that decimals in PK is weird mainly for persisted tables, and for cases where the Decimal values come from computed amounts. So it's a good rule of thumb to have. However, neither of these is true in this scenario.

Your proposed fix of having a single column as PK and a different key for querying has the following advantages:
- Doesn't risk rounding problems
- Allows for evolving the grouping record in case you have to group by something else later on

The first advantage is not really applicable in this scenario, since the values come from the fixed VAT % without any operations, they are a discrete set of values that the Decimal can take.

The second advantage is applicable, however if such requirement comes (not very likely), we can use a different temporary table and obsolete the previous one. This is a tradeoff.

Now your proposed solution has this disadvantage:
- Extra bookkeeping and code performance impact: I'm not certain that temp tables can define non-primary keys and have the lookups use them, so the proposal is effectively adding more code that does the same with some potential performance impact (granted, this is minimal, the size of this table should be small).

Alternatively, to ensure discreteness we could use values like VAT Product Posting Group and VAT Business Posting Groups, that correspond to a %. However these can be edited by a user, and for summarizing an invoice we should use what was used for that invoice at the moment of posting.

I've also checked around in the code base and there's some precedent in using VAT % as primary keys in buffer tables. See https://github.com/microsoft/BCApps/blob/eeb62962c2a8cbaa8fa9f66c8ae44c102b2e3ee6/src/Layers/ES/BaseApp/Local/Finance/ReceivablesPayables/SalesPurchBookVATBuffer.Table.al#L10 or https://github.com/microsoft/BCApps/blob/eeb62962c2a8cbaa8fa9f66c8ae44c102b2e3ee6/src/Layers/IT/BaseApp/Local/Finance/VAT/Ledger/VATRegisterBuffer.Table.al#L9 or https://github.com/microsoft/BCApps/blob/eeb62962c2a8cbaa8fa9f66c8ae44c102b2e3ee6/src/Layers/NA/BaseApp/Local/Finance/SalesTax/SalesTaxAmountDifference.Table.al#L4

So I'll keep this solution.

src/Apps/DK/OIOUBL/test/src/OIOUBLERMElecDocumentSales.Codeunit.al
```diff
@@ -347,6 +347,270 @@ codeunit 148053 "OIOUBL-ERM Elec Document Sales"
         VerifyAllowanceCharge(DocumentNo, ExpectedResult);
     end;
 
+    [Test]
+    [HandlerFunctions('MessageHandler')]
```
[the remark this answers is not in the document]
mynjj: existing handler

ventselartur (dismissed):

alexei-dobriansky:

<details>

This PR changes the DK OIOUBL export so header invoice discounts and line discounts use the real line tax category instead of a hard-coded `ReverseCharge` category. It also splits document allowances and document tax subtotals by `(TaxCategory, VAT %)` and adjusts line tax totals so the line amount and header allowance reconcile.

The change addresses the described schematron failure: every allowance tax category now comes from the affected posted lines, and the document `TaxTotal` contains matching subtotals. The sales and service invoice paths are aligned, and the credit memo paths use the same shared grouping helper. The new tests cover standard VAT, mixed VAT rates, zero-rated VAT, reverse charge, and the main sales/service invoice discount paths.
</details>

<details>

**S1** - *Mixed VAT credit memos lack coverage*
The grouping change also runs for sales and service credit memos. Add one mixed VAT credit memo test, or explain why invoice coverage is enough. This will protect the changed credit memo paths from future regressions.
</details>

<details>

**Risk:** this is a regulatory XML export path, so wrong tax categories or totals can make posted sales and service documents fail OIOUBL validation. The change is broad inside the DK OIOUBL export, but it is mostly shared helper logic and preserves the existing public event signatures. I did not find a BaseApp event dependency in this PR.

**Necessity:** the linked bug and work item discussion describe a valid customer-impacting validation failure after the OIOUBL schematron update. There was no `[AI-REPRO]` comment on the work item, so I used the PR description, work item fields, and work item discussion to assess the scenario.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9749 round=1 by=alexei-dobriansky at=2026-08-03T10:08:13.082Z lastSha=76223e6073854f4d5765b1e4cee4aca8781e4a41 suggestions=S1

alexei-dobriansky:

<details>

The single commit since round 1 fixes the `GetGLN()` test helper in `OIOUBLElecServiceDocument.Codeunit.al`. It now returns a hardcoded valid GLN (`3974567891234`) when no customer with a GLN exists in the current test run, instead of failing with an unhandled error. It also updates the old-style `SETFILTER`/`FINDFIRST` calls to the current casing. The fix is correct and improves test robustness in isolated environments.
</details>

<details>

| S1 | Mixed VAT credit memos lack coverage | Not addressed | No author reply; the new commit only touches the GLN helper. |

</details>

<details>

*None — the new commit is a test helper robustness fix with no new issues.*
</details>

<details>

**Risk:** The GLN fallback value is hardcoded in a test helper. If a future test needs to assert that the GLN in the generated XML matches a specific customer record, the fallback value could mask a real lookup failure. The risk is confined to the test layer and does not affect any production code.

**Necessity:** The GLN fix is a reasonable defensive change for CI environments where customer data may not be present. The original scenario assessment from round 1 is unchanged.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=1 system=github pr=9749 round=2 by=alexei-dobriansky at=2026-08-06T13:16:23Z lastSha=b5cb55d96a2c039fdadbe163c907095e6af34faa reviewKey=bc02dd84db5e80babcc7c5cf263fd17fe627cd8c7522f6f75f0014a5373a269b suggestions=S1@e554032b:notaddressed parentRound=1

ventselartur (approved):

djukicmilica (approved):

Groenbech96 (approved):

PR #9751, opened 2026-07-27
Extract invoice date into Invoice Date field and recover due date fro…

Explorer986:
## What & why

Invoice Date on the purchase draft came through blank because the extractors only wrote the invoice's date into Document Date, never into the Invoice Date field. This maps the extracted date into Invoice Date too, in both the ADI and the MLLM readers, so it is no longer empty.

Also recovers the Due Date on credit memos and some invoices where it sits under PaymentMeans instead of at the top level. The MLLM reader now falls back to payment_means.payment_due_date when the top level due date is missing, matching what the PEPPOL handler already does, and the field was added to the UBL schema so the model has a slot for it.

## Linked work

Fixes [AB#642335](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/642335)

## How I validated this

- [ ] I added or updated tests for the new behavior or explained below why none are needed.

- Tested on a real invoice through and confirmed Invoice Date shows in the extracted data.

## Risk & compatibility

Low. It only fills a field that used to be blank and adds a due date fallback. Document Date mapping is untouched so nothing downstream at finalisation changes. The extra schema key is additive. No permissions, upgrade or telemetry impact.

### DIFF
--- a/src/Apps/W1/EDocument/App/.resources/Prompts/EDocMLLMExtraction-SystemPrompt.md
+++ b/src/Apps/W1/EDocument/App/.resources/Prompts/EDocMLLMExtraction-SystemPrompt.md
@@ -6,6 +6,7 @@ EXTRACTION RULES:
 3. Dates: YYYY-MM-DD format
 4. Extract ALL invoice lines with sequential IDs starting from "1"
 5. Quantity: use "1" only if no quantity column exists on the document
+6. Due date: put the payment due date in "due_date" wherever it appears on the document
 
 CUSTOMER vs VENDOR IDENTIFICATION:
 The JSON structure includes pre-filled accounting_customer_party data. This is OUR company — the buyer receiving the invoice. Use this to distinguish between customer and vendor on the document:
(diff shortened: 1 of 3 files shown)


djukicmilica (dismissed):

ventselartur (dismissed):

Groenbech96:
Have we run the evals with this change to ensure no regression?

src/Apps/W1/EDocument/App/.resources/AITools/ubl_example.json
mynjj, line 6: We already have this due date, adding an additional date field for something we are not properly modeling on the PurchaseDoc just adds another field for confusion.

The extraction behavior is controlled by the prompt sent to the LLM, if you have found that adding this field captures the missing date, you can probably also prompt to put this field wherever the due date is found (additional guidance in the MLLM extraction prompt, for example to say "whether the due date is for a payment or for the document, store it in due date" or something like this)

mynjj (changes requested):

mynjj (approved):

Groenbech96 (approved):

PR #9772, opened 2026-07-28
[PA][Pioneer]: Numeric line fields mis-extracted (discount, quantity, unit of measure)

Explorer986:
## What & why

When a scanned invoice line has no quantity or a zero line total, the importer used to force the quantity to 1 and invent a discount from the price. That created phantom lines and wrong discounts. Now we keep a genuine zero quantity, only derive quantity from total over price when it makes sense, and never fabricate a discount. Unresolved units like HUR now carry through to the BC unit of measure field instead of showing blank.

## Linked work

Fixes [AB#642334](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/642334)

## How I validated this

- Imported the Spanish invoice in CRONUS ES. The zero hour lines now show quantity 0 instead of 1, and the Docencia line shows 50 as on the document.
- Confirmed no discount is invented when the line total already matches price times quantity.

## Risk & compatibility

The unresolved unit fallback puts the raw code like HUR straight into the BC unit of measure field in View Extracted Data Page.

### DIFF
--- a/src/Apps/W1/EDocument/App/src/Processing/Import/StructureReceivedEDocument/EDocMLLMSchemaHelper.Codeunit.al
+++ b/src/Apps/W1/EDocument/App/src/Processing/Import/StructureReceivedEDocument/EDocMLLMSchemaHelper.Codeunit.al
@@ -162,13 +162,19 @@ codeunit 6232 "E-Doc. MLLM Schema Helper"
                     GetDecimal(NestedObj, 'value', TempLine.Quantity);
                     GetString(NestedObj, 'unit_code', MaxStrLen(TempLine."Unit of Measure"), TempLine."Unit of Measure");
                 end;
-                if TempLine.Quantity <= 0 then
-                    TempLine.Quantity := 1;
 
                 if GetNestedObject(LineObj, 'price', NestedObj) then
                     GetDecimal(NestedObj, 'price_amount', TempLine."Unit Price");
 
                 GetDecimal(LineObj, 'line_extension_amount', TempLine."Sub Total");
+
+                if TempLine.Quantity < 0 then
+                    TempLine.Quantity := 0;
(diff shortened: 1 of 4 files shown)


src/Apps/W1/EDocument/App/src/Processing/Import/StructureReceivedEDocument/EDocMLLMSchemaHelper.Codeunit.al
```diff
@@ -162,14 +162,17 @@ codeunit 6232 "E-Doc. MLLM Schema Helper"
...
-                    TempLine.Quantity := 1;
 
                 if GetNestedObject(LineObj, 'price', NestedObj) then
                     GetDecimal(NestedObj, 'price_amount', TempLine."Unit Price");
 
                 GetDecimal(LineObj, 'line_extension_amount', TempLine."Sub Total");
 
+                if TempLine.Quantity < 0 then
+                    TempLine.Quantity := 0;
+                if (TempLine.Quantity = 0) and (TempLine."Unit Price" <> 0) and (TempLine."Sub Total" <> 0) then
```
mynjj: and what about the case when qty is zero and any of the others is zero? I thought we discussed assigning 1 in such case

Explorer986: Thanks, updated my logic to keep quantity = 1 when any of the Unit Price or Sub Total is non-zero

mynjj (dismissed):

mynjj (approved):

ventselartur (approved):

Groenbech96 (approved):

PR #9774, opened 2026-07-28
[Extensibility Request] issue 30378: add OnItemVariantOnAfterGetRecordOnBeforePrintOnlyIfSalesCheck event in Item Sales Statistics report

AleksandricMarko:
## Summary
When "Only Items with Sales" and "Breakdown By Variant" are both enabled on the Item Sales Statistics report (10135), the base report decides, per item variant, whether to skip a row based solely on `Item."Sales (Qty.)"` and the private `PrintOnlyIfSales` flag. Extensions that compute their own sales figures (for example alternate unit-of-measure quantities) cannot influence this decision because `PrintOnlyIfSales` has no accessible getter and the skip is evaluated before any extension trigger runs. This PR exposes the already-computed skip decision through a new integration event at the variant level so subscribers can fold in their own figures.

Source issue repository: microsoft/AlAppExtensions; issue number: 30378

## Changes Made
- `OnItemVariantOnAfterGetRecordOnBeforePrintOnlyIfSalesCheck` - new integration event raised in the `Item Variant` `OnAfterGetRecord` trigger after `Item.CalcFields` and before the `CurrReport.Skip()` check; the inline skip condition now assigns a `SkipRecord` boolean that the event can override via `var`, exposing `Item`, `ItemVariant`, and `PrintOnlyIfSales` to subscribers.

Fixes [AB#644154](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/644154)


AleksandricMarko:
@copilot
in  local procedure OnItemVariantOnAfterGetRecordOnBeforePrintOnlyIfSalesCheck(var Item: Record Item; ItemVariant: Record "Item Variant"; PrintOnlyIfSales: Boolean; var SkipRecord: Boolean) paremeter Itemvariant should be with VAR

qasimikram (dismissed):
S1 - Clearing SkipRecord here does not control the final decision: the same trigger later calls CurrReport.Skip when all standard quantity/value fields are zero. That is precisely possible for alternate-UOM extension sales. Move the publisher after the return calculation or expose the final skip decision, and test zero standard sales with nonzero extension sales.

S2 - The parent Item trigger can skip before the Item Variant dataitem runs, so this publisher depends on the companion item-level hook. Deliver the hooks together or otherwise ensure the variant event is reachable for the documented scenario.

S3 - Pass ItemVariant by value. Allowing mutation after SetRange and CalcFields can combine values calculated for one variant with a different variant code used by later logic.

PredragMaricic (dismissed):
The event is placed after the variant sales FlowFields are calculated and before the existing skip decision, initializes SkipRecord for every variant, and exposes exactly the state needed by the linked extensibility request. No blocking findings.

alexei-dobriansky:

<details>

Adds a new integration event `OnItemVariantOnAfterGetRecordOnBeforePrintOnlyIfSalesCheck` in report 10135 "Item Sales Statistics". The event fires in the `Item Variant` dataitem's `OnAfterGetRecord` trigger after `CalcFields` computes `Sales (Qty.)` and before the `CurrReport.Skip()` decision, allowing subscribers to override whether a variant row is suppressed when "Only Items with Sales" is enabled.

The implementation is correct: it extracts the skip condition into a local `SkipRecord` boolean, raises the event passing `Item` (var), `ItemVariant` (var), `PrintOnlyIfSales` (value), and `SkipRecord` (var), then branches on the returned `SkipRecord`. The subscriber can therefore prevent a skip (showing the row) or force a skip (hiding it). The `[IntegrationEvent(false, false)]` attribute and empty body are standard.
</details>

<details>

**S1** - *Pass ItemVariant by value, not by var*
`ItemVariant` is passed `var` but the caller does not read back any mutation. Allowing subscribers to modify the record after `SetRange` and `CalcFields` have already executed can silently corrupt profit and quantity calculations later in the same trigger. Pass it without `var` — the subscriber only needs to read its fields to decide the skip.

**S2** - *Consider a shorter event name*
`OnItemVariantOnAfterGetRecordOnBeforePrintOnlyIfSalesCheck` is 57 characters with three stacked "On" segments. A name like `OnItemVariantAfterGetRecordBeforeSkipCheck` or `OnBeforeSkipItemVariant` would be clearer and still follow the BC pattern `On<DataItem>On<Trigger>On<Before|After><Step>`. Not blocking, but improves discoverability.

**S3** - *Add a matching event at the Item level*
The parent `Item` dataitem has the same pattern (`if ("Sales (Qty.)" = 0) and PrintOnlyIfSales then CurrReport.Skip()`) with no event. When the item-level skip fires, the variant dataitem never runs, so this new event is unreachable for those items. Delivering both events together completes the scenario.
</details>

<details>

**Risk:** Low regression surface — the change only refactors an inline condition into a local boolean and raises a publisher with an empty body; existing behaviour is preserved when no subscriber is present. Passing `ItemVariant` by `var` is the only concrete risk (S1). The event signature becomes public API once shipped; renaming later is a breaking change. No performance concern — the event fires once per variant row.

**Necessity:** The scenario is real and concrete: extensions that compute alternate sales figures (e.g. alternate UOM) cannot suppress or un-suppress variant rows today because the skip decision is internal. Without this change, partners must copy the entire report. Scope is minimal and appropriate for a single-event extensibility request.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9774 round=1 by=alexei-dobriansky at=2026-08-03T09:30:00Z lastSha=b299581ff94a17c8e8d18b9b62cad753f1f1f517 suggestions=S1,S2,S3

AleksandricMarko:

alexei-dobriansky:

<details>

Since round 1 the PR added a second integration event, `OnAfterGetRecordItemOnBeforePrintOnlyIfSalesCheck`, in the parent `Item` dataitem's `OnAfterGetRecord`, next to the existing variant-level event. This addresses the round-1 gap where an item-level skip could hide the variant rows before the variant event ran.

The item-level hook is correct: it sets `SkipRecord`, raises the event, and skips only when the subscriber leaves `SkipRecord` true; nothing else in that trigger calls `CurrReport.Skip()`. The variant-level hook, however, is still not enough to deliver the documented alternate-UOM scenario, because a second, unconditional skip runs later in the same trigger (see S4). The `var ItemVariant` parameter (S1) and the long event names (S2) from round 1 are unchanged.
</details>

<details>

| ID  | Title                                         | Status        | Author response |
|-----|-----------------------------------------------|---------------|-----------------|
| S1  | Pass ItemVariant by value, not by var         | Not addressed | - |
| S2  | Consider a shorter event name                 | Not addressed | - |
| S3  | Add a matching event at the Item level        | Addressed     | Added `OnAfterGetRecordItemOnBeforePrintOnlyIfSalesCheck` in `8ce61a8` |
</details>

<details>

**S4** - *Variant event does not control the final skip*
In the `Item Variant` `OnAfterGetRecord` trigger, a second skip runs after the event: `if (Item."Sales (Qty.)" = 0) and (QuantityReturned = 0) and (Item."Sales (LCY)" = 0) and (Item."COGS (LCY)" = 0) then CurrReport.Skip();`. A subscriber that has only alternate-UOM sales sets `SkipRecord := false` at the event, but this later check still hides the row because all standard fields are zero. Move the event after this final skip calculation, or expose this decision too, so the documented scenario works for variants.

Carried forward from round 1 (still open):

**S1** - *Pass ItemVariant by value, not by var*
`ItemVariant` is still passed `var`. A subscriber can change the record after `SetRange` and `CalcFields` have run, which can mix values from one variant with a different variant code used later in the trigger. Pass it without `var`, since the subscriber only needs to read its fields.

**S2** - *Consider a shorter event name*
The variant event name is still 57 characters with three stacked "On" segments. A shorter name (for example `OnItemVariantBeforeSkipCheck`) is clearer. This is not blocking, but the name becomes public API once shipped and cannot be changed later without a breaking change.
</details>

<details>

**Risk:** Low regression surface for existing behaviour - both events have empty bodies and default to the current skip when no subscriber is present. The main risk is functional: as placed, the variant event cannot deliver the documented alternate-UOM scenario (S4), so partners may still be forced to copy the report. Both event signatures become public API once shipped, so the `var ItemVariant` parameter (S1) and the long names (S2) are locked in after merge. No performance concern - each event fires once per row.

**Necessity:** The scenario is real and important: extensions that compute their own sales figures cannot influence the per-item and per-variant skip today. The item-level hook now closes the round-1 reachability gap. The scope stays minimal and appropriate for an extensibility request, but the variant path must actually let a subscriber show the row for the change to meet its stated goal.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=1 system=github pr=9774 round=2 by=alexei-dobriansky at=2026-08-05T10:04:26Z lastSha=8ce61a8c277bd6a71f674b82096a40466a24fb72 reviewKey=6de0623ef1e00e7090d2ad87b9952aa773cb4c6bc0ec72988f6f34c3a4680a8f suggestions=S1@51304fd9:notaddressed,S2@69c8b960:notaddressed,S3@dbb5e5f2:addressed,S4@48190fba:new parentRound=1

alexei-dobriansky:
Please check the suggestion S4.

alexei-dobriansky (dismissed):

alexei-dobriansky:

<details>

Since round 2, the variant-level event was moved after the final zero-sales, zero-return, zero-amount skip calculation. The trigger now builds one `SkipRecord` value, raises `OnItemVariantOnAfterGetRecordOnBeforePrintOnlyIfSalesCheck`, and only then calls `CurrReport.Skip()`.

This placement fixes the earlier gap. A subscriber can now keep a variant row visible after all base skip inputs have been calculated, and the existing behavior stays the same when there is no subscriber.
</details>

<details>

**Fit:** Strong

The change now matches the requested variant-level extension point: extensions can see the final skip decision before the report skips the row. The item-level event from the previous round is still present, so the parent item can also be kept visible when variant rows need to run.
</details>

<details>

| ID  | Title                                         | Status        | Author response / evidence |
|-----|-----------------------------------------------|---------------|----------------------------|
| S1  | Pass ItemVariant by value, not by var         | Not addressed | The event signature still passes `var ItemVariant`. |
| S2  | Consider a shorter event name                 | Not addressed | The variant event name is unchanged. |
| S3  | Add a matching event at the Item level        | Addressed     | The item-level event is still present before the item-level skip. |
| S4  | Variant event does not control the final skip | Addressed     | The current commit moves the variant event after the final skip calculation and gates `CurrReport.Skip()` with `SkipRecord`. |
</details>

<details>

*None - the new commit only addresses the previous variant skip blocker.*

Carried forward from earlier rounds:

**S1** (🟠 Moderate): *Pass ItemVariant by value, not by var*
`ItemVariant` is still passed by `var`. A subscriber can change the current variant record even though the event only needs it as read context. Pass `ItemVariant` by value and keep only `SkipRecord` as `var`.

**S2** (🟡 Minor): *Use a shorter event name*
The variant event name is still long and repeats `On` segments. A shorter name such as `OnItemVariantBeforeSkipCheck` is easier to use. This is easier to change before the event is shipped.
</details>

<details>

**Risk:** Low for existing behavior because both events are empty by default and the base skip outcome is preserved when no subscriber changes `SkipRecord`. The current variant placement is correct: it runs after `CalcFields`, the return adjustment, and the final skip calculation. The remaining risk is public API shape, because the `var ItemVariant` parameter and the long event name become hard to change after release.

**Necessity:** The extension point is needed because extensions cannot otherwise affect the report's item and variant skip decisions. The scope is narrow and appropriate for an extensibility request.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=9774 round=3 by=alexei-dobriansky at=2026-08-25T10:04:56Z lastSha=320311b412980b6f519ee708467c2a5fab6737ac reviewKey=a7940d6fd8d5f1f98286ad608cf143c511902f666039b0881b495e7e42281c67 suggestions=S1@51304fd9:notaddressed,S2@69c8b960:notaddressed,S3@dbb5e5f2:addressed,S4@48190fba:addressed parentRound=2

alexei-dobriansky (approved):

PredragMaricic (approved):

PR #9783, opened 2026-07-28
[Master] Purchase Return Order Posting Fails for Manually Reserved Lot-Tracked Items with Exact Cost Reversing Mandatory Enabled: "Applies-to Entry must not be filled out when reservations exist in Item Ledger Entry Entry No.='XXX'"

Shikhverma:
[Bug 644262](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/644262): [Master][All-E] Purchase Return Order Posting Fails for Manually Reserved Lot-Tracked Items with Exact Cost Reversing Mandatory Enabled: "Applies-to Entry must not be filled out when reservations exist in Item Ledger Entry Entry No.='XXX'"InitialCommit

Fixes [AB#644262](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/644262)

Issue:
A lot-tracked purchase return (created via "Get Posted Document Lines to Reverse" and reserved against the receipt entry) cannot be posted when "Exact Cost Reversing Mandatory" is enabled — it fails with "Appl.-to Item Entry must have a value in Reservation Entry…", and the manual fix then fails with "Applies-to Entry must not be filled out when reservations exist…".

Cause:
Reserving the return line against the item ledger entry rebuilds the reservation entry without the exact-cost reversing link (Appl.-to Item Entry), which posting then requires.

Solution:
When the return line is fully reserved against an item ledger entry and Appl.-to Item Entry is blank, derive it from the reserved entry during posting, so the exact-cost link is restored without manual input.


v-ankitgoyal (approved):

PredragMaricic (approved):
The fallback derives the apply-to item entry from the opposite item-ledger reservation only in the exact-cost/item-tracking flow, preserving the existing behavior otherwise. The focused lot-tracked purchase-return test covers the reported regression. No blocking findings.

PR #9796, opened 2026-07-28
Show a message when an e-document file cannot be viewed or no data could be extracted

Explorer986:
## What & why

When a Payables Agent e-document came from a file that is not a valid invoice, or a corrupt PDF, clicking View PDF or View extracted data threw a runtime error. This adds small guards so both actions show a clear message when there is nothing to display or the file cannot be read.

## Linked work

Fixes [AB#638785](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/638785)

## How I validated this

- Corrupt PDF: View PDF now shows "the file cannot be displayed" message instead of the client render error.
- Valid non-invoice PDF: View extracted data shows a clear "no extracted data" message, and View PDF still opens the file.
- Normal invoice: both views work as before.

## Risk & compatibility

Low. Only adds guards and user messages on two view actions. No schema, data, or permission impact.

### DIFF
--- a/src/Apps/W1/EDocument/App/src/Processing/EDocImport.Codeunit.al
+++ b/src/Apps/W1/EDocument/App/src/Processing/EDocImport.Codeunit.al
@@ -329,8 +329,10 @@ codeunit 6140 "E-Doc. Import"
         EDocumentDataStorage: Record "E-Doc. Data Storage";
         IStructuredFormatReader: Interface IStructuredFormatReader;
     begin
+        if (EDocument."Structured Data Entry No." = 0) or (not EDocumentDataStorage.Get(EDocument."Structured Data Entry No.")) then
+            Error(NoExtractedDataErr);
+
         IStructuredFormatReader := EDocument."Read into Draft Impl.";
-        EDocumentDataStorage.Get(EDocument."Structured Data Entry No.");
         IStructuredFormatReader.View(EDocument, EDocumentDataStorage.GetTempBlob());
     end;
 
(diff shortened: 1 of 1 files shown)


src/Apps/W1/EDocument/App/src/Document/EDocument.Table.al
```diff
@@ -502,7 +502,11 @@ table 6121 "E-Document"
         EDocDataStorage: Record "E-Doc. Data Storage";
         IEDocFileFormat: Interface IEDocFileFormat;
     begin
-        EDocDataStorage.Get(Rec."Unstructured Data Entry No.");
+        if (Rec."Unstructured Data Entry No." = 0) or (not EDocDataStorage.Get(Rec."Unstructured Data Entry No.")) then begin
+            Message(NoSourceFileMsg);
```
mynjj: use `Error` instead, more consistent with similar validations throughout the product

src/Apps/W1/EDocument/App/src/Processing/Import/FileFormat/EDocPDFFileFormat.Codeunit.al
```diff
@@ -16,7 +16,37 @@ codeunit 6191 "E-Doc. PDF File Format" implements IEDocFileFormat
...
-        File.ViewFromStream(TempBlob.CreateInStream(), FileName, true)
+        if not IsDisplayablePdf(TempBlob) then begin
+            Message(CannotDisplayFileMsg);
+            exit;
+        end;
+
+        File.ViewFromStream(TempBlob.CreateInStream(), FileName, true);
+    end;
+
+    local procedure IsDisplayablePdf(TempBlob: Codeunit "Temp Blob"): Boolean
```
mynjj: we shouldn't dive into the internals of the PDF to gate this action, let's use the fact that it can't be processed by our pipeline (errored when attempting MLLM extraction). See what are the records inserted for that (the side effects of the failure) and propose a different way of gating this instead that relies on those records.

Alternatively, just add the gate to the other action, and let the View PDF fail as it is failing right now, it's not a bad error message. I think this is a better solution.

mynjj (changes requested):
See suggested changes.

src/Apps/W1/EDocument/App/src/Processing/Import/FileFormat/EDocPDFFileFormat.Codeunit.al
Explorer986: updated

src/Apps/W1/EDocument/App/src/Document/EDocument.Table.al
Explorer986: Thanks Joshua, updated.

mynjj (approved):

Groenbech96 (approved):

PR #9797, opened 2026-07-28
Bug 641388: Fix WIP transfer remainder over-creation when open line quantity is reduced

ChethanT:
## Summary
- Bug 641388: the regression test `WIPTransferRemainderCreatedWhenOpenLineQuantityReduced` (added by the bug 639382 fix) started failing once CI actually ran it, and was temporarily disabled to unblock the pipeline.
- Root cause: `GetOpenWIPTransferLineQtyBase` in `Subc. Create Transf. Order` summed `"Quantity (Base)"` on open (unposted) WIP transfer lines. But `"Qty. per Unit of Measure"` is force-set to `0` for WIP transfer lines (see `Subc. Transfer Line`.`OnValidate("Transfer WIP Item")`), which makes `"Quantity (Base)"` always evaluate to `0` for these lines regardless of the actual quantity. So after reducing an open WIP transfer line's quantity and re-running **Create Transfer Order to Subcontractor**, the existing (reduced) quantity was never counted as "already covered", and the report inserted a brand-new line for the full amount again instead of only the remainder.
- Fix: compute the open quantity from `Quantity` (which is correctly maintained) instead of the always-zero `"Quantity (Base)"`, and convert it to the item's base UOM ourselves using the same Purchase Line unit-of-measure factor used elsewhere in the report (`CalcPurchLineQtyBase` / `InsertWIPTransferLine`).
- Re-enabled the test by removing its entry from `src/DisabledTests/Subcontracting_Test/Subcontracting_Test.DisabledTest.json`.

## Test
Re-enabled and validated \WIPTransferRemainderCreatedWhenOpenLineQuantityReduced\ ([SCENARIO 639382], codeunit 149911). Also ran the full `Subc. WIP Trans. Create Test` (149911) suite (19 tests) — all pass, no regressions.

- Baseline (before fix): test FAILS — `Assert.AreEqual failed. Expected:<6> (Decimal). Actual:<8> (Decimal). Re-running Create Transfer Order must create the remaining WIP quantity after the open line was reduced.`
- After fix: test PASSES, and the full 19-test suite for codeunit 149911 passes.

Fixes [AB#641388](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/641388)

🤖 Generated with GitHub Copilot

### DIFF
--- a/src/Apps/W1/Subcontracting/App/src/Transfer/SubcCreateTransfOrder.Report.al
+++ b/src/Apps/W1/Subcontracting/App/src/Transfer/SubcCreateTransfOrder.Report.al
@@ -571,6 +571,7 @@ report 99001501 "Subc. Create Transf. Order"
         TransferLineToCheck.SetRange("Subc. Operation No.", PurchaseLine."Operation No.");
         TransferLineToCheck.SetRange("Derived From Line No.", 0);
         TransferLineToCheck.SetRange("Transfer WIP Item", true);
+        TransferLineToCheck.SetRange("Subc. Return Order", false);
         TransferLineToCheck.CalcSums(Quantity);
         if TransferLineToCheck.Quantity = 0 then
             exit(0);
--- a/src/Apps/W1/Subcontracting/App/src/Transfer/SubcCreateTransfOrder.Report.al
+++ b/src/Apps/W1/Subcontracting/App/src/Transfer/SubcCreateTransfOrder.Report.al
@@ -552,14 +552,17 @@ report 99001501 "Subc. Create Transf. Order"
             exit(false);
 
         PostedWIPQtyBase := GetWIPQtyBase(PurchaseLine, TransferToLocationCode);
-        OpenWIPLineQtyBase := GetOpenWIPTransferLineQtyBase(PurchaseLine);
+        OpenWIPLineQtyBase := GetOpenWIPTransferLineQtyBase(PurchaseLine, ProdOrderLine);
 
         exit((PostedWIPQtyBase + OpenWIPLineQtyBase) < ExpectedQtyBase);
     end;
 
-    local procedure GetOpenWIPTransferLineQtyBase(PurchaseLine: Record "Purchase Line"): Decimal
+    local procedure GetOpenWIPTransferLineQtyBase(PurchaseLine: Record "Purchase Line"; ProdOrderLine: Record "Prod. Order Line"): Decimal
     var
         TransferLineToCheck: Record "Transfer Line";
+        Item: Record Item;
+        UOMManagement: Codeunit "Unit of Measure Management";
+        QtyPerUom: Decimal;
     begin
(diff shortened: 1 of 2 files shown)


ChethanT:
## Pipeline investigation & follow-up fix

**Failing check found:** Classify team ownership (run [30397409117](https://github.com/microsoft/BCApps/actions/runs/30397409117/job/90403707851)) — the only failing check on this PR; not required for merge.

**Root cause:** This is a pull_request_target workflow (.github/workflows/ownership-classification.yml, executed from the base branch, not this PR's branch) that dispatches a workflow_dispatch event to microsoft/BCAppsTriage's ownership-classification.yml. At the time this run executed, that target workflow did not expose a workflow_dispatch trigger (HTTP 422: Workflow does not have 'workflow_dispatch' trigger). This is an external, transient infra issue unrelated to any code in this PR — confirmed because every run of the same workflow on other PRs/branches (including main) since that time has succeeded. Since the workflow definition is pulled from the base branch, no change to this branch could have affected it.

**Action taken:** Re-ran the failed job via the GitHub API (
erun-failed-jobs) — it now passes, confirming the upstream issue has since resolved and no code change was needed for this check.

## Additional fix (independent review finding)

While validating the existing bug fix, an independent code review pass flagged a latent filter gap exposed by the original fix: GetOpenWIPTransferLineQtyBase didn't exclude open WIP **return** transfer lines ("Subc. Return Order" = true), which share the same key fields as outbound WIP transfer lines. Before this PR's fix, that gap was harmless because CalcSums("Quantity (Base)") always evaluated to 0 for every WIP line regardless of match. Now that the fix correctly sums Quantity, an open return transfer line could be miscounted as open outbound quantity, masking the need to create the outbound remainder in some scenarios.

Pushed a follow-up commit (b12647084) adding SetRange("Subc. Return Order", false) to that filter, mirroring the direction filter already used on the return-order side (WIPReturnTransferLineAlreadyExists).

**Validation:** Compiled and published both the App and Test app via l-mcp-action.sh; ran the full Subc. WIP Trans. Create Test codeunit (149911) locally — **19/19 tests pass**, including the return-order scenarios (WIPReturnTransferOrderCreatedWithCorrectLocations, DeleteReturnTransferOrderAndRecreateSucceeds) and the original regression test (WIPTransferRemainderCreatedWhenOpenLineQuantityReduced).

🤖 Generated with GitHub Copilot

AleksandricMarko (approved):

JakovljevicDusan (approved):

PredragMaricic (approved):
No blocking findings. The open WIP quantity conversion mirrors the existing purchase-line base-quantity calculation, excludes return transfer lines, and re-enables the focused regression test.

qasimikram (approved):
The fix correctly accounts for reduced open WIP transfer quantities by converting the maintained Quantity value to base UOM, excludes return transfer lines, and re-enables the targeted regression test. The implementation is consistent with the report's existing purchase-line conversion and WIP insertion logic. All reported checks are successful.

PR #9802, opened 2026-07-28
Split the legacy test stage into four buckets

Groenbech96:
> **Stacked on #9736.** This branch contains #9736 plus the four-bucket change, so the diff below includes both until #9736 merges — **the reviewable delta is the last commit** (`perf(tests): split the legacy stage into four buckets`). It is stacked because the bucket lists reference the `Tests-ERM-*` / `Tests-SCM-*` apps that #9736 creates.
>
> It targets `main` rather than `copilot/split-tests-scm` for a practical reason: `.github/workflows/PullRequestHandler.yaml` only triggers on `pull_request` when the **base** is `main`, `releases/*` or `features/*`, so a PR stacked on a `copilot/*` branch gets no build at all.

## Summary

Splits the legacy test stage from two buckets into four. Measured on run `30426830747`: the
legacy critical path drops **82.9 → 49.1 min median (−41%)**, and **156.6 → 49.1 min (−69%)**
against the unsplit baseline. Cost is **+57% runner-hours**.

This is the largest single improvement in the series — larger than the `Tests-SCM` split, the
`Tests-ERM` split and the tenant increase delivered individually.

## Why buckets and not tenants

Buckets are **separate jobs on separate runners with separate containers**, so they run fully in
parallel — run `30354787707` used 44 distinct runners with Bucket1 and Bucket2 starting in the same
minute. A bucket adds *real, contention-free capacity*. A tenant does not: it adds concurrency inside
one container, and #9736 measured the cost of that directly — splitting `Tests-ERM` raised Bucket1's
concurrency and made every other app in that bucket 10–19% slower, inflating bucket work +27% for a
net −5%.

Both buckets were already tenant-bound before this change, so more tenants had nothing left to give.

## What

- `TestConfiguration.json` gains `LegacyTests-Bucket3` and `LegacyTests-Bucket4`.
- `.github/AL-Go-Settings.json` and all 22 `build/projects/Test Apps <country>/.AL-Go/settings.json`
  gain the two new build modes and their `bucketNumber` conditional blocks.
- Assignment is LPT over measured per-app times from run `30354787707`, each list ordered
  longest-first for the FIFO dispatch loop.

**No code changes were needed.** `Get-AppNamesForBucket` already resolves `"LegacyTests-Bucket$N"`
generically, and the Uncategorized path discovers every `LegacyTests-*` key.

| Bucket | apps | work | largest app |
|---|---|---|---|
| 1 | 12 | 194.4 min | `Tests-Workflow` 51.7 |
| 2 | 9 | 156.5 min | `Tests-SCM` 50.9 |
| 3 | 13 | 194.0 min | `AlCosting` 45.6 |
| 4 | 15 | 156.5 min | `Tests-SCM-Warehouse` 46.0 |

The split was constrained to keep every app in its original demo-data class — see
**[Finding](#finding-additionaldemodatatypes-is-a-canary-not-test-data)** below, which shows that
constraint does not actually exist.

## Measured result — run `30426830747`

88 legacy jobs, 22 per bucket. **18 of 22 countries produced all four buckets**; `AU`, `CZ`, `DK` and
`US` each lost one job (two runners died in `Checkout`, two hit the canary flake below) and are
excluded from the medians rather than reported understated.

Critical path is the test-phase wall time of the slowest bucket, read from the test-result artifacts,
so it excludes container setup and queueing.

| | median critical path | worst country |
|---|---|---|
| no split, 3 tenants, 2 buckets | 156.6 min | 180.7 min |
| SCM split, 6 tenants | 87.1 min | 98.7 min |
| SCM + ERM split (#9736) | 82.9 min | 106.5 min |
| **+ four buckets (this PR)** | **49.1 min** | **59.4 min** |

**−41% median and −44% worst-case against two buckets**; **−69% / −67% against the unsplit
baseline.** All 18 countries improved, and the spread narrowed from 145–181 min to 39–59 min.

### It beat the model, because the tests themselves got faster

The model predicted the split of *work*. It did not predict that the same tests would also run
faster, because each container now carries less concurrent load. Identical apps, identical test
counts, W1:

| App | 2 buckets | 4 buckets | |
|---|---|---|---|
| `Tests-Workflow` | 51.7 min | 31.6 min | **−39%** |
| `Tests-ERM-Application` | 34.8 min | 23.4 min | **−33%** |
| `Tests-SCM` | 50.9 min | 37.5 min | **−26%** |
| `Tests-SCM-Planning` | 29.3 min | 23.4 min | **−20%** |
| `AlCosting` | 45.6 min | 41.0 min | −10% |

Total W1 test work fell **701.5 → 661.0 min (−6%)** for exactly the same tests — the mirror image of
#9736, where adding tenants inside one container inflated work by +27%. This confirms the diagnosis
that the legacy stage was **contention-limited, not parallelism-limited**.

## Cost

| | legacy jobs | runner-hours | longest legacy job |
|---|---|---|---|
| 2 buckets (`30354787707`) | 44 | 129.2 | 214.7 min |
| **4 buckets (`30426830747`)** | **88** | **202.9** | **163.5 min** |

**+57% runner-hours for −24% on the longest legacy job.** Less than the doubled job count implies,
because each job is shorter, but it is the real price of this change. Per-bucket job wall medians:
150.3 / 134.8 / 143.5 / 141.9 min.

**This is the end of the bucket lever.** Setup now dominates — ~97 min of a ~142 min job is container
creation and app publish — so a fifth bucket would buy roughly 10 min of test phase against another
22 jobs of fixed overhead. Remaining gains are in reducing setup or in the pathological codeunits
(`AlCosting` is 41 min in a *single* codeunit).

## Finding: `additionalDemoDataTypes` is a canary, not test data

Found while triaging this PR's build, where `CZ`/`Bucket3` and `DK`/`Bucket3` failed setup with:

```
VAT period with the same starting date already exists.
"VAT Return Period CZL"(TableExtension 31066).OnBeforeInsert
```

`Invoke-AdditionalLegacyDemoData` in `build/scripts/ImportTestDataInBcContainer.ps1` creates a
**throwaway company**, runs the legacy DemoTool against it, and deletes it in `finally` — all *before*
`New-TestCompany`, which wipes every company and recreates the real one. Its docstring is explicit:
*"from a clean slate, this verification cannot interfere with the demo data the tests actually run
against."*

So the setting is a per-country validation canary with **no effect on the data tests run against**.
Two consequences:

1. **Bucket3 should not have inherited it.** Running the canary twice per country adds no signal and
   doubles exposure to a flaky path — 22 executions per build became 44, which caused the two
   failures above. `CZ`/`Bucket1` and `DK`/`Bucket1` passed the identical check in the same run, so
   the path is nondeterministic rather than broken.
2. **The demo-data "class" constraint this PR was built around does not exist.** Apps can be
   rebalanced across any bucket freely. Constrained, the four buckets are
   194.4 / 194.0 / 156.5 / 156.5; unconstrained they balance to **175.4 min each** — roughly another
   10% off the critical path.

**This also affects #9736.** That PR deliberately declined a full LPT rebalance of its two buckets,
on the stated grounds that moving an app between buckets would change the dataset it tests against.
That justification is wrong, and #9736 left about **9 min** of balancing on the table
(388.4 / 313.1 against a balanced 351 / 351). Its measured results are unaffected — only the
rationale for not rebalancing.

## TODO before merge

- [ ] Remove the `LegacyTestsBucket3` block from `.github/AL-Go-Settings.json` so the canary runs once per country again.
- [ ] Rebalance all four buckets with unconstrained LPT over measured per-app times (target ~175.4 min each).
- [ ] Correct the rebalancing rationale in #9736 and take the ~9 min there.
- [ ] Raise the CZ/DK Evaluation demo-data flake separately — the canary is correctly catching a real pre-existing localization bug.

## Tests

Adds six Pester tests over the bucket configuration. These failure modes are **silent**, which is why
they are worth guarding:

- an app listed in two buckets runs its tests twice;
- an app listed in **no** bucket does not error — it falls through to the **Uncategorized** build mode
  with a different `testType` and `companyName`.

Both new guards were negative-tested by injecting a duplicate app and a non-contiguous bucket number,
and both failed as intended. `Get-AppNamesForBucket` previously had no test coverage at all.

`build/scripts` Pester suite: **124/124 pass**.

## Follow-ups not in this PR

- **Re-tune `numberOfTenantsForTesting` downward.** With four buckets each bucket is *app-bound*
  rather than tenant-bound — `work/6` is 26–32 min against a largest app of 45.6–51.7 — so tenants 5
  and 6 are largely idle. 4–5 tenants should give the same floor with less contention.
- **Split `Tests-Workflow` and `AlCosting`.** In the two-bucket layout these had 30 and 38 min of
  slack and splitting them was worth nothing. At this bucket size they become the binding constraint
  for the first time. `AlCosting` is a single codeunit, so it needs a source-level split.


aholstrup1:
We considered it but it would increase the compute footprint quite a bit, so for now I don't really see it as an option.

PR #9807, opened 2026-07-29
Deduct already-reversed prepayment GST on AU sales prepayment credit memos

### LINKED ISSUE #8896: [Bug]: Incorrect Prepayment GST Calculation (Only in AU)
### Describe the issue

Credit / re-invoice for order lines with more than one prepayment will calculate incorrect GST when the final invoice is re-posted.

> [!NOTE]
> The bug only exists in AU.

### Expected behavior

Prepmt. VAT Base Amt. also needs to be updated here:
for Credit Memo Document Type (blue)
`SalesLine."Prepmt. VAT Base Amt." = SalesLine."Prepmt. VAT Base Amt." - TempOriginalSalesLine."Prepmt. VAT Base Amt."`
 and otherwise (Red)
`SalesLine."Prepmt. VAT Base Amt." = SalesLine."Prepmt. VAT Base Amt." + TempOriginalSalesLine."Prepmt. VAT Base Amt."`

Image

### Steps to reproduce

NOTE: need to have two lines on SO, so that the SO is not finalised and deleted when the first 'final' invoice is posted.
 

1. SO with two item lines 
    - ensure that GST is calculated for the order lines
3. Post prepayment invoice 1 (20% or both lines)
4. Post cash receipt and apply to prepayment invoice
5. Change prepayment to 50% on one line.
6. Post prepayment invoice 2
7. Post cash receipt and apply to prepayment invoice
8. Post final invoice for the item line with two prepayments    - select serial number
    - enter qty to ship, qty to invoice for one item line
8. SO is partially shipped and invoiced
    - GST on invoice is correct.

Create CR/Adj Note
9. Copy Document -> select final invoice
> [!NOTE]
> On the unposted credit, it looks like the item is credited at full amount, no prepayment reversal is made. 
> GST is for the full amount
> Preview posting does show the prepayment reversal entries
 
10. When the credit is posted, the prepayment line is generated
    - the credit then reflects the lesser amount.
    - Ledger entries are OK

> [!NOTE]
> Qty. to Ship and Qty. to Invoice are re-instated on the sales order line when the credit is posted.
> This is how we are able to do the posting of the second invoice and create the incorrect GST entries
 
11. Post final invoice 2 for the item line with two prepayments
    - select serial number
    - enter qty to ship, qty to invoice for one item line
    - GST for prepayment reversal is incorrect
    - Posted sales invoice shows correct amount
    - But the customer ledger entry and the GL Entries are incorrect

### Additional context

This is the issue [#1776](https://github.com/microsoft/BusinessCentralApps/issues/1776) from https://github.com/microsoft/BusinessCentralApps/
PR of Proposed Solution: [#1880](https://github.com/microsoft/BusinessCentralApps/pull/1880/changes/8202f018ba279a61f93fa38b2c987efec5d28be8) from https://github.com/microsoft/BusinessCentralApps/

- [ ] I will provide a fix for a bug

Franco111000:
#### What and why

AU only: a sales prepayment credit memo posted after a partial final invoice reverses the full prepayment GST instead of only the part that has not been deducted yet, so the customer ledger entry and the G/L entries carry too much GST when the order is re-invoiced.

The APAC layer already has a mechanism for exactly this: during final invoice posting, `SavePrepmtVATDeducted` collects the GST reversed by the prepayment deduction lines into a buffer, and the credit memo read sites subtract Sales Line fields 28006 `Prepmt. VAT Amount Deducted` and 28007 `Prepmt. VAT Base Deducted` from the prepaid GST (`SalesPostPrepayments` 1200-1201 and 1250-1251, gated by Full GST on Prepayment plus the credit memo document type). On the purchase side the mechanism is complete: `Purch.-Post` drains its buffer onto the purchase lines through `DividePrepmtVATDeducted` and `DistributeAmount` (7488-7545, called at 8428). On the sales side only the first half exists: the buffer is filled (`SalesPost` 7769, called 1156) and nothing ever drains it, so both fields stay at zero for the life of every document and the subtraction never subtracts anything.

This PR completes the sales side by porting `DividePrepmtVATDeducted` and `DistributeAmount` from `Purch.-Post` and calling the former from `PostUpdateOrderLine`, mirroring the purchase call site. Three deliberate deviations from a verbatim port, each needed for the port to be correct:

1. **Sign normalisation in `SavePrepmtVATDeducted`.** The prepayment deduction lines are built with Quantity = -1 on both sides, so their amounts are negative. The purchase saver negates them into positive values (`PurchPost` 7493-7494); the sales saver did not. Both credit memo read sites subtract the fields, so the stored values must be positive. The sales saver now negates exactly like the purchase one. This edit is inert for existing behaviour: the two buffer fields it feeds had no reader on the sales side until this PR.
2. **Per-line reset of `DeductedVATBaseAmount` in `UpdateVATOnLines`.** The variable is procedure-scoped, assigned only on the Full GST credit memo path, and consumed on every Full GST line. While field 28007 was always zero this was harmless; with the field populated, one line's deducted base would leak into the next line in the same run. Reset to zero at the top of each iteration.
3. **Clean cycle on consumption.** The credit memo branch of `UpdateSalesDocument` already rebases `Prepmt. Amt. Incl. VAT` and `Prepayment Amount` and clears the other per-cycle prepayment fields (2000-2004). Both deducted fields are cleared there too, so a prepayment that is re-invoiced after a credit memo starts a clean deduction cycle instead of subtracting the previous cycle's deduction again. The purchase side does not do this and keeps document-lifetime accumulators; aligning purchase the same way may be worth a follow-up, but changing purchase behaviour is out of scope here.

One more small deviation: the ported procedure resets the line number buffer before deleting it. The purchase original deletes that buffer while the filter from the last loop iteration is still active, which only clears the last group.

Out of scope: the fix proposed in the issue description (extending `RestorePrepmtAmounts` with `Prepmt. VAT Base Amt.`) is not part of this change. That procedure saves and restores a field that `UpdateVATOnLines` deliberately rewrites per posting, and under Full GST the rewrite keeps the field unchanged from the second prepayment onward, so an unconditional restore would double it. The root cause analysis is on the issue.

#### Linked work

Fixes #8896

Area label: the issue carries **Finance**. I do not have label permissions, so a maintainer would need to set it.

#### How I validated

- [x] Traced the full mechanism on both sides: buffer fill (`SalesPost` 1156/7769 vs `PurchPost` 1210/7488), drain (`PurchPost` 7510/8428, sales previously absent), read sites (`SalesPostPrepayments` 1200/1250, `PurchasePostPrepayments` 866/916), and the temp-to-database flush (`ModifyTempLine`, wholesale `TransferFields`, so the drained values persist).
- [x] Verified the global `Currency` used by `DistributeAmount` is initialised before `PostUpdateOrderLine` runs (`GetCurrency` at 346, `FinalizePosting` at 372), matching the purchase flow.
- [x] Confirmed fields 28006/28007 have no other reader or writer anywhere in the repository, so behaviour changes only on the Full GST prepayment credit memo path.
- [x] Added a regression test (`GSTOnSalesPrepmtCrMemoAfterPartialFinalInvoice` in `ERM GST On Prepayments II`): prepayment invoice, paid, partial final invoice, then prepayment credit memo under Full GST on Prepayment. It asserts the deducted GST amount and base on the order line against the posted deduction lines, and that the credit memo GST equals the prepaid GST minus the deducted GST. The existing suite cannot catch this defect: every current Full GST credit memo test posts the credit memo before any final invoice, so the fields are still zero at the read sites either way.
- [ ] Full test suite run is pending the workflow approval on this PR; I could not execute the AU suite locally.

#### Risk & compatibility

- Behaviour radius: posting results change only for APAC builds with Full GST on Prepayment enabled for the line's VAT posting setup, on a sales prepayment credit memo posted after a final invoice. Both read sites are gated by `IsFullGST` plus the credit memo document type, and nothing else reads these fields.
- Data radius: order lines that deduct a prepayment now carry values in fields 28006/28007 (previously always zero). The fields are `Editable = false` and appear on no page, report, or API in the repository.
- The existing Full GST tests that post two prepayment invoices or a prepayment credit memo without a prior final invoice are unaffected: on those paths the fields are zero at the read sites, which is exactly today's behaviour.

### DIFF
--- a/src/Layers/APAC/BaseApp/Sales/Posting/SalesPost.Codeunit.al
+++ b/src/Layers/APAC/BaseApp/Sales/Posting/SalesPost.Codeunit.al
@@ -7771,8 +7771,8 @@ codeunit 80 "Sales-Post"
         if SalesLine.Type <> SalesLine.Type::" " then
             if SalesLine."Prepayment Line" then begin
                 TempPrepmtVATAmtBuf.Get(0, '', SalesLine."Line No.");
-                TempPrepmtVATAmtBuf.Amount := SalesLine."Amount Including VAT" - SalesLine.Amount;
-                TempPrepmtVATAmtBuf."VAT Base Amount" := SalesLine."VAT Base Amount";
+                TempPrepmtVATAmtBuf.Amount := -SalesLine."Amount Including VAT" + SalesLine.Amount;
+                TempPrepmtVATAmtBuf."VAT Base Amount" := -SalesLine."VAT Base Amount";
                 TempPrepmtVATAmtBuf.Modify();
             end else
                 if not SalesLine."System-Created Entry" then
@@ -7786,6 +7786,44 @@ codeunit 80 "Sales-Post"
                               TempPrepmtVATAmtBuf."Prepmt Amt to Deduct" + SalesLine."Prepmt Amt to Deduct";
                             TempPrepmtVATAmtBuf.Modify();
                         end;
+    end;
+
+    local procedure DividePrepmtVATDeducted(var SalesLine: Record "Sales Line")
+    var
+        PrepmtVATReminder: Decimal;
+        PrepmtVATBaseReminder: Decimal;
+        Ratio: Decimal;
+    begin
+        if TempPrepmtVATAmtBuf.FindSet() then
+            repeat
+                TempPrepmtLineNoBuf.SetRange("New Line Number", TempPrepmtVATAmtBuf."Line No.");
+                if TempPrepmtLineNoBuf.FindSet() then
+                    repeat
+                        SalesLine.Get(SalesLine."Document Type", SalesLine."Document No.", TempPrepmtLineNoBuf."Old Line Number");
+                        Ratio := SalesLine."Prepmt Amt to Deduct" / TempPrepmtVATAmtBuf."Prepmt Amt to Deduct";
+                        DistributeAmount(
+                          TempPrepmtVATAmtBuf.Amount, Ratio,
+                          SalesLine."Prepmt. VAT Amount Deducted", PrepmtVATReminder);
+                        DistributeAmount(
+                          TempPrepmtVATAmtBuf."VAT Base Amount", Ratio,
+                          SalesLine."Prepmt. VAT Base Deducted", PrepmtVATBaseReminder);
+                        SalesLine.Modify();
+                    until TempPrepmtLineNoBuf.Next() = 0;
+            until TempPrepmtVATAmtBuf.Next() = 0;
+        TempPrepmtVATAmtBuf.DeleteAll();
+        TempPrepmtLineNoBuf.Reset();
+        TempPrepmtLineNoBuf.DeleteAll();
+    end;
+
+    local procedure DistributeAmount(AmountToDistibute: Decimal; Ratio: Decimal; var Amount: Decimal; var Reminder: Decimal)
+    var
+        DistibutedAmount: Decimal;
+        DistibutedAmountRnded: Decimal;
+    begin
+        DistibutedAmount := Reminder + AmountToDistibute * Ratio;
+        DistibutedAmountRnded := Round(DistibutedAmount, Currency."Amount Rounding Precision");
+        Reminder := DistibutedAmount - DistibutedAmountRnded;
+        Amount := Amount + DistibutedAmountRnded;
     end;
 
     local procedure UpdateTaxForPostedDoc(var SalesHeader: Record "Sales Header")
@@ -8608,6 +8646,7 @@ codeunit 80 "Sales-Post"
         OnBeforePostUpdateOrderLine(SalesHeader, TempSalesLineGlobal, SuppressCommit, SalesSetup);
 
         ResetTempLines(TempSalesLine);
+        DividePrepmtVATDeducted(TempSalesLine);
         TempSalesLine.SetRange("Prepayment Line", false);
         TempSalesLine.SetFilter(Quantity, '<>0');
         OnPostUpdateOrderLineOnAfterSetFilters(TempSalesLine);
--- a/src/Layers/APAC/BaseApp/Sales/Posting/SalesPostPrepayments.Codeunit.al
+++ b/src/Layers/APAC/BaseApp/Sales/Posting/SalesPostPrepayments.Codeunit.al
@@ -1172,6 +1172,7 @@ codeunit 442 "Sales-Post Prepayments"
         PrepmtAmtToInvTotal := SalesLine."Prepmt. Line Amount" - SalesLine."Prepmt. Amt. Inv.";
         if SalesLine.FindSet() then
             repeat
+                DeductedVATBaseAmount := 0;
                 PrepmtAmt := PrepmtAmount(SalesLine, DocumentType);
                 if PrepmtAmt <> 0 then begin
                     if IsFullGST(SalesLine) then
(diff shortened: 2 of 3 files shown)


alexei-dobriansky:

<details>

This PR completes the AU sales prepayment GST deduction path by writing the already-deducted prepayment GST amount and base onto the sales line, then subtracting those values when a sales prepayment credit memo is posted. The main posting change mirrors the existing APAC purchase flow, including the same buffer split and rounding pattern, and the sign change in `SavePrepmtVATDeducted` matches the purchase-side convention. The reset in `UpdateVATOnLines` also looks needed because the deducted base is line-specific. I did not find an event or public API signature change. The code path is plausible, but the regression test stops before the final posted-entry symptom from the linked bug, which leaves a sensitive posting result unproven.
</details>

<details>

**S1** - *Regression test stops before reported posting error*
Please extend the new regression test to continue through the re-invoice step and assert the posted Customer Ledger Entry and G/L/GST entries. The linked bug reports the wrong posted entries after that second final invoice, but the current test stops at the prepayment credit memo amount. This is posting and GST logic, so the test should cover the final financial entries before merge.
</details>

<details>

**Risk:** The changed path affects AU Full GST on Prepayment posting amounts for sales orders with prepayment credit memos after a partial final invoice. The implementation is narrow and reuses the purchase-side mechanism, but it changes stored sales-line deduction fields that were previously always zero and feeds posted tax amounts. I found no linked `AB#` ADO work item in the PR or linked GitHub issue, so I used GitHub issue #8896 as the bug source.

**Necessity:** The scenario is valid and important because incorrect GST can reach customer ledger entries and G/L entries. The scope is appropriate for a bug fix, but the test must prove the reported final posting result, not only the intermediate credit memo calculation.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9807 round=1 by=alexei-dobriansky at=2026-08-03T10:08:13.420Z lastSha=02862454c5eb84583b52322d00fb6ed12551b5c6 suggestions=S1

Franco111000:
Thanks @alexei-dobriansky, S1 addressed in f05253b85b.

The test no longer stops at the prepayment credit memo. It now verifies the credit memo clears both deducted fields, re-invoices the prepayment, posts the remaining final invoice, and asserts the posted result three ways:

1. The prepayment GST reversed by the final invoice equals the GST charged by the re-invoiced prepayment, both measured from the posted documents.
2. The customer ledger entry amount matches the posted invoice amount, which is exactly the invariant the issue reports as broken ("Posted sales invoice shows correct amount. But the customer ledger entry and the GL Entries are incorrect").
3. The G/L credit total and the item line GST entry carry the remaining half amounts, asserted through the same VerifyGLAndGSTSalesEntry helper the sibling partial-invoice test uses.

The continuation also directly exercises the clean-cycle reset in UpdateSalesDocument: without it, the second final invoice would subtract the first cycle's deduction again, and assertion 1 would fail.

On the missing AB# reference: the PR is linked to GitHub issue #8896 only; I have no ADO work item for it.

alexei-dobriansky:

# **Recommendation: Accept**

<details>

The new commit extends the regression test past the prepayment credit memo into the re-invoice cycle. It now verifies that the deducted GST fields are cleared before re-invoicing, that the second final invoice reverses the GST from the re-invoiced prepayment, and that the posted Customer Ledger Entry plus G/L/GST entries match the expected remaining amounts. This directly covers the gap from round 1 and exercises the clean-cycle behavior introduced by the fix.
</details>

<details>

| S1 | Regression test stops before reported posting error | Addressed | The reply says the test now continues through re-invoicing and checks the customer ledger, G/L, and GST entries. The incremental diff confirms that coverage was added. |
</details>

<details>

*None - the new commit only addresses the prior test coverage blocker.*
</details>

<details>

**Risk:** The production-code risk is unchanged from round 1: this is narrow AU Full GST on Prepayment posting logic for sales prepayment credit memos after a partial final invoice. The new test reduces regression risk by checking the later posted entries that the linked GitHub bug reports. I still found no linked `AB#` ADO work item, so this review uses GitHub issue #8896 as the bug source.

**Necessity:** The scenario remains important because incorrect GST can reach customer ledger entries and G/L entries. The scope is still appropriate for a bug fix, and the added test now proves the final posting result that was previously missing.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9807 round=2 by=alexei-dobriansky at=2026-08-03T17:10:49.713Z lastSha=f05253b85b7b7e8d19a5843ce0fdfb32be221644 suggestions=S1:addressed parentRound=1

PR #9810, opened 2026-07-29
[master]-Edit in Excel fails with false duplicate error when creating new Item Variants

MSNehaNawal:
[Bug 644606](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/644606): [ALL-E] Edit in Excel fails with false duplicate error when creating new Item Variants

Fixes [AB#644606](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/644606)

Issue - Edit in Excel fails with false duplicate error when creating new Item Variants

Cause-PR 240753 (Item Variant Attributes feature) added an OnValidate trigger to the Code field on the Item Variants page (5401) that calls CurrPage.Update(true) when a Code is first entered on a new row. This forces the record to be persisted immediately.
During Edit in Excel, publishing a new variant runs through the OData insert handler. The forced CurrPage.Update(true) inserts the record first, then the OData handler attempts to insert the same record again, producing a false duplicate-key error.

Solution-Scope the forced update so it runs only for interactive clients and is skipped for OData/web-service sessions (Edit in Excel is OData V4).

### DIFF
--- a/src/Layers/W1/BaseApp/Inventory/Item/ItemVariants.Page.al
+++ b/src/Layers/W1/BaseApp/Inventory/Item/ItemVariants.Page.al
@@ -7,6 +7,7 @@ namespace Microsoft.Inventory.Item;
 using Microsoft.Inventory.Item.Attribute;
 using Microsoft.Inventory.Item.Catalog;
 using Microsoft.Inventory.Item.Picture;
+using System.Environment;
 using System.Text;
 
 page 5401 "Item Variants"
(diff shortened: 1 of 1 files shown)


v-ankitgoyal (approved):

PredragMaricic (approved):

PredragMaricic (changes requested):
S1 (High): Add regression coverage for the noninteractive insertion path. This fix depends on client-type detection to avoid CurrPage.Update(true) during Edit in Excel/OData publishing, but the PR has no test proving that a new Item Variant is inserted exactly once through that path while interactive page behavior remains intact.

PredragMaricic (approved):

PR #9816, opened 2026-07-29
Reuse apps already published in the BC artifact database

Groenbech96:
## What & why

Every `Test Apps *` build restores the BC artifact database — in which all apps are already published, installed and synchronized — and then throws it away, so AL-Go republishes ~300 apps one at a time before a single test runs. `Install-App Packages` alone takes ~46 minutes.

```
# build/scripts/NewBcContainer.ps1 (before)
# Clean the container for all apps. Apps will be installed by AL-Go
foreach($app in $installedApps) { UnInstall-BcContainerApp ...; Unpublish-BcContainerApp ... }
```

The apps in the artifact are built by the NAV build from a known BCApps commit. The companion NAV change records it in the artifact manifest as `bcAppsCommit`, so this PR can diff that commit against the commit being built, work out which apps actually changed, and refresh **only** those.

**Retained apps are uninstalled but left published and synchronized** — exactly the state AL-Go's publish step produces today. That keeps two invariants intact: `ImportTestDataInBcContainer.ps1` retains full control of install order (which the `Legacy` type depends on), and test discovery — driven by `runTestsInAllInstalledTestApps` from the *installed* apps — is unchanged. Only the publish/sync work is skipped.

No extra storage, nothing to upload or invalidate: the artifact is downloaded and restored by the build anyway.

**Off by default.** Nothing changes until `"useArtifactBaseline": true` is set in `.github/AL-Go-Settings.json`.

Design, rejected storage alternatives and the full fallback matrix: `build/scripts/ArtifactBaseline.md`.

## Only the owning app, never its dependents

The first revision of this PR expanded the refresh set down the dependency graph, via `BuildOptimization`'s `Get-AffectedApps`, and also consulted `fullBuildPatterns`. Both were the wrong tool.

Those exist to answer **"which tests must we run?"** — a change in a dependency can change a dependent's behaviour, so expanding downstream is correct there. **"Which published binaries are still valid?"** is a different question. An app whose own source did not change has identical source at both commits, is already published, installed and synchronized in the restored database, and still resolves its dependencies because BCApps pins them at `<major>.<minor>.0.0`, so a refreshed Base Application at `29.0.2147483647.x` still satisfies it.

Measured on a real 16-commit, 3-day window (`1e77e9c2..28ff6645`, 432 changed files, 841 apps):

| | apps to republish | reused |
| --- | --- | --- |
| expanded to dependents (previous revision) | 841 | **0 (0.0%)** |
| the app that owns the change | **10** | **831 (98.8%)** |

`src/Layers/W1/BaseApp` changed in that window, and everything depends on Base Application. Base Application changes in nearly every multi-day window, so the previous revision would have fired approximately never — on top of `fullBuildPatterns` vetoing the whole thing on `build/Packages.json` and `src/DisabledTests/*` before any app mapping happened.

Per container, after intersecting with the apps the container actually holds:

```
Test Apps W1   chain W1        container 232 apps    republish  8    keep 224
Test Apps DK   chain W1 -> DK  container 258 apps    republish 10    keep 248
```

## Country layers go through the view chain

A country layer folder carries no `app.json`. `PreCompileApp.ps1` calls `New-GDLView`, which composes `src/Views/<CC>` from a chain declared in `src/Layers/.config/views_config.json`, and compiles the localized app from that view:

```
W1   -> [W1]           DK -> [W1, DK]
NA   -> [W1, NA]       US -> [W1, NA, US]
DACH -> [W1, DACH]     AT -> [W1, DACH, AT]
```

A change to `src/Layers/<L>/<AppFolder>/**` now marks `<AppFolder>` for republish **only when `L` is in `chain(CC)`**, resolving app identity by walking the chain from the most specific layer back to the base. For `Test Apps W1` the RU/IT/ES/NA/APAC layer changes (316 of the 432 files in the example) are genuinely irrelevant; for `Test Apps DK` a `src/Layers/DK/BaseApp` change *does* invalidate that container's Base Application. Ignoring country layers wholesale would publish a stale localized Base Application — a correctness bug, not a missed optimization.

The country comes from the artifact url (`.../sandbox/29.0.53247.0/w1` → `w1`), which is the country the container is built for.

## What is deliberately not an invalidator

A retained app is **never compiled by this build** — its binary is fixed inside the artifact — so no repository-side setting can retroactively change it. Rulesets and disabled-test lists are diagnostic and test-selection inputs: they can fail a build, they cannot make a published `.app` wrong.

The artifact and platform pins matter most here. In the example window they are the only `build/**` and `.github/**` changes:

```
.github/AL-Go-Settings.json   artifact   29.0.53094.0 -> 29.0.53221.0
build/Packages.json           BCPlatform 29.0.53093.0 -> 29.0.53216.0
```

That is `UpdateBCArtifactVersion` housekeeping, roughly weekly. Bumping a pin downloads a **newer** artifact carrying a **newer** `bcAppsCommit`, which makes the diff *smaller*. Treating those files as a full-refresh trigger meant the commit that improves this optimization was the one that switched it off.

## Linked work

⚠️ **No approved issue yet** — this is a draft and needs one before review.

Companion change: NAV PR `251642`, **shipped 2026-07-31**. Verified in a published artifact: `sandbox/29.0.53247.0/w1` and `.../dk` both carry `"bcAppsCommit": "1e77e9c2e8d4c58e45e32be2876550dbb68b773c"`, which is a real commit on BCApps `main`. The repo-level `base` artifact is not stamped, but no `Test Apps` project uses it — `Test Apps W1` sets `country: w1` and each localization project sets its own.

## How I validated this

- [x] I built the affected app(s) locally with no new analyzer warnings. — *n/a: no AL app is touched, this is build tooling only.*
- [x] I ran the change in Business Central and confirmed it behaves as expected. — **not done. This is the main gap; see below.**

- **65 Pester tests** in `build/scripts/tests/ArtifactBaseline.Test.ps1`, all passing: commit resolution, country extraction from the artifact url, layer-chain resolution (`W1`, `DK`, `AT`, `US`, unknown), the explicit *no dependent expansion* contract, per-country layer attribution in both directions, the inert-path list, unattributable and unrecognized paths, the ratio guard, the unpublish/keep split, stale-state rejection and every fallback.
- **Full `build/scripts` suite: 181/182.** The one failure is the pre-existing `BuildOptimization` "50 vs 51 affected apps" assertion — it fails identically on an untouched `main` (a newly added `Payables Agent` app), and is untouched here.
- **Replayed against the real delta**: the artifact commit read from the live `w1` manifest, diffed against `main`, produces the 10 apps above and the 8/10 per-container split.
- Verified against a real artifact on disk that app names match repo `app.json` names, and that BCApps dependencies are pinned at `<major>.<minor>.0.0` so a reused app at the artifact's version still satisfies apps built at `29.0.2147483647.x`.
- **Not yet done: a real CI run.** This code path has still never executed in CI. The actual saving, and `-upgrade`/install behavior against artifact-installed apps, must be measured on a real `Test Apps` build before this is enabled.

## Risk & compatibility

- **Everything is behind `useArtifactBaseline`, which is not set.** With it unset, both hooks behave exactly as before.
- **Fallbacks to today's full clean**: no commit in the manifest; commit unreachable (a shallow CI clone is fetched once, a complete clone is never made shallow); `git diff` failure; a build mode that changes compilation (derived from `conditionalSettings.preprocessorSymbols`, i.e. `Clean`); a changed `src/` file that maps to no app; a layer change whose country has no resolvable view chain; an unrecognized path; or >75% of apps changed.
- **Provenance**: reused apps are the NAV-built binaries at the artifact version rather than the ones this build compiled — same source commit, different compiler invocation. Anything asserting on exact app versions would notice.
- **Container parity**: apps this repo does not produce, and apps that must never be in a test container (`Library - No Transactions`, `Prevent Metadata Updates Library`), are always unpublished, so the container holds what it holds today.
- **Per-country validation.** The chain rule is exercised by unit tests for W1 and DK, but only a real CI run per country will prove the localized Base Application is refreshed when it should be. Enable W1 first.
- **Freshness**: the payoff shrinks as the artifact ages. `UpdateBCArtifactVersion` already keeps it current — and, unlike the previous revision, that bump no longer disables the optimization.
- **Follow-up (out of scope)**: demo data generation is now the biggest remaining fixed cost, and the artifact database already contains a populated CRONUS company.

[AB#644840](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/644840)


Groenbech96:
Abandoning. Without very fresh artifacts, it would not work.

PR #9836, opened 2026-07-29
Reset legacy test container DB in bulk instead of per-app uninstall loop

aholstrup1:
## What & why

Legacy test jobs spend ~19-24 min of every container setup uninstalling and unpublishing the ~150 apps pre-installed in the BC image, one at a time, before AL-Go publishes the repo-built apps. The bottleneck is not schema work but per-app container-session overhead (~8s each round-trip), so it cannot be optimized by tuning the loop.

This replaces that loop with a single bulk operation: drop and recreate the application database, which removes every app at once, then rebuild only what AL-Go needs to take over.

## Approach

New `build/scripts/ResetBcContainerDatabase.psm1` exposes `Reset-BcContainerApplicationDatabase`, which:
- extracts the System Application from the container's own published copy first (BCApps overrides the platform artifact, which strips System App from disk, so this is the only source that survives),
- drops the app + tenant databases and creates an empty application database,
- re-imports the developer license (dropped with the DB), re-publishes System App (required for the SUPER permission set under `UsePermissionSetsFromExtensions=true`), and creates a SUPER user,
- restores multitenancy with a fresh, empty `default` tenant.

`build/scripts/NewBcContainer.ps1` resolves the license deterministically from the artifact manifest and calls the reset. The existing removal loop stays but now runs over the single remaining System App.

## Linked work

Fixes [AB#646519](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/646519)

## How I validated this

- Ran the reset end to end on a fresh multitenant BC 28 container locally: full reset in ~3.6 min, ending with the default tenant Operational, System App installed, Multitenant=True, and `AllowAppDatabaseWrite=True`.
- Validated in CI (run 30431641630): the reset succeeds and 15,020 tests run with no read-only / DemoTool / Media-insert errors. On the same DE LegacyTestsBucket2 job, container creation dropped from 24.3 min to 9.8 min (about -14.5 min, -60%). Scaled across 110 self-hosted jobs that is roughly -27 compute-hours per full build.
- Two issues found and fixed during local validation: the default tenant must be mounted with `-allowAppDatabaseWrite` (otherwise demo-data generation cannot insert Media records), and the in-container database scriptblocks must run under Windows PowerShell 5.1 (`-usePwsh $false`) because the NAV management cmdlets / `Invoke-Sqlcmd` hit .NET assembly-load conflicts under the container's pwsh7.
- No AL tests added: this is a CI/build-infrastructure change to PowerShell setup scripts, not product code.

Note: the remaining known-flaky test failures in this run are pre-existing and are tolerated on real PR builds (the Test Tolerance feature only applies when `GITHUB_EVENT_NAME == 'pull_request'`, not on the `workflow_dispatch` run used above).

## Risk & compatibility

- Only affects the CI container-setup path (`build/scripts`); no product code changes.
- Behavior depends on multitenant containers and on the developer license shipping inside the BC artifact (resolved via `manifest.json`). The reset throws early if the System App is not published or the license is missing, so a misconfigured container fails fast rather than silently.
- End state for AL-Go is unchanged (empty writable default tenant, no repo apps published yet); the downstream publish/test phases are untouched.


spetersenms (dismissed):

aholstrup1:
Closing this without merging.

The optimization worked as designed (bulk application-database drop instead of the ~150-app uninstall/unpublish loop, verified: 2 unpublish operations instead of 288), but the performance case no longer holds.

When this was measured in July, container creation was ~24 min and the bulk reset brought it to ~9.8 min, about -14.5 min per job. Re-measured today against a current main CI/CD run (32083412334), the runners have gotten much faster and the old per-app loop now costs only ~11.5 min:

| Phase | This branch | main baseline |
| --- | --- | --- |
| Creating Container | 653 s | 688 s |

That is a ~35 second difference. Whole-job comparisons across DE Bucket1/Bucket2, FR Bucket2 and AU Default ranged from -7 min to +5 min, which is within normal run-to-run variance.

Since the remaining benefit does not justify the added complexity (bulk reset module, license re-import, System Application extract/republish, SUPER user recreation, pre-created company for per-company install triggers), it is not worth maintaining. Closing.

AB#646519

PR #9482, opened 2026-07-15
Make service-invoked System Application facades callable by any user (Privacy Notice + Page Summary Provider)

onbuyuka:
**Work item:** [AB#641612](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/641612)

## Summary

Two System Application public **facade** codeunits are reachable from system-trigger subscribers in low-permission (web-service / OData / MCP / Copilot / agent) sessions, but declare only `Access = Public` — no `InherentEntitlements` / `InherentPermissions` — even though the impl codeunits they delegate to declare both. Entering such a facade requires the calling user to hold Execute permission on it; regular (non-SUPER) users don't, so they hit `NavPermissionException` (`18023703`). This makes them callable by any authenticated user by declaring the inherent grants, matching their impls.

| Facade (fixed) | Impl (already inherent `X`) | Reached from |
| --- | --- | --- |
| CU 1563 `Privacy Notice` | CU 1565 `Privacy Notice Impl.` | `Copilot Capability Impl.IsCapabilityActive` -> privacy notice, via `System Action Triggers.GetCopilotCapabilityStatus`/`GetCopilotCapabilityInfo` |
| CU 2718 `Page Summary Provider` | CU 2717 `Page Summary Provider Impl.` | `Agent Task Impl` subscriber to `System Action Triggers.GetPageSummary`; also directly OData/web-service exposed |

## Problem (Privacy Notice)

Non-admin users hit `NavPermissionException 18023703` ("CodeUnit 1563 Privacy Notice Execute: System Application") when the platform evaluates a Copilot capability:

```
Codeunit 7774 "Copilot Capability Impl".IsCapabilityActive
  -> Codeunit 1563 "Privacy Notice".GetPrivacyNoticeApprovalState   <-- fails entering here
     -> Codeunit 1565 "Privacy Notice Impl.".CheckPrivacyNoticeApprovalState
```

It surfaced via the platform MCP feature gate, which evaluates a Copilot capability for every MCP request.

## Root cause

Along each chain, both the caller/subscriber and the impl declare inherent execute, but the **public facade in the middle does not**:

| Codeunit | InherentEntitlements / InherentPermissions |
| --- | --- |
| 7774 `Copilot Capability Impl` (caller) | `X` / `X` |
| **1563 `Privacy Notice`** (facade) | **neither — only `Access = Public`** |
[...]

PR #9837, opened 2026-07-29
Grant inherent read on Privacy Notice tables for service-invoked capability checks

onbuyuka:
## Problem

A follow-up to #9482. That PR fixed the facade **Execute** denial on the MCP capability path; this fixes the next denial one level deeper — a **tabledata Read** (`NavPermissionException 18023807`) on the **Privacy Notice** platform table.

The MCP feature gate calls `GetCopilotCapabilityStatus` during **session initialization, before the company is opened** (`McpServerInitializationMiddleware` runs `ThrowIfNotEnabledAsync` ahead of `InitializeSessionAsync` → `OpenCompanyAsync`). Role / permission-set permissions resolve against the *open* company; with no company yet, they collapse to `None`. The capability path then reads:

- `Privacy Notice` (2000000237) via `Privacy Notice Impl.CheckPrivacyNoticeApprovalState` (`PrivacyNotice.Get`), and
- `Privacy Notice Approval` (2000000238) via the `Enabled`/`Disabled` FlowFields (`CalcFormula = exist("Privacy Notice Approval" ...)`).

Both are denied for any non-SUPER user. Telemetry confirms: `Entitlements for object TableData 2000000237: Read, Insert, Modify, IndirectDelete. Role permissions: None. ... IsTenantAdmin: False` — the entitlement allows Read, but nothing supplies the grant because roles are `None`.

## Why inherent permissions (and not a permission set)

Because roles resolve to `None` in this pre-company context, a permission-set grant can't help — it's never evaluated. **Inherent permissions** are capped by license ∩ entitlement only (no role/company dependency), so the object self-grants within what the entitlement already allows.

`Privacy Notice` / `Privacy Notice Approval` are **platform tables** and can't declare table-level inherent permissions. (By contrast, `Copilot Settings` (7775) — also read on this path — already declares `InherentPermissions = rimdX` at the table level and needs no change; that's why its read already succeeds.) So the read is granted from the codeunit instead.

## Fix

On `Privacy Notice Impl.` (1565), matching the existing pattern in `User Login Time Tracker Impl.` (9013) and `Azure AD Plan Impl.` (9018):

- `[InherentPermissions(PermissionObjectType::TableData, Database::"Privacy Notice", 'r')]` and the same for `"Privacy Notice Approval"` on `CheckPrivacyNoticeApprovalState` — the self-sufficient, entitlement-capped grant, collected as the read executes in that method's scope.
- `Permissions` property: `"Privacy Notice" = im` → `rim`, and add `"Privacy Notice Approval" = r` — the indirect read the check ANDs against.

## Notes
- Work item: [AB#641612](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/641612)
- Diagnosed from a production session trace (MCP Copilot Chat init, `WebServiceClient`, non-admin user).

### DIFF
--- a/src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
+++ b/src/System Application/App/Privacy Notice/src/PrivacyNoticeImpl.Codeunit.al
@@ -13,7 +13,8 @@ codeunit 1565 "Privacy Notice Impl."
     InherentEntitlements = X;
     InherentPermissions = X;
     Permissions = tabledata Company = r,
-                  tabledata "Privacy Notice" = im;
+                  tabledata "Privacy Notice" = rim,
+                  tabledata "Privacy Notice Approval" = r;
 
     var
         EmptyGuid: Guid;
@@ -118,6 +119,8 @@ codeunit 1565 "Privacy Notice Impl."
         exit(CheckPrivacyNoticeApprovalState(PrivacyNoticeId, true));
     end;
 
+    [InherentPermissions(PermissionObjectType::TableData, Database::"Privacy Notice", 'r')]
+    [InherentPermissions(PermissionObjectType::TableData, Database::"Privacy Notice Approval", 'r')]
     procedure CheckPrivacyNoticeApprovalState(PrivacyNoticeId: Code[50]; SkipCheckInEval: Boolean): Enum "Privacy Notice Approval State"
     var
         Company: Record Company;


onbuyuka:
Closing in favor of a platform-level fix.

The inherent read cannot be granted from BCApps: `Privacy Notice` (2000000237) and `Privacy Notice Approval` (2000000238) are **platform** system tables, so the `[InherentPermissions(PermissionObjectType::TableData, Database::"Privacy Notice", 'r')]` method attribute fails to compile with **AL0720** ("application object ... could not be found in the current extension"). The 371 existing precedents all reference BCApps-owned tables (e.g. `User Login` = 9008, `User Plan` = 9005), which is why they compile.

The classic `Permissions` property does compile with these tables but is role-gated (`indirect &= rolePermissions`), so it is a no-op when the pre-company capability check resolves roles to `None`.

The fix is moving to BC-Platform: declaring table-level `InherentPermissions` on the two platform table definitions, exactly like sibling system tables (`Tenant License State`, `Record Link`, `User Property`) and how `Copilot Settings` (7775) self-grants. Work item AB#641612 will follow that PR.

PR #9841, opened 2026-07-30
Indent the case block under the IsHandled guard

Explorer986:
## What & why

Small formatting follow up to #9421. In FinancialVoidCheck the case block runs under `if not IsHandled then` but was left at the same indentation as the guard. This indents the whole case block one level so it reads correctly. Pure whitespace, no behavior change.

## Linked work

[AB#640242](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/640242)

## How I validated this

- [x] The change is whitespace only, git diff --ignore-all-space is empty.

- Confirmed the diff is pure reindentation of the case block in all four layers (APAC, NA, RU, W1). No logic touched.

## Risk & compatibility

None. Reindentation only, no code changes.


AleksanderGladkov (approved):

mynjj (approved):
what a crazy way of rendering the diff as it shows in github

PR #9850, opened 2026-07-30
Fix VariantCode lost during Contract Price Update proposal

djukicmilica:
## Why

When using the **Recent Item Prices** method in Subscription Contract Price Update, the system proposes the base item price instead of the variant-specific price. The `VariantCode` from the Subscription Header (Service Object) is never passed to the temporary Sales/Purchase Line used for price calculation, causing the pricing engine to ignore variant-specific prices entirely.

Reported via ICM LiveSite ticket — reproduced on version 28.3.

## Summary

- **Fixed** `CreateTempSalesLine` overload (line ~160) to pass `ServiceObject."Variant Code"` instead of chaining through an intermediate overload that hardcoded an empty string
- **Fixed** `CreateTempPurchaseLine` overload (line ~229) — same issue, now passes `ServiceObject."Variant Code"`
- **Added** missing `TempPurchaseLine."Variant Code" := VariantCode` assignment in the full purchase line overload (line ~257) — the parameter was accepted but never used

[AB#644750](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/644750)


mynjj (approved):

ventselartur (approved):

AleksanderGladkov (approved):

stephanweidmann:
isn't the unit code also missing?

PR #9863, opened 2026-07-30
[Extensibility Request] issue 29074: expose PurchLineQty in OnSumPurchLines2OnBeforeDivideAmount

AleksandricMarko:
## Summary
4PS Construct implements an "Amounts only" purchasing mode that processes purchase lines by monetary amount rather than physical quantity, which requires adjusting the calculated quantity before the amount-division logic runs. In `SumPurchLines2`, the `OnSumPurchLines2OnBeforeDivideAmount` event fires right before `DivideAmount` but does not expose the calculated `PurchLineQty`, so subscribers have no way to influence the quantity used for the division. This change adds `PurchLineQty` to that event by reference so subscribers can adjust it before `DivideAmount` runs.

Source issue repository: microsoft/ALAppExtensions; issue number: 29074

## Changes Made
- `OnSumPurchLines2OnBeforeDivideAmount` - added a new `var PurchLineQty: Decimal` parameter at the end of the event signature and passed the local `PurchLineQty` variable at the call site in `SumPurchLines2`, so subscribers can adjust the quantity before `DivideAmount` executes. Propagated to the same-named event in every country layer that has its own copy of `PurchPost.Codeunit.al` (APAC, BE, CH, ES, FI, GB, IT, NA, RU).

Fixes [AB#597602](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/597602)

### DIFF
--- a/src/Layers/APAC/BaseApp/Purchases/Posting/PurchPost.Codeunit.al
+++ b/src/Layers/APAC/BaseApp/Purchases/Posting/PurchPost.Codeunit.al
@@ -3941,7 +3941,7 @@ codeunit 90 "Purch.-Post"
                             PurchLineQty := PurchLine."Qty. to Receive"
                 end;
 
-                OnSumPurchLines2OnBeforeDivideAmount(PurchHeader, PurchLine, QtyType);
+                OnSumPurchLines2OnBeforeDivideAmount(PurchHeader, PurchLine, QtyType, PurchLineQty);
                 DivideAmount(PurchHeader, PurchLine, QtyType, PurchLineQty, TempVATAmountLine, TempVATAmountLineRemainder);
                 OnSumPurchLines2OnAfterDivideAmount(PurchHeader, PurchLine, QtyType, PurchLineQty, TempVATAmountLine, TempVATAmountLineRemainder);
                 PurchLine.Quantity := PurchLineQty;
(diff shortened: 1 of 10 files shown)


PredragMaricic (approved):
No blocking findings. The by-reference PurchLineQty parameter is exposed at the correct point before DivideAmount and is propagated consistently across all country-layer copies.

alexei-dobriansky:

# **Recommendation: Accept**

<details>

Adds a `var PurchLineQty: Decimal` parameter to the existing `OnSumPurchLines2OnBeforeDivideAmount` integration event in Codeunit 90 "Purch.-Post", across all 10 country layers. This enables subscribers to modify the computed purchase-line quantity before `DivideAmount` runs — needed for "Amounts only" purchasing scenarios where quantity must be overridden.

The value is fully computed at the call site (the case/switch block that assigns `PurchLineQty` completes immediately before the event fires). The parameter is `var`, which is correct — the entire purpose is to allow subscribers to override it before `DivideAmount` consumes it. The sibling `OnAfterDivideAmount` event already exposes `PurchLineQty` (by value), confirming naming and position are consistent. The change is mechanical and identical across all layers.

Adding a parameter to an already-published event is technically a compile-time break for existing subscribers. However, BCApps policy accepts this pattern for extensibility requests (confirmed by commit 5dbc86e9c1, 409230ab79, and others merged in the same wave). Partners must update their subscriber signatures on the next major version — this is expected.
</details>

<details>

None.
</details>

<details>

**Risk:** Low for the platform. Existing partner subscribers of this event will get a compile error on the next BC major version and must add the new parameter — this is a known, accepted consequence of in-place event extensions in BCApps. No runtime behavioral change for subscribers that do not modify the new parameter. No performance impact.

**Necessity:** Without this parameter, a subscriber cannot override `PurchLineQty` before `DivideAmount`. The `PurchLine` record parameter does not carry the computed quantity (it is a local variable derived from multiple fields depending on `QtyType`). The scenario (amounts-only purchasing) is real and concrete (GitHub issue #29074, ADO AB#597602).
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9863 round=1 by=alexei-dobriansky at=2026-08-03T09:30:00Z lastSha=8fca122eb108747457ef45e060b75881eaa3e22c suggestions=none

alexei-dobriansky (approved):

PR #8848, opened 2026-06-27
Sync from BCAppsPrivate + NAV (efe2d9954d)

mazhelez:
Publishes the latest BCAppsPrivate and NAV source to BCApps as two reviewable commits.

**Commits**
- `bdea607174` Sync from BCAppsPrivate 3e8428800b (build/CI, rulesets, DisabledTests, tools, docs, AL-Go country project configs)
- `c61488e0be` Sync from NAV efe2d9954d (Layers, GDL, DemoTool, Apps overlay, build metadata, version tokens, country settings)

**Notes**
- AL-Go-Settings.json merge conflict resolved (kept Private runner settings, took newer artifact, single incrementalBuilds block).
- Verified: 0 leftover version tokens; all platform values 29.0.0.0; sync-state.json untouched.

[AB#612711](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/612711)

PR #9881, opened 2026-07-31
Restore the full-sum sign decision in Apply-to-Oldest application

### LINKED ISSUE #9529: [Bug]: Regression in BC 28.2: posting an invoice for a customer/vendor with Application Method = "Apply to Oldest" applies the invoice to **other open invoices (same sign)**, closing them and accumulating their amounts onto the new invoice
**Affected version:** Business Central 28.2+ (verified on application build 28.2.50931.51034 OnPrem W1 and 28.2.50931.52241 SaaS production; **source inspection of the latest available build 28.3.52162.52641 confirms the defective code is unchanged and the issue is still present**)
**Works correctly in:** 28.0 (verified on 28.0.46665.48549) and, per code inspection, all versions before the 28.2 change
**Area:** Base Application, `Codeunit 12 "Gen. Jnl.-Post Line"`, procedures `PrepareTempCustLedgEntry` and `PrepareTempVendLedgEntry`
**Severity suggestion:** High / data-integrity — silently produces incorrect applications during normal posting; no error is raised; recovery requires manual Unapply of potentially thousands of entries

## Summary

In 28.2 an early-exit optimization (`SufficientEntriesFound`) was added to the "weighing" loop in `PrepareTempCustLedgEntry` (and `PrepareTempVendLedgEntry`). The loop now stops summing open entries as soon as the partial sum offsets the new document, but the subsequent decision `TempOldCustLedgEntry.SetRange(Positive, RemainingAmount < 0)` still uses that now **partial** sum. Because the temporary buffer iterates in key order (`Customer No., Applies-to ID, Open, Positive, Due Date`) where `Positive = false` sorts first, **credit memos/payments are summed first**. Whenever their total exceeds the new invoice amount, the partial sum turns negative and the filter resolves to `Positive = TRUE` — so the new **invoice is applied to older open invoices (same sign)**, closing them one by one in due-date order. After the filtered set is exhausted, `FindNextOldCustLedgEntryToApply` clears the filter and continues, consuming the remaining opposite-sign entries as well. The net effect: a single posted invoice can close **every open entry** of the account and accumulate the entire balance onto itself under one due date.

The trigger condition on 28.2 is mundane: *customer (or vendor) with Application Method = "Apply to Oldest" + open credit memos/payments totaling more than the newly posted invoice*. The account's overall balance can be perfectly normal (net debit).

## Steps to reproduce (minimal, 4 journal postings, W1)

1. Create a customer; set **Application Method = Manual** (so setup entries stay open).
2. In a general journal post, with blank Applies-to fields, balancing to a G/L account:
   - Invoice **F1** = 1,000 (posting date D1)
   - Invoice **F2** = 1,000 (D2 > D1)
   - Invoice **F3** = 1,000 (D3 > D2)
   - Credit Memo **CM1** = −2,500 (D4)
   All four entries remain open. Net balance = **+500 (debit)**.
3. On the customer card set **Application Method = Apply to Oldest**.
4. Post a new invoice **F4** = 2,000 (D5, blank Applies-to). Note: CM1 (2,500) > F4 (2,000).

## Expected result

F4 is applied against CM1 (the only opposite-sign entry): F4 closed, CM1 remaining −500, **F1–F3 untouched**. This is exactly what 28.0/28.1 do.

## Actual result on 28.2

**F1, F2 and F3 are closed by invoice F4** (invoice applied to invoices), applications continue into CM1, and F4 ends with an incorrect remaining amount. The receivable of the closed invoices is silently consolidated onto F4 with F4's due date; per-invoice due-date tracking is destroyed. No error or warning is raised.

## Root cause (code diff 28.0 → 28.2)

`Codeunit 12, PrepareTempCustLedgEntry` (identical change in `PrepareTempVendLedgEntry`):

```al
// 28.0:
repeat
    ...
    RemainingAmount += TempOldCustLedgEntry."Remaining Amount";
until TempOldCustLedgEntry.Next() = 0;
TempOldCustLedgEntry.SetRange(Positive, RemainingAmount < 0);

// 28.2 (new):
repeat
    ...
    RemainingAmount += TempOldCustLedgEntry."Remaining Amount";
    if (Cust."Application Method" = Cust."Application Method"::"Apply to Oldest") and
       (RemainingAmount * NewCVLedgEntryBuf."Remaining Amount" <= 0)
    then
        SufficientEntriesFound := true;
until (TempOldCustLedgEntry.Next() = 0) or SufficientEntriesFound;   // <-- early exit
TempOldCustLedgEntry.SetRange(Positive, RemainingAmount < 0);        // <-- decision on PARTIAL sum
```

Chain of events for the repro above (new invoice +2,000):

1. Temp buffer iterates with `Positive = false` first → CM1 is summed first: `RemainingAmount = 2,000 − 2,500 = −500`.
2. `RemainingAmount * NewRemaining ≤ 0` → `SufficientEntriesFound := true` → loop exits before summing the invoices.
3. `SetRange(Positive, −500 < 0)` → `SetRange(Positive, TRUE)` → the applying set = **the open invoices** (same sign as the new document!).
4. `ApplyCustLedgEntry` applies F4 to F1→F2→F3 (due-date order); when the positives are exhausted, `FindNextOldCustLedgEntryToApply` clears the Positive filter (line `TempOldCustLedgEntry.SetRange(Positive); Find('-')`) and, since the product check is negative, continues into CM1 as well.

On 28.0 the loop sums **all** entries → `RemainingAmount = +500` → `SetRange(Positive, FALSE)` → only CM1 is applied (correctly, capped at 2,000).

Note there is also a pre-existing (much rarer) variant of the same defective decision on ≤28.1: when the account's **full** net balance has the opposite sign of the new document (e.g. open payments > open invoices, then posting an invoice), `SetRange(Positive, RemainingAmount < 0)` likewise selects same-sign entries. 28.2 merely made the defect reachable with everyday data.

## Consistency argument

The same procedure **forbids** same-sign application on the manual path: for `Applies-to Doc. No.` it runs `OldCustLedgEntry.TestField(Positive, not NewCVLedgEntryBuf.Positive)`. The automatic Apply-to-Oldest path lacks any such guard, so the system automatically performs an application that a user is explicitly prevented from making manually.

## Verification matrix (automated AL tests, same test app on both containers)

| Scenario (new invoice posted, blank Applies-to) | 28.0.46665 | 28.2.50931 |
|---|---|---|
| Open CMs > new invoice, net balance positive (repro above) | ✅ pass | ❌ **fails — invoices closed by invoice** |
| Same, vendor side | ✅ pass | ❌ fails |
| Net balance opposite to new document (payments > invoices) | ❌ fails (pre-existing variant) | ❌ fails |
| Payment spread oldest-first, prepayment exclusion, FCY, real order/credit-memo posting flows (10 scenarios) | ✅ pass | ✅ pass |

## Suggested fix

Either compute the full sum before the sign decision (restore 28.0 behavior), or — more robustly — restrict the applying set for automatic Apply-to-Oldest to entries of the **opposite sign** of the new document (mirroring the `TestField(Positive, not ...)` rule already enforced on the manual path), e.g. `TempOldCustLedgEntry.SetRange(Positive, NewCVLedgEntryBuf."Remaining Amount" < 0)`.

## Reporter

GoPro doo (Microsoft partner, Belgrade, Serbia) — delegated admin for the affected tenant. Contact: nikola.jovicic@gopro.rs. 

### Expected behavior

F4 is applied against CM1 (the only opposite-sign entry): F4 closed, CM1 remaining −500, **F1–F3 untouched**. This is exactly what 28.0/28.1 do.

## Steps to reproduce (minimal, 4 journal postings, W1) 
1. Create a customer; set **Application Method = Manual** (so setup entries stay open).
2. In a general journal post, with blank Applies-to fields, balancing to a G/L account:   - Invoice **F1** = 1,000 (posting date D1)
   - Invoice **F2** = 1,000 (D2 > D1)
   - Invoice **F3** = 1,000 (D3 > D2)
   - Credit Memo **CM1** = −2,500 (D4)
   All four entries remain open. Net balance = **+500 (debit)**.
3. On the customer card set **Application Method = Apply to Oldest**.
4. Post a new invoice **F4** = 2,000 (D5, blank Applies-to). Note: CM1 (2,500) > F4 (2,000).

### Additional context

Attached AL test app **"ApplyToOldest Repro"** (`app.json` + `codeunit 50100 "ApplyToOldest Repro"`, 4 tests) depends **only** on Microsoft test libraries (`Library Assert`, `Tests-TestLibraries`) — no third-party code whatsoever. Each test is the 4-posting scenario from "Steps to reproduce" (customer + vendor, both variants), asserting the standard-correct outcome.

**Verified on a clean Cronus W1 OnPrem 28.2.50931.51034 container with NO third-party extensions installed** (Microsoft test framework only):

| Test | 28.0.46665 (vanilla) | 28.2.50931 (vanilla) |
|---|---|---|
| `CustInvoiceWithLargeOpenCreditMemoMustNotCloseInvoices` | pass | **FAIL** — `Expected 1,000, Actual 0` (older invoice closed by the new invoice) |
| `VendInvoiceWithLargeOpenCreditMemoMustNotCloseInvoices` | pass | **FAIL** — `Expected −1,000, Actual −500` |
| `CustInvoiceWithNetCreditBalanceMustNotCloseInvoices` | FAIL (pre-existing variant) | **FAIL** |
| `VendInvoiceWithNetDebitBalanceMustNotCloseInvoices` | FAIL (pre-existing variant) | **FAIL** |

Attachments:  `Repro app.zip` with `Repro Test (BaseApp only).al`, `

[Repro app.zip](https://github.com/user-attachments/files/30082114/Repro.app.zip)

Franco111000:
## What & why

Since 28.2, posting a document for a customer or vendor with **Application Method = "Apply to Oldest"** can apply the new document to open entries of the **same sign**, closing them and accumulating their amounts onto the new document (issue #9529). A posted invoice can end up with a Remaining Amount larger than its own Amount, and per-entry due-date tracking is destroyed. No error is raised.

**Root cause.** In codeunit 12, the weighing loops in `PrepareTempCustLedgEntry`, `PrepareTempVendLedgEntry` and `PrepareTempEmplLedgEntry` feed the sign decision `SetRange(Positive, RemainingAmount < 0)`. That decision needs the net of the new document plus **all** open entries: it selects the ledger side that the application loop may consume uncapped (`FindAmtForAppln` applies the full remaining amount of each entry when the Positive filter is set, and caps with `ABSMin` once the filter is cleared). 28.2 added a `SufficientEntriesFound` early exit that stops the loop as soon as the running sum offsets the new document. Because the buffer key iterates `Positive = false` entries first, a positive new document (customer invoice or refund, vendor payment or refund) whose opposite-sign open total exceeds it trips the exit inside the negative block, before the positive entries pull the sum back. The truncated sum flips the decision, and the new document consumes old same-sign entries uncapped.

**Fix.** Remove the early exit in all three procedures, restoring the pre-28.2 weighing semantics. The decision is a property of the whole open-entry set, so any early exit leaves it on a partial sum in some conditions: a payment discount can exceed a partially credited invoice's remaining amount, the stored `Positive` flag can disagree with the current remaining sign on overshot entries (which this very defect produces), and the employee loop raises `OnPrepareTempEmplLedgEntryOnBeforeUpdateRemainingAmount`, which lets subscribers adjust the summed amount. The decision must see the full sum.

**Why not the opposite-sign filter suggested in the issue.** `SetRange(Positive, NewCVLedgEntryBuf."Remaining Amount" < 0)` fixes the reported case but breaks the invariant that the selected side must be fully absorbable: with an open invoice of 1,000, an open credit memo of 5,000 and a new invoice of 2,000, it consumes the credit memo uncapped and leaves the new invoice permanently open at a remaining amount of -2,000. The full-sum rule picks the side that the netting can exhaust without overshoot.

**Origin.** The early exit entered through the NAV sync published in this repo as 30ce9d7f71 (`features/BCAppsMigration`, "Sync application code from NAV repo (#109)") and reached main in the layer import 748fdaa43f (#8848). It shipped with one companion test, `ApplyToOldestPartialPaymentAppliesToOldestFirst`, which covers only the single-sign case, where the exit is harmless. This PR restores the three procedure bodies to their state at `30ce9d7f71^` byte for byte, in all 11 layer copies (W1, APAC, BE, CH, ES, FI, FR, IT, NA, NO, RU).

**Out of scope, called out for transparency.** There is a rarer pre-existing variant on <= 28.1: when the account's full net balance opposes the new document, the full-sum decision also selects same-sign entries. That behavior is unchanged here (it is the overshoot-protection trade-off described above and predates 28.2). Tests 3 and 4 of the reporter's attached repro app target that variant and still fail after this fix, by design. The payment-discount and stale-`Positive` edges above are likewise pre-existing and out of scope.

Credit to @nikolajovicic for the precise root-cause analysis and the attached repro app; direction agreed in the issue thread.

## Linked work

Fixes #9529

## How I validated this

- The change is the exact inverse of the hunks that introduced the early exit: all 33 edited regions (3 procedures x 11 layers) are byte-identical to their state at `30ce9d7f71^`, verified by direct comparison against that commit.
- This repo's CI compiles the W1 layer only, so the 10 country copies are outside the build. To rule out drift, the loop regions were hash-compared across all 11 copies before the edit (identical) and after the edit (identical again). The W1 compile is covered by the PR build.
- I have no local BC environment for this repo, so the runtime behavior was verified by tracing the application flow (`ApplyCustLedgEntry`, `FindAmtForAppln`, `FindNextOldCustLedgEntryToApply`) with concrete numbers for the repro scenario, the netting scenario and the vendor mirror, and locked in as tests:
  - `ApplyToOldestInvoiceDoesNotApplyToOtherInvoices` (codeunit 134006, W1 + DACH copy): the reported defect. Under the current code the three older invoices are closed by the new invoice and the new invoice stays open above its own amount; with the fix the credit memo is closed, the new invoice is closed at 0, and only the oldest invoice absorbs the credit memo excess.
  - `ApplyToOldestPaymentNettingAgainstMixedEntries` (codeunit 134006, W1 + DACH copy): pins the mixed-sign netting end state, which is identical before and after this change.
  - `ApplyToOldestPaymentDoesNotApplyToOtherPayments` (codeunit 134004, W1 + CH + IT copies): the vendor mirror of the defect (vendor payments are the positive direction there).
- The existing `ApplyToOldestPartialPaymentAppliesToOldestFirst` (single-sign) takes the same decision path before and after this change and stays green.

## Risk & compatibility

- Behavior changes only for Apply-to-Oldest application decisions on accounts holding open entries of both signs, reached from journal posting and from the Apply Posted Entries flows. Single-sign situations produce the same filter as before.
- Performance: this restores the pre-28.2 cost of one Remaining Amount `CalcFields` per qualifying open entry during Apply-to-Oldest application on mixed-sign accounts. The weighing loop is entered whenever at least one open entry has the same sign as the new document; it is skipped when every open entry is opposite-sign. A correct optimization would need a decision that does not depend on the truncated sum; I am happy to explore that separately if there is interest.
- The regression ships in 28.2 and later. Should this be cherry-picked to the 28.x release branches, as was done for #8941 (#9328 - #9332)?
- Area: Finance (matching the issue; I cannot set labels).


alexei-dobriansky:

<details>

This PR removes the `SufficientEntriesFound` early exit from the customer, vendor, and employee Apply-to-Oldest preparation loops in codeunit 12 across the layer copies. That restores the sign decision to use the full sum of the new entry plus all eligible old entries, instead of a partial sum that can select same-sign entries after mixed-sign balances.

The changed customer and vendor logic matches the root cause in GitHub issue #9529. The new tests cover the reported customer case, the vendor mirror, and a mixed-entry customer case. The employee procedure has the same changed decision and the same event-adjusted sum path, but there is no matching employee regression test in this PR.
</details>

<details>

**S1** - *Add employee regression coverage*
`PrepareTempEmplLedgEntry` changes the same Apply-to-Oldest sign decision as the customer and vendor paths. `ERMApplyUnapplyEmployee.Codeunit.al` already has Apply-to-Oldest coverage, so add the same mixed-sign regression test for employee ledger entries. This is finance posting logic, so the employee path should not rely only on customer and vendor tests.
</details>

<details>

**Risk:** The regression surface is finance posting for Apply-to-Oldest on accounts with mixed-sign open entries. The change is narrow and does not change public signatures or events, but it changes how much data the weighing loop reads and restores the pre-28.2 full scan. The customer and vendor tests reduce risk for the reported paths; the employee path remains untested.

**Necessity:** The linked GitHub bug is detailed and important: the current behavior can silently apply documents to same-sign entries and damage remaining amounts and due-date tracking. There is no linked ADO `AB#` work item in the PR body, so I derived bug mode from GitHub issue #9529 and the diff. The scope is right, but the employee change needs direct coverage before merge.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9881 round=1 by=alexei-dobriansky at=2026-08-03T10:03:14.844Z lastSha=0e45364cc4894c21a3e4de331c0e0dbfb301046c suggestions=S1

Franco111000:
Thanks for the review! S1 addressed in 8012bb6fe4: added `ApplyToOldestPaymentDoesNotApplyToOtherPayments` to codeunit 134114 "ERM Apply Unapply Employee", next to the existing Apply-to-Oldest employee test and mirrored into the CZ copy of that codeunit. Same mixed-sign shape as the customer and vendor tests: three reimbursement payments, a larger expense posted last but back-dated before them (with a setup assert proving it is still open), then a new payment that must settle the expense only, with the excess reducing the oldest payment and the other payments untouched. Same even-integer amount scheme, so all expected values are whole numbers.

alexei-dobriansky:

# **Recommendation: Accept**

<details>

The new commit adds the missing employee Apply-to-Oldest mixed-sign regression test in the W1 and CZ employee test codeunits. The test mirrors the customer and vendor regression shape from round 1: older payments, a larger back-dated expense, and a new payment that must close only the expense and leave the other payments open.

This addresses the round-1 gap. The test exercises the employee path that changed in PrepareTempEmplLedgEntry, uses whole expected amounts, checks the setup state before the final posting, and verifies both the closed opposite-sign entry and the untouched same-sign entries after posting.
</details>

<details>

| S1 | Add employee regression coverage | Addressed | Author said commit 8012bb6fe4 adds the mirrored W1 and CZ employee regression tests. |
</details>

<details>

*None - changes only addressed the previous suggestion.*
</details>

<details>

**Risk:** The new commit only adds tests, so it does not expand runtime behavior. The remaining PR risk is still the finance posting Apply-to-Oldest change reviewed in round 1, now covered for customer, vendor, and employee paths.

**Necessity:** The linked GitHub issue describes an important data-integrity bug. There is still no linked ADO AB# work item in the PR body, so I could not read an ADO Bug or Slice; I reviewed from GitHub issue #9529 and the diff. The added employee test is necessary and closes the only round-1 blocker.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9881 round=2 by=alexei-dobriansky at=2026-08-03T17:11:12.306Z lastSha=8012bb6fe48ab31ed8a9dac28fd3e15ac8c93342 suggestions=S1:addressed parentRound=1

src/Layers/W1/BaseApp/Finance/GeneralLedger/Posting/GenJnlPostLine.Codeunit.al
```diff
@@ -4962,11 +4956,7 @@ codeunit 12 "Gen. Jnl.-Post Line"
                     if PaymentToleranceMgt.CheckCalcPmtDiscCVVend(NewCVLedgEntryBuf, TempOldVendLedgEntry, 0, false, false) then
                         TempOldVendLedgEntry."Remaining Amount" -= TempOldVendLedgEntry.GetRemainingPmtDiscPossible(NewCVLedgEntryBuf."Posting Date");
                     RemainingAmount += TempOldVendLedgEntry."Remaining Amount";
-                    if (Vend."Application Method" = Vend."Application Method"::"Apply to Oldest") and
-                       (RemainingAmount * NewCVLedgEntryBuf."Remaining Amount" <= 0)
-                    then
-                        SufficientEntriesFound := true;
-                until (TempOldVendLedgEntry.Next() = 0) or SufficientEntriesFound;
+                until TempOldVendLedgEntry.Next() = 0;
```
dcenic, line 4959: This will reintroduce the severe performance issue that is affecting customers who have many (tens of thousands of) open ledger entries for one or very few vendors, and use 'Apply to Oldest'.
We should try to keep the performance optimization and address the bug that it introduced.

dcenic (changes requested):

src/Layers/W1/BaseApp/Finance/GeneralLedger/Posting/GenJnlPostLine.Codeunit.al
dcenic: @Franco111000 I opened an ICM for this regression, and I am working on a fix that will fix it and also keep the performance optimization of 'Apply-to-Oldest'.

dcenic: @Franco111000 I made a new pull request https://github.com/microsoft/BCApps/pull/10227 in which I:

- copied your new regression tests from this PR
- made sure to fix it while keeping the performance optimization (take sign into consideration when looping and summing)
- added one more test to make sure the performance optimization is still applied.

Franco111000:
Superseded by #10227 (merged as efa0b87caa), which fixes the regression while keeping the Apply-to-Oldest optimization and carries these regression tests forward. Closing as committed. Thanks @dcenic for the quick turnaround and the collaboration on this one.

PR #9893, opened 2026-08-03
[Main]-The Reservation Entry does not exist error when creating a Purchase Order from a Sales Order

DeepsShukla:
Workitem [Bug 643358](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/643358): [ALL-E] "The Reservation Entry does not exist. Identification fields and values: Entry No.='9' Positive='No'" error when creating a Purchase Order from a Sales Order.

Fixes [AB#643358](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/643358)

### DIFF
--- a/src/Layers/W1/BaseApp/Inventory/Tracking/ItemTrackingManagement.Codeunit.al
+++ b/src/Layers/W1/BaseApp/Inventory/Tracking/ItemTrackingManagement.Codeunit.al
@@ -577,14 +577,6 @@ codeunit 6500 "Item Tracking Management"
         CopyItemTracking(FromRowID, ToRowID, SwapSign, false);
     end;
 
-    procedure CopyItemTracking(FromRowID: Text[250]; ToRowID: Text[250]; SwapSign: Boolean; SkipReservation: Boolean; NewReservationStatus: Enum "Reservation Status")
-    var
-        ReservEntry: Record "Reservation Entry";
-    begin
-        ReservEntry.SetPointer(FromRowID);
-        ReservEntry.SetPointerFilter();
-        CopyItemTracking3(ReservEntry, ToRowID, SwapSign, SkipReservation, NewReservationStatus);
-    end;
 
     procedure CopyItemTracking(FromRowID: Text[250]; ToRowID: Text[250]; SwapSign: Boolean; SkipReservation: Boolean)
     var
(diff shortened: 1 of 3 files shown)


alexei-dobriansky:

<details>

This PR fixes a "Reservation Entry does not exist" error that happens when creating a Purchase Order from a Sales Order via Order Planning. The root cause is that CopyItemTracking was copying item tracking entries with Reservation Status::Reservation, which requires paired (positive+negative) reservation entries. Since only one side was created, downstream code failed when looking for the counterpart.

The fix removes the NewReservationStatus parameter and always sets Prospect status on copied entries. This is correct — Prospect entries do not require paired matching and are the standard status for tracking-only entries created by planning. The logic in CopyItemTracking3 is straightforward and all call sites now consistently use Prospect. The root cause is properly addressed.
</details>

<details>

**S1** - *Obsolete the removed public overload instead of deleting it*
The procedure CopyItemTracking(FromRowID: Text[250]; ToRowID: Text[250]; SwapSign: Boolean; SkipReservation: Boolean; NewReservationStatus: Enum "Reservation Status") is a public member of codeunit 6500. Deleting it breaks any partner extension that calls it. Mark it [Obsolete('Removed ...', '27.0')] and have it delegate to the 4-parameter overload. Even though it was added recently (June 2026), it may already be compiled against.

**S2** - *Replace the deleted test with one that verifies the fix*
The test SalesOrderPlanMakePurchOrderWithSerialTracking is deleted because its assertion (Reservation Status::Reservation) matched the broken behavior. A new test should verify the same scenario end-to-end and assert that a Purchase Order is created without error and that tracking entries exist with Prospect status. The test infrastructure (Library - Item Tracking, Library - Planning, page handlers) is already present in codeunit 137088.
</details>

<details>

**Risk:** The regression surface is moderate. The change affects every code path that copies item tracking via CopyItemTracking3 (order planning, transfer orders, general item tracking copy). If any caller legitimately needed Reservation status, it would now get Prospect — but the only caller that used the NewReservationStatus parameter was SalesLinePlanning, and that usage was itself the bug. The removed public overload is a compatibility risk for compiled extensions. No event signatures are changed.

**Necessity:** High. The bug causes a hard error blocking Purchase Order creation from Sales Orders with serial/lot tracking. There is no workaround other than removing tracking before planning. The fix scope is appropriate for a bug fix.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9893 round=1 by=alexei-dobriansky at=2026-08-03T09:33:00Z lastSha=8fdc10d2ac8b3050265f7995a8b1fb1c935bf87e suggestions=S1,S2

PR #9898, opened 2026-08-03
[Main]-The Reservation Entry does not exist error when creating a Purchase Order from a Sales Order

DeepsShukla:
Workitem [Bug 644925](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/644925): [master] [ALL-E] "The Reservation Entry does not exist. Identification fields and values: Entry No.='9' Positive='No'" error when creating a Purchase Order from a Sales Order.

Fixes [AB#644925](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/644925)

This Pull Requested is created to revert back the changes of [Pull Request 242051](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_git/NAV/pullrequest/242051): [Master]- Creating Purchase order directly from Sales Order card page, assigning the serial no. Reservation entry doesn’t get created.

### DIFF
--- a/src/Layers/W1/BaseApp/Inventory/Tracking/ItemTrackingManagement.Codeunit.al
+++ b/src/Layers/W1/BaseApp/Inventory/Tracking/ItemTrackingManagement.Codeunit.al
@@ -577,14 +577,17 @@ codeunit 6500 "Item Tracking Management"
         CopyItemTracking(FromRowID, ToRowID, SwapSign, false);
     end;
 
+#if not CLEAN29
+    [Obsolete('Removed and will be deleted in future release.', '29.0')]
     procedure CopyItemTracking(FromRowID: Text[250]; ToRowID: Text[250]; SwapSign: Boolean; SkipReservation: Boolean; NewReservationStatus: Enum "Reservation Status")
     var
         ReservEntry: Record "Reservation Entry";
     begin
         ReservEntry.SetPointer(FromRowID);
         ReservEntry.SetPointerFilter();
-        CopyItemTracking3(ReservEntry, ToRowID, SwapSign, SkipReservation, NewReservationStatus);
+        CopyItemTracking3(ReservEntry, ToRowID, SwapSign, SkipReservation);
     end;
+#endif
 
     procedure CopyItemTracking(FromRowID: Text[250]; ToRowID: Text[250]; SwapSign: Boolean; SkipReservation: Boolean)
     var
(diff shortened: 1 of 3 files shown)


PredragMaricic (approved):

alexei-dobriansky:

<details>

This PR fixes a severe regression (Bug 644925): after opening **Create Purchase Orders** from a Sales Order (the Order Planning path via page 1328), a "The Reservation Entry does not exist" error appears and table 337 is left corrupted, blocking all later sales orders. The fix reverts an earlier change (NAV PR 242051, Bug 618883). It removes the `NewReservationStatus` parameter from `CopyItemTracking3`, so item tracking is again copied with the `Prospect` status instead of `Reservation`, and it reverts `Sales Line-Planning.OnInsertDemandLinesOnCopyItemTracking` to the default overload.

The revert is internally consistent. `SalesLinePlanning` was the only caller that passed `Reservation` status, so restoring `Prospect` matches the behavior that shipped correctly in 27.0, 26.x and 18.x. The 5-parameter public overload is kept and marked `[Obsolete(..., '29.0')]` under `#if not CLEAN29`, which is the right way to remove a public method without breaking partner extensions. The root cause (requisition lines created with `Reservation` status that are later deleted inconsistently) is addressed by no longer setting that status.
</details>

<details>

**S1** - *Add a regression test for the corruption scenario*
This change touches reservation entries, which is a data-sensitive area, and the bug corrupts table 337 for every later sales order. Please add a test that runs the Order Planning path (Calculate Order Plan, then Create Purchase Orders / Cancel) for a lot-tracked item with no inventory, and asserts that reservation entries stay valid and the flow can be re-run. The `SCM Order Planning - III` codeunit already has `LibraryPlanning`, `LibraryItemTracking`, and the needed page handlers, so adding this test is mechanical.

**S2** - *Confirm Bug 618883 does not come back*
The removed test `SalesOrderPlanMakePurchOrderWithSerialTracking` covered Bug 618883 (serial reservation entries created on the purchase line). Reverting to `Prospect` status may reintroduce that bug. Please confirm what happens to 618883 after this revert, and if it still works, keep a test that proves it; if it regresses, state this trade-off in the PR description.

**S3** - *Obsolete overload now ignores its status parameter*
The obsolete 5-parameter `CopyItemTracking` still takes `NewReservationStatus` but no longer uses it, so it always copies with `Prospect`. External callers that pass `Reservation` will silently get different behavior until CLEAN29, and the unused parameter may raise a compiler warning. Consider noting this in the `[Obsolete]` message so callers understand the parameter has no effect.
</details>

<details>

**Risk:** The change is a narrow revert limited to `Item Tracking Management` and `Sales Line-Planning`. `SalesLinePlanning` is the only caller passing `Reservation` status, so blast radius is small; other callers (`GetSalesOrders`, `PurchGetDropShpt`, production and job flows) already used `Prospect` and are unaffected. The main compatibility point is the obsoleted public overload (S3). The main untested risk is the 618883 trade-off (S2): without a test we cannot confirm both scenarios work at once.

**Necessity:** Clearly necessary. Bug 644925 is Severity 2 with a 10/10 repro, reproduces on standard demo data with no extensions, and leaves persistent data corruption that blocks all later orders. It is a regression introduced in 28.x that works in 27.0/26.x/18.x, so reverting the offending change through this channel is appropriate. The scope matches a bug fix. The one gap is proof: a regression test (S1) is feasible and expected in this sensitive area, which is why this is Request Changes rather than Accept.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9898 round=1 by=alexei-dobriansky at=2026-08-03T17:19:58Z lastSha=c61675cfe981d26cb12c9aaf177aa4f929fd8a11 suggestions=S1,S2,S3

v-praghav (approved):

PR #9900, opened 2026-08-03
[Test] table 12116 "Withholding Tax": add positive (standard field bl…

Explorer986:
## What & why
**Follow Up PR**
Adds tests for the Withholding Tax `OnModify` guard (table 12116). The guard keeps reported/paid entries immutable for standard fields but allows edits to extension fields these tests lock that behaviors in so it can't regress.

## Linked work

[AB#640860](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/640860)

## How I validated this

- `ModifyingStandardFieldOnReportedWithholdingTaxIsBlocked` reported entry, change a standard field, expect the "cannot be modified" error. Passes.
- `ModifyingOnlyExtensionFieldOnReportedWithholdingTaxIsAllowed` reported entry, change only an extension field (id ≥ 50000, added via a small test table extension), the edit goes through. Passes.
- Test-only change, still need to run the IT test suite in a container locally before merge.

## Risk & compatibility

Tests only, no product code touched. Adds one test table extension (field 144300) scoped to the IT test app.

### DIFF
--- a/src/Layers/IT/Tests/Local/ITCU2015UnitTest.Codeunit.al
+++ b/src/Layers/IT/Tests/Local/ITCU2015UnitTest.Codeunit.al
@@ -1399,6 +1399,29 @@ codeunit 144021 "IT - CU 2015 Unit Test"
         // Verification is done inside ErrorPageHandler
     end;
 
+    [Test]
+    [Scope('OnPrem')]
+    procedure ModifyingStandardFieldOnReportedWithholdingTaxIsBlocked()
+    var
+        WithholdingTax: Record "Withholding Tax";
+    begin
+        // [FEATURE] [Withholding Tax]
+        // [SCENARIO] Modifying a standard field on a reported withholding tax entry is blocked
+        Initialize();
+
+        // [GIVEN] A reported withholding tax entry
+        WithholdingTax.Get(CreateWithholdingTaxWithAU001006WithEmptyNonTaxable(CreateVendor(), "Withholding Tax Reason"::A, 0, WorkDate(), WorkDate()));
+        WithholdingTax.Reported := true;
+        WithholdingTax.Modify(false);
+
+        // [WHEN] A standard field is modified and the record is saved
+        WithholdingTax."External Document No." := 'MODIFIED';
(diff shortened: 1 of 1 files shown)


AleksanderGladkov (approved):

mynjj (approved):

alexei-dobriansky:

<details>

This PR adds a regression test that saves a reported withholding tax entry, changes a standard field, and verifies that `Modify(true)` is blocked.

The test covers the protected path in `ITCU2015UnitTest.Codeunit.al:1402-1422`. It does not cover the new behavior from `WithholdingTax.Table.al:248-256` and `:347-368`, where a change limited to an extension field must succeed. The current test would still pass if `StandardFieldModified` always returned `true`, so it cannot detect a regression in the behavior this follow-up is meant to secure.
</details>

<details>

**S1** - *Test the allowed extension-field path*
The test at `src/Layers/IT/Tests/Local/ITCU2015UnitTest.Codeunit.al:1402-1422` covers only the blocked standard-field path. Add a test table extension and verify that changing only its field on a persisted Reported or Paid entry succeeds with `Modify(true)`; this is the new behavior in `src/Layers/IT/BaseApp/Local/Finance/WithholdingTax/WithholdingTax.Table.al:248-256`.
</details>

<details>

**Risk:** This PR changes tests only, so it has no runtime regression surface. The risk is false assurance around a compliance guard: the added test proves that standard data stays locked, but it does not prove that partner extension data remains editable. The PR description also lists an extension-field test and test table extension that are absent from the actual diff.

**Necessity:** Regression coverage is needed because the earlier product change deliberately relaxed the `OnModify` guard for extension fields on reported or paid entries. The blocked-path test is useful, but the allowed extension-field path is the core new scenario and is feasible to test in this test app.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9900 round=1 by=alexei-dobriansky at=2026-08-04T11:00:00Z lastSha=913d573ca2c850e2d4c7b00094339d75a8306b02 suggestions=S1

PR #9901, opened 2026-08-03
Backport #9430: [28.x] E-Document ADI import crashes with untrappable JSON error when ADI response lacks outputs.1 structure

ventselartur:
Backport of #9430 for [AB#644118](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/644118)

### DIFF
--- a/src/Apps/W1/EDocument/App/src/Helpers/EDocumentJsonHelper.Codeunit.al
+++ b/src/Apps/W1/EDocument/App/src/Helpers/EDocumentJsonHelper.Codeunit.al
@@ -8,13 +8,20 @@ codeunit 6121 "EDocument Json Helper"
 {
     Access = Internal;
 
+    var
+        MalformedAdiResponseTxt: label 'ADI response is missing the expected ''%1'' property; returning an empty object.', Locked = true;
+        TelemetryCategoryTxt: label 'E-Document Matching Assistance', Locked = true;
+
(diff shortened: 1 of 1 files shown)


alexei-dobriansky:
**Problem-solution fit: Partial**

This is a faithful code port of #9430 into 28.x, and it does prevent the missing-key crash. But it still leaves the malformed-shape case open, so the stated ADI crash fix is not complete.

### Suggestions
- **S1** (🔴 High) Wrong JSON token types still crash the helper - The new checks only handle missing keys. GetInnerObject() and GetHeaderFields() still call AsObject() without JsonToken.IsObject(), so payloads like outputs: [] or fields: "x" can still raise the same untrappable runtime error. Treat wrong token types the same as missing properties, or reuse a helper that validates both presence and shape.

### Risk & necessity
This backport is worth taking because the current 28.x code can crash on malformed ADI responses in a user-facing import path. If it ships as-is, some malformed responses will still crash instead of degrading gracefully, so the risk is reduced but not removed.

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=9901 round=1 by=alexei-dobriansky at=2026-08-26T22:29:55Z lastSha=716bb4ee5fcca206f57a7958b19e72192817cb2f reviewKey=74c586ec096b03026908d3c4b408e97a9492c23bcf287a018e052aedbc1ea94a suggestions=S1@7a1c2e4f

PR #9913, opened 2026-08-03
[Shopify] Fix and re-enable UnitTestCalcPriceUsesCurrentWorkDate price-calc test

onbuyuka:
## What & why

Re-enables `UnitTestCalcPriceUsesCurrentWorkDate` (codeunit 139605 "Shpfy Product Price Calc. Test") for bug 642194 (stale WorkDate cache in the SingleInstance price-calc codeunit). The test was disabled NAV-side during uptake because its setup helper `CreateDatedAllCustDiscPriceList` (codeunit 139603) failed. Two separate problems surfaced once it ran end-to-end in NAV, both fixed here:

1. **Duplicate `Price List Line No.`** — the helper inserted two lines relying on the `Line No.` `AutoIncrement` field. Because `ShpfyInitializeTest.CreateShop()` commits mid-test while the `GenerateGUID` No. Series (the price-list code source) rolls back, the AutoIncrement sequence desyncs from committed data and the second insert reused an existing `Line No.` (`... Line No.='2' already exists`). Fixed by assigning `Line No.` explicitly via the base-app-standard `SetNextLineNo()` before each insert.

2. **Resolved price came back `0`** — the helper used dated *"All Customers" discount* price list lines on top of the item card price. In NAV, that path resolved the sales line `Unit Price` (and therefore `Line Amount`) to `0`, so the assertion failed (`Expected <102.4> ... actual <0>`). Switched the helper to dated *price* lines (`Amount Type = Price`): the applicable dated line sets the sales line `Unit Price` directly, so the resolved amount depends solely on the Work Date — which is exactly the behavior bug 642194 needs to exercise. Lines are activated explicitly and the header is activated after insert.

## Linked work

Fixes [AB#642673](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/642673)

## How I validated this

- AL language server reports **0 errors / 0 warnings** for both changed files; the only project-level error is the pre-existing, unrelated `MockAzureKeyVaultSecretProvider` DotNet gap in `ShpfyTestShopify.Codeunit.al` (local-env only, present on `main`).
- A private NAV uptake of this branch was run: the test now **passes** end-to-end.

## Test plan

- `UnitTestCalcPriceUsesCurrentWorkDate` runs green in the NAV uptake (both the up-to-boundary and from-boundary Work Date branches assert the expected dated price).
- Sibling tests in codeunit 139605 (`UnitTestCalcPriceTest`, `UnitTestCalcPriceTestNewPricing`) are unaffected — the shared init codeunit change only adds an explicit `Line No.` and switches the dated helper's amount type.

## Follow-up

Once this is uptaken, remove the NAV-side disable entry `App/DisabledTests/ShpfyProductPriceCalcTest.DisabledTest.json` to fully re-enable the test.


darjoo (approved):

gggdttt:
If we want to re-enable the test, should we delete the corresponding disable json file first?

gggdttt (approved):

PR #9934, opened 2026-08-04
Fix Show Capacity reference in Power BI Manufacturing semantic model

### LINKED ISSUE #9932: [Bug]: Manufacturing App – Capacity Measures Fails due to custom Capacity Unit of Measure codes
### Summary

Capacity-related visuals in the Manufacturing Power BI App return blank values when custom capacity unit of measure codes have been created and used in the **Show Capacity In** field on the Manufacturing Setup page. The bug is expressed when Show Capacity In is used to lookup values in the Time Factors table in the semantic model.

### Problem Description
Several measures use and depend on the selected capacity unit of measure from the Manufacturing Setup table:

`VAR SelectedUOM =
    SELECTEDVALUE('Manufacturing Setup'[Show Capacity In])`

The selected value is used in a lookup against the Time Factors table:

`VAR TimeFactor =
    LOOKUPVALUE(
        'Time Factors'[Time Factor],
        'Time Factors'[Cap Unit of Measure Type],
        SelectedUOM
    )`

The Time Factors[Cap Unit of Measure Type] column contains values like (Hours, Minutes, Seconds, Days), causing the lookup to fail if the user has created Capacity Unit of Measure records with custom Code values, such as TIMER. 

### Expected behavior

The capacity measures should work as expected regardless of the Code used in Show Capacity In. 

### Steps to reproduce

1. Create a new Capacity Unit of Measure with the Code = TIMER
2. On the Manufacturing Setup page, set **Show Capacity In** to the new unit of measure code.
3. Refresh the Manufacturing semantic model

Image

Image

### Additional context

_No response_

JNGZ:
## Linked work

Fixes #9932 
Fixes [AB#646115](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/646115)

## How I validated this

- Consumed the revised Manufacturing Setup query into an unchanged Manufacturing semantic model without issue.
- Updated the affected measures and verified that the semantic model and dependent reports load as expected.

## Risk & compatibility

- None in Business Central, as the changes are additive
- Low risk in Power BI. If a user updates the semantic model **without** updating the connector app in Business Central the dependent measures will break.


alexei-dobriansky:

<details>

The fix repairs the Manufacturing Power BI capacity measures, which return blank when the `Show Capacity In` value on Manufacturing Setup is a custom Capacity Unit of Measure code (for example `TIMER`). The root cause is a type mismatch: the measures looked up `Time Factors[Cap Unit of Measure Type]` (values Hours/Minutes/Seconds/Days) using the Capacity Unit of Measure **code**, so any code that is not literally one of those type names failed the lookup.

The change is correct and targeted at the root cause. The API query `Manufacturing Setup - PBI API` now adds a nested `Capacity Unit of Measure` dataitem linked on `code = Show Capacity In` and exposes its `code` and `type` columns; the semantic model adds matching `code`/`type` columns and every measure now looks up by `'Manufacturing Setup'[type]`. I confirmed all 21 measure references across Work Center, Machine Center, Capacity Ledger Entries, Prod Order Capacity Need and Production Orders were switched consistently, with no leftover `[Show Capacity In]` lookups. The existing `TestManufacturingSetup` still passes because it configures a valid capacity unit, so the new inner join matches and `showCapacityIn` is unchanged.
</details>

<details>

**S1** - *Assert the new type column in the API test*
`TestManufacturingSetup` already creates a Capacity Unit of Measure with a random code and Type `Minutes` (the exact bug scenario) but only checks `showCapacityIn`. Extend `VerifyManufacturingSetup` to also assert the returned `type` equals the Capacity Unit of Measure Type. This is the only automated check of the fix's core mechanism, and the test infrastructure is already in place.

**S2** - *Inner join hides the setup row when no unit matches*
The nested `capacityUnitOfMeasure` dataitem uses a default inner join. If `Show Capacity In` is blank or points to a missing Capacity Unit of Measure, the whole `manufacturingSetup` row drops and the API returns no rows. Consider a left outer join so `showCapacityIn` still comes back and the measures degrade gracefully.

**S3** - *Display text now shows the type, not the code*
The `Show Capacity In UOM` measure now returns the type name (for example `Show capacity in Minutes`) instead of the configured code (for example `TIMER`). Please confirm this visible change is intended.
</details>

<details>

**Risk:** Low and contained to the Manufacturing Power BI app. The Business Central change is additive (two new API columns), and the semantic model changes are DAX/column edits that cannot be unit-tested in AL. The main behavioral risk is the inner join in S2 and the display change in S3. Note also that the `Show Capacity In UOM` value now depends on the Type enum name matching `Time Factors[Cap Unit of Measure Type]`; the `Minutes` case is exercised by the test, but other enum values (e.g. `100/Hour`) are not verified. The `Validate link to issues` CI check is currently failing and should be resolved before merge.

**Necessity:** Justified. Issue #9932 describes a clear, reproducible defect where standard capacity visuals break for any custom capacity unit code - a realistic customer configuration. The scope matches a bug fix. There is no linked ADO work item (only GitHub issue #9932); the issue is detailed enough to review from, but the missing work item and failing link check should be confirmed by a maintainer.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9934 round=1 by=alexei-dobriansky at=2026-08-04T06:02:03Z lastSha=922f488cdecd3bd214a620d5334742519e7ba4eb suggestions=S1,S2,S3

JNGZ:
@alexei-dobriansky thanks for the review. 

I had a closer look at the model's behaviour when setting the **SqlJoinType** property on the nested data item. 

What I observed was the following:

1. **No SqlJoinType defined**:  single row returned with a default type of **Milliseconds**
2. **LeftOuterJoin**: single row returned with a default type of **Milliseconds**
3. **InnerJoin**; no rows returned 

image
image

It appears the default join is actually LeftOuterJoin, which the [documentation also suggests](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-query-links-joins#set-the-sqljointype-property).

Happy to set the SqlJoinType property explicitly if you think it's a safer approach. 

I also, experimented with setting **Minutes** as a fallback value **if Show Capacity In is blank**. This way the report is still viable, and and user has an option to change the setup later on. 

Keen to hear your thoughts too @mynjj

mynjj:
> @alexei-dobriansky thanks for the review.
> 
> I had a closer look at the model's behaviour when setting the **SqlJoinType** property on the nested data item.
> 
> What I observed was the following:
> 
> 1. **No SqlJoinType defined**:  single row returned with a default type of **Milliseconds**
> 2. **LeftOuterJoin**: single row returned with a default type of **Milliseconds**
> 3. **InnerJoin**; no rows returned
> 
> image image
> It appears the default join is actually LeftOuterJoin, which the [documentation also suggests](https://learn.microsoft.com/en-us/dynamics365/business-central/dev-itpro/developer/devenv-query-links-joins#set-the-sqljointype-property).
> 
> Happy to set the SqlJoinType property explicitly if you think it's a safer approach.
> 
> I also, experimented with setting **Minutes** as a fallback value **if Show Capacity In is blank**. This way the report is still viable, and and user has an option to change the setup later on.
> 
> Keen to hear your thoughts too @mynjj

Thanks for verifying, you are a responsible one 😄 this is one of the common hallucinations in models, 🤖s think that the default union type in AL queries is inner, when it's not. But you proved them otherwise.

On another topic, your change is the first one that will make use of the new deployment mechanism! I recently merged some changes so that we no longer depend on AppSource as a delivery mechanism of our apps. Instead it will be directly in Business Central for users to deploy (see this blog post https://yzhums.com/72067/ for evaluation companies, but it will be made available in general)

Whenever we need to make an update and surface it within Business Central, we will need to update the corresponding "manifest" of the report. This is specially useful in this scenario where both the API and the report need to evolve at the same time.

So the steps to do that are:
- Update the manifest (the codeunit implementing the "Power BI Deployable Report" interface), for this app it is the PBIManufacturingApp.Codeunit.al. Bump the version number.
- Update the corresponding PBIX, now in the `.resources` folder. This should be the updated report that matches the AL changes.

And that's it, when the customers get the new version of the AL app, they will be prompted to install the update of their Manufacturing report.

I'll do this steps on your PR.

And on another topic, for this fix in particular, seems like the hardcoded table that it looks up against `Time Factors` doesn't have the blank value of the enum

image

I'm also committing adding it (and while I'm at it, making it a bit more TMDL-friendly).

mynjj (dismissed):

ventselartur (dismissed):

alexei-dobriansky:

<details>

Since round 1, three commits were pushed: the maintainer added the missing blank entry to the `Time Factors` lookup table (addressing the gap where an unset Capacity Unit of Measure Type would not match), rewrote the table source in a more TMDL-friendly `#table(...)` form, bumped the semantic model version to 2 in `PBIManufacturingApp.Codeunit.al`, and added the missing `using Microsoft.Manufacturing.Capacity;` namespace to the query file.

These changes are correct and complete the fix end-to-end. The blank entry `{" ", 1}` matches the BC enum's blank option name, so an unset Capacity Unit of Measure Type will now resolve to a time factor of 1 instead of a blank lookup. The namespace addition resolves the implicit dependency on `Capacity Unit of Measure`. The version bump will trigger the new deployment mechanism so users are prompted to update the report when the AL app is updated.
</details>

<details>

|----|-------|--------|----------------|
| S1 | Assert the new type column in the API test | Not addressed | No test change in new commits |
| S2 | Inner join hides the setup row when no unit matches | Disputed | Author confirmed with BC documentation that the default join is LeftOuterJoin, not InnerJoin; maintainer verified with screenshots |
| S3 | Display text now shows the type, not the code | Not addressed | No explicit reply, but validation section confirms the updated measures work as expected; appears intentional |
</details>

<details>

No new blocking issues. One low-priority note:

**S4** - *Time Factors now covers all enum values including blank*
The new row `{" ", 1}` uses a single space, which matches the BC option field blank value. All five named enum values (Seconds, Minutes, 100/Hour, Hours, Days) are now present. The 100/Hour row uses `36000` ms per 100 operations — confirm this is the intended time factor for that unit (Hours = 3,600,000 ms, so 100/Hour = 36,000 ms per 100 operations is correct).
</details>

<details>

**Risk:** Low. The AL change is additive (namespace import only), the version bump triggers an update prompt but does not break existing deployments, and the TMDL rewrite is purely cosmetic (same data, cleaner syntax). The blank Time Factor row prevents a null-match for users who have not configured a Capacity Unit of Measure, which is a safe default. S1 (missing type assertion in the test) remains the only gap: the test creates the right scenario but does not verify the `type` column value returned by the API, so a regression in the join or column mapping would not be caught automatically.

**Necessity:** Unchanged from round 1. The fix is justified and the new commits bring the PR to a shippable state. S3 (display change from code to type name) appears intentional based on the validation description; if it is not, it should be confirmed before merge.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=1 system=github pr=9934 round=2 by=alexei-dobriansky at=2026-08-07T11:36:44Z lastSha=8a91f230306b4df0eb790f58db6d43a33d48c09c reviewKey=03dd2b3f0ae76e9d168e86e61f2164cb920f6a6349c2e660ee9c8924e8055203 suggestions=S1@086adbfa:notaddressed,S2@389d20f2:disputed,S3@5d62b5b4:notaddressed,S4@2d6b0e25:new parentRound=1

PredragMaricic (changes requested):
The fallback key does not match the value emitted by the API for a blank or unmatched setup.

`Capacity Unit of Measure.Type` is an enum whose value 0 has internal name `" "` but caption `Milliseconds`. The query's left outer join preserves the Manufacturing Setup row when no matching unit exists, but the imported `type` value is formatted as `Milliseconds`. `Time Factors` now adds `{ " ", 1 }`, so `LOOKUPVALUE` still finds no matching factor and the affected measures remain blank for that fallback case.

Please key the fallback row by the actual imported value (or normalize the API/model consistently), and extend `TestManufacturingSetup()` to assert the new `code` and `type` fields plus the blank/unmatched case so this boundary is covered.

The failing CH legacy job appears unrelated: its failures are in Service Reports line sorting and VAT setup.

PR #9948, opened 2026-08-04
Move Contact Sync User validation to table subscribers

Naveen26423:
## Linked work

Fixes [AB#645036](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/645036)

## How I validated this


src/Layers/W1/BaseApp/CRM/Outlook/ContactSyncUserSubscriber.al
```diff
@@ -0,0 +1,17 @@
+namespace Microsoft.CRM.Outlook;
+codeunit 7100 "Contact Sync UserSubscriber"
+{
```
darjoo, line 8: Make this `access=internal`. It'll be easier to get rid of this CU in the future if we do not need it.

src/Layers/W1/BaseApp/CRM/Outlook/O365BidirectionalSync.Codeunit.al
```diff
@@ -461,6 +478,17 @@ codeunit 7106 "O365 Bidirectional Sync"
...
     end;
 
+    local procedure IsApprovedGraphRequestUri(UriToValidate: Text): Boolean
+    var
+        NormalizedUri: Text;
+    begin
+        if UriToValidate = '' then
+            exit(false);
+
+        NormalizedUri := LowerCase(DelChr(UriToValidate, '<>', ' '));
```
darjoo: This should use the URI module like in the ContactSyncUser.Table.al

src/Layers/W1/BaseApp/CRM/Outlook/ContactSyncUserSubscriber.al
```diff
@@ -0,0 +1,17 @@
...
+    [EventSubscriber(ObjectType::Table, Database::"Contact Sync User", 'OnBeforeInsertEvent', '', false, false)]
```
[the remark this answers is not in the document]
darjoo: It is correct about another extension running running after this one. However, it is incorrect that OnInsert/OnModify is guaranteed to run.

Fx Rec.Insert(false) will cause the trigger to not run.

darjoo (dismissed):

darjoo (dismissed):

alexei-dobriansky:

<details>

This PR adds validation around the Contact Sync User table and the O365 contact sync flow. It tries to ensure that only the current user can change that user's sync record, and that stored or returned Graph URLs are approved before they are used.

The ownership check is a targeted change, and there is no BaseApp publisher dependency because the new subscribers use table events. The URL protection is not complete: the helper used here only checks the URI host, telemetry still writes token-bearing Graph state URLs, and the current W1 build fails before runtime validation can be trusted.
</details>

<details>

**S1** - *URL validation checks only the host*
`ValidateIntegrationURL` only compares hosts, so `http://graph.microsoft.com/...` and other non-v1.0 Graph paths can pass. Add an explicit HTTPS and `https://graph.microsoft.com/v1.0/` prefix check before storing or using a delta, next, or request URL. This is needed because the change is meant to protect Graph delta URLs.

**S2** - *Telemetry still logs token URLs*
`0000QT9` and `0000QTG` still write the full request, nextLink, or delta URL to telemetry. Graph nextLink and deltaLink values carry state tokens, so log only a fixed message or endpoint type. This leaves the token exposure scenario partly open.

**S3** - *Build fails on unused using*
`ContactSyncUser.Table.al` adds `using System.Security.User;`, but the W1 build reports AL0792 for it. Remove the unused using directive so the app compiles. This is a merge blocker even before runtime testing.

**S4** - *Add regression tests for security rules*
No test file changes cover the new table subscribers or URL checks. Add tests for own-user versus other-user writes and for invalid schemes or paths, using the existing Contact Sync or URI test areas. This area is security-sensitive and the behavior is testable.

**S5** - *Persist invalid delta URL cleanup*
`GetContacts` calls `UpdateDeltaUrl(FolderId, '')`, but `UpdateDeltaUrl` skips blank links. Let blank input clear the field or remove the call. Otherwise a cancelled or failed fallback keeps the bad delta URL for the next run.
</details>

<details>

**Risk:** this is a security and privacy-sensitive Contact Sync change. It affects all inserts and modifies of table 7121, and it controls which Graph delta, next, and request URLs can be stored, logged, and used for HTTP calls. A wrong fix can either keep leaking Graph state tokens or block valid contact sync records.

**Necessity:** the linked Deliverable describes a delta sync URL token exposure fix, so the change is needed. The scope is the right area, but it is not complete until URL validation is strict, token-bearing telemetry is removed, the build is green, and regression tests cover the security rules.
</details>

---
> [AI-PR-REVIEW] version=1 system=github pr=9948 round=1 by=alexei-dobriansky at=2026-08-04T14:04:21Z lastSha=00516bc1b33d43e95b676ef8b2f243b4f0622f7a suggestions=S1,S2,S3,S4,S5

alexei-dobriansky:

<details>

This PR adds validation for Contact Sync User delta URLs, enforces record ownership through table event subscribers, and checks Microsoft Graph request, nextLink, and deltaLink URLs before use.

The security scenario is valid and important, but the fix is incomplete. The fallback path for a rejected stored delta URL calls `UpdateDeltaUrl(FolderId, '')`, while `UpdateDeltaUrl` still ignores empty values, so the bad URL remains in table 7121. The same sync flow can still write full Graph URLs to telemetry message text, which keeps the token exposure risk open.
</details>

<details>

**S1** - *Clear rejected delta URLs persistently*
UpdateDeltaUrl(FolderId, '') does not change the record because UpdateDeltaUrl only writes non-empty links. Allow an explicit blank value, or add a clear helper, so a rejected delta URL is removed from Contact Sync User.

**S2** - *Stop logging full Graph URLs*
Remove full Graph request and nextLink values from telemetry message text. These URLs can contain delta tokens, so checking the host is not enough to make them safe to log.
</details>

<details>

**Risk:** High. `O365BidirectionalSync.Codeunit.al` handles stored Microsoft Graph delta URLs and paging links; the current diff still leaves rejected stored links in place and still has telemetry paths that include full Graph URLs.

**Necessity:** Required. This is a security-sensitive Contact Sync URL handling change, and the scope is right, but it must prevent unsafe URLs from being called, stored, and logged.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=1 system=github pr=9948 round=1 by=alexei-dobriansky at=2026-08-04T15:47:57.268266Z lastSha=1e9632ad0d393154cbc0aa46478bd3a951dba84b reviewKey=975131379297012f6819ef0c122353f11fca81b35214bd1e6e0aab5d83a6b3ab suggestions=S1@6ad323a1,S2@7cda4d9b

PR #9958, opened 2026-08-04
[Main] -Detail Trial Balance - Running Balance Incorrect When "Include Closing Entries" Is Disabled

DhavalMore88:
[Bug 645090](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/645090): [master] [ALL-E] Detail Trial Balance - Running Balance Incorrect When "Include Closing Entries" Is Disabled

[AB#645090](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/645090)

**Issue**: When running the Detail Trial Balance report with "Include Closing Entries Within the Period" disabled, closing entries were excluded from the printed lines but their amounts still contributed to the "Balance" (running balance) column, causing the displayed balance to not reconcile with the entries actually shown on the report.

**Cause**: In the "G/L Entry" OnAfterGetRecord trigger in DetailTrialBalance.Report.al, GLBalance := GLBalance + Amount executed unconditionally before the closing-entry check (if ClosingEntry and not PrintClosingEntries then CurrReport.Skip(),so a skipped/hidden closing entry's amount was still accumulated into the running balance.

**Solution**: Reordered the trigger to compute ClosingEntry and perform CurrReport.Skip() before the GLBalance := GLBalance + Amount line, matching the pattern already used for the reversed-entries and corrections-only filters — so a closing entry that isn't printed no longer contributes to the balance.

### DIFF
--- a/src/Layers/BE/BaseApp/Finance/GeneralLedger/Reports/DetailTrialBalance.Report.al
+++ b/src/Layers/BE/BaseApp/Finance/GeneralLedger/Reports/DetailTrialBalance.Report.al
@@ -160,18 +160,12 @@ report 4 "Detail Trial Balance"
                         if not PrintReversedEntries and Reversed then
                             CurrReport.Skip();
 
-                        GLBalance := GLBalance + Amount;
-                        if ("Posting Date" = ClosingDate("Posting Date")) and
-                           not PrintClosingEntries
-                        then begin
-                            "Debit Amount" := 0;
-                            "Credit Amount" := 0;
-                        end;
+                        ClosingEntry := "Posting Date" = ClosingDate("Posting Date");
 
-                        if "Posting Date" = ClosingDate("Posting Date") then
-                            ClosingEntry := true
-                        else
-                            ClosingEntry := false;
+                        if ClosingEntry and not PrintClosingEntries then
+                            CurrReport.Skip();
+
+                        GLBalance := GLBalance + Amount;
 
                         NumberOfGLEntryLines += 1;
                     end;
--- a/src/Layers/BE/Tests/Report/ERMFinancialReports.Codeunit.al
+++ b/src/Layers/BE/Tests/Report/ERMFinancialReports.Codeunit.al
@@ -1635,6 +1635,63 @@ codeunit 134982 "ERM Financial Reports"
         end;
     end;
 
+    [Test]
+    [HandlerFunctions('RHDetailTrialBalance')]
+    [Scope('OnPrem')]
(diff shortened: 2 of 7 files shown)


alexei-dobriansky:

# **Recommendation: Accept**

<details>

The Detail Trial Balance report was adding every G/L entry to the running balance before deciding whether to exclude closing entries. When "Include Closing Entries Within the Period" was disabled, the old code zeroed out the Debit/Credit display fields for closing entries but had already accumulated their amounts into `GLBalance`. The fix reorders the logic: `ClosingEntry` is evaluated first, and if the entry is a closing entry that should be excluded, `CurrReport.Skip()` is called immediately — before `GLBalance := GLBalance + Amount`. This means excluded entries are skipped entirely (they do not appear in the report dataset and do not affect the running balance), which is the correct behavior. The fix addresses the root cause directly: balance accumulation was happening unconditionally, and it now only happens for entries that pass the inclusion checks.
</details>

<details>

No suggestions.
</details>

<details>

**Risk:** Low. The change only affects the Detail Trial Balance report (`DetailTrialBalance.Report.al`) and only when "Include Closing Entries Within the Period" is set to false. Non-closing entries and the case where closing entries are included are not affected — the `CurrReport.Skip()` branch is not taken in those paths. There is a minor behavior change: the old code printed closing-entry rows with zero Debit/Credit amounts; the new code omits them entirely. This is the intended behavior (a report option called "Include Closing Entries" should hide the rows, not show them with zeros). The `ClosingEntry` report column remains correctly set for rows that do appear. The fix is applied in both the W1 and BE layers, which is the complete set of layers that extend this report.

**Necessity:** The bug causes the running balance column to include closing-entry amounts even when those entries are supposed to be excluded, producing incorrect totals in a financial report. This is a correctness issue in a finance area (general ledger reporting), and the fix is well-scoped to the exact lines that caused the problem. Tests are added in all five affected test layers (W1, BE, CZ, DE, IT), covering the exact scenario from the bug report.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=1 system=github pr=9958 round=1 by=alexei-dobriansky at=2026-08-06T13:10:00Z lastSha=09938f8840e94a0dcf644fd870cec9220fd9077e reviewKey=20ff86549c5260adf4c32382b0052b4088169b49adac1c18e0c02bb93c7a6c4b suggestions=

djukicmilica (approved):

v-praghav (approved):

PR #9971, opened 2026-08-05
[Master]-Contoso - The binding of codeunit 11489 was unsuccessful. The codeunit has already been bound. - Copy

neeleshsinghal:


PredragMaricic (changes requested):
S1 (High): The linked bug is caused by manual subscribers remaining bound when demo-data generation fails. This change only avoids one specific `NET INCOME` lookup failure. `US Contoso Localization` still binds codeunit 11489 in `OnBeforeGeneratingDemoData` and only unbinds it in `OnAfterGeneratingDemoData`, so any other exception still leaves the subscriber bound and the next run fails with the same error. Please fix the binding lifecycle and add a regression test that forces generation to fail, then verifies a second run does not report an already-bound codeunit.

S2 (High): `GetAccountNo` previously guaranteed a valid account number or raised an explicit lookup error. If neither the temporary mapping nor a G/L Account name matches, the new implementation falls through and returns blank, allowing invalid account references to propagate. Preserve explicit failure when both lookups miss.

neeleshsinghal:
> S1 (High): The linked bug is caused by manual subscribers remaining bound when demo-data generation fails. This change only avoids one specific `NET INCOME` lookup failure. `US Contoso Localization` still binds codeunit 11489 in `OnBeforeGeneratingDemoData` and only unbinds it in `OnAfterGeneratingDemoData`, so any other exception still leaves the subscriber bound and the next run fails with the same error. Please fix the binding lifecycle and add a regression test that forces generation to fail, then verifies a second run does not report an already-bound codeunit.
> 
> S2 (High): `GetAccountNo` previously guaranteed a valid account number or raised an explicit lookup error. If neither the temporary mapping nor a G/L Account name matches, the new implementation falls through and returns blank, allowing invalid account references to propagate. Preserve explicit failure when both lookups miss.

s1 fixed but generally test was not found for demodata
s2 fixed

PredragMaricic (approved):
Reviewed the idempotent subscription-binding change for rerunning Contoso generation after an earlier failure.

v-ankitgoyal (approved):

haoranpb (changes requested):
This doesn't look right, why was it failing to begin with?

neeleshsinghal:
> This doesn't look right, why was it failing to begin with?

Hi haoranpb@outlook.com,
This was passing earlier. Requeuing it. If not, this fix please suggest what are other option we can proceed. We did not find this type of case handled anywhere on Business central. Please suggest how can we proceed further.

haoranpb:
> > This doesn't look right, why was it failing to begin with?
> 
> Hi [haoranpb@outlook.com](mailto:haoranpb@outlook.com), This was passing earlier. Requeuing it. If not, this fix please suggest what are other option we can proceed. We did not find this type of case handled anywhere on Business central. Please suggest how can we proceed further.

Sorry I was a bit vague on the previous message. I meant the code changes do not look right, see the comments on the linked bug for more details.

PR #10017, opened 2026-08-06
[Master] - Bug 644751 Moving an expense report line to another report wrongly reports a duplicate

v-rohangarg20:
Workitem Bug 644751: [master] Moving an expense report line to another report wrongly reports a duplicate

Fixes [AB#644751](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/644751)

Issue: When an expense report line is moved from one open expense report to another, the target report immediately raises a "duplicate expense" rule violation, even though the line was relocated (not copied). No real duplicate exists — it is the same logical line in a new report.

Cause: In ExpenseRuleValidation, ExistDuplicateInExpenseReportLine looks for matching lines (Expense Ext. Doc. No., Expense Date, Merchant Name, Amount) across ALL expense reports and only excludes the current record by SystemId. During a move the relocated line gets a new record/SystemId while the original still exists (or is matched by content), so the check treats the moved line as a duplicate of its own source line.

Solution: Update the duplicate check so a line being moved is not compared against its own source line — exclude the originating line (not just by SystemId of the new record) when scanning for duplicates, so ExistDuplicateInExpenseReportLine only flags genuinely different lines. Moving a line to another open report no longer reports a false duplicate, while true duplicates within/across reports and posted reports are still detected. Added/updated tests to cover the move scenario.

### DIFF
--- a/src/Apps/W1/ExpenseAgent/app/src/ExpenseReport/Tables/ExpenseReportLine.Table.al
+++ b/src/Apps/W1/ExpenseAgent/app/src/ExpenseReport/Tables/ExpenseReportLine.Table.al
@@ -1969,10 +1969,11 @@ table 6907 "Expense Report Line"
         ExpenseReport.CopyReportLineAttachments(SourceDocNo, SourceLineNo, NewLine."Document No.", NewLine."Line No.");
 
         NewLine.UpdateAmounts();
-        NewLine.ApplyRule(false, true);
-        NewLine.Modify();
 
         Rec.Delete(true);
+
+        NewLine.ApplyRule(false, true);
+        NewLine.Modify();
 
         if NewLine."Expense No." <> '' then
             if Expense.Get(NewLine."Expense No.") then begin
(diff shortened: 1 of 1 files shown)


BardurKnudsen (approved):

alexei-dobriansky:

<details>

This PR moves `Rec.Delete(true)` before `NewLine.UpdateAmounts()` and `NewLine.ApplyRule(false, true)` in `MoveToExpenseReport`. That matches the bug: the duplicate scan compares by receipt, date, merchant, and amount, so the source line must be gone before the target line is validated. The change is narrow and should still detect real duplicates in other open or posted expense reports, because only the moved source line is removed before validation.
</details>

<details>

**S1** - *Add the missing move duplicate test*
Add a regression test that moves a line whose receipt, date, merchant, and amount match only its source line, then assert that the target line has no duplicate rule violation. This path deletes and recreates expense report data, and the PR currently changes only production code.
</details>

<details>

**Risk:** The regression surface is `MoveToExpenseReport` on table `Expense Report Line`. It deletes the source line, copies related records, recalculates amounts, reapplies rules, and updates the linked `Expense`, so an incorrect order can leave wrong rule violations or move report data incorrectly. There is no public API or BaseApp event dependency in this diff.

**Necessity:** The bug is valid and important because moving a line to another open report should not report the same logical line as a duplicate. The fix is small and in the right procedure, but this data-integrity flow needs a regression test for the exact false-duplicate scenario before merge.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=10017 round=1 by=alexei-dobriansky at=2026-08-23T07:07:21.414Z lastSha=bb543fee56c44cddacd756a45873e5f00359538d reviewKey=a21f02e758df5270686c3ac3282574ec38dc25f191ff556ff3ecc51a26782384 suggestions=S1@8d9ef7e7

MonicaPAhuja (approved):

djukicmilica (approved):

PR #10020, opened 2026-08-06
[Master] - Bug 641894: [Expense Agent] [Rules] Mileage expense should not validate require receipt number rule

v-rohangarg20:
Workitem Bug 641894: [master] [Expense Agent] [Rules] Mileage expense should not validate require receipt number rule

Fixes [AB#641894](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/641894)

Issue: When "Require receipt number" (Receipt No. Mandatory) is enabled in Expense Agent Setup, a mileage expense / expense report line without a receipt number wrongly raises a rule violation. Mileage is distance-based and never has a receipt, so the line can never satisfy the rule and is incorrectly blocked.

Cause: In ExpenseRuleValidation the receipt-number-mandatory check only tested whether "Expense Ext. Doc. No." was blank. It did not exclude mileage expenses, so the rule was applied to every expense type including Mileage (Expense Detail Required = Mileage).

Solution: Guard the require-receipt-number rule so it is skipped for mileage. In both InitialExpenseValidation and InitialExpenseReportLineValidation the check now also verifies "Expense Detail Required" <> Mileage before raising the violation, consistent with the existing Merchant Name Mandatory rule. Mileage expenses/report lines no longer trigger a false "receipt number required" violation, while non-mileage expenses still enforce the rule. Added/updated tests to cover the mileage case.


src/Apps/W1/ExpenseAgent/app/src/MasterData/Pages/ExpenseRuleCard.Page.al
```diff
@@ -75,4 +81,26 @@ page 6901 "Expense Rule Card"
             }
         }
     }
+
+    trigger OnAfterGetRecord()
+    begin
+        SetMerchantRequirementsVisibility();
+    end;
```
BardurKnudsen: Not needed - this just means that you call the function twice every time

src/Apps/W1/ExpenseAgent/app/src/MasterData/Pages/ExpenseRuleCard.Page.al
```diff
@@ -53,6 +58,7 @@ page 6901 "Expense Rule Card"
             group(Merchant)
             {
                 Caption = 'Merchant Requirements';
+                Visible = MerchantRequirementsVisible;
```
BardurKnudsen: Consider using the 'Enabled' property instead of hiding the field. It would cause flickering (if it even works) when you browse the card using the left/right arrows.

src/Apps/W1/ExpenseAgent/app/src/MasterData/Tables/ExpenseRuleHeader.Table.al
```diff
@@ -56,10 +61,24 @@ table 6927 "Expense Rule Header"
         field(40; "Required Specific Merchant"; Boolean)
         {
             Caption = 'Required Specific Merchant';
+
+            trigger OnValidate()
+            begin
+                if not Rec."Required Specific Merchant" then
```
BardurKnudsen, line 67: Looks like the field name should be "Allow Specific Merchant" and not "Required ..." ?

BardurKnudsen (dismissed):

src/Apps/W1/ExpenseAgent/app/src/MasterData/Pages/ExpenseRuleCard.Page.al
v-rohangarg20: Updated. Thanks!

src/Apps/W1/ExpenseAgent/app/src/MasterData/Pages/ExpenseRuleCard.Page.al
v-rohangarg20: Updated. Thanks!

src/Apps/W1/ExpenseAgent/app/src/MasterData/Tables/ExpenseRuleHeader.Table.al
v-rohangarg20: This field has already been introduced in the releases branch. Renaming it would be a breaking change.

MonicaPAhuja (approved):

BardurKnudsen (approved):

PR #10023, opened 2026-08-06
[Master]-Customer Ledger Bill Entries display default Customer Posting Group instead of Alternative Customer Posting Group set on Sales Documents in the Spanish version.

neeleshsinghal:
Fixes [AB#644753](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/644753)

WorkItem [Bug 644753](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/644753): [master]-[ALL-E]-Customer Ledger Bill Entries display default Customer Posting Group instead of Alternative Customer Posting Group set on Sales Documents in the Spanish version.

**ISSUE**:- Customer Ledger Bill Entries display default Customer Posting Group instead of Alternative Customer Posting Group set on Sales Documents in the Spanish version.

**CAUSE**:- InvoiceSplitPayment.SplitSalesInv creates the Bill and Close Invoice Gen. Journal Line records without setting the Posting Group field. As a result, when the lines are posted through GenJnlPostLine.PostCust, the empty posting group defaults to the customer's current Customer Posting Group instead of using the posting group from the original invoice, which can result in incorrect customer ledger postings.

**SOLUTION** :- The invoice split posting logic was updated to explicitly set the journal line posting group from the sales document’s Customer Posting Group in all relevant bill-creation paths.

### DIFF
--- a/src/Layers/ES/BaseApp/Local/Finance/ReceivablesPayables/InvoiceSplitPayment.Codeunit.al
+++ b/src/Layers/ES/BaseApp/Local/Finance/ReceivablesPayables/InvoiceSplitPayment.Codeunit.al
@@ -124,6 +124,7 @@ codeunit 7000005 "Invoice-Split Payment"
             GenJnlLine."Reason Code" := SalesHeader."Reason Code";
             GenJnlLine."External Document No." := GenJnlLineExtDocNo;
             GenJnlLine.Validate("Currency Code", SalesHeader."Currency Code");
+            GenJnlLine."Posting Group" := SalesHeader."Customer Posting Group";
             GenJnlLine.Amount := -TotalAmount;
             GenJnlLine."Amount (LCY)" := -TotalAmountLCY;
             GenJnlLine."System-Created Entry" := true;
@@ -194,6 +195,7 @@ codeunit 7000005 "Invoice-Split Payment"
         GenJnlLine."Reason Code" := SalesHeader."Reason Code";
         GenJnlLine."External Document No." := GenJnlLineExtDocNo;
         GenJnlLine.Validate("Currency Code", SalesHeader."Currency Code");
+        GenJnlLine."Posting Group" := SalesHeader."Customer Posting Group";
         GenJnlLine."System-Created Entry" := true;
         GenJnlLine."On Hold" := SalesHeader."On Hold";
         GenJnlLine."Source Code" := SourceCode;
--- a/src/Layers/ES/Tests/Local/CarteraPosting.Codeunit.al
+++ b/src/Layers/ES/Tests/Local/CarteraPosting.Codeunit.al
@@ -1597,6 +1597,76 @@ codeunit 147305 "Cartera Posting"
             StrSubstNo(FieldErr, GLEntry.FieldCaption("Credit Amount"), GLEntry.TableCaption()));
     end;
 
+    [Test]
+    procedure BillCustLedgEntryUsesAlternativePostingGroupFromSalesInvoice()
+    var
+        CustLedgerEntry: Record "Cust. Ledger Entry";
+        Customer: Record Customer;
+        CustomerPostingGroup2: Record "Customer Posting Group";
+        SalesHeader: Record "Sales Header";
+        SalesInvoiceNo: Code[20];
+    begin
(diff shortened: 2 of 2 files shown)


src/Layers/ES/BaseApp/Local/Finance/ReceivablesPayables/InvoiceSplitPayment.Codeunit.al
```diff
@@ -124,6 +124,7 @@ codeunit 7000005 "Invoice-Split Payment"
             GenJnlLine."Reason Code" := SalesHeader."Reason Code";
             GenJnlLine."External Document No." := GenJnlLineExtDocNo;
             GenJnlLine.Validate("Currency Code", SalesHeader."Currency Code");
+            GenJnlLine."Posting Group" := SalesHeader."Customer Posting Group";
```
djukicmilica, line 127: should we validate posting group?

alexei-dobriansky:

# **Recommendation: Accept**

<details>

This PR fixes a bug in the Spanish (ES) Cartera posting flow where the `SplitSalesInv` procedure in `InvoiceSplitPayment` created Gen. Journal Line records for both the close-invoice entry and the bill entry without setting the `Posting Group` field. When a sales document uses an alternative Customer Posting Group, the omission caused `GenJnlPostLine` to fall back to the customer's default posting group, resulting in incorrect Customer Ledger Entry postings for bills.

The fix adds `GenJnlLine."Posting Group" := SalesHeader."Customer Posting Group"` to both Gen. Journal Line construction paths in `SplitSalesInv`. The Purchase counterpart (`SplitPurchInv`) already had the equivalent assignment (`PurchHeader."Vendor Posting Group"`) in both its paths, confirming this is the right pattern and that the Sales side was simply missing it. The fix directly addresses the root cause: with the assignment in place, `GenJnlPostLine` uses the posting group the document was created with, not the customer's current default.

</details>

<details>

No suggestions. The fix is correct, targeted, and the tests cover the key scenario.

</details>

<details>

**Risk:** The change affects only the ES country layer (`src/Layers/ES`). It is a two-line addition, strictly additive, and mirrors code that already exists in `SplitPurchInv`. The only scenarios that could behave differently after the fix are those where a sales document is posted with an alternative Customer Posting Group and payment method `Create Bills = true` — exactly the broken scenario. All other scenarios are unaffected because when the posting group on the Sales Header equals the customer's default, the new assignment sets the same value that `GenJnlPostLine` would have defaulted to anyway. No public API, event signature, or table schema changes.

**Necessity:** The scenario is clearly described, reproducible from the PR body, and confirmed by the two new tests. Without this fix, any ES customer using alternative Customer Posting Groups on sales documents will get bill ledger entries posted against the wrong receivables account, which is a financial correctness issue. The fix is appropriately scoped and the right channel for this type of targeted country-layer correction.

</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=1 system=github pr=10023 round=1 by=alexei-dobriansky at=2026-08-06T13:08:36Z lastSha=80dd80573965a7861b51f8bded99b0ab80d8887f reviewKey=ae37116096bd6d0c15f75f1b70f839f562929c50b0baca7bf96eac625d09a3ed suggestions=

src/Layers/ES/BaseApp/Local/Finance/ReceivablesPayables/InvoiceSplitPayment.Codeunit.al
neeleshsinghal, line 127: I implemented this by following the purchase logic and applying the same approach to the sales scenario.

djukicmilica: thanks

djukicmilica (approved):

v-praghav (commented):
Reviewed the changes against the Microsoft BC patterns. No actionable findings: the Sales Header Customer Posting Group is copied in both generated journal-line paths, matching the existing purchase-side implementation, and the added tests cover alternative and default posting-group behavior.

v-praghav (approved):

PR #10055, opened 2026-08-07
Fix: migrating a document attachment deletes shared Tenant Media

Groenbech96:
## Status

The RED commit is verified: run [31368546162](https://github.com/microsoft/BCApps/actions/runs/31368546162) failed in **all 24 Unit Tests jobs** with exactly the two predicted assertions, while `DeleteFromInternalRemovesMediaWhenNotShared` passed and every `Build Apps` job was green. The fix is now on the branch (`d5b46d51d1`).

## The defect

`DA External Storage Impl.DeleteFromInternalStorage` deletes the `Tenant Media` row unconditionally after moving an attachment to external storage:

```al
if TenantMedia.Get(DocumentAttachment."Document Reference ID".MediaId()) then begin
    TenantMedia.Delete();                        // <-- global act, decided per record
    DocumentAttachment.MarkAsDeletedInternally();
```

`Tenant Media` is **shared storage**. `Document Attachment Mgmt.CopyAttachments` / `CopyAttachmentsForPostedDocs` copy attachments onto posted documents with `TransferFields` + `Insert`, which copies the *media reference* rather than the bytes, and `Document Attachment.OnInsert` never re-imports content. So Vendor -> Purchase Header -> Purch. Inv. Header attachments all end up referencing a single `Tenant Media` row.

Deleting that row on behalf of one attachment destroys the content of every attachment copied from it. Those attachments are then skipped **silently** by `UploadToExternalStorage`:

```al
if not DocumentAttachment."Document Reference ID".HasValue() then
    exit(false);        // no telemetry - this guard sits above LogFeatureUsed()
```

`Media.HasValue()` also verifies the media *exists in the database*, so it returns `false` and the record is left with `Stored Externally = false`, a blank `External File Path`, and no internal content. Opening it then raises `Tenant Media does not exist. ID='{00000000-0000-0000-0000-000000000000}'` from `GetAsTempBlob`.

Note the app already guards the *external* blob against exactly this hazard - `"Skip Delete On Copy" := ToDocumentAttachment."Stored Externally"` protects `DeleteFromExternalStorage`. The equivalent protection was never applied to internal `Tenant Media`.

## Field impact

Observed on a production tenant migrating ~5,200 attachments in one company:

| | |
|---|---|
| Records iterated by report 8752 | 5,226 |
| Uploaded successfully | 4,810 |
| **Silently skipped, content destroyed** | **416** |

Skips were **0/min** while processing Vendor (23), Item (27), Sales Header (36), Purchase Header (38) and Sales Invoice Header (112), then rose to 30-60/min for the entire Purch. Inv. Header (122) block - i.e. the copy targets started failing only once their sources had been migrated.

## Tests added

`DA Ext. Storage Impl. Tests`:

| Test | Asserts | On the RED commit |
|---|---|---|
| `DeleteFromInternalKeepsMediaSharedWithCopiedAttachment` | Shared `Tenant Media` survives; the copy keeps its content | **FAIL** |
| `DeleteFromInternalRemovesMediaWhenNotShared` | Space is still reclaimed when the attachment is the sole owner | PASS |
| `UploadSucceedsForCopiedAttachmentAfterSourceIsMigrated` | After the source is migrated, the copy still uploads and gets its own external file | **FAIL** |

The middle test is expected to pass both before and after the fix. It is there to prove the fix does not simply stop reclaiming database space, which is the whole point of the Move operation.

A helper `CreateCopyOfDocumentAttachment` mirrors how `Document Attachment Mgmt` copies attachments, and asserts the shared-media precondition, so the tests cannot pass for the wrong reason if that copy behaviour ever changes.

## The fix

```al
OtherDocumentAttachment.SetRange("Document Reference ID", DocumentAttachment."Document Reference ID");
if OtherDocumentAttachment.Count() <= 1 then
    if TenantMedia.Get(DocumentAttachment."Document Reference ID".MediaId()) then
        TenantMedia.Delete();
```

Index-backed by the existing `key(Key2; "Document Reference ID")` on table 1173. `Count() > 1` rather than excluding self, since the primary key is five fields and "more than one owner" is the actual question.

[AB#646505](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/646505)


Groenbech96:
@StefanSosic can you take a look at this change and let me know what you think.

nikolakukrika (dismissed):

StefanSosic:
> Stefan Sosic (@StefanSosic) can you take a look at this change and let me know what you think.

Taking a look on it today! Thanks

Groenbech96:
> > Stefan Sosic (Stefan Sosic (@StefanSosic)) can you take a look at this change and let me know what you think.
> 
> Taking a look on it today! Thanks

@StefanSosic are you ready to approve?

StefanSosic (approved):

StefanSosic:
> > > Stefan Sosic (Stefan Sosic (@StefanSosic)) can you take a look at this change and let me know what you think.
> > 
> > Taking a look on it today! Thanks
> 
> @StefanSosic are you ready to approve?

Yeah, sorry to keep you waiting. Wanted to test it together with another bug fix. All good. Thank you

JesperSchulz (dismissed):

JesperSchulz:

JesperSchulz (dismissed):

JesperSchulz (approved):

haoranpb (approved):

gggdttt (approved):

aholstrup1 (approved):

PR #10063, opened 2026-08-09
Fix the shadowed header cache in reminder line GetCurrencyCodeFromHeader

### LINKED ISSUE #8850: [Bug]: Issued/Reminder Line GetCurrencyCodeFromHeader()
### Describe the issue

The function GetCurrencyCodeFromHeader in table 298 "Issued Reminder Line" and table 296 "Reminder Line" gets the Currency Code from the header.

Image

But the statement in line 351 is never true, because the locale variable is always empty and it reads for every Amount field and line the header new.

[https://github.com/microsoft/BusinessCentralApps/issues/1875](url)

### Expected behavior

Only read once the Header.

### Steps to reproduce

Set a breakpoint at the line in the code and open a Issued Reminder or Reminder.

### Additional context

The local variable should be deleted.

Franco111000:
## What & why

`GetCurrencyCodeFromHeader()` on table 296 "Reminder Line" and table 298 "Issued Reminder Line" declares a local record that shadows the same-named global. The cache check on the first line therefore always compares against a freshly initialized local, never matches, and the header is re-read from the database on every single call (issue #8850).

The procedure is not called occasionally. It is the `AutoFormatExpression` of the amount fields on both tables and of the amount columns on the reminder reports: **230 usages across 26 files**, covering the W1, ES, FI, GB and NO copies of the two tables, the `Reminder` and `Reminder Test` reports in APAC, BE, DACH, ES, FI, FR, GB, NA, NO and W1, and report extension 10584 in the GB `ReportsGB` app. Every amount rendered on a reminder line or a reminder report triggers one `Get`, so a reminder with N lines costs roughly 4N header reads to display where 1 would do.

**The fix is not the same on the two tables, and that is deliberate.**

- **Table 298 "Issued Reminder Line"** (W1, FI, GB, NO): the shadowing local is simply deleted, so the existing global finally does its job. This is the fix proposed in the issue and it matches Microsoft's own pattern on posted tables that already have no shadow, for example `PurchRcptLine.GetCurrencyCodeFromHeader()` and `SalesShipmentLine.GetCurrencyCodeFromHeader()`. It is safe here because the `IssuedReminderHeader` global on this table carries no contract: it is only ever read straight after an unconditional `Get`, and every writer of Issued Reminder Header after issuance touches non-currency fields.

- **Table 296 "Reminder Line"** (W1, ES, FI, GB, NO): the same plain deletion would introduce a real defect, so this fix adds a dedicated global used only by this procedure instead. On the editable table the `ReminderHeader` global is a shared cache with an invariant: `GetReminderHeader()` treats a matching `"Reminder No."` as proof that `ProcessReminderHeader()` has already run, and `ProcessReminderHeader()` is the only writer of the `Currency` global and the only place four header `TestField` checks are made. Letting the display path populate that shared global would break the invariant: the lines part formats amounts before any validation runs, `GetReminderHeader()` would then hit the cache, `ProcessReminderHeader()` would be skipped, and the subsequent `Amount` validation would round with an `"Amount Rounding Precision"` of 0 while the four header checks silently never fire. A separate `CurrencyCodeReminderHeader` global gives the identical fetch-count win with no coupling to that contract, and a short comment in the variable block records why it exists.

Implementing this through `GetReminderHeader()` was also considered and rejected: it hard-`Get`s, so a missing header would raise an error during rendering where the procedure currently returns a blank string, and its `TestField` checks would fire while merely formatting a list.

Credit to @FSchneiderYav, who reported this and had already written the plain-deletion fix as PR #1894 in the BusinessCentralApps repository, closed by that repository's retirement rather than on its merits. I offered to carry it forward in the issue thread and am doing so here, with the table 296 shape changed for the reason above.

## Linked work

Fixes #8850

## How I validated this

- **Why no new test.** Under the dedicated-cache shape on table 296 the change is behavior-preserving by construction: the procedure's return value is the same in every reachable state, and the new global is touched by nothing else. The behavior actually worth protecting is an ordering property (format the lines part, then edit an amount, and observe that rounding precision is still initialized), and the repository has no existing coverage of that ordering to extend. I did not want to claim the existing reminder suites cover this, because they do not: `ERMFinancialReportsII` and `ERMInvoiceandReminder` drive fresh record variables and cannot reach the format-then-validate sequence. If you would like a regression test anyway, the shape I would write is a `TestPage` test on page 434 "Reminder" driving the `ReminderLines` host control, since page 435 is a ListPart and cannot be opened standalone. Happy to add it if you prefer that over the argument above.
- **How the behavior was verified.** I have no local Business Central environment for this repository, so I checked the reasoning by tracing the code rather than by running it. The chain the table 296 shape protects is: `GetReminderHeader()` treats a matching `"Reminder No."` as proof that `ProcessReminderHeader()` has already run; `ProcessReminderHeader()` is the only writer of the `Currency` record and the only place the four header `TestField` checks happen; and the `Amount`, `Remaining Amount` and line fee validations each round against `Currency."Amount Rounding Precision"` immediately after calling `GetReminderHeader()`. Since page 435 renders the `AutoFormatExpression` fields before any of those validations run, that is the ordering a shared-global fix would break and the dedicated global leaves untouched. If you would like this confirmed on a live environment before merging, I am glad to run whichever repro you prefer.
- **Layer-copy verification, since CI does not cover it.** This repository's build compiles `src/Layers/W1/BaseApp` only; the ES, FI, GB and NO copies have no app.json and no AL-Go project references them, and `VerifyMiappSync` checks that the files exist, not what they contain. So the eight country files in this PR get **no compiler coverage from the PR build** and I verified them by hand instead. Each file was edited independently at its own offsets, never by copying W1 over a country copy, because the copies carry exclusive content (NO field 10601 "Account Code", FI field 32000000 "Reference No.", the GB key guard, the ES "VAT+EC %" field). To prove no country content was disturbed, I captured `diff -u` of W1 against each country copy before and after the edit and asserted the divergence is unchanged:

  | Country copy | Divergent lines before | After |
  |---|---:|---:|
  | ES Reminder Line | 2 | 2 |
  | FI Reminder Line | 4 | 4 |
  | GB Reminder Line | 2 | 2 |
  | NO Reminder Line | 33 | 33 |
  | FI Issued Reminder Line | 4 | 4 |
  | GB Issued Reminder Line | 2 | 2 |
  | NO Issued Reminder Line | 14 | 14 |

  I also asserted that each of the nine files loses exactly one header-record declaration (the shadowing local) and that the five table 296 files gain exactly one `CurrencyCodeReminderHeader` declaration, in the global variable block.
- All nine files are required. `VerifyMiappSync` fails a W1-only changelist for these tables, so trimming this to W1 is not an option.

## Risk & compatibility

- I am deliberately not claiming "no observable behavior change". The precise statement is: the procedure returns the same value in every reachable state, and on table 296 the dedicated global keeps the display path from touching business state. What does change is how often the header is read, which is the point of the fix.
- **One accepted cosmetic limitation.** Any key-only cache, including Microsoft's existing `GetReminderHeader()`, keeps returning the cached header until the key changes. The reachable consequence: after changing a header's currency and running Suggest Reminder Lines, an already-open lines part can format amounts with the previous currency until the page instance is rebuilt. This is display formatting only, and every validation path stays protected, because `OnInsert` performs a raw `Get` and a currency change deletes the lines through Undo. I chose to document this rather than add invalidation, since fixing it would mean a broader change than the issue asks for. Say the word if you would rather have it addressed here.
- The fix assumes the platform reuses the record instance across `AutoFormatExpression` evaluations, which is the issue's own premise and what the reporter observed. If some path formats on throwaway instances, the change degrades to a harmless no-op there.
- **Out of scope, flagged for transparency.** Tables 303 "Finance Charge Memo Line" and 305 "Issued Fin. Charge Memo Line" carry the identical shadowed local in their `GetCurrencyCode()`, across six more files (W1 for both, plus CZ, ES and NO Finance Charge Memo Line and NO Issued Fin. Charge Memo Line). Table 303 has the same shared-global invariant as 296 and would need the dedicated-cache treatment; table 305 is a plain win like 298. I left them out because #8850 is scoped to the reminder tables. Happy to file a follow-up issue and fix them the same way if you want them covered.
- Area: Finance, matching the issue. I cannot set labels myself.

### DIFF
--- a/src/Layers/ES/BaseApp/Sales/Reminder/ReminderLine.Table.al
+++ b/src/Layers/ES/BaseApp/Sales/Reminder/ReminderLine.Table.al
@@ -667,6 +667,9 @@ table 296 "Reminder Line"
         ReminderTerms: Record "Reminder Terms";
         ReminderLevel: Record "Reminder Level";
         ReminderHeader: Record "Reminder Header";
+        // Used only by GetCurrencyCodeFromHeader(). The shared ReminderHeader global cannot be reused there:
+        // GetReminderHeader() takes a matching "No." as proof that ProcessReminderHeader() already ran.
+        CurrencyCodeReminderHeader: Record "Reminder Header";
         ReminderLine: Record "Reminder Line";
         ReminderLine2: Record "Reminder Line";
         ReminderEntry: Record "Reminder/Fin. Charge Entry";
@@ -893,14 +896,12 @@ table 296 "Reminder Line"
     /// </summary>
     /// <returns>The currency code from the header, or blank if not found.</returns>
     procedure GetCurrencyCodeFromHeader(): Code[10]
-    var
-        ReminderHeader: Record "Reminder Header";
     begin
-        if "Reminder No." = ReminderHeader."No." then
-            exit(ReminderHeader."Currency Code");
+        if "Reminder No." = CurrencyCodeReminderHeader."No." then
+            exit(CurrencyCodeReminderHeader."Currency Code");
 
-        if ReminderHeader.Get("Reminder No.") then
-            exit(ReminderHeader."Currency Code");
+        if CurrencyCodeReminderHeader.Get("Reminder No.") then
+            exit(CurrencyCodeReminderHeader."Currency Code");
 
         exit('');
     end;
--- a/src/Layers/FI/BaseApp/Sales/Reminder/IssuedReminderLine.Table.al
+++ b/src/Layers/FI/BaseApp/Sales/Reminder/IssuedReminderLine.Table.al
@@ -351,8 +351,6 @@ table 298 "Issued Reminder Line"
     /// </summary>
     /// <returns>The currency code from the header, or empty string if not found.</returns>
     procedure GetCurrencyCodeFromHeader(): Code[10]
-    var
-        IssuedReminderHeader: Record "Issued Reminder Header";
     begin
         if "Reminder No." = IssuedReminderHeader."No." then
             exit(IssuedReminderHeader."Currency Code");
--- a/src/Layers/FI/BaseApp/Sales/Reminder/ReminderLine.Table.al
+++ b/src/Layers/FI/BaseApp/Sales/Reminder/ReminderLine.Table.al
@@ -671,6 +671,9 @@ table 296 "Reminder Line"
         ReminderTerms: Record "Reminder Terms";
         ReminderLevel: Record "Reminder Level";
         ReminderHeader: Record "Reminder Header";
+        // Used only by GetCurrencyCodeFromHeader(). The shared ReminderHeader global cannot be reused there:
+        // GetReminderHeader() takes a matching "No." as proof that ProcessReminderHeader() already ran.
+        CurrencyCodeReminderHeader: Record "Reminder Header";
         ReminderLine: Record "Reminder Line";
         ReminderLine2: Record "Reminder Line";
         ReminderEntry: Record "Reminder/Fin. Charge Entry";
@@ -897,14 +900,12 @@ table 296 "Reminder Line"
     /// </summary>
     /// <returns>The currency code from the header, or blank if not found.</returns>
     procedure GetCurrencyCodeFromHeader(): Code[10]
-    var
-        ReminderHeader: Record "Reminder Header";
     begin
-        if "Reminder No." = ReminderHeader."No." then
-            exit(ReminderHeader."Currency Code");
+        if "Reminder No." = CurrencyCodeReminderHeader."No." then
+            exit(CurrencyCodeReminderHeader."Currency Code");
 
-        if ReminderHeader.Get("Reminder No.") then
-            exit(ReminderHeader."Currency Code");
+        if CurrencyCodeReminderHeader.Get("Reminder No.") then
+            exit(CurrencyCodeReminderHeader."Currency Code");
 
         exit('');
     end;
--- a/src/Layers/GB/BaseApp/Sales/Reminder/IssuedReminderLine.Table.al
+++ b/src/Layers/GB/BaseApp/Sales/Reminder/IssuedReminderLine.Table.al
@@ -349,8 +349,6 @@ table 298 "Issued Reminder Line"
     /// </summary>
     /// <returns>The currency code from the header, or empty string if not found.</returns>
     procedure GetCurrencyCodeFromHeader(): Code[10]
-    var
-        IssuedReminderHeader: Record "Issued Reminder Header";
     begin
         if "Reminder No." = IssuedReminderHeader."No." then
             exit(IssuedReminderHeader."Currency Code");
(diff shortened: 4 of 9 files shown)


alexei-dobriansky:

# **Recommendation: Accept**

<details>

`GetCurrencyCodeFromHeader()` on table 296 "Reminder Line" and table 298 "Issued Reminder Line" declares a local record variable that shadows the same-named global. Because the local is freshly initialized on every call, the key-comparison cache check on the first line always fails, and a database `Get` is issued on every call. Since `GetCurrencyCodeFromHeader()` is the `AutoFormatExpression` of every amount field on both tables and on the reminder reports, each rendered reminder line triggers multiple redundant header reads. This PR fixes that.

The fix correctly uses two different approaches for the two tables. For table 298 (Issued Reminder Line), removing the local variable is safe because the `IssuedReminderHeader` global has no contract with other procedures — `LookupCustomerLedgerEntry` always calls `.Get()` unconditionally, so it does not rely on any prior state. This matches the pattern already used in `PurchRcptLine.GetCurrencyCodeFromHeader()`. For table 296 (Reminder Line), a plain deletion of the local would let the display path populate the shared `ReminderHeader` global, which `GetReminderHeader()` treats as proof that `ProcessReminderHeader()` has already run. Bypassing `ProcessReminderHeader()` would skip four `TestField` checks and leave `Currency."Amount Rounding Precision"` uninitialised when the amount validation triggers run. The dedicated `CurrencyCodeReminderHeader` global gives the same performance gain with no coupling to that invariant. The reasoning is correct and the code matches the stated intent.

All nine country-layer files (W1, ES, FI, GB, NO for table 296; W1, FI, GB, NO for table 298) are edited consistently. The author verified the country-layer divergence is unchanged, which is appropriate since CI does not compile the country copies.

</details>

<details>

*None.*

</details>

<details>

**Risk:** The change is narrow and does not alter the return value of `GetCurrencyCodeFromHeader()` in any reachable state. The only observable difference is how often the header record is read. The shared-global invariant on table 296 is preserved by the dedicated variable. The one accepted limitation — an already-open lines part can display amounts in the previous currency until the page instance is rebuilt after a currency change — is display-only and does not affect validation or posting paths. There are no public API or event signature changes.

**Necessity:** The fix is clearly needed. With 230 usages across 26 files, every rendered reminder page and report pays a database read per amount column per line. The root cause (local variable shadowing the global) is concrete, the repro is easy to understand, and the fix is the minimal correct change. Finance Charge Memo tables 303 and 305 carry the same bug and are out of scope by the author's own note; that is a reasonable boundary for issue #8850 and worth a follow-up issue.

</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=1 system=github pr=10063 round=1 by=alexei-dobriansky at=2026-08-10T08:18:15Z lastSha=84ed9e1d643a770edfcaa31078db18d3e69dc064 reviewKey=403cbb326c3f4600877d47eda5b84de9e2576c239891364ca28ff17bd8c39890 suggestions=

AndersLarsenMicrosoft (approved):

PR #10067, opened 2026-08-09
[Quality Management] Enable Bin code selection for transfer disposition in workflow response

JakovljevicDusan:
## What & why

Synchronize behavior with manual disposition form quality inspection.

## Linked work

Fixes [AB#624735](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/624735)

### DIFF
--- a/src/Apps/W1/Quality Management/app/src/Workflow/QltyWorkflowRespOptions.PageExt.al
+++ b/src/Apps/W1/Quality Management/app/src/Workflow/QltyWorkflowRespOptions.PageExt.al
@@ -386,14 +386,12 @@ pageextension 20403 "Qlty. Workflow Resp. Options" extends "Workflow Response Op
                             QltyWorkflowResponse: Codeunit "Qlty. Workflow Response";
                         begin
                             QltyWorkflowResponse.SetStepConfigurationValue(Rec, QltyWorkflowResponse.GetWellKnownKeyLocation(), QltyLocationCode);
-                            if not QltyShouldShowGrpTransfer then begin
-                                QltyShowBinCode := true;
-                                if DestinationLocation.Get(QltyLocationCode) then begin
-                                    QltyShowBinCode := DestinationLocation."Bin Mandatory";
-                                    if QltyBinCode <> '' then
-                                        if not DestinationBin.Get(QltyLocationCode, QltyBinCode) then
-                                            QltyBinCode := '';
-                                end;
+                            QltyShowBinCode := true;
+                            if DestinationLocation.Get(QltyLocationCode) then begin
+                                QltyShowBinCode := DestinationLocation."Bin Mandatory";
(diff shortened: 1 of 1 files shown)


alexei-dobriansky:

<details>

This PR fixes the Quality Management workflow response options page so that bin code selection works when the workflow response action is "Create Transfer Order". Previously, the bin visibility and validation logic (`QltyShowBinCode` flag, bin clearing on location change) was wrapped in `if not QltyShouldShowGrpTransfer then begin ... end`, which skipped the entire block for the transfer disposition. The PR removes this guard in two places: the `OnValidate` trigger of the location field and the `SetLocationAndBinCode` procedure.

The fix is correct. When `QltyShouldShowGrpTransfer = true`, the page is showing the Transfer destination group and the bin code field should be enabled based on whether the destination location has bin mandatory. The old guard was preventing this, making bin selection unavailable for transfer workflow responses even when the destination location requires bins.

The `OnValidate` change also preserves the bin-clearing logic (if the previously selected bin does not exist in the new location, clear it), which is correct behavior.

</details>

<details>

**S1** - *Pre-existing semicolon after `then` in `SetLocationAndBinCode` makes the guard ineffective*

In `SetLocationAndBinCode`, the line `if Location.Get(QltyLocationCode) then;` has a semicolon immediately after `then`, so the next line `QltyShowBinCode := Location."Bin Mandatory";` runs unconditionally even when `Location.Get` returns false. When the location does not exist, `Location."Bin Mandatory"` is `false`, so `QltyShowBinCode` ends up `false` (overriding the `true` set two lines above). This bug is pre-existing and not introduced by this PR, but the PR touches this block — consider fixing the semicolon so the guard actually works: `if Location.Get(QltyLocationCode) then QltyShowBinCode := Location."Bin Mandatory";`.

</details>

<details>

**Risk:** Low. The change is confined to a single page extension in the Quality Management app (W1). It affects only UI state (`QltyShowBinCode` visibility flag and bin code clearing on location change). No financial posting, no table writes, no event signatures touched.

**Necessity:** High. Without the fix, the bin code field is always invisible and ignored when setting up a transfer workflow response, making it impossible to configure a bin destination even for bin-mandatory locations. The manual disposition form already supports this (per the PR description), and the workflow response form should be consistent.

</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=1 system=github pr=10067 round=1 by=alexei-dobriansky at=2026-08-10T07:51:15Z lastSha=ca6dcddcb653b32d47a92802d35c4233b52ce591 reviewKey=e1f40790ba7c1b2dbd59d1739f025d63e209dc54c409a2fe304df77634d4c512 suggestions=S1@97951d10

alexei-dobriansky (approved):

attilatoury (approved):

PR #10072, opened 2026-08-10
[main] - Serial and Lot number mismatch when item is produced via Production Order and moved to inventory through an Internal Put-away

MSNehaNawal:
[Bug 642316](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/642316): [ALL-E] Serial and Lot number mismatch when item is produced via Production Order and moved to inventory through an Internal Put-away

Fixes [AB#642316](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/642316)

Issue - Serial and Lot number mismatch when item is produced via Production Order and moved to inventory through an Internal Put-away

Cause - The put-away validation was skipped as time of registration.As a result, the lot requirement was never enforced at registration time and a blank lot slipped through.

Solution - Added a data-driven check in WhseActivityRegister.CheckLines → CheckItemTrackingRequiredForPutAway that compares the put-away line against the actual posted Item Ledger Entry instead of the configuration flags. For each Put-away line carrying a Serial No., it resolves the open Item Ledger Entry by Item / Variant / Location / Serial No.; if that entry has a Lot No., the line's Lot No. must match it:

### DIFF
--- a/src/Layers/W1/BaseApp/Warehouse/Activity/WhseActivityRegister.Codeunit.al
+++ b/src/Layers/W1/BaseApp/Warehouse/Activity/WhseActivityRegister.Codeunit.al
@@ -1816,6 +1816,8 @@ codeunit 7307 "Whse.-Activity-Register"
                 if Location."Bin Mandatory" then
                     CheckBinRelatedFields(GlobalWhseActivLine);
 
+                CheckItemTrackingRequiredForPutAway(GlobalWhseActivLine);
+
                 OnAfterCheckWhseActivLine(GlobalWhseActivLine);
 
                 if ((GlobalWhseActivLine."Activity Type" = GlobalWhseActivLine."Activity Type"::Pick) or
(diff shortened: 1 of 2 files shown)


alexei-dobriansky:

<details>

This PR fixes a serial/lot number mismatch that occurred when registering an Internal Put-away created from a Production Order output. The root cause was that the lot requirement was never enforced at registration time, so a Put-away line with a blank Lot No. could slip through even when the underlying Item Ledger Entry carried a Lot No.

The fix adds CheckItemTrackingRequiredForPutAway inside the CheckLines loop. For each Put-away line that carries a Serial No., it resolves the open Item Ledger Entry by Item / Variant / Location / Serial No. and calls TestField("Lot No.", ...) if that entry has a Lot No. The logic is placed before OnAfterCheckWhseActivLine, so it cannot be bypassed through that event. An OnBeforeCheckItemTrackingRequiredForPutAway event with the standard IsHandled pattern is provided for partner extensibility. The approach is correct: it checks actual posted data rather than configuration flags, which is exactly what the description says was missing.

</details>

<details>

**S1** - *Test covers Purchase Order, not the reported Production Order scenario*
The bug was triggered by a Production Order output moved through an Internal Put-away, but the new test uses a Purchase Order and a regular warehouse receipt Put-away. The fix logic is generic and should work for both paths, but the test does not directly verify the reported scenario. Consider adding a second test (or extending the existing one) that creates a production order output, generates an Internal Put-away, clears the Lot No. on a line, and confirms that registration is blocked.

</details>

<details>

**Risk:** The new check runs inside the CheckLines loop for every Put-away line that carries a Serial No. For locations without put-away activity or items without serial tracking the code exits immediately after the first two guards, so performance impact is negligible. The TestField call will block registration for any Put-away line where the Lot No. on the line does not match the open ILE's Lot No.; this is intentional, but it is a new hard stop that could surface in edge cases not covered by the test (for example, a Put-away created before a corrective adjustment changes the ILE's Lot No.). The OnBeforeCheckItemTrackingRequiredForPutAway event gives partners a clean escape hatch. No public API or table schema changes are made.

**Necessity:** The mismatch described in the bug leads to permanently incorrect Warehouse Bin Content and Item Ledger Entries, which is a data integrity issue that requires manual correction. The fix is narrow in scope, targets exactly the Put-away registration path, and provides both a guard and an extensibility event. The change is necessary and appropriately scoped.

</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=1 system=github pr=10072 round=1 by=alexei-dobriansky at=2026-08-11T09:36:32Z lastSha=5077847aaf5097d231a69e0a3259e66c5263e77f reviewKey=322ef141f6e11ecdd74d38b97a0c07b111e5585f6c03a445469d6eff1c0ba961 suggestions=S1@ab1e01f8

alexei-dobriansky (approved):

neeleshsinghal (approved):

PredragMaricic (approved):
The registration-time lot validation prevents the item-ledger and warehouse-bin mismatch, and the regression test covers the blocked blank-lot put-away.

PR #10083, opened 2026-08-10
[main]- Negative Shipped and Invoiced Quantities appear on Sales Order Lines after using the Copy Document from a Sales Credit Memo.

MSNehaNawal:
[Bug 646375](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/646375): [ALL-E] Negative Shipped and Invoiced Quantities appear on Sales Order Lines after using the Copy Document from a Sales Credit Memo.

Fixes [AB#646375](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/646375)

Issue-
When a corrective Sales Credit Memo is created via Copy Document from a Posted Sales Invoice, and that invoice contains the same item on multiple lines (e.g. shipped from several sales orders and invoiced together via Get Shipment Lines), posting the credit memo produces negative Quantity Shipped / Quantity Invoiced on one sales order line while another order line is left unchanged.

Root cause-
Copy Document does not populate Appl.-from Item Entry on the credit memo lines, so SalesCrMemoLine.GetSalesInvoiceLine falls back to matching the source invoice line only by Item No. + Quantity >= X and takes FindFirst(). With duplicate items, every credit memo line resolves to the same first invoice line, so the corresponding order line is decremented repeatedly (going negative) and the other order line is never corrected.

Solution-
GetSalesInvoiceLine:
Added an overload that accepts a temporary "used invoice lines" buffer. The new FindUnusedSalesInvoiceLine helper skips invoice lines already consumed by earlier credit memo lines. Returns a blank line when no unused candidate remains.
UpdateSalesOrderLineIfExist now keeps a temporary buffer of resolved invoice lines, passes it into the resolver, and records each successfully consumed invoice line so it cannot be reused for a subsequent credit memo line.

### DIFF
--- a/src/Layers/APAC/BaseApp/Sales/History/CorrectPostedSalesInvoice.Codeunit.al
+++ b/src/Layers/APAC/BaseApp/Sales/History/CorrectPostedSalesInvoice.Codeunit.al
@@ -1028,6 +1028,7 @@ codeunit 1303 "Correct Posted Sales Invoice"
     var
         SalesCrMemoLine: Record "Sales Cr.Memo Line";
         SalesInvoiceLine: Record "Sales Invoice Line";
+        TempUsedSalesInvoiceLine: Record "Sales Invoice Line" temporary;
     begin
         SalesCrMemoLine.SetLoadFields("Document No.", "No.", "Appl.-from Item Entry", Quantity, "Variant Code");
         SalesCrMemoLine.SetRange("Document No.", SalesCreditMemoNo);
@@ -1036,9 +1037,12 @@ codeunit 1303 "Correct Posted Sales Invoice"
         if SalesCrMemoLine.FindSet() then
             repeat
                 Clear(SalesInvoiceLine);
-                SalesCrMemoLine.GetSalesInvoiceLine(SalesInvoiceLine);
-                if SalesInvoiceLine."Line No." <> 0 then
+                SalesCrMemoLine.GetSalesInvoiceLine(SalesInvoiceLine, TempUsedSalesInvoiceLine);
+                if SalesInvoiceLine."Line No." <> 0 then begin
                     UpdateSalesOrderLinesFromCreditMemo(SalesInvoiceLine, SalesCrMemoLine);
+                    TempUsedSalesInvoiceLine := SalesInvoiceLine;
+                    if TempUsedSalesInvoiceLine.Insert() then;
+                end;
             until SalesCrMemoLine.Next() = 0;
     end;
 
--- a/src/Layers/APAC/BaseApp/Sales/History/SalesCrMemoLine.Table.al
+++ b/src/Layers/APAC/BaseApp/Sales/History/SalesCrMemoLine.Table.al
@@ -1248,6 +1248,13 @@ table 115 "Sales Cr.Memo Line"
 
     internal procedure GetSalesInvoiceLine(var SalesInvoiceLine: Record "Sales Invoice Line")
     var
+        TempUsedSalesInvoiceLine: Record "Sales Invoice Line" temporary;
+    begin
+        GetSalesInvoiceLine(SalesInvoiceLine, TempUsedSalesInvoiceLine);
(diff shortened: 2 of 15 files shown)


alexei-dobriansky:

# **Recommendation: Accept**

<details>

When a corrective Sales Credit Memo is created via Copy Document from a Posted Sales Invoice, `Appl.-from Item Entry` is not populated on the credit memo lines. This causes `GetSalesInvoiceLine` to fall back to a `FindFirst()` match by Item No. and Quantity. If the same item appears on multiple invoice lines (e.g., shipped from several orders and invoiced together via Get Shipment Lines), every credit memo line resolves to the **same first invoice line**, so that order line receives all the quantity decrements (going negative) while the other order line is never updated.

The fix introduces a temporary buffer (`TempUsedSalesInvoiceLine`) in `UpdateSalesOrderLineIfExist` to track which invoice lines have already been consumed by earlier credit memo lines. A new `FindUnusedSalesInvoiceLine` helper skips buffer entries when scanning for a match, ensuring each invoice line is used at most once. The new overload of `GetSalesInvoiceLine` accepts this buffer, and the original single-argument overload delegates to it with an empty buffer, preserving backward compatibility.

The fallback path also gains a `Variant Code` filter and an exact-quantity pass (before falling back to `>= quantity` and then any quantity), both of which make the matching more precise. The `Clear(SalesInvoiceLine)` at the end of the fallback path is a defensive improvement that ensures callers always see an empty record when no candidate is found, rather than relying on the side effect of a failed `FindFirst()`.

The logic is correct. The code anchor in `UpdateSalesOrderLineIfExist` inserts into the buffer only after `UpdateSalesOrderLinesFromCreditMemo` succeeds, so a line that fails to produce a credit correction (e.g., no matching order) is not recorded as "used" — another credit memo line can still try it. The `if TempUsedSalesInvoiceLine.Insert() then;` pattern correctly silences duplicate-key errors for cases where the item-ledger-entry path coincidentally resolves two credit memo lines to the same invoice line (the buffer is consulted only in the fallback path, so this does not block the item-ledger-entry path).

The change covers all country-layer variants of `SalesCrMemoLine.Table.al` and `CorrectPostedSalesInvoice.Codeunit.al` consistently.
</details>

<details>

No suggestions.
</details>

<details>

**Risk:** Low. The fix touches the fallback path inside `GetSalesInvoiceLine` — the path that runs only when `Appl.-from Item Entry = 0` (i.e., Copy Document scenario, not the normal correction flow). The normal flow (via Item Ledger Entry) is unchanged. The added `Variant Code` filter is a narrow tightening; it could theoretically change behavior for historic data where a credit memo line's variant differs from the matched invoice line, but such data would reflect incorrect prior matching. The temp-buffer approach is bounded to a single call of `UpdateSalesOrderLineIfExist`, so there are no cross-posting or state-leak risks.

**Necessity:** High. Negative Quantity Shipped / Quantity Invoiced on an open Sales Order are data integrity issues that can cascade into incorrect availability, inventory valuation, and warehouse processing. The scenario (duplicates merged via Get Shipment Lines, credit via Copy Document) is a documented customer flow. The fix is appropriately scoped to the exact root cause.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=1 system=github pr=10083 round=1 by=alexei-dobriansky at=2026-08-11T09:44:08Z lastSha=704ffa8079cf0b7e12307ddcd563f330b11b0322 reviewKey=c745fb92d4a2847db421c19e796e7596dfc082cd3698cd89b3e3f9a78bea26b0 suggestions=

alexei-dobriansky (approved):

neeleshsinghal (approved):

PredragMaricic (approved):
The source invoice-line matching now prevents line reuse across copied credit memo lines, and the regression test covers the negative shipped and invoiced quantity scenario.

PR #10087, opened 2026-08-10
Make AA0189 an error in the base ruleset

Groenbech96:
Removes the `AA0189` override from `src/rulesets/base.ruleset.json`.

The ruleset declares `"generalAction": "Error"`, so removing the explicit `"action": "Warning"` entry promotes AA0189 to an error for the base application.


JesperSchulz:
Nice catch. I made that catch too 😉 As well as 18 others in #10088. Closing this one as a dupe.

PR #10119, opened 2026-08-11
Fixes #8851

### LINKED ISSUE #8851: [Bug]: table 7326 "Whse. Worksheet Line" - TemplateSelection with different PageID
### Describe the issue

If the function TemplateSelection in table 7326 "Whse. Worksheet Line" is called with a page ID different from the standard Page ID. The Whse. Worksheet Template is with the standard page ID created.

https://github.com/microsoft/BusinessCentralApps/issues/1359

### Expected behavior

The Whse. Worksheet Template should be created with the Page ID from the input.

### Steps to reproduce

Database without a Whse. Worksheet Template.
Call Function TemplateSelection with PageID <>7351 and PageTemplate = Movement.

### Additional context

```
    procedure TemplateSelection(PageID: Integer; PageTemplate: Option "Put-away",Pick,Movement; var WhseWkshLine: Record "Whse. Worksheet Line"; var WhseWkshSelected: Boolean)
    var
        WhseWkshTemplate: Record "Whse. Worksheet Template";
    begin
        WhseWkshSelected := true;

        WhseWkshTemplate.Reset();
        WhseWkshTemplate.SetRange("Page ID", PageID);
        WhseWkshTemplate.SetRange(Type, PageTemplate);

        case WhseWkshTemplate.Count of
            0:
                begin
                    WhseWkshTemplate.Init();
                    WhseWkshTemplate.Validate(Type, PageTemplate);
-                    **WhseWkshTemplate.Validate("Page ID");**
+                  **WhseWkshTemplate.Validate("Page ID", PageID);**
                    WhseWkshTemplate.Name :=
                      Format(WhseWkshTemplate.Type, MaxStrLen(WhseWkshTemplate.Name));
                    WhseWkshTemplate.Description := StrSubstNo(Text008, WhseWkshTemplate.Type);
                    WhseWkshTemplate.Insert();
                    Commit();
                end;
            1:
                WhseWkshTemplate.FindFirst();
            else
                WhseWkshSelected := Page.RunModal(0, WhseWkshTemplate) = Action::LookupOK;
        end;
        if WhseWkshSelected then begin
            WhseWkshLine.FilterGroup := 2;
            WhseWkshLine.SetRange("Worksheet Template Name", WhseWkshTemplate.Name);
            WhseWkshLine.FilterGroup := 0;
            if OpenFromBatch then begin
                WhseWkshLine."Worksheet Template Name" := '';
                Page.Run(WhseWkshTemplate."Page ID", WhseWkshLine);
            end;
        end;
    end;
```

FSchneiderYav:
## What & why

If the function TemplateSelection in table 7326 "Whse. Worksheet Line" is called with a page ID different from the standard Page ID. The Whse. Worksheet Template is with the standard page ID created.

## Linked work

Fixes #8851

### DIFF
--- a/src/Layers/W1/BaseApp/Warehouse/Worksheet/WhseWorksheetLine.Table.al
+++ b/src/Layers/W1/BaseApp/Warehouse/Worksheet/WhseWorksheetLine.Table.al
@@ -1099,7 +1099,7 @@ table 7326 "Whse. Worksheet Line"
                 begin
                     WhseWkshTemplate.Init();
                     WhseWkshTemplate.Validate(Type, PageTemplate);
-                    WhseWkshTemplate.Validate("Page ID");
+                    WhseWkshTemplate.Validate("Page ID", PageID);
                     WhseWkshTemplate.Name :=
                       Format(WhseWkshTemplate.Type, MaxStrLen(WhseWkshTemplate.Name));
                     WhseWkshTemplate.Description := StrSubstNo(Text008, WhseWkshTemplate.Type);


alexei-dobriansky:

<details>

This PR fixes a bug in `TemplateSelection` (table 7326 "Whse. Worksheet Line") where a new `Whse. Worksheet Template` was always created with the standard page ID instead of the `PageID` parameter passed to the function.

The root cause is clear: `Validate("Page ID")` with no argument re-validates the field at its post-`Init()` value of `0`. The `"Page ID"` field's `OnValidate` trigger checks `if "Page ID" = 0 then Validate(Type)`, which then overwrites the field with the hardcoded standard page for that template type (e.g. page 7351 "Movement Worksheet"). Changing the call to `Validate("Page ID", PageID)` passes a non-zero value, so the trigger exits without overwriting it. The fix is correct and minimal.

</details>

<details>

**S1** - *Same bug exists in two sibling warehouse tables*
`BinCreationWorksheetLine.Table.al` (line ~404) and `WarehouseJournalLine.Table.al` (line ~1174) both call `Validate("Page ID")` with no argument in their own `TemplateSelection` zero-count branches. The same trigger logic applies, so they will also ignore a custom `PageID` and create the template with the standard page. Consider applying the same one-line fix to those two tables.

**S2** - *No automated test for the custom-PageID scenario*
The fixed path (count = 0, non-standard PageID) is not covered by a test. A test that calls `TemplateSelection` with a page ID different from the default and then checks that the created template's `"Page ID"` matches the input would prevent this from regressing. The area has existing warehouse tests, so this is feasible.

</details>

<details>

**Risk:** The change is a one-liner in a single zero-count branch. The only callers that reach this path are those providing a non-standard PageID when no template exists yet; all three existing standard callers (Put-away, Pick, Movement worksheets) already have a template, so they hit the `count = 1` branch and are not affected. Regression surface is narrow.

**Necessity:** The bug is confirmed by the linked issue (#8851) and a cross-linked external report. Any extension or integration that opens a custom warehouse worksheet page with `TemplateSelection` will silently get a template pointing to the wrong page, breaking navigation. The fix is correct, scoped, and needed.

</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=1 system=github pr=10119 round=1 by=alexei-dobriansky at=2026-08-11T09:41:51Z lastSha=abfd4b725e9d1f607b043364f68f94cb5ce11778 reviewKey=9130f758200331974118dc82d58942a1aa3dfc12ad5eb91dd1716f9c5ef35a2f suggestions=S1@a3549ff5,S2@8e8e07b2

pri-kise (changes requested):
The requested tests are missing

alexei-dobriansky:
@pri-kise please check S1

alexei-dobriansky:

<details>

The linked issue reports that TemplateSelection in table 7326 creates a Whse. Worksheet Template with the standard page ID when the caller passes a custom page ID. The diff changes the zero-template path to validate "Page ID" with the input PageID, so the following Insert() should store the page that was used in the lookup filter. This addresses the root cause for the table 7326 scenario; the remaining risk is that the persistent creation path has no regression coverage.
</details>

<details>

**S1** - *Add a custom page ID regression test*
Add a test for TemplateSelection with no existing Whse. Worksheet Template, PageTemplate = Movement, and a non-standard page ID. Assert that the created template stores that page ID, then call TemplateSelection again and verify it does not create a duplicate or fail. This path writes setup data, so the regression should be covered before merge.

**S2** - *Clarify scope for similar template code*
The same creation pattern still exists in related TemplateSelection routines, such as warehouse journal and bin creation worksheet templates. If custom page ID support is meant only for table 7326, say that in the PR or linked issue. If it is a general contract, include the matching fixes in this PR or a linked follow-up.
</details>

<details>

**Risk:** The regression surface is narrow: first-time creation of a warehouse worksheet template for callers that pass a custom page ID. There is no public signature or event change, and no performance concern. If this is wrong, the code can persist the wrong setup data and later calls can either open the wrong page or try to create the same template name again.

**Necessity:** The linked issue gives a clear and valid scenario, and the one-line code change is the right shape for table 7326 because the filter already uses the input PageID. The change is needed for custom page callers, but the missing test leaves the data-writing path unguarded. Related TemplateSelection routines still use the old pattern, so the intended scope should be clear.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=10119 round=1 by=alexei-dobriansky at=2026-08-12T07:59:23Z lastSha=abfd4b725e9d1f607b043364f68f94cb5ce11778 reviewKey=c0b27a8312f3405d3e995a414dd4fb7b68ee33443a13df80e0e0e7d47142a890 suggestions=S1@8e8e07b2,S2@a3549ff5

alexei-dobriansky (changes requested):
.

alexei-dobriansky (changes requested):
.

PredragMaricic (changes requested):
S1 - High confidence: add the required regression coverage.

The one-line change correctly preserves the caller-supplied `PageID` when creating a new `Whse. Worksheet Template`, but this is a persistent setup-data behavior change with no automated test. The linked issue's approved scope explicitly requires coverage that:

1. Creates a Movement template through `TemplateSelection` with a custom page ID and asserts the stored `"Page ID"` is the custom value.
2. Calls `TemplateSelection` again and verifies that the existing template is reused without a duplicate-key error.
3. Confirms the standard page-ID path remains unchanged.

Without these tests, the zero-template creation branch and its interaction with the template `OnInsert` trigger remain unguarded. The code fix itself is minimal and appears correct for the reported scenario.

PR #10163, opened 2026-08-12
All emails in folder were processed at first activation

mynjj:
In folder mode we were processing all old emails at activation, this fix sets a date filter (that doesn't move) by using the date at which the agent was enabled. Additionally a defensive validation to ensure no 0D goes in non-folder mode is also added (shouldn't happen, but still)

Fixes [AB#638782](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/638782)

### DIFF
--- a/src/Apps/W1/EDocumentConnectors/Microsoft365/app/src/OutlookProcessing.Codeunit.al
+++ b/src/Apps/W1/EDocumentConnectors/Microsoft365/app/src/OutlookProcessing.Codeunit.al
@@ -95,12 +95,15 @@ codeunit 6385 "Outlook Processing"
         TempFilters."Load Attachments" := true;
         TempFilters."Load Headers" := true;
         TempFilters."Max No. of Emails" := GetMaxNoOfEmails();
-        // In folder mode, dedup is enforced server-side by the category-exclude filter only.
-        // Applying "Earliest Email" would drop emails moved into the folder after that floor.
-        if not FolderConfigured then
+        if not FolderConfigured then begin
+            if OutlookSetup."Last Sync At" = 0DT then // Defensive, this should not happen due to the OnValidate then enabling.
+                OutlookSetup."Last Sync At" := CurrentDateTime();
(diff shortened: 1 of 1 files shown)


alexei-dobriansky:

<details>

This PR changes Outlook folder retrieval to use `Enabled At` as a fixed lower bound, so first activation does not process every old email in the selected folder. It also adds a fallback for non-folder mode when `Last Sync At` is `0DT`.

The folder-mode part matches the bug evidence: `Earliest Email` becomes a stable filter, while `Last Sync At` still does not advance for configured folders. The new non-folder fallback is unsafe because it changes the record only in memory and then uses that value to build the Graph filter.
</details>

<details>

**S1** - *Persist the fallback sync time*
The `0DT` fallback only changes the local `OutlookSetup` variable. If no email is returned, the value is not saved, so the setup can keep filtering from the current time and stop importing mail. Save the fallback value with `Modify()`, or move this recovery to a path that persists it before retrieval.

**S2** - *Add a regression test for folder filtering*
There is no test that verifies folder mode sends `Enabled At` as `Earliest Email`. Add coverage for a configured folder with old and new messages, so this bug does not come back.
</details>

<details>

**Risk:** This code controls which mailbox messages become e-documents. A wrong lower bound can either create many unwanted e-documents from old mail, or skip new imports after a `0DT` sync state. The change does not alter public APIs or event publishers.

**Necessity:** The bug is valid: on first activation, folder mode can read all unprocessed messages in the folder because the category filter is the only bound. A fixed activation-time floor is a reasonable scope, but the fallback path must be safe before merge.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=10163 round=1 by=alexei-dobriansky at=2026-08-12T07:52:59Z lastSha=b629735263209b1d924c449a06595e3121f38881 reviewKey=2becaac61350b70f9bdee1031ca3e1b99c00389cefec5ca1e72d3943dd4c6526 suggestions=S1@e3166f3b,S2@b20e7f58

darjoo (approved):

dcenic (approved):

PR #10209, opened 2026-08-12
Promote AA0471, AA0472 and AA0474 to Error in the base ruleset

JesperSchulz:
Continues the ruleset hardening for [AB#640773](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/640773) by promoting three analyzer rules from `None` back to `Error` in `src/rulesets/base.ruleset.json`.

Stacked on #10133.

## What changed

| Rule | Distinct violations | How it is handled |
| --- | --- | --- |
| `AA0472` | 0 | Override removed |
| `AA0471` | 1 | Fixed in code, override removed |
| `AA0474` | 5 | Fixed in code, override removed |

All six violations are in the Withholding Tax app, and all are genuine omissions rather than exceptions worth accepting, so they are fixed rather than suppressed.

`AutoFormatType = 1` means the value is an amount, and the analyzer requires a matching `AutoFormatExpression` naming the currency. In `Withholding Tax Entry` and `WHT Empl. Ledger Entry Ext` the neighbouring amount fields already specify `AutoFormatExpression = Rec."Currency Code"`, and the fields flagged here had simply been missed, so they now do the same. `WHT Threshold Accumulator` has no currency field at all because it only ever accumulates local currency amounts, so it uses an empty expression, which is the standard way to express LCY.

Fixing the table field also resolved the `AA0471` report on `Withholding Tax Entries`, since the page control inherits its formatting from the underlying field.

## Corrected scope

The first revision of this PR also promoted `AA0473` and `AS0112`, on the strength of local measurements that found 4 and 3 violations respectively. CI disagreed, and CI was right: **886** `AA0473` and **630** `AS0112` diagnostics across the 44 `Build Apps` jobs.

The local scan had simply not reached far enough. It had covered 398 of 878 projects when it was stopped, and the dependency-ordered pass before it could not build large parts of `src/Apps` at all, so entire areas — `Master Data Mgt.`, `BC14 Historical Data`, remittance advice and others — were never examined. A rule looked clean because nothing had looked at the code that violates it.

Both rules therefore keep their overrides and need real remediation in their own PRs.

## How the surviving numbers were established

From the failed CI run itself, which compiles everything. Tallying `##[error]` diagnostics across all 44 `Build Apps` jobs gives `AA0471` 36 and `AA0474` 180 occurrences, but those are the same handful of W1 sites recompiled once per country: 1 and 5 distinct sites. `AA0472` does not appear in any job.

The six sites were then reproduced locally against the updated ruleset to get exact line numbers, and the Withholding Tax app now compiles clean (exit 0) with CodeCop, AppSourceCop, PerTenantExtensionCop and UICop enabled.

Related to [AB#640773](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/640773)

### DIFF
--- a/src/Apps/W1/WithholdingTax/app/src/Employee/History/WHTThresholdAccumulator.Table.al
+++ b/src/Apps/W1/WithholdingTax/app/src/Employee/History/WHTThresholdAccumulator.Table.al
@@ -57,12 +57,16 @@ table 6794 "WHT Threshold Accumulator"
         {
             Caption = 'Accumulated Base Amount';
             AutoFormatType = 1;
+            // The accumulator only ever holds local currency amounts.
+            AutoFormatExpression = '';
             ToolTip = 'Specifies the accumulated base amount for the withholding tax threshold accumulator.';
         }
         field(9; "Accumulated WHT Amount"; Decimal)
         {
             Caption = 'Accumulated Withholding Tax Amount';
             AutoFormatType = 1;
+            // The accumulator only ever holds local currency amounts.
+            AutoFormatExpression = '';
             ToolTip = 'Specifies the accumulated withholding tax amount for the withholding tax threshold accumulator.';
         }
     }
--- a/src/Apps/DK/OIOUBL/app/src/Management/OIOUBLTaxGroupBuffer.Table.al
+++ b/src/Apps/DK/OIOUBL/app/src/Management/OIOUBLTaxGroupBuffer.Table.al
@@ -15,6 +15,9 @@ table 13673 "OIOUBL-Tax Group Buffer"
         {
             Caption = 'Tax Category ID';
         }
+        // Temporary buffer used only to aggregate amounts for the OIOUBL XML export.
+        // The fields are never shown in the UI, so AutoFormatType is not meaningful here.
+#pragma warning disable AA0473
         field(2; "OIOUBL-VAT %"; Decimal)
         {
             Caption = 'VAT %';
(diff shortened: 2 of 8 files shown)


nikolakukrika (dismissed):

nikolakukrika (approved):

PredragMaricic (approved):

src/Apps/W1/WithholdingTax/app/src/Employee/History/WHTThresholdAccumulator.Table.al
```diff
@@ -57,12 +57,16 @@ table 6794 "WHT Threshold Accumulator"
         {
             Caption = 'Accumulated Base Amount';
             AutoFormatType = 1;
+            // The accumulator only ever holds local currency amounts.
```
BardurKnudsen, line 60: Do we want this kind of comments in properties?

AndersLarsenMicrosoft (approved):

BardurKnudsen (approved):

src/Apps/W1/WithholdingTax/app/src/Employee/History/WHTThresholdAccumulator.Table.al
JesperSchulz: Nah, you're right. We probably don't want those. But now that the build has passed and I have my approvals, I'm going to check this on 😂 I'll keep an eye on this going forward.

PR #9420, opened 2026-07-14
Manage next-action ownership for issues and pull requests

JesperSchulz:
## What & why

Manage deterministic next-action ownership for every open issue and pull request with exactly one of `Turn: Microsoft`, `Turn: Partner`, or `Turn: Blocked`. The trusted default-branch workflow reacts to issue/PR lifecycle, human conversation, review, and inline-review activity; preserves explicit blocked state; and never touches unrelated labels.

Team actors must be human and have effective write, maintain, or admin repository permission. Team approvals return the next action to Microsoft only when no team reviewer's latest opinionated review still requests changes. Comment-only and pending reviews cannot mask an earlier change request; approval or dismissal clears that reviewer's request.

Fork review events use a zero-write producer and privileged `workflow_run` consumer because direct review-event tokens are read-only. The producer uploads one raw artifact containing only event kind, repository, PR number, and review/comment ID. The consumer resolves the trusted originating PR before job-level concurrency, bounds and validates the raw artifact, binds it to the exact first-attempt source event/actor/time, and re-fetches canonical API data before label mutation.

The workflows never create or update label definitions. `docs/features/next-action-owner/DESIGN.md` contains the one-time provisioning commands, behavior, security design, and operational backfill using `ManageNextActionOwner.yaml`.

## Linked work

[AB#642155](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/642155)

## How I validated this

- [x] I read the full diff and it contains only changes I intended.
- [ ] I built the affected app(s) locally with no new analyzer warnings. *(N/A: GitHub workflows only.)*
- [ ] I ran the change in Business Central and confirmed it behaves as expected. *(N/A: GitHub workflows only.)*
- [x] I added or updated tests for the new behavior, or explained below why none are needed.

**What I tested and the outcome**

- Parsed and structurally checked both workflow YAML files, exact display names, permissions, triggers, raw artifact settings, and 40-character action pins.
- Verified the `workflow_run` trigger and all runtime allowlists use `Capture PR review for next action owner` and `.github/workflows/CaptureNextActionReview.yaml` exactly.
- Compiled all three inline `github-script` bodies as async JavaScript functions.
- Syntax-checked the documented provisioning/backfill commands with Git Bash and verified dispatch targets `ManageNextActionOwner.yaml`.
[...]

PR #10228, opened 2026-08-13
Route pull request reviews to the owning application team

JesperSchulz:
## What & why

The default `CODEOWNERS` entry points every path at `@microsoft/dynamics-365-business-central`, which has **123 members**. Every pull request therefore requests review from 123 people — a request addressed to everyone is owned by nobody.

The measurable consequence today: **149 of 233 ready pull requests have no review decision at all**. GitHub itself cannot say who owes the review, so no filter, query or report can either.

## Where the mapping comes from

The team rules are **generated** from the routing data the BCApps triage agent already uses to classify issues and pull requests:

- `microsoft/BCAppsTriage` → `plugins/triage/skills/triage/scripts/ownership/ownership-rules.js`
- `microsoft/BCAppsTriage` → `plugins/triage/skills/triage/scripts/ownership/ownership-resolver.js`

which are in turn seeded from the Business Central Ownership Matrix.

That agent already decides which team owns an issue or pull request. Deriving `CODEOWNERS` from the same data means **the team label an item gets and the reviewer GitHub requests cannot disagree** — and there is no second copy of the mapping to drift.

### Structure

The generated block is ordered **broad to narrow**, because `CODEOWNERS` resolves last-match-wins and the resolver it mirrors is itself layered:

| Tier | Contains |
|---|---|
| 1 | Tree defaults — W1 apps and layers to Integration, every country tree to Finance |
| 2 | Base Application areas for W1 and every localization layer, plus the Finance sub-ledgers that sit inside SCM area folders (`Sales/Reminder`, `Purchases/Payables`) |
| 3 | Other W1 layer trees, including per-area test ownership |
| 4 | W1 apps that are not Integration, plus the per-area `PowerBIReports` folders |
| 5 | Demo datasets, which follow the functional `DemoData/<Area>` folder rather than the app |
| 6 | Documented path overrides (for example the Italian Service Declaration) |
| 7 | Reconciliation pending — see below |

Worth calling out: **a team appears in more than one tier.** Grouping every rule for a team together reads better, but it cannot reproduce the resolver's precedence, so correctness wins.

## Linked work

[AB#642155](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/642155) — the "team ownership" half of that slice; #9420 is the "whose turn" half.

## How I validated this

- [ ] I built the affected app(s) locally with no new analyzer warnings. *(N/A: `CODEOWNERS` and Markdown only.)*
- [ ] I ran the change in Business Central and confirmed it behaves as expected. *(N/A: no product code.)*
- [x] I added or updated tests for the new behavior, or explained below why none are needed. *(No test hook exists for `CODEOWNERS`; validated by exhaustive simulation instead.)*

I implemented a `CODEOWNERS` resolver (last-match-wins, gitignore-style patterns) and ran the triage agent's own `teamFromPath()` (plus its path-based overrides) over **every one of the 46,437 tracked files**, then compared.

| Check | Result |
|---|---|
| Generated block reproduces the triage resolver's path decision | **44,796 of 44,796 files (100%)** |
| Pre-existing **non-default** ownership changed vs `main` | **0 files** |
| Files moved off the 123-member default owner | **37,694** |

Where the final file differs from the resolver, it is always a specialized rule lower in the file deliberately taking precedence: `app.json` to the app team (877 files), the Copilot and Developer Tools carve-outs inside the System Application (170), `dotnet.al` to App Security (33), `*.Entitlement.al` to Integration (24), plus `*.ps1`, `/build` and `/.github` to Engineering Systems. That is the intended "Integration owns the System Application, with sub-sections owned by others" shape.

**Cross-checked against the team labels the triage agent actually applied**, over 425 labelled pull requests:

| Outcome | PRs | Share |
|---|---|---|
| Single-team PR, CODEOWNERS matches the label | 312 | 73.4% |
| Multi-team PR, labelled team is the dominant owner | 68 | 16.0% |
| Multi-team PR, labelled team owns a minority of files | 13 | 3.1% |
| CODEOWNERS never requests the labelled team | 25 | 5.9% |
| No team owner at all | 7 | 1.6% |

**CODEOWNERS requests the labelled team on 92.5% of pull requests.** Exact one-to-one agreement is neither expected nor desirable: the agent picks a single team per pull request, whereas `CODEOWNERS` requests every owner of every touched file, so a pull request spanning two teams correctly gets two reviewers. The agent also reads file content (namespace, object id) which takes precedence over path, and `CODEOWNERS` can only ever see the path.

Re-verified after rebasing onto current `main`.

## Risk & compatibility

**This makes the three teams required approvers for their paths.** `microsoft-production-ruleset` sets `require_code_owner_review: true`, so these entries gate merges — this is the change that needs your sign-off, not the file itself.

Review load from currently open, non-draft pull requests that have a team owner (187 of them, of which **53 touch more than one team** and will therefore need approval from each):

| Team | Members | Open PRs | Per member |
|---|---|---|---|
| Finance | 8 | 91 | 11.4 |
| SCM | 9 | 73 | 8.1 |
| Integration | 10 | 90 | 9.0 |

Integration is the biggest change here: it previously owned almost nothing in `CODEOWNERS` and now owns the System Application, Business Foundation, the migration and connector apps, and every unlisted W1 app.

**Two disagreements are deliberately left unresolved.** The triage rules and today's `CODEOWNERS` genuinely disagree in two places. Taking ownership away from a team is not something this change should do silently, so current ownership is preserved in tier 7 and the disagreement is recorded:

- **`PowerBIReports`** — the triage agent resolves its per-area report folders to the area owner (`App/Sales` → SCM); `CODEOWNERS` has owned the whole app as Finance since #9752. 112 files.
- **`ServiceManagement`** — `appFolderRules` says SCM; `CODEOWNERS` says Finance. 2 files.

**Three findings for whoever maintains `ownership-rules.js`** (none of them blocking this pull request):

1. **Intrastat** is forced to SCM by an override keyed on the *namespace* `Microsoft.Inventory.Intrastat` and object id 4810, explicitly so "PRs and issues agree" — but `appFolderRules.Intrastat` says `Finance`. A path-only consumer such as `CODEOWNERS` cannot see the namespace, so it routes Intrastat to Finance. A path rule would close the gap.
2. **`BaseApp/Projects`** — `baseAppAreaRules.Projects` says `Finance`, while `subAreaToTeam['Projects']` and the ownership matrix both say SCM. The agent resolves SCM via object id (Job, table 167). The two data sections contradict each other.
3. **Country E-Document labels look stale.** Seven of the 25 disagreements are country E-Document apps labelled `Integration`. `ownership-rules.js` contains an override routing those to Finance whose comment notes it "deliberately supersedes the earlier #9172 decision". `CODEOWNERS` matches the current rule; those labels predate it and would benefit from a re-run.

**Do not hand-edit the generated block.** Edits are lost on the next regeneration and silently diverge from issue and pull request triage. Change `ownership-rules.js` in `BCAppsTriage` instead. The generator reads those modules directly, so it belongs alongside them in `BCAppsTriage` rather than in this repo.

**Follow-up requiring a repository administrator (not in this PR):** the `Official Branches` ruleset still has a `file_patterns: ["*"]` rule requiring one approval from the same 123-member team. Until that is replaced with the same team patterns, the *merge gate* stays anonymous even though the *request* is now routed correctly. It should follow this PR, so routing is proven before the gate is tightened.

Reversible: revert the commit and resolution returns to exactly the current behaviour.

**Reviewers:** please sanity-check the tiers for your own area, confirm the two tier-7 disagreements, and confirm `d365-bc-finance-1` is the correct Finance group (a `d365-bc-finance` team also exists, with 6 members).

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -5,6 +5,469 @@
 # Default owners
 * @microsoft/dynamics-365-business-central
 
+# ---------------------------------------------------------------------------
+# Application team ownership (GENERATED - do not hand-edit)
+#
+# These rules are generated from the routing data the BCApps triage agent
+# already uses to classify issues and pull requests:
+#   microsoft/BCAppsTriage
+#     plugins/triage/skills/triage/scripts/ownership/ownership-rules.js
+#     plugins/triage/skills/triage/scripts/ownership/ownership-resolver.js
+# which are themselves seeded from the Business Central Ownership Matrix.
+#
+# They reproduce the path-based part of that resolver (its path overrides plus
+# teamFromPath) exactly: every one of the 44,796 tracked files the resolver can
+# place resolves to the same team here. Keeping one source means a label applied
+# by the triage agent and the reviewer requested by GitHub cannot disagree.
+#
+# To change ownership, edit ownership-rules.js in BCAppsTriage and regenerate.
+# Do not hand-edit the block below; hand edits will be lost and will silently
+# diverge from issue and pull request triage.
+#
+# The rules are ordered broad-to-narrow because CODEOWNERS resolves
+# last-match-wins, so a team can appear in more than one tier. The specialized
+# cross-cutting rules further down (App Security, App Permissions, app.json,
+# AL-Go and build files, Code Review Tools) still take precedence over all of
+# them.
+#
+# See docs/process/ReviewRouting.md.
+# ---------------------------------------------------------------------------
+
+# TIER 1 - tree defaults.
+/src/Apps/W1/ @microsoft/d365-bc-integrations
+/src/Layers/W1/ @microsoft/d365-bc-integrations
+/src/Apps/APAC/ @microsoft/d365-bc-finance-1
+/src/Layers/APAC/ @microsoft/d365-bc-finance-1
+/src/Apps/AT/ @microsoft/d365-bc-finance-1
+/src/Layers/AT/ @microsoft/d365-bc-finance-1
+/src/Apps/AU/ @microsoft/d365-bc-finance-1
+/src/Layers/AU/ @microsoft/d365-bc-finance-1
+/src/Apps/BE/ @microsoft/d365-bc-finance-1
+/src/Layers/BE/ @microsoft/d365-bc-finance-1
+/src/Apps/CA/ @microsoft/d365-bc-finance-1
+/src/Layers/CA/ @microsoft/d365-bc-finance-1
+/src/Apps/CH/ @microsoft/d365-bc-finance-1
+/src/Layers/CH/ @microsoft/d365-bc-finance-1
+/src/Apps/CZ/ @microsoft/d365-bc-finance-1
+/src/Layers/CZ/ @microsoft/d365-bc-finance-1
+/src/Apps/DACH/ @microsoft/d365-bc-finance-1
+/src/Layers/DACH/ @microsoft/d365-bc-finance-1
+/src/Apps/DE/ @microsoft/d365-bc-finance-1
+/src/Layers/DE/ @microsoft/d365-bc-finance-1
+/src/Apps/DK/ @microsoft/d365-bc-finance-1
+/src/Layers/DK/ @microsoft/d365-bc-finance-1
+/src/Apps/ES/ @microsoft/d365-bc-finance-1
+/src/Layers/ES/ @microsoft/d365-bc-finance-1
+/src/Apps/FI/ @microsoft/d365-bc-finance-1
+/src/Layers/FI/ @microsoft/d365-bc-finance-1
+/src/Apps/FR/ @microsoft/d365-bc-finance-1
+/src/Layers/FR/ @microsoft/d365-bc-finance-1
+/src/Apps/GB/ @microsoft/d365-bc-finance-1
+/src/Layers/GB/ @microsoft/d365-bc-finance-1
+/src/Apps/IN/ @microsoft/d365-bc-finance-1
+/src/Layers/IN/ @microsoft/d365-bc-finance-1
+/src/Apps/IS/ @microsoft/d365-bc-finance-1
+/src/Layers/IS/ @microsoft/d365-bc-finance-1
+/src/Apps/IT/ @microsoft/d365-bc-finance-1
+/src/Layers/IT/ @microsoft/d365-bc-finance-1
+/src/Apps/MX/ @microsoft/d365-bc-finance-1
+/src/Layers/MX/ @microsoft/d365-bc-finance-1
+/src/Apps/NA/ @microsoft/d365-bc-finance-1
+/src/Layers/NA/ @microsoft/d365-bc-finance-1
+/src/Apps/NL/ @microsoft/d365-bc-finance-1
+/src/Layers/NL/ @microsoft/d365-bc-finance-1
+/src/Apps/NO/ @microsoft/d365-bc-finance-1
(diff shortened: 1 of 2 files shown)


docs/process/ReviewRouting.md
```diff
@@ -0,0 +1,195 @@
...
+
+Because code owner review is required, these entries make each team a required approver for its
+paths. From currently open, non-draft pull requests that have a team owner — 187 of them, of which
+**53 touch more than one team** and therefore need approval from each:
+
+| Team | Members | Open PRs | Per member |
+|---|---|---|---|
+| Finance | 8 | 91 | 11.4 |
+| SCM | 9 | 73 | 8.1 |
+| Integration | 10 | 90 | 9.0 |
```
Aleyenda, line 164: 6 FTEs and 1 student worker. Not sure where the 10 comes from :)

Aleyenda (approved):

docs/process/ReviewRouting.md
```diff
@@ -0,0 +1,195 @@
...
+
+### What this costs the teams
+
+Because code owner review is required, these entries make each team a required approver for its
+paths. From currently open, non-draft pull requests that have a team owner — 187 of them, of which
+**53 touch more than one team** and therefore need approval from each:
+
+| Team | Members | Open PRs | Per member |
+|---|---|---|---|
+| Finance | 8 | 91 | 11.4 |
```
AndersLarsenMicrosoft: the numbers are 6 fte plus me

docs/process/ReviewRouting.md
JesperSchulz: It's looking at the GitHub team. Probably it hasn't been updated yet. Also, it includes Klaus and me.

CODEOWNERS
```diff
@@ -5,6 +5,469 @@
...
+/src/Layers/*/BaseApp/HumanResources/ @microsoft/d365-bc-scm
+/src/Layers/*/BaseApp/Integration/ @microsoft/d365-bc-integrations
+/src/Layers/*/BaseApp/Inventory/ @microsoft/d365-bc-scm
+/src/Layers/*/BaseApp/Invoicing/ @microsoft/d365-bc-scm
+/src/Layers/*/BaseApp/Manufacturing/ @microsoft/d365-bc-scm
+/src/Layers/*/BaseApp/Modules/ @microsoft/d365-bc-integrations
+/src/Layers/*/BaseApp/OtherCapabilities/ @microsoft/d365-bc-integrations
+/src/Layers/*/BaseApp/Permissions/ @microsoft/d365-bc-integrations
+/src/Layers/*/BaseApp/Pricing/ @microsoft/d365-bc-scm
+/src/Layers/*/BaseApp/Projects/ @microsoft/d365-bc-finance-1
```
AndersLarsenMicrosoft: SCM (project was jobs earlier on))

CODEOWNERS
```diff
@@ -5,6 +5,469 @@
...
+/src/Layers/*/BaseApp/Purchases/ @microsoft/d365-bc-scm
+/src/Layers/*/BaseApp/RoleCenters/ @microsoft/d365-bc-integrations
+/src/Layers/*/BaseApp/Sales/ @microsoft/d365-bc-scm
```
AndersLarsenMicrosoft: The folder have some short cummings (Prepayment and IC) are finance features that are a part of sales

aholstrup1 (approved):

PR #10236, opened 2026-08-13
Add Recalculate to expense vat specification

Alexander-Ya:
## Linked work

Fixes #

## How I validated this


Alexander-Ya:
Not needed

PR #10253, opened 2026-08-14
Avoid starting a write transaction under `ShowCanCreateAgent`

mynjj:
See bug 646707 for the platform side, explaining why we are calling ShowCanCreateAgent after the `Commit` when we have 3rd party code like:

```
procedure ChangeUserName()  
var  
    User: Record User;  
    ShowUserQst: Label 'Show User?';  
begin  
    User.FindFirst();  
    User."Full Name" := 'New Full Name';  
    User.Modify();  
    Commit();  
    if Confirm(ShowUserQst, false) then  
        Page.RunModal(Page::"User Card", User);  
end;  
```

In the meantime, we ensure that no write transactions are started when computing `ShowCanCreateAgent`.

Fixes [AB#641849](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/641849)


AleksanderGladkov (approved):

alexei-dobriansky:

<details>

This fixes a bug (AB#641849) where computing whether a new Payables Agent can be created starts a write transaction. The platform can call `ShowCanCreateAgent` right after a `Commit` in third-party code, and starting a new write transaction there is not safe. The fix adds an `UpdateSetup` overload to `GetAgent` and an `InsertIfMissing` overload to `Table.GetSetup`, and routes the read-only path (`AllowCreateNewAgent` → `GetAgent(Agent, false)` → `GetSetup(false)`) so it neither inserts the setup record nor calls `Modify`.

The change addresses the root cause, not just the symptom: both write points inside `GetAgent` (the `GetSetup` insert and the trailing `Modify`) are now guarded by the flag. The parameterless `GetAgent(Agent)` and `GetSetup()` overloads still insert/modify, so every existing write-path caller keeps its old behavior. The read path still returns the correct answer because agent existence is resolved by security id and then by user name, and only persistence is skipped. `GetSetup(false)` correctly `Clear`s `Rec` and returns `false` when no record exists, so the caller reads a blank (null security id) record safely.
</details>

<details>

**S1** - *Sibling caller still writes during page open*
`CanShowAgentActions` (line 269) still calls `GetAgent(Agent)` with the write-enabled default, and it runs from `OnOpenPage` of the Purchase Invoice List and Agent Task List. When no agent exists yet, that path still inserts the setup record and calls `Modify` during page render. Please confirm this is intended; if the goal is to avoid write transactions on read/render surfaces in general, this caller could also use `GetAgent(Agent, false)`.

**S2** - *Missing trailing newline at end of file*
`PayablesAgentSetup.Table.al` ends without a trailing newline (the diff shows `\ No newline at end of file`). Add a final newline to match the repository convention and keep future diffs clean.
</details>

<details>

**Risk:** Low. The new behavior is opt-in through the added boolean overloads; the parameterless `GetAgent`/`GetSetup` keep the previous insert-and-modify behavior, so callers other than `AllowCreateNewAgent` (e.g. `CanShowAgentActions`, the OCV codeunit, upgrade) are unchanged. The one behavior change is scoped to `AllowCreateNewAgent`, which now only reads. There is no public API break: the old signatures remain, and the new overloads are additive.

**Necessity:** Justified. Starting a write transaction while the platform is computing agent-creation visibility after a partner `Commit` is a real defect with a clear repro in the linked work item. The scope is tight and matches the bug. No automated test is included; this area (Payables Agent setup) has no existing test project and the change does not touch a posting, financial, or data-integrity path, so the missing test is a non-blocking gap rather than a merge blocker.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=10253 round=1 by=alexei-dobriansky at=2026-08-17T01:11:23Z lastSha=842656f6c3227e1716bec974efca17dca8660a18 reviewKey=b5340025357521caff1be299df70a6e7ca094c094e557d127a35365000da7cfe suggestions=S1@fd50407a,S2@dfefd4da

AndersLarsenMicrosoft (approved):

PR #10265, opened 2026-08-15
[Validation] Test header-only API authentication

t-prda:
## Validation scope

Tests explicit Basic Authorization without setting `HttpWebRequest.Credentials`.

Only the 57 APIV2 methods that returned 401 in the clean-tenant run are enabled. This draft exists to determine whether challenge credentials overwrite the explicit header; the durable result will be integrated into #10085.

[AB#646383](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/646383)


t-prda:
Validation complete: header-only Basic authentication did not change the result; all 57 targeted methods still returned 401. The experiment is not being integrated.

PR #10295, opened 2026-08-17
#10288 - Always load attachments

### LINKED ISSUE #10288: [Bug]: Inline-only email attachments are skipped because `hasAttachments` is used as a fetch gate
### Describe the issue

The Outlook/Graph email integration uses the message's `hasAttachments` property to decide whether to fetch attachments at all. However, `hasAttachments` only reflects regular attachments — it excludes inline attachments (e.g. images pasted into the email body). As a result, messages containing only inline attachments report `hasAttachments: false`, causing the attachment-fetch step to be skipped even though the attachments are genuinely present and retrievable via `$expand=attachments`. This means inline-only attachments are silently lost.

### Expected behavior

Attachments should be retrieved for any message that actually has them, regardless of whether they're inline or regular. Specifically, I expected the attachment-fetch step to run unconditionally (not gated on `hasAttachments`), so that a message containing only an inline image would still have that attachment fetched via `$expand=attachments` — with any distinction between inline and regular attachments handled afterward, per-attachment, using `isInline` and `contentId`/`cid:` correlation, rather than being used upfront to decide whether to fetch at all.

### Steps to reproduce

1. In Business Central, using a feature that processes inbound email via the Outlook/Graph integration (e.g. an AI agent such as the Sales Order Agent or Payables Agent), have an external sender send an email where the only attachment is an inline image — for example, a screenshot pasted directly into the body of the email, with no separate file attached.
2. Let Business Central pick up and process the incoming email through the normal inbound flow.
3. Observe that the email is processed as if it has no attachments — the inline image is not retrieved, stored, or surfaced anywhere in BC.
4. Enable diagnostic/trace logging (or debug) around the attachment-fetching logic in the Outlook API helper codeunit, and confirm that the `hasAttachments` value returned for this message is `false`, causing the codeunit to skip the attachment-fetch call.
5. As a control, send a second test email to the same BC inbound flow, this time with a regular (non-inline) file attachment alongside or instead of the inline image. Confirm that in this case `hasAttachments` is `true` and the attachment is fetched and processed correctly.
6. Comparing steps 3–5 confirms the defect is specific to messages containing *only* inline attachments: the codeunit's reliance on `hasAttachments` as a pre-fetch gate causes those attachments to be silently dropped, even though they are present and retrievable.

### Additional context

_No response_

g-Bergsma:
## What & why

Fixes a defect in the Outlook/Graph email attachment retrieval logic where the hasAttachments property on a Graph message is used as a gate before fetching attachments. hasAttachments only reflects regular (non-inline) attachments, so messages containing only inline attachments (e.g. a screenshot pasted into the email body) report hasAttachments: false and are skipped entirely — even though the attachment is present and retrievable via $expand=attachments. This change removes the hasAttachments gate and instead always fetches attachments, applying inline/regular distinction afterward via isInline and contentId/cid: correlation.

## Linked work

Fixes #10288

## How I validated this

- New unit tests in OutlookAPIHelperTests.Codeunit.al pass locally; full module test suite green.

## Risk & compatibility

- Behavior change, not breaking change: Messages that previously had attachments silently skipped (inline-only case) will now have those attachments fetched and processed. Any downstream logic that assumed "no attachments" for these messages should be reviewed, but this is a bug fix restoring intended behavior, not an API surface change.
- Performance: Removing the hasAttachments gate means the attachment-fetch call ($expand=attachments or equivalent) now runs unconditionally, including for messages with genuinely zero attachments. This adds a Graph API call that was previously skipped in that case — likely negligible, but worth noting for high-volume inbound email scenarios.
- No data upgrade impact: This is a runtime logic change only; no schema, table, or field changes involved. 
- No permission changes: Does not touch permission sets or object access.
- No breaking changes to public interfaces: Internal fix to the Outlook API helper codeunit's fetch logic; no changes to published/public methods, events, or extensibility points expected (confirm this holds once you see the actual diff).
- Telemetry: No new telemetry added.
- Follow-up work: None anticipated beyond this fix.

### DIFF
--- a/src/Apps/W1/Email - Outlook REST API/app/src/EmailOutlookAPIHelper.Codeunit.al
+++ b/src/Apps/W1/Email - Outlook REST API/app/src/EmailOutlookAPIHelper.Codeunit.al
@@ -435,8 +435,7 @@ codeunit 4509 "Email - Outlook API Helper"
             Body := KeepLastMessageOnly(Body);
         EmailMessage.Create('', Subject, Body, HTMLBody, not Filters.GetBypassBodySanitization());
 
-        if HasAttachments then
-            AddAttachmentsToMessage(EmailJsonObject, EmailMessage);
+        AddAttachmentsToMessage(EmailJsonObject, EmailMessage);
 
         if Filters."Load Headers" then
             SetMessageHeaders(EmailJsonObject, EmailMessage);
--- a/src/Apps/W1/Email - Outlook REST API/test/HttpResponseFiles/RetrieveEmailWithInlineAttachmentsOnly.txt
+++ b/src/Apps/W1/Email - Outlook REST API/test/HttpResponseFiles/RetrieveEmailWithInlineAttachmentsOnly.txt
@@ -1,0 +1,34 @@
+{
+  "@odata.count": 1,
+  "value": [
+    {
+      "id": "test-message-id-006",
+      "conversationId": "test-conversation-id-006",
+      "subject": "Test Inline Only Attachments",
+      "receivedDateTime": "2026-04-21T10:00:00Z",
+      "sentDateTime": "2026-04-21T09:55:00Z",
+      "sender": {
+        "emailAddress": {
+          "name": "Test Sender",
+          "address": "sender@test.com"
+        }
+      },
+      "body": {
+        "contentType": "HTML",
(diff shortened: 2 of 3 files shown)


g-Bergsma:
@microsoft-github-policy-service agree company="Boltrics Professionals BV"

PR #10299, opened 2026-08-17
[Master] - Slice 638119: [Expense Agent] Mileage date range rate setup

ViditGupta1277:


src/Apps/W1/ExpenseAgent/app/src/Common/Enums/ExpenseVehicleType.Enum.al
```diff
@@ -0,0 +1,17 @@
+// ------------------------------------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+namespace Microsoft.ExpenseAgent;
+
+enum 6989 "Expense Vehicle Type"
```
BardurKnudsen: Is it really best to hard-code vehicle types? Shouldn't data like this be in a separate table named "Expense Vehicle Type"? What if the customer wants to add 'Electric Car' or 'Bicycle' or 'Tuk-tuk'? Or 'Boat' if it's a country with a lot of rivers or a large lake?

BardurKnudsen (commented):
I think we should reconsider how we define vehicle types

src/Apps/W1/ExpenseAgent/app/src/Common/Enums/ExpenseVehicleType.Enum.al
ViditGupta1277: A separate table has been created. Please check

BardurKnudsen (approved):

src/Apps/W1/ExpenseAgent/app/src/MasterData/Tables/MileageRateSetup.Table.al
```diff
@@ -0,0 +1,167 @@
...
+            ToolTip = 'Specifies the first date the mileage rate is effective.';
+
+            trigger OnValidate()
+            begin
+                TestField("Starting Date");
+                CheckDates();
+                CheckOverlappingRate();
+            end;
+        }
+        field(5; "Ending Date"; Date)
```
BardurKnudsen, line 61: Fine, but often we don't have end-dates on this kind of rates. Unless we expect it do be discontinued at some point. Whenever we add a new rate with a new start-date, we need to maintain the previous rate's ending date. But let's keep it if want the option to discontinue it.

BardurKnudsen (approved):

alexei-dobriansky:
## Good Sense Review

Recommendation: Accept w/ Suggestions

### Suggestions

🟠 S1 - Implement the documented generic vehicle-rate fallback.

`src/Apps/W1/ExpenseAgent/app/src/MasterData/Tables/MileageRateSetup.Table.al` documents a blank Vehicle Type as the generic rate used when no vehicle-specific rate matches, but `FindEffectiveRate` filters `Vehicle Type` to the requested value only. When an expense has a vehicle type and only a blank/generic mileage rate exists, the lookup misses that row and `GetStandardRateOfMileage` falls back to the setup standard rate instead. The new test at `ExpenseMileageTest.Codeunit.al` also exposes the mismatch: the comments say the generic rate should be used, but the assertion expects `GenericRate + 1` (the standard rate). Please add a fallback lookup for blank Vehicle Type before falling back to the standard setup rate, and update the test expectation to lock in the intended generic-rate behavior.

I did not cast a formal review vote.

> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10299 round=1 by=alexei-dobriansky at=2026-08-26T22:24:49Z lastSha=07f10ddd8ee5848f627c79a578d4218570bb0a1f reviewKey=fcfd67a7175abd082e615a0ab46417acdbea978e7b01bd06adc4827232b87ab8 suggestions=S1@945ffcf2af42

PR #10300, opened 2026-08-17
Consolidate minor release ruleset and fix strict mode branch detection

aholstrup1:
## What & why

Two related build-configuration changes for the strict mode / minor-release breaking-changes check:

1. **Consolidate the minor-release ruleset.** `build/projects.json` had the `ruleSetPathMinorRelease` property spread across several different files with inconsistent paths and casing, and `ruleSetPath` likewise pointed at multiple different files. This unifies them: every `ruleSetPathMinorRelease` now points to a new shared `src/rulesets/baseminorrelease.ruleset.json`, and every `ruleSetPath` points to `src/rulesets/base.ruleset.json`. The new ruleset includes `base.ruleset.json` plus the minor-release breaking-change rules (AS0077, AS0078, AS0102). `PreCompileApp.ps1` is updated to resolve the new file name for the strict mode check.

2. **Fix strict mode detection on release-branch CI/CD.** On a push build against a release branch, `Test-IsStrictModeEnabled` resolved the branch via `Get-CurrentBranch`, which used `git rev-parse --abbrev-ref HEAD`. In GitHub Actions the checkout is a detached HEAD, so that returns the literal `HEAD`, which never matches `^releases/\d+\.\d+$`. The result was strict mode being silently skipped on release branches (only PR builds worked, via `GITHUB_BASE_REF`). `Get-CurrentBranch` now returns `GITHUB_REF_NAME` when it is set, falling back to git only for local runs.

Centralizing the fix in `Get-CurrentBranch` also corrects the same latent detached-HEAD issue in the baseline-storage `main` branch check.

## Linked work

Fixes #

[AB#647333](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/647333)

## How I validated this

- Verified all 493 `ruleSetPathMinorRelease` and all 550 `ruleSetPath` entries in `projects.json` now resolve to a single path each, and that `projects.json` remains valid JSON.
- Parse-checked the edited PowerShell modules.
- Simulated branch resolution: push on `releases/28.4` now resolves to `releases/28.4` and matches the strict mode regex (previously `HEAD`, no match); PR into a release branch still works via `GITHUB_BASE_REF`; `main` and `releases/NN.x` correctly stay disabled (no numeric StrictMode tag).
- No AL tests added; changes are build-script/config only.

## Risk & compatibility

- Behavior change: strict mode breaking-changes validation will now actually run on release-branch push builds where it was previously being skipped. This is the intended fix, but it may surface breaking-change analyzer errors that were not being reported before.
- The old `src/rulesets/minorrelease.ruleset.json` is now unreferenced. It was left in place in this PR; it can be removed in a follow-up if desired.

### DIFF
--- a/build/projects.json
+++ b/build/projects.json
@@ -23,7 +23,7 @@
       "dvdFolder": "Applications\\TestFramework\\TestLibraries\\Any",
       "runStaticCodeAnalysis": true,
       "ruleSetPath": "$env:INETROOT\\App\\BCApps\\src\\rulesets\\base.ruleset.json",
-      "ruleSetPathMinorRelease": "$env:INETROOT\\App\\BCApps\\src\\rulesets\\minorrelease.ruleset.json",
+      "ruleSetPathMinorRelease": "$env:INETROOT\\App\\BCApps\\src\\rulesets\\baseminorrelease.ruleset.json",
       "repositoryUrl": "https://github.com/microsoft/BCApps"
     },
     "Library Assert": {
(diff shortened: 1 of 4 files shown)


mazhelez (approved):

JesperSchulz (approved):

JesperSchulz (approved):
Looks fine. Approved.

nikolakukrika (approved):

PR #10305, opened 2026-08-17
Truncate Shopify PO Number to External Document No. length

### LINKED ISSUE #10243: [Bug]: Shopify Connector: order sync fails when Shopify PO Number exceeds 35 characters (missing truncation in ShpfyProcessOrder)
### Describe the issue

`Codeunit "Shpfy Process Order".CreateHeaderFromShopifyOrder` assigns the Shopify PO number to the sales header without truncation:

```al
SalesHeader.Validate("External Document No.", ShopifyOrderHeader."PO Number");
```

https://github.com/microsoft/BCApps/blob/main/src/Apps/W1/Shopify/App/src/Order%20handling/Codeunits/ShpfyProcessOrder.Codeunit.al (currently line ~133)

`Sales Header."External Document No."` is `Code[35]`, while the PO number field on `Shpfy Order Header` holds longer values coming from Shopify (B2B checkout PO numbers routinely exceed 35 characters — customers paste internal reference strings). When that happens, order creation fails at runtime with:

> The length of the string is 45, but it must be less than or equal to 35 characters.

The Shopify order then sits in error state and requires manual PO editing + reprocessing. Every retry fails identically, so for affected merchants this is a recurring daily-operations burden.

### Expected behavior

The PO number is truncated to the target field length (like other string fields mapped in the same procedure), so order creation succeeds. The full PO number remains available on the order in Shopify. Proposed one-line fix:

```al
SalesHeader.Validate("External Document No.", CopyStr(ShopifyOrderHeader."PO Number", 1, MaxStrLen(SalesHeader."External Document No.")));
```

### Steps to reproduce

1. BC SaaS with the Shopify Connector, connected shop with B2B / draft-order flow.
2. Create a Shopify order whose PO number is longer than 35 characters (e.g. `TEST-PO-1234567890-ABCDEFGHIJKLMNOPQRSTUVWXYZ`).
3. Sync the order and run **Create Sales Document** (or let the order sync job process it).
4. Order processing fails with the string-length error; no sales document is created.

### Additional context

Observed in production on BC 27 (SaaS) with Shopify Connector 27.x and 28.3; the assignment is unchanged on current `main`. All surrounding header fields are length-safe — this one assignment is the outlier. We currently work around it with a per-tenant extension subscribing to `Shpfy Order Events.OnBeforeProcessSalesDocument` and truncating there.

I will provide a fix for this bug (one-line change above) once the issue is approved.

seanschade:
Truncate Shopify PO Number to External Document No. length in ShpfyProcessOrder

The PO number coming from Shopify can exceed 35 characters, while Sales Header "External Document No." is Code[35]. The direct Validate call made order creation fail with a string-length error for any B2B order with a long PO number. Truncate with CopyStr/MaxStrLen like the other string mappings in this procedure.

Fixes #10243

## Linked work

Fixes #

## How I validated this

### DIFF
--- a/src/Apps/W1/Shopify/App/src/Order handling/Codeunits/ShpfyProcessOrder.Codeunit.al
+++ b/src/Apps/W1/Shopify/App/src/Order handling/Codeunits/ShpfyProcessOrder.Codeunit.al
@@ -130,7 +130,7 @@ codeunit 30166 "Shpfy Process Order"
             SalesHeader."Shpfy Order Id" := ShopifyOrderHeader."Shopify Order Id";
             SalesHeader."Shpfy Order No." := ShopifyOrderHeader."Shopify Order No.";
             SalesHeader.Validate("Document Date", ShopifyOrderHeader."Document Date");
-            SalesHeader.Validate("External Document No.", ShopifyOrderHeader."PO Number");
+            SalesHeader.Validate("External Document No.", CopyStr(ShopifyOrderHeader."PO Number", 1, MaxStrLen(SalesHeader."External Document No.")));
             if ShopifyOrderHeader."Due Date" <> 0D then
                 SalesHeader.Validate("Due Date", ShopifyOrderHeader."Due Date");
             if OrderMgt.FindTaxArea(ShopifyOrderHeader, ShopifyTaxArea) and (ShopifyTaxArea."Tax Area Code" <> '') then


alexei-dobriansky:
**Recommendation: Accept with Suggestions**
**Problem-solution fit: Strong**

This change fixes the reported failure in the right place and uses the standard CopyStr(..., MaxStrLen(...)) pattern already used around the same mapping code. I did not find evidence that Shopify document linkage depends on External Document No.; it is keyed by Shopify order fields instead.

### Suggestions
- **S1** (🟠 Moderate) Add a regression test for long PO numbers - Add a Shopify order test with a PO number longer than 35 characters and assert sales document creation succeeds. This is the exact path that used to fail, and the fix is small enough to regress quietly later.

### Risk & necessity
Not merging leaves affected Shopify B2B orders stuck in an error state and forces manual cleanup. Merging is low risk: truncation is bounded from the target field length, and the main downstream effect looks limited to possible duplicate customer-facing references rather than wrong Shopify document matching.

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10305 round=1 by=alexei-dobriansky at=2026-08-26T22:29:57Z lastSha=e5606c71e8c6e9c9a44cba18b22b9614aca91f6e reviewKey=6f86a4d316edc57c1e9efb9bb9955116a61097a191072f439ec2e7663fb4724f suggestions=S1@3a4c9e12

PR #10310, opened 2026-08-17
Bug 647080: [28.x] Fix: migrating a document attachment deletes shared Tenant Media

JesperSchulz:
#### Summary

Backport of #10055 to `releases/28.x` for the September release (28.5).

`DA External Storage Impl.DeleteFromInternalStorage` deleted the `Tenant Media` row unconditionally after moving an attachment to external storage. `Tenant Media` is shared storage: `Document Attachment Mgmt.CopyAttachments` copies the media *reference* rather than the bytes, so Vendor → Purchase Header → Purch. Inv. Header attachments all end up referencing a single row. Deleting that row on behalf of one attachment destroyed the content of every attachment copied from it.

Those attachments were then skipped **silently** by `UploadToExternalStorage` — the `HasValue()` guard sits above `LogFeatureUsed()`, so no telemetry was emitted — leaving `Stored Externally = false`, a blank `External File Path` and no internal content. Opening one raises `Tenant Media does not exist. ID='{00000000-0000-0000-0000-000000000000}'` from `GetAsTempBlob`.

On a production tenant migrating 5,226 attachments in one company, 4,810 uploaded successfully and 416 were silently skipped with their content destroyed.

The fix only deletes the `Tenant Media` row when no other `Document Attachment` references it, index-backed by the existing `key(Key2; "Document Reference ID")` on table 1173.

#### Work Item(s)

Fixes [AB#647080](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/647080) (backport of [AB#646505](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/646505))

#### Cherry-pick

Cherry-pick of 473bb07683d69a4637515209367837b444b3a43e (`-x`). The AL change is byte-identical to the change merged to `main` (2 files, 155 insertions, 7 deletions).

**One deliberate omission.** The original PR also registered the *External Storage - Document Attachments* test app in the build, via `build/groups.json` and 24 × `build/projects/Apps <CC>/.AL-Go/settings.json`. Those files do not exist on `releases/28.x`, which still uses the earlier build layout:

- there is no `build/groups.json` on this branch (and no equivalent grouping file — `AllExtensions` appears nowhere on `releases/28.x`);
- `build/projects/Apps (W1)/.AL-Go/settings.json` declares `testFolders` with wildcards (`../../../src/Apps/W1/*/Test`), which already matches `src/Apps/W1/External Storage - Document Attachments/Test`.

So the test app is picked up automatically here and no registration change is needed.

#### Validation

Tests in `DA Ext. Storage Impl. Tests`:

| Test | Asserts |
|---|---|
| `DeleteFromInternalKeepsMediaSharedWithCopiedAttachment` | Shared `Tenant Media` survives; the copy keeps its content |
| `DeleteFromInternalRemovesMediaWhenNotShared` | Space is still reclaimed when the attachment is the sole owner |
| `UploadSucceedsForCopiedAttachmentAfterSourceIsMigrated` | After the source is migrated, the copy still uploads and gets its own external file |

The first and third fail before the fix and pass after; the middle one passes both before and after, proving the fix does not stop reclaiming database space.

### DIFF
--- a/src/Apps/W1/External Storage - Document Attachments/Test/src/DAExtStorageImplTests.Codeunit.al
+++ b/src/Apps/W1/External Storage - Document Attachments/Test/src/DAExtStorageImplTests.Codeunit.al
@@ -7,6 +7,7 @@ namespace Microsoft.ExternalStorage.DocumentAttachments.Test;
 
 using Microsoft.ExternalStorage.DocumentAttachments;
 using Microsoft.Foundation.Attachment;
+using System.Environment;
 using System.ExternalFileStorage;
 using System.TestLibraries.ExternalFileStorage;
 using System.TestLibraries.Utilities;
@@ -17,6 +18,7 @@ codeunit 136820 "DA Ext. Storage Impl. Tests"
     Subtype = Test;
     TestPermissions = Disabled;
     Permissions = tabledata "Document Attachment" = rimd,
+                  tabledata "Tenant Media" = r,
                   tabledata "DA External Storage Setup" = rimd;
 
     var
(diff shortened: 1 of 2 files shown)


JesperSchulz:
Closing and reopening due to GitHub outage.

darjoo (approved):

Aleyenda (approved):

PR #10336, opened 2026-08-18
[Main]- Incorrect calculation of single-level capacity and material cost for Stockkeeping Units (SKU) when running cost changes in the Standard Cost Worksheet and implementing the Standard Cost change

DeepsShukla:
Workitem [Bug 637759](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/637759): [ALL-E] Incorrect calculation of single-level capacity and material cost for Stockkeeping Units (SKU) when running cost changes in the Standard Cost Worksheet and implementing the Standard Cost change.

Fixed [AB#637759](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/637759)

Issue: Incorrect calculation of single-level capacity and material cost for Stockkeeping Units (SKU) when running cost changes in the Standard Cost Worksheet and implementing the Standard Cost change.
Cause: The Implement Standard Cost Change report unconditionally copied the item’s updated cost shares to its SKUs, ignoring the Load SKU Cost on Manufacturing setting.
Solution: Skip SKU cost updates in this mode and instruct users to recalculate SKU costs using Calc. Standard Cost on the Stockkeeping Unit Card.

### DIFF
--- a/src/Layers/W1/BaseApp/Manufacturing/StandardCost/ImplementStandardCostChange.Report.al
+++ b/src/Layers/W1/BaseApp/Manufacturing/StandardCost/ImplementStandardCostChange.Report.al
@@ -343,7 +343,7 @@ report 5855 "Implement Standard Cost Change"
         StdCostWkshName := NewStdCostWkshName;
     end;
 
-    local procedure GetMessage() TheMsg: Text[250]
+    local procedure GetMessage() TheMsg: Text
     var
         Item: Record Item;
         MachCtr: Record "Machine Center";
--- a/src/Layers/W1/BaseApp/Manufacturing/StandardCost/ImplementStandardCostChange.Report.al
+++ b/src/Layers/W1/BaseApp/Manufacturing/StandardCost/ImplementStandardCostChange.Report.al
@@ -11,6 +11,7 @@ using Microsoft.Inventory.Item;
 using Microsoft.Inventory.Journal;
 using Microsoft.Inventory.Location;
 using Microsoft.Manufacturing.MachineCenter;
+using Microsoft.Manufacturing.Setup;
 using Microsoft.Manufacturing.WorkCenter;
 using Microsoft.Pricing.Calculation;
 using Microsoft.Pricing.PriceList;
@@ -72,6 +73,7 @@ report 5855 "Implement Standard Cost Change"
                 LockTable();
 
                 RevalJnlCreated := false;
+                LoadSKUCostOnMfg := LoadSKUCostOnManufacturing();
             end;
         }
     }
@@ -189,6 +191,8 @@ report 5855 "Implement Standard Cost Change"
         WorkCtrCostsUpdated: Boolean;
         ResCostsUpdated: Boolean;
         SKUCostsUpdated: Boolean;
+        SKUCostUpdateSkipped: Boolean;
+        LoadSKUCostOnMfg: Boolean;
         NoMessage: Boolean;
         HideDuplWarning: Boolean;
 #pragma warning disable AA0074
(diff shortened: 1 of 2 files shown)


v-ankitgoyal (dismissed):

alexei-dobriansky:

<details>

The `Implement Standard Cost Change` report (report 5855) copied the item's new standard cost onto its stockkeeping units even when Manufacturing Setup `"Load SKU Cost on Manufacturing"` is enabled, which overwrote the SKU's own independently calculated cost. The fix reads that setup once in `OnPreDataItem` and, in `UpdateSKU`, skips the SKU cost update when the setting is on, then appends a message telling the user to run the SKU standard cost calculation on the Stockkeeping Unit Card.

The fix is correct and targeted. The setup is read before `OnAfterGetRecord` calls `UpdateSKU`, so the flag is set in time; the skip happens only when SKUs exist; the `OnUpdateSKUOnBeforeValidateStandardCost` event no longer fires in this mode, which is the intended behavior change, not a break. Report 5855 exists only in the W1 layer, so no other layer needs the same change. Three tests cover the skipped update, an SKU with its own routing keeping its cost, and the instruction message.
</details>

<details>

**S1** - *Final message can exceed its 250-char limit*
`GetMessage()` returns `TheMsg: Text[250]`. On line 381 the new instruction (about 110 characters) is appended after the item, SKU, machine center, work center and resource captions plus the revaluation-journal text. A worksheet run that updates many of these can push the total over 250 characters, and assigning past that length raises a runtime error at the very end of the report. Consider widening the return type to an unbounded `Text` (or building the message in a `TextBuilder`) so the report cannot fail while only showing its summary.
</details>

<details>

**Risk:** The change is limited to report 5855 and only removes an incorrect SKU write, so it cannot post wrong ledger amounts; it changes behavior only when `"Load SKU Cost on Manufacturing"` is enabled. The one robustness concern is the message-length limit above, which is pre-existing but made more likely by the added text. Users who currently rely on the old (incorrect) SKU overwrite will now need to recalculate SKU cost themselves, which is exactly what the message explains.

**Necessity:** The scenario is valid and clear - copying item cost onto SKUs ignores a documented setup and corrupts independently maintained SKU costs. The fix is the right size for a bug, sits on the correct extension point, and is covered by focused tests with exact-value assertions. Only the non-blocking message-length item keeps this from a plain Accept.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=10336 round=1 by=alexei-dobriansky at=2026-08-20T12:28:36Z lastSha=4067c33e163035bc780265ffd308e35fc80cfcb6 reviewKey=3da1a84a0e15f073f187c91563424d9f5173bd03c82064e06236076b5454317a suggestions=S1@d2fbd238

alexei-dobriansky:

# **Recommendation: Accept**

<details>

Round 2 has one new non-merge commit. It changes `GetMessage()` in report 5855 from `Text[250]` to unbounded `Text`, which removes the message-length risk raised in round 1.

This directly addresses the previous concern. The procedure is local and is only used to pass text to `Message(GetMessage())`, so the wider return type does not change a public API or event contract.
</details>

<details>

| S1 | Final message can exceed its 250-char limit | Addressed | Fixed in `9c4e61f` by widening `TheMsg` to unbounded `Text`. |
</details>

<details>

</details>

<details>

**Risk:** The round-2 change is very small and local to report 5855. It reduces risk by preventing the final summary message from failing or being limited by `Text[250]`; it does not add new database writes, posting behavior, or integration-surface changes.

**Necessity:** The original bug scenario remains valid: SKU costs should not be overwritten when Manufacturing Setup says SKU manufacturing costs are loaded separately. The round-2 change is necessary because the new instruction text made the old 250-character return type too small.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=10336 round=2 by=alexei-dobriansky at=2026-08-21T06:05:38Z lastSha=9c4e61f6856eb0af5d7d09eb9d76951152aeb0d2 reviewKey=20aa297117243c55b75dc8894bd48ad8b51b9395c98fa2f6fd7eaaa3c9aa5bc3 suggestions=S1@d2fbd238:addressed parentRound=1

alexei-dobriansky (approved):

neeleshsinghal (approved):

PR #10337, opened 2026-08-18
Let subscribers override the half-year new year date in SetHalfYearConventionMethod

Explorer986:
## What & why

Codeunit 5611 raises OnBeforeSetHalfYearConventionMethod, but NewYearDate was passed by value and then always overwritten with the accounting period start. So a subscriber could set it, but it never took effect.

This keeps the existing event signature unchanged and appends a trailing var NewYearDateOverride parameter. When a subscriber sets it, the base uses that date instead of the accounting period start. With no subscriber the behavior is unchanged. The same change is applied to every copy of codeunit 5611, so the event behaves the same across the W1, FR, GB, IT, and RU layers.

Fixes [AB#646465](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/646465)

## How I validated this

- [x] I read the full diff, and it contains only changes I intended.

No test added since this is an additive event parameter with existing coverage.

## Risk & compatibility

Additive and non-breaking. The existing event keeps its shipped signature, a trailing var NewYearDateOverride parameter is appended, and existing subscribers that omit it continue to compile.

### DIFF
--- a/src/Layers/FR/BaseApp/FixedAssets/Depreciation/CalculateNormalDepreciation.Codeunit.al
+++ b/src/Layers/FR/BaseApp/FixedAssets/Depreciation/CalculateNormalDepreciation.Codeunit.al
@@ -682,9 +682,10 @@ codeunit 5611 "Calculate Normal Depreciation"
         AccountingPeriod: Record "Accounting Period";
         Result: Boolean;
         IsHandled: Boolean;
+        NewYearDateOverride: Date;
     begin
         IsHandled := false;
-        OnBeforeSetHalfYearConventionMethod(FADeprBook, DeprMethod, Year365Days, FirstDeprDate, NewYearDate, Result, IsHandled);
+        OnBeforeSetHalfYearConventionMethod(FADeprBook, DeprMethod, Year365Days, FirstDeprDate, NewYearDate, Result, IsHandled, NewYearDateOverride);
         if IsHandled then
             exit(Result);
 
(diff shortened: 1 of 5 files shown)


AndersLarsenMicrosoft (dismissed):

AleksanderGladkov (approved):

alexei-dobriansky:

<details>

This PR lets subscribers to `OnBeforeSetHalfYearConventionMethod` provide a `NewYearDateOverride`, then uses that date instead of the first matching accounting period start in all five BaseApp layer copies of codeunit 5611. The change is coherent and the event publisher and call site were checked together in W1, FR, GB, IT, and RU. The default path still uses the accounting period date, so existing behavior stays the same when no subscriber sets the new parameter. The uncovered part is the new subscriber-driven depreciation result, which can change fixed-asset depreciation amounts.
</details>

<details>

**S1** - *Add coverage for the override path*
Add a regression test where an event subscriber sets `NewYearDateOverride` and depreciation uses that date. This code changes a fixed-asset depreciation path, so the new financial behavior should be covered before merge. The existing `ERMFixedAssets` half-year depreciation setup looks like the right test base.
</details>

<details>

**Risk:** The regression surface is fixed-asset depreciation with half-year convention enabled. A subscriber can now change `NewYearDate`, which feeds `CalcHalfYearConventionDepr()` and can change depreciation amounts. Adding a trailing event parameter is compatible with existing subscribers, and the same publisher/signature change is applied across W1, FR, GB, IT, and RU.

**Necessity:** The PR description gives a valid bug scenario: the existing event exposed `NewYearDate` by value before it was overwritten, so subscribers had no usable way to affect this date without taking over the whole method. The scope is targeted, but the new financial path needs a test because an incorrect override would produce wrong depreciation.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=10337 round=1 by=alexei-dobriansky at=2026-08-18T13:11:48Z lastSha=55611ea0de749590a12aa0b84096ff668813917d reviewKey=cccaf2e680188d24592cce1fd51f366787364112f22954d2736a108c7a0dbca6 suggestions=S1@55f54792

ventselartur (approved):

PR #10276, opened 2026-08-17
[master] Renewal Term field on the Create Contract Renewal Quote page does not accept the entered value such as 12M

neeleshsinghal:
Workitem : [Bug 647017](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/647017): [master] [all-e][FTE][SaaS] Renewal Term field on the Create Contract Renewal Quote page does not accept the entered value such as '12M'

Fixes [AB#647017](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/647017)

**Issue:** On the Create Contract Renewal Quote page, entering a Renewal Term (e.g. 12M) resets to blank when leaving the field.

**Cause:** OnAfterGetRecord always reloaded the persistent Subscription Line via LoadServiceCommitmentForContractLine, overwriting the just-entered value held in the temporary buffer after CurrPage.Update.

**Solution:** Read the current row from the temporary buffer first and only fall back to loading the persistent record when it isn't buffered, preserving the entered Renewal Term.

PR #10351, opened 2026-08-18
[28.x] Renewal Term field on the Create Contract Renewal Quote page does not accept the entered value such as 12M- #10276

neeleshsinghal:
Workitem : [Bug 645198](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/645198): [all-e][FTE][SaaS] Renewal Term field on the Create Contract Renewal Quote page does not accept the entered value such as '12M'

Fixes [AB#645198](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/645198)

**Issue:** On the Create Contract Renewal Quote page, entering a Renewal Term (e.g. 12M) resets to blank when leaving the field.

**Cause:** OnAfterGetRecord always reloaded the persistent Subscription Line via LoadServiceCommitmentForContractLine, overwriting the just-entered value held in the temporary buffer after CurrPage.Update.

**Solution:** Read the current row from the temporary buffer first and only fall back to loading the persistent record when it isn't buffered, preserving the entered Renewal Term.

### DIFF
--- a/src/Apps/W1/Subscription Billing/App/Contract Renewal/Pages/ContractRenewalSelection.Page.al
+++ b/src/Apps/W1/Subscription Billing/App/Contract Renewal/Pages/ContractRenewalSelection.Page.al
@@ -280,7 +280,8 @@ page 8006 "Contract Renewal Selection"
     trigger OnAfterGetRecord()
     begin
         Rec.GetServiceObject(ServiceObject);
-        Rec.LoadServiceCommitmentForContractLine(TempServiceCommitment);
+        if not TempServiceCommitment.Get(Rec."Subscription Line Entry No.") then
+            Rec.LoadServiceCommitmentForContractLine(TempServiceCommitment);
         RenewalTerm := TempServiceCommitment."Renewal Term";
         RenewalTermEnabled := TempServiceCommitment."Subscription Header No." <> '';
 
--- a/src/Apps/W1/Subscription Billing/Test/Contract Renewal/ContractRenewalTest.Codeunit.al
+++ b/src/Apps/W1/Subscription Billing/Test/Contract Renewal/ContractRenewalTest.Codeunit.al
@@ -740,6 +740,51 @@ codeunit 139692 "Contract Renewal Test"
         until SalesLine.Next() = 0;
     end;
 
+    [HandlerFunctions('ExchangeRateSelectionModalPageHandler,MessageHandler')]
+    [Test]
+    procedure RenewalTermEnteredValueIsRetainedOnPage()
+    var
+        ServiceCommitment: Record "Subscription Line";
+        CurrentRenewalTerm: DateFormula;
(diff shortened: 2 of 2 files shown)


alexei-dobriansky:

<details>

The Create Contract Renewal Quote page reset the Renewal Term to blank when the user left the field. The fix changes `OnAfterGetRecord` to read the current row from the temporary buffer first (`TempServiceCommitment.Get(Rec."Subscription Line Entry No.")`) and only load the persistent Subscription Line when the row is not buffered.

This addresses the root cause. The old code always called `LoadServiceCommitmentForContractLine`, which overwrote the value the user had just typed into the buffer after `CurrPage.Update`. Reading the buffer first returns the pending value, so the entered term is shown. The `else` fallback keeps the old behavior for lines that are not buffered (for example a line where `Get` fails), so unbuffered lines are unaffected. The buffer is filled in `InitTempServiceCommitment` from a direct record copy; the displayed money fields still match the previous load because those are stored fields, and the only extra step in the old load path (`CalcFields(Quantity)`) is for a field this page does not show.
</details>

<details>

**S1** - *Test uses the same value on both lines*
The test sets the same `NewRenewalTerm` on the first and the next line, and it does not go back to the first line to check its value again. The fix depends on reading the correct row from the buffer by `Subscription Line Entry No.`. Please set a different Renewal Term on each line and, after moving to the next line, check that the first line still keeps its own value. This tests per-line buffering, not just that one value stays.
</details>

<details>

**Risk:** The change touches one page trigger in the Subscription Billing module (`ContractRenewalSelection.Page.al`) and does not change the apply path (`TransferChangedValuesFromBufferToSubscriptionLine`), so the regression surface is small. All displayed fields on the row now come from the buffer copy instead of a fresh load; for buffered lines these are stored fields with the same values, and the only load-only step (`CalcFields(Quantity)`) is for a field this page does not display, so there is no visible difference. This is a page/UX display fix, not a posting or financial path.

**Necessity:** The scenario is a clear, reproducible customer-facing defect (entered Renewal Term appears to be rejected), so the fix is warranted. The change is minimal and correctly scoped to a bug fix, with a regression test added.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=10351 round=1 by=alexei-dobriansky at=2026-08-18T19:09:57Z lastSha=97aad9a061e1cdb83aeb4578df715d5f0edd6130 reviewKey=a9d2eabffdb1ec8d5a47396f3c74f1e7ac12062e3de398118dfc0ead2493840e suggestions=S1@08aca1da

v-ankitgoyal (approved):

AndersLarsenMicrosoft (approved):

AndersLarsenMicrosoft (approved):

PR #10363, opened 2026-08-18
Bug 646360: Validate introduced object IDs against allowed ranges in the PR gate

ChethanT:
## Summary
Adds a pull-request / merge-group gate check in the existing `TestObjectIdsAndManifests` action that **fails when a newly added production AL file declares an object ID outside the allowed first-party ranges** `1..49999` and `99000750..99001048`.

## Root cause
The gate validated test-object ranges, duplicate object IDs, app IDs, and manifests, but did not prevent new production objects from using ranges assigned to partners or ISVs. This allowed the Subcontracting app to overlap with LS Retail, which was discovered only later and caused substantial rework.

## Fix
- Validates object declarations only in **newly added AL files** (`git diff --diff-filter=A`); modified, renamed, moved, deleted, and existing files are intentionally outside this gate.
- Uses merge-base-to-head change detection so changes on the target branch are not misclassified as PR additions.
- Covers W1 and country apps through `$allPaths`.
- Excludes test applications using AL-Go `testFolders`, nearest `app.json` test-app classification, and `Subtype = Test` as supplementary detection.
- Covers ID-bearing object types including extensions, enums, and permission sets.
- Aggregates all offending objects and reports type, ID, name, and path before failing.
- Fails explicitly when required PR change detection cannot be completed instead of silently disabling the check.

## Test
Adds focused regression coverage for allowed range boundaries, invalid W1 and country additions, non-codeunit objects in test apps, manifest and AL-Go test-folder classification, wildcard test folders, ignored non-AL/out-of-scope files, aggregated failures, and added-file Git filtering.

[AB#646360](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/646360)

🤖 Generated with GitHub Copilot


ChethanT:
Blocked by https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/644164

build/scripts/AppObjectValidation.psm1
```diff
@@ -323,8 +337,271 @@ function GetObjectId
...
+        [Parameter(Mandatory = $true)] [string[]] $RelativeSourcePaths,
+        [string] $ObjectTypePattern = 'codeunit|page|table|query|report|xmlport'
+    )
+
+    Push-Location -Path $RepositoryRoot
+    try {
+        & git cat-file -e "$Commitish^{commit}" 2>$null
+        if ($LASTEXITCODE -ne 0) {
+            Write-Host "Base commit '$Commitish' is not available locally. Attempting to fetch it from origin."
+            & git fetch --no-tags --quiet origin $Commitish 2>$null
```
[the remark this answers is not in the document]
ChethanT: Fixed in `eac90c57`:
```powershell
$fetchRef = "+refs/heads/$branchName`:refs/remotes/origin/$branchName"
& git fetch --no-tags --quiet origin $fetchRef
```
The fallback now fetches the remote branch into the local tracking ref that is subsequently resolved. Added a regression test that starts without `origin/main` and verifies both object loading and tracking-ref creation.

*Should have been caught during implementation - a local remote-tracking name cannot be reused as a remote-side fetch ref.*

Useful detection signal: fallback Git fetch paths should distinguish the remote source ref from the local destination ref whenever later code resolves a remote-tracking name.

PredragMaricic (dismissed):

JesperSchulz (dismissed):

.github/actions/TestObjectIdsAndManifests/action.ps1
```diff
@@ -81,4 +81,35 @@ $expectedPlatformVersionPrevious = "$($currentMajor - 1).0.0.0"
...
+}
+else {
+    $allowedRangesText = ($AllowedObjectIdRanges | ForEach-Object { "$($_.From)..$($_.To)" }) -join ', '
+    Write-Host "Validating introduced object IDs against base commit '$baseCommitSha' (allowed ranges: $allowedRangesText)."
+
+    $repositoryRoot = (Resolve-Path -Path (Get-BaseFolder)).Path
+    $relativeRangeCheckPaths = @($w1OnlyPaths | ForEach-Object { (Resolve-Path -Path $_).Path.Substring($repositoryRoot.Length).TrimStart('\', '/') })
+
+    $currentObjects = Get-FilesCollection -SourceCodePaths $w1OnlyPaths -ObjectTypePattern $ObjectTypePattern
+    $baseObjects = Get-ObjectCollectionAtCommit -Commitish $baseCommitSha -RepositoryRoot $repositoryRoot -RelativeSourcePaths $relativeRangeCheckPaths -ObjectTypePattern $ObjectTypePattern
```
aholstrup1: This seems like a lot of added code just to only run the check on the changed files. 

I'm considering if we should take this out for now? The other checks in the script run against the full changeset. Seems odd that we start changing that mid-script 🤷‍♂️ 

Additionally, what happens if you make a change to this script? Say you reduce the range. The check will pass because it doesn't validate any objects, but it may start failing once people make changes to those objects.

.github/actions/TestObjectIdsAndManifests/action.ps1
```diff
@@ -81,4 +81,35 @@ $expectedPlatformVersionPrevious = "$($currentMajor - 1).0.0.0"
...
+# Test-ApplicationTestTypes -SourceCodePaths $w1OnlyPaths -Exceptions $allowedUncategorizedTests
+
+# Test that newly introduced object IDs are within the allowed ranges.
+$AllowedObjectIdRanges = @(
+    [PSCustomObject]@{ From = 1;        To = 49999 },
+    [PSCustomObject]@{ From = 99000750; To = 99001048 }
+)
+
+# Object signatures ("<object type> <object id>") explicitly allowed to be outside the ranges above.
+$AllowedOutOfRangeObjects = @()
```
mazhelez: Why is this needed if it's empty?
What is the case where we add a value here?

build/scripts/AppObjectValidation.psm1
```diff
@@ -323,8 +337,277 @@ function GetObjectId
...
+    not contain the base SHA, the function falls back to 'origin/<GITHUB_BASE_REF>'. Returns $null when no base
+    can be determined.
+    .PARAMETER EventName
+    The GitHub event name. Defaults to the GITHUB_EVENT_NAME environment variable.
+    .PARAMETER EventPath
+    The path to the GitHub event payload JSON file. Defaults to the GITHUB_EVENT_PATH environment variable.
+    .PARAMETER BaseRef
+    The base ref (target branch) name. Defaults to the GITHUB_BASE_REF environment variable.
+#>
+function Get-PullRequestBaseSha {
```
mazhelez: That function doesn't look like it should belong here.

ChethanT: Addressed in `101e2156`. PR/Git change detection now lives in `BuildOptimization.psm1`; `AppObjectValidation.psm1` is limited to classifying files and validating object IDs. The unrelated shared helper was not moved.

.github/actions/TestObjectIdsAndManifests/action.ps1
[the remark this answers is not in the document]
ChethanT: Addressed in `101e2156`. Added-file validation now uses `$allPaths`, so newly added AL files under country apps are covered as well as W1. Existing country-app files are not scanned because the gate intentionally applies only to new files.

ChethanT: Addressed in `101e2156`. The gate now intentionally validates only newly added AL files (`git diff --diff-filter=A`) rather than scanning all existing objects on every PR. Change detection compares the merge base with the PR head, and fails explicitly if the changed-file set cannot be determined. This keeps the gate focused on preventing new invalid files without making unrelated PRs responsible for historical objects or later range-policy changes.

build/scripts/AppObjectValidation.psm1
```diff
@@ -323,8 +337,277 @@ function GetObjectId
...
+#>
+function Test-IntroducedObjectIDsAreInAllowedRange {
+    param(
+        [Parameter(Mandatory = $true)] [object] $CurrentObjects,
+        [Parameter(Mandatory = $true)] [object] $BaseObjects,
+        [Parameter(Mandatory = $true)] [object[]] $AllowedRanges,
+        [string[]] $AllowedOutOfRangeObjects = @()
+    )
+
+    $introducedObjects = Get-IntroducedObjects -CurrentObjects $CurrentObjects -BaseObjects $BaseObjects -ExcludeTestObjects
```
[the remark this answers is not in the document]
ChethanT: Addressed in `101e2156`. The gate now validates only newly added AL files and excludes test applications using AL-Go `testFolders`, with nearest-manifest test-app classification and `Subtype = Test` as supplementary detection. This covers non-codeunit test objects such as tables, pages, enums, and extensions instead of relying only on codeunit subtype.

.github/actions/TestObjectIdsAndManifests/action.ps1
ChethanT: Addressed in `101e2156`. The empty `AllowedOutOfRangeObjects` exception mechanism has been removed; the added-file-only scope makes that placeholder unnecessary.

.github/actions/TestObjectIdsAndManifests/action.ps1
```diff
@@ -81,4 +93,27 @@ $expectedPlatformVersionPrevious = "$($currentMajor - 1).0.0.0"
 Test-ApplicationManifests -Path $allPaths -ExpectedAppVersion "$($currentMajorMinor).0.0" -ExpectedPlatformVersions @($expectedPlatformVersionCurrent, $expectedPlatformVersionPrevious)
 
 # Test that we are not adding new uncategorized tests (W1 only) - Disabled for now
-# Test-ApplicationTestTypes -SourceCodePaths $w1OnlyPaths -Exceptions $allowedUncategorizedTests
\ No newline at end of file
+# Test-ApplicationTestTypes -SourceCodePaths $w1OnlyPaths -Exceptions $allowedUncategorizedTests
+
+# Test object IDs declared in newly added production AL files.
+$AllowedObjectIdRanges = @(
```
JesperSchulz, line 99: Please consider all objects listed here: https://microsoft.ghe.com/bic/BC-Wiki/blob/master/Engineering/Application/CAL-Application-development/GDL-Unique-ID-(UID)-Ranges.md

In the "Objects and fields" section, you will find a few more allowed ranges, which are outside of the two ranges you specified.

JesperSchulz (changes requested):
We need to open up a few more ranges, than the two specified.

.github/actions/TestObjectIdsAndManifests/action.ps1
ChethanT: Addressed in `e5bfa1c8b3`. Added the additional allowed ranges from the GDL Unique ID (UID) Ranges wiki page (Objects and fields section): 2000000-2000059, 3005000-3011499, 5000000-5000099, 5000250-5055999, 5001900-5001999, 5005100-5005399, 7000000-7000099, 11000000-11000499, 15000000-15000499, 31000000-31000099, 32000000-32000199.

JesperSchulz (approved):

.github/actions/TestObjectIdsAndManifests/action.ps1
```diff
@@ -10,12 +11,23 @@ $w1Apps = @(Join-Path $sourceCodeFolder "Apps\W1")
...
+        ForEach-Object { Join-Path $_.FullName 'BaseApp' } |
+        Where-Object { Test-Path -Path $_ }
+)
 
 # Build path sets for different validations
 # Some folders (e.g. Layers) only exist on main and not on release branches, so filter out
 # any paths that do not exist to avoid Get-ChildItem failing on missing directories.
 [string[]] $w1OnlyPaths = @($baseFolders + $w1Apps + $w1Layers | Where-Object { Test-Path -Path $_ })
 [string[]] $allPaths = @($baseFolders + $allApps + $w1Layers | Where-Object { Test-Path -Path $_ })
+[string[]] $objectIdValidationPaths = @(
```
aholstrup1, line 25: Could we just validate everything under the src folder?

.github/actions/TestObjectIdsAndManifests/action.ps1
aholstrup1: Sounds like the opposite of what the comment was about 🤔

.github/actions/TestObjectIdsAndManifests/action.ps1
mazhelez, line 25: How about approaching the problem differently: instead of scanning all .al files, how about verifying only the `idRanges` property in all of the app.json files?

.github/actions/TestObjectIdsAndManifests/action.ps1
```diff
@@ -81,4 +93,38 @@ $expectedPlatformVersionPrevious = "$($currentMajor - 1).0.0.0"
 Test-ApplicationManifests -Path $allPaths -ExpectedAppVersion "$($currentMajorMinor).0.0" -ExpectedPlatformVersions @($expectedPlatformVersionCurrent, $expectedPlatformVersionPrevious)
 
 # Test that we are not adding new uncategorized tests (W1 only) - Disabled for now
-# Test-ApplicationTestTypes -SourceCodePaths $w1OnlyPaths -Exceptions $allowedUncategorizedTests
\ No newline at end of file
+# Test-ApplicationTestTypes -SourceCodePaths $w1OnlyPaths -Exceptions $allowedUncategorizedTests
+
+# Test object IDs declared in newly added production AL files.
+$AllowedObjectIdRanges = @(
```
PredragMaricic, line 99: S1 - The PR summary still says the only allowed ranges are `1..49999` and `99000750..99001048`, but this production list now permits 11 additional ranges. Please update the description so the stated policy matches the implemented gate. The tests also hardcode only the original two ranges; consider single-sourcing or directly testing this production policy list in follow-up.

PredragMaricic (approved):
The added-file gate is fail-closed, covers the intended production roots and object types, excludes test applications through multiple metadata paths, and has focused regression coverage. No blocking correctness issue found.

S1: Please update the PR summary: it still names only 1..49999 and 99000750..99001048, while the implementation now allows 11 additional reviewed ranges. The tests hardcode only the original two ranges, so consider single-sourcing or directly testing the production policy list in follow-up.

.github/actions/TestObjectIdsAndManifests/action.ps1
ChethanT: updated further tests will be implemented in fix for https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/646360

.github/actions/TestObjectIdsAndManifests/action.ps1
ChethanT: Will be addressed in https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/646360

ChethanT: Will be addressed in https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/646360

.github/actions/TestObjectIdsAndManifests/action.ps1
ChethanT: Will be addressed in https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/646360

aholstrup1 (approved):
Approved but please do address the follow up work as part of AB#646360 👍

PR #10370, opened 2026-08-19
[Master]-Bug 647018: Withholding tax for Employees and GL Accounts

ViditGupta1277:
Fixes [AB#647018](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/647018)

WHT for Employees will be primarily used for expense management, but it should be designed to use for other cases as well (other payments to employees, such as dividends, stocks compensations, project-based or professional services...).
So, it must be possible to have WHT product posting group inherited from GL account to the General Journal line when used GL as a balance account.

### DIFF
--- a/src/Apps/W1/ExpenseWithholdingTax/app/src/WHTExpenseCategoryMgt.Codeunit.al
+++ b/src/Apps/W1/ExpenseWithholdingTax/app/src/WHTExpenseCategoryMgt.Codeunit.al
@@ -5,6 +5,7 @@
 namespace Microsoft.ExpenseTaxIntegration;
 
 using Microsoft.ExpenseAgent;
+using Microsoft.Finance.GeneralLedger.Account;
 using Microsoft.Finance.GeneralLedger.Journal;
 using Microsoft.Finance.GeneralLedger.Setup;
 using Microsoft.WithholdingTax;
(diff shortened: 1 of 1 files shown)


alexei-dobriansky:

<details>

This PR adds an event subscriber so that an Employee general journal line whose balancing account is a G/L Account inherits its `Wthldg. Tax Prod. Post. Group` from that balancing G/L account, when no `Expense Category` is set.

The subscriber binds to `OnAfterAccountNoOnValidateGetGLBalAccount`, which the `Gen. Journal Line` table raises only inside the `Bal. Account Type = "G/L Account"` branch of `GetGLBalAccount`, so `GLAccount` is always the balancing G/L account. The guard exits unless the line is an Employee line, and exits when an `Expense Category` is already set, so the expense-category value keeps priority and the G/L account only fills the gap. This mirrors the existing `AssignBalGLAccValue` subscriber in the Purchase `Withholding Tax Jnl Subscriber`, so the event choice and the assignment match the established pattern. It is an additive subscriber and changes no public signature.
</details>

<details>

**S1** - *No test covers the new default*
The `Expense Withholding Tax` app has no test project, so this new default has no automated coverage. The field it sets drives the withholding tax posting setup lookup and the employee WHT calculation, so a regression could post the wrong withholding tax. Please add a test that validates `Bal. Account No.` on an Employee line with a G/L balancing account and checks the resulting `Wthldg. Tax Prod. Post. Group`.

**S2** - *Bal. Account Type check can never be true*
This event fires only when `Bal. Account Type` is `G/L Account`, so the `Bal. Account Type <> Employee` part of the condition is always true. You can simplify the guard to check only `Account Type = Employee`.
</details>

<details>

**Risk:** Small, additive change in one W1 add-on codeunit. It only sets a value during `Bal. Account No.` validation when the account is an Employee and no `Expense Category` is set, so the blast radius is limited to Employee journal lines balanced to a G/L account. The value feeds `WithholdingTaxPostingSetup.Get` and the employee WHT calculation, so a wrong default would affect withholding amounts; the field stays user-visible and can be overridden. No public API or event signature changes. The primary-account (G/L) case is not handled, which matches the Employee-primary scenario in the title.

**Necessity:** The change is needed so the WHT Prod. Posting Group defaults correctly for employee expense lines that balance to a G/L account, matching the Vendor/Purchase behavior. The scope is right for a bug fix (18 lines, one file). Without it the group must be entered by hand and withholding tax can be missed.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=10370 round=1 by=alexei-dobriansky at=2026-08-19T07:08:55Z lastSha=90b38d38fafbe9ca93c236ea7569f7d9492f7553 reviewKey=a84e3d6c4bff685b66d4f1ad14f5d29acc951d229d7292c9ea0ed4e83c85c086 suggestions=S1@42e3055c,S2@b116be16

Alexander-Ya (approved):
Approved, but tests are missing

AndersLarsenMicrosoft (approved):

PR #10412, opened 2026-08-19
Fixes in Excel reports for Aged Accounts.

mynjj:
Both the receivables and payables reports had the following problems:

- The aged buckets were not covering the whole history for a customer, only bounded to the end date supplied and the starting date according to the number of periods. Therefore a customer that historically had a debt that is before the inspected period, would not show up.
    - A new bucket from `0D` to the next date is added
    - The filtering for skipping customers/vendors without due balance is considering all the history, not only those that didn't have activity on the period (this is in line with the legacy report 120)
- The logic to get the bucket that a given date belongs to was wrong for the ending dates, the list had entries sorted descending, hence the first comparison would always win, resulting in weird period ends. 
    - Instead we keep the same computation approach for the start date, but return an index; since the index corresponds to the ending period, we retrieve the corresponding one.
- Due date filter was removed: a customer as of a given date (say 19th Aug 2016) that has some balance due posted in the 17th but not yet due until the 19th of Sept, should appear on the report, the customer owes something, even if not yet in effect. This is in agreement with legacy report 120, but in contrast to that one, we don't put these entries under a special bucket for "not yet due".

Tests have been added for each of these flows. Removed the `AgedAccountsReceivableExcelReportExportsAsPerPeriodCount` tests, that weren't really testing anything and conflicted with the Due Date scenario.

Xlsx was not changed, but I verified the new bucket with 0D renders correctly.

Fixes [AB#642030](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/642030)

### DIFF
--- a/src/Apps/W1/ExcelReports/App/src/Financials/EXRAgedAccPayableExcel.Report.al
+++ b/src/Apps/W1/ExcelReports/App/src/Financials/EXRAgedAccPayableExcel.Report.al
@@ -307,7 +307,6 @@ report 4403 "EXR Aged Acc Payable Excel"
 
     local procedure InitReport()
     var
-        FirstStartDate: Date;
         WorkingEndDate: Date;
         WorkingStartDate: Date;
         i: Integer;
@@ -331,12 +330,15 @@ report 4403 "EXR Aged Acc Payable Excel"
             WorkingStartDate := CalcDate(PeriodLength, WorkingStartDate);
             WorkingEndDate := CalcDate(PeriodLength, WorkingEndDate);
         until i >= PeriodCount;
-        FirstStartDate := WorkingStartDate;
 
-        VendorAgingData.SetAutoCalcFields("Net Change (LCY)");
-        VendorAgingData.SetRange("Date Filter", FirstStartDate, EndingDate);
-        if SkipZeroBalanceVendors then
+        PeriodStarts.Add(0D);
+        PeriodEnds.Add(WorkingEndDate);
+
+        VendorAgingData.SetRange("Date Filter", 0D, EndingDate);
+        if SkipZeroBalanceVendors then begin
+            VendorAgingData.SetAutoCalcFields("Net Change (LCY)");
             VendorAgingData.SetFilter("Net Change (LCY)", '<>0');
+        end;
     end;
 
     local procedure InsertAgingData(var Vendor: Record "Vendor")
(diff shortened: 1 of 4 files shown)


alexei-dobriansky:

# **Recommendation: Accept**

<details>

This fixes the Aged Accounts Receivable and Aged Accounts Payable Excel reports so they include older open entries, include not-yet-due open entries, and use matching period start and end buckets. The receivable and payable changes are aligned, and the shared buffer now chooses one bucket index before reading both dates. The added tests cover the main receivable and payable flows, including the skip-zero-balance option and the boundary case where the old logic mixed period dates.
</details>

<details>

No suggestions.
</details>

<details>

**Risk:** This affects report output for open customer and vendor ledger entries, but it does not change posting or persisted ledger data. The main regression surface is Excel report bucket assignment and which entries are exported. The PR adds focused tests for those paths, and no BaseApp event dependency was introduced.

**Necessity:** The linked bug and PR description describe incorrect aged report output for valid historical and not-yet-due entries. The scope is limited to the two affected Excel reports, their shared buffer, and matching tests, so the change is justified.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=10412 round=1 by=alexei-dobriansky at=2026-08-19T19:10:31Z lastSha=f2b8830c3ca6df1bfdf73adfe4e401a9ed657975 reviewKey=6794bd6f4d18bf7e7893ffecffe7050984fc537e7d2dcffa14423825e5f8db83 suggestions=none

src/Apps/W1/ExcelReports/Test/src/AgedAccountsExcelReports.Codeunit.al
```diff
@@ -254,115 +254,232 @@ codeunit 139555 "Aged Accounts Excel Reports"
...
+        Variant: Variant;
+        RequestPageXml: Text;
+        PeriodStartText: Text;
+        PeriodEndText: Text;
+        PeriodStart: Date;
+        PeriodEnd: Date;
+        FirstPeriodStart: Date;
+        SecondPeriodStart: Date;
+    begin
+        // [SCENARIO] Period Start Date and Period End Date are always read from the same bucket.
```
ventselartur, line 400: Please, give it a try next time :) https://microsoft.ghe.com/bic/BC-agentic-workspace/blob/master/bc-ai-artifacts/application/common/skills/bc-test-implementor/SKILL.md

Explorer986 (approved):

src/Apps/W1/ExcelReports/App/src/Financials/EXRAgedAccPayableExcel.Report.al
```diff
@@ -331,12 +330,15 @@ report 4403 "EXR Aged Acc Payable Excel"
...
-        VendorAgingData.SetAutoCalcFields("Net Change (LCY)");
-        VendorAgingData.SetRange("Date Filter", FirstStartDate, EndingDate);
-        if SkipZeroBalanceVendors then
+        PeriodStarts.Add(0D);
+        PeriodEnds.Add(WorkingEndDate);
+
+        VendorAgingData.SetRange("Date Filter", 0D, EndingDate);
+        if SkipZeroBalanceVendors then begin
+            VendorAgingData.SetAutoCalcFields("Net Change (LCY)");
             VendorAgingData.SetFilter("Net Change (LCY)", '<>0');
```
dcenic, line 340: Long term, we should look for how to get rid of this. This is a SetFilter on a FlowField, meaning that it will not be the SQL Server who does the query - it will be the NST (because this FlowField doesn't exist as a SQL table field).

dcenic (approved):

PR #10439, opened 2026-08-20
[Master]-Inconsistent Validation of Non-Deductible VAT % for Purchase Lines Sharing the Same VAT Identifier - Copy

neeleshsinghal:


alexei-dobriansky:

<details>

The fix makes Non-Deductible VAT % validation symmetric on purchase lines that share a VAT Identifier. Before, only setting a non-zero % was checked (`CheckNonDeductibleVATPctIsAllowed`), so the error depended on the order lines were entered. The PR adds `CheckNonDedVATPctConsistencyForZeroLine()` in the `"Non-Deductible VAT %"` OnValidate to also block setting a line to 0 when a sibling line with the same VAT Identifier still has a non-zero %.

The direction of the check is correct and the same change is applied across all 14 Purchase Line country layers with a test. But the new local procedure does not mirror the existing validator (`NonDedVATImpl.Codeunit.al` -> `CheckNonDeductibleVATPctIsAllowed`) in three ways: it skips blank VAT Identifier, it reports the wrong line in the error, and it bypasses the extensibility hook. The first of these leaves the order-dependent inconsistency in place for blank-identifier lines.
</details>

<details>

**S1** - *Zero check skips blank VAT Identifier*
Line 10296 exits early when `"VAT Identifier" = ''`, but the existing non-zero validator does not skip blank identifiers. So a line can still be set to 0 while a sibling line with a blank VAT Identifier keeps a non-zero Non-Deductible VAT %. This re-creates the same order-dependent inconsistency the PR is trying to remove, just for blank-identifier lines. Remove this early exit so both directions treat blank the same way.

**S2** - *New check bypasses the validation hook*
The existing `CheckNonDeductibleVATPctIsAllowed` raises `OnBeforeCheckNonDeductibleVATPctIsAllowed` with an `IsHandled` flag, so a partner can replace this validation. The new zero-direction check has no such hook and always runs. A subscriber that suppresses the non-zero check would still be blocked on the zero case. Route the zero check through the same event so behavior stays consistent.

**S3** - *Error reports the current line, not the conflicting one*
The existing check reports the found (conflicting) line's `"VAT Bus. Posting Group"` and `"VAT Prod. Posting Group"`. Line 10307 reports the current line (`Rec`) instead. The added test does not catch this because both lines use the same posting groups. Use `PurchLine`'s posting groups so the message points to the line that conflicts.

**S4** - *Duplicate logic instead of reusing the validator*
The zero-direction logic is copied as a local procedure into all 14 Purchase Line layer files, while the matching non-zero logic lives once in `NonDedVATImpl.Codeunit.al`. Consider adding the zero-direction check next to `CheckNonDeductibleVATPctIsAllowed` in that codeunit and calling it from the field. That single move also fixes S1, S2, and S3 in one place.
</details>

<details>

**Risk:** The change is limited to the `"Non-Deductible VAT %"` OnValidate path, so the regression surface is small and it only adds a new error, never changes posted amounts directly. The main risk is the incomplete/divergent validation described above: blank-identifier documents remain inconsistent, and the missing hook can block partners who previously overrode this validation. Non-Deductible VAT % feeds VAT and cost amounts, so an inconsistent set of lines in one VAT Identifier group can still post inconsistent non-deductible VAT.

**Necessity:** The scenario is valid and clear (inconsistent validation depending on line entry order) and worth fixing; a test is included on a financial path, which is appropriate. The scope is right for a bug fix. The recommendation is Request Changes only because S1 leaves part of the same bug unfixed and diverges from the established validator; S2-S4 are consistency and structure improvements.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=10439 round=1 by=alexei-dobriansky at=2026-08-20T12:14:06Z lastSha=1494e837f7df6d43c99ede066395eb49d4ceaaec reviewKey=1d6077f3213173409df0525f02693edfcfb78f57925deff6461cd47c6be4b6f7 suggestions=S1@d1bbd0f3,S2@44c8aee0,S3@d0c961fa,S4@67f0fce8

alexei-dobriansky:

<details>

The round-2 commit changes the new zero-direction lookup from `FindFirst()` to `IsEmpty()` in the Purchase Line layers to satisfy the analyzer warning. It does not change the validation decision: the code still raises an error when another line in the same document has the same VAT Identifier and a different Non-Deductible VAT %.

The linked work items describe a valid bug: purchase lines with the same VAT Identifier should be rejected consistently, regardless of the order in which deductible and non-deductible lines are entered. The latest commit does not address the prior correctness issues. Blank VAT Identifiers are still skipped, the existing validation hook is still bypassed, the error still reports the current line instead of the conflicting line, and the duplicated local logic remains in the Purchase Line layer files.
</details>

<details>

| S1 | Zero check skips blank VAT Identifier | Not addressed | No reply. The early exit for blank `VAT Identifier` is still present. |
| S2 | New check bypasses the validation hook | Not addressed | No reply. The new zero-direction check still does not route through `OnBeforeCheckNonDeductibleVATPctIsAllowed`. |
| S3 | Error reports the current line, not the conflicting one | Not addressed | No reply. The latest commit changed the lookup to `IsEmpty()`, but the error still uses the current line's posting groups. |
| S4 | Duplicate logic instead of reusing the validator | Not addressed | No reply. The same local validation logic is still copied into the Purchase Line layer files. |
</details>

<details>

*None - the new commit only changes `FindFirst()` to `IsEmpty()` and does not introduce a distinct new finding beyond the unresolved prior suggestions.*
</details>

<details>

**Risk:** The round-2 change is narrow and only affects an existence check in the new validation path. The remaining risk is the unresolved behavior from round 1: blank-identifier lines can still keep inconsistent Non-Deductible VAT %, partner overrides of the existing validator are not respected in the zero case, and the error text can still point at the wrong line. This is a VAT/financial data-integrity path, so leaving the asymmetric validation in place can still allow inconsistent document data before posting.

**Necessity:** The scenario is still valid and worth fixing. The work item and test show that the validation should be independent of line entry order. The scope is right for a bug fix, but the implementation still needs to align the zero-direction check with the established validator before merge.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=10439 round=2 by=alexei-dobriansky at=2026-08-20T18:09:18Z lastSha=5eb4451bd470ee8343b9a2c812a48aa3a13daf69 reviewKey=3bc0ca28bbd4f32c0c103e28b037c4b1ab8ac0ef2f091fd921ada7097f98f02f suggestions=S1@d1bbd0f3:notaddressed,S2@44c8aee0:notaddressed,S3@d0c961fa:notaddressed,S4@67f0fce8:notaddressed parentRound=1

PR #10442, opened 2026-08-20
[MAIN]-Bug 647352 Issue with lot no assignment after posting a transfer order

v-praghav:
## What & why

**Issue**: Transfer shipment for LOT0002 incorrectly reduces the remaining quantity of LOT0001.
**Cause**: Reservation-based application does not validate lot/serial tracking, and fallback filtering only considers Specific Tracking.

## Linked work

[AB#647352](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/647352)

Fixes #
**Solution**: Reject mismatched reservations and apply transfer tracking filters so the outbound entry consumes the correct lot

## How I validated this


v-ankitgoyal (dismissed):

alexei-dobriansky:

# **Recommendation: Accept**

<details>

This PR fixes Bug 647352, where posting a transfer order with manually selected lot tracking could reduce the remaining quantity on a different lot. The change makes transfer application filtering use the transfer tracking requirements, not only specific tracking, so `ApplyItemLedgEntrySetFilters` keeps the selected serial or lot when the item uses transfer tracking. The same targeted change is applied to the W1 and localized `Item Jnl.-Post Line` copies, and the new W1 and CZ tests cover the warehouse shipment plus transfer receipt path from the bug scenario.
</details>

<details>

No suggestions.
</details>

<details>

**Risk:** This is inventory posting code, so the regression surface is sensitive. The change is narrow: it only changes filtering for item ledger entries whose entry type is Transfer, and only when transfer serial or lot tracking is required. It does not add or change public events or subscribers, so no BaseApp event publisher dependency needed verification. The PR includes regression tests in W1 and CZ, and the GitHub checks visible on the PR were successful.

**Necessity:** The change is needed because the old fallback application path could ignore lot transfer tracking when the item was not lot-specific. Without the fix, a valid transfer can consume the wrong lot and leave incorrect remaining quantity by lot. The scope matches the bug and updates the country-layer copies that contain the same posting code.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=10442 round=1 by=alexei-dobriansky at=2026-08-21T06:05:39.113Z lastSha=ea0b05288338585d22a3254baf80b7259daf0c8e reviewKey=51f2a234b3a644b73b28ad6f7662db549d4aceeeb2f99b268fc75023942f2b55 suggestions=

alexei-dobriansky (dismissed):

src/Layers/W1/BaseApp/Inventory/Posting/ItemJnlPostLine.Codeunit.al
```diff
@@ -1734,6 +1734,11 @@ codeunit 22 "Item Jnl.-Post Line"
         ItemTrackingSetup2.CopyTrackingFromItemTrackingCodeSpecificTracking(ItemTrackingCode);
         ItemTrackingSetup2.CopyTrackingFromItemLedgerEntry(FromItemLedgEntry);
 
+        if FromItemLedgEntry."Entry Type" = FromItemLedgEntry."Entry Type"::Transfer then begin
+            ItemTrackingSetup2."Serial No. Required" := GlobalItemTrackingSetup."Serial No. Required";
+            ItemTrackingSetup2."Lot No. Required" := GlobalItemTrackingSetup."Lot No. Required";
```
PredragMaricic, line 1739: This transfer-specific override handles serial and lot tracking, but it leaves `ItemTrackingSetup2."Package No. Required"` at the package-specific-tracking value copied above. Package tracking participates in this same filter through `Package Management`'s subscriber to `OnAfterSetTrackingFilterFromItemTrackingSetupIfRequired`, and `GlobalItemTrackingSetup` carries `"Package Transfer Tracking"` in `"Package No. Required"`. Therefore a package-all-tracked transfer can still select and apply an item ledger entry from the wrong package. Please copy `GlobalItemTrackingSetup."Package No. Required"` here as well and cover that transfer scenario; apply the same change to the localization copies.

v-praghav: Inclued package Changes

alexei-dobriansky:

<details>
<summary><strong>What changed since round 1</strong></summary>

Round 1 (Accept, no suggestions) covered the transfer application filter fix that keeps the selected serial or lot when the item uses transfer tracking. Since then, one new author commit ("Include Package Changes") refactors that fix into an extensibility point and widens it to package tracking:

- The inline transfer assignment now raises a new integration event `OnAfterSetTrackingFilterFromItemTrackingSetupIfRequired`, added inside the `Transfer` branch of `ApplyItemLedgEntrySetFilters` in the W1 copy and all five localized copies (APAC, CH, ES, IT, RU).
- A new subscriber in codeunit 6516 `Package Management` copies `"Package No. Required"` from the global setup into the applied tracking setup, so a transfer application also honors the package requirement, not only serial/lot.

</details>

<details>
<summary><strong>Assessment of the round-2 change</strong></summary>

The refactor is correct and idiomatic. The event is a `[IntegrationEvent(false, false)]` local publisher, the subscriber signature matches, and `var ItemTrackingSetup` is the parameter the subscriber writes to. Placing the publisher inside the existing `Transfer` entry-type block keeps package handling consistent with the serial/lot handling it sits next to. Publishing the event in every localized copy of codeunit 22 is the right call, because the single W1 `Package Management` subscriber then fires regardless of which country layer is active. Moving the package-specific logic into the package codeunit (rather than hardcoding another field in codeunit 22) follows Business Central's separation of concerns.

Regression risk stays low and in the same class as round 1: the change only narrows the item ledger entry application filter for `Transfer` entries when package tracking is required, mirroring the serial/lot path.

</details>

<details>

**S1 - Add a regression test for the package-tracking transfer path (Medium).** The new event and subscriber extend the transfer application filter to also require the package number, but the added tests only exercise lot tracking, and the test codeunit has no package scenario. Add a test that posts a transfer for an item using package tracking and asserts the correct package and remaining quantity, so the package path is protected the same way the lot path now is. This is a non-blocking suggestion, not a merge blocker: the shared event-and-filter mechanism is already exercised by the lot and serial transfer tests, and only the package requirement flag differs.

</details>

<details>
<summary><strong>Prior round status</strong></summary>

Round 1 raised no suggestions, so there is nothing to mark addressed or disputed. S1 is new in this round and applies to the package change added after round 1.

</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=10442 round=2 by=alexei-dobriansky at=2026-08-22T12:06:29.783Z lastSha=f5eb85c4392089c032d49d963fe7c9d3d8bf7799 reviewKey=d4df77c10a3c602dc0b351486f2a5818ae7dfed15a4eb8f8e994e46a4d8e72a7 suggestions=S1@3deffd4d parentRound=1

v-ankitgoyal (approved):

src/Layers/W1/BaseApp/Inventory/Tracking/PackageManagement.Codeunit.al
```diff
@@ -2051,6 +2051,12 @@ codeunit 6516 "Package Management"
             OldItemLedgEntry.TestField("Package No.", ItemLedgEntry."Package No.");
     end;
 
+    [EventSubscriber(ObjectType::Codeunit, Codeunit::"Item Jnl.-Post Line", 'OnAfterSetTrackingFilterFromItemTrackingSetupIfRequired', '', false, false)]
+    local procedure ItemJnlPostLineOnAfterSetTrackingFilterFromItemTrackingSetupIfRequired(var ItemTrackingSetup: Record "Item Tracking Setup"; GlobalItemTrackingSetup: Record "Item Tracking Setup")
+    begin
+        ItemTrackingSetup."Package No. Required" := GlobalItemTrackingSetup."Package No. Required";
```
PredragMaricic, line 2057: **S1 - Add package-transfer regression coverage (Medium)**

This subscriber extends the fix beyond the reported lot scenario so transfer application also filters by package number, but the added tests only exercise lot tracking. Please add a package-tracked transfer test with two package entries and verify that posting consumes the selected package while leaving the other package's remaining quantity unchanged. The implementation looks correct, so I consider this non-blocking, but the new package-specific path should have direct regression coverage.

PredragMaricic (approved):
The transfer application fix is correct and appropriately scoped. Transfer entries now inherit the active serial/lot requirements before item-ledger application filtering, preserving the manually selected lot, and the W1/CZ regression tests cover the reported warehouse-shipment and receipt path. The complete CI matrix is green.

**S1 - Package-transfer regression coverage (Medium, non-blocking):** The follow-up package subscriber extends the behavior to package tracking without a direct package-transfer test. I left an inline suggestion requesting coverage with two packages and verification of remaining quantity by package.

**S2 - Event naming (Low, non-blocking):** `OnAfterSetTrackingFilterFromItemTrackingSetupIfRequired` fires before the filter is applied and actually extends preparation of `ItemTrackingSetup`. Consider renaming it to describe that position, such as `OnAfterSetTrackingSetupForTransfer`, so subscribers do not infer that filtering has already occurred.

v-praghav:
Duplicate PR , Changes already queued to Merge in MAIN

PR #10443, opened 2026-08-20
[Subscription Billing] Sales-Explode BOM fails for foreign-currency

### LINKED ISSUE #10441: [Bug]: [W1] [Subscription Billing] Sales-Explode BOM fails for foreign-currency customers when Sales Subscription Line reads uninserted Sales Line
### Describe the issue

Exploding a BOM (Codeunit `"Sales-Explode BOM"`) on a sales line for a customer that invoices in a foreign currency fails with a Sales Line `"does not exist"` runtime error.

Table 8068 `"Sales Subscription Line"` caches the originating sales line in the global variable `SalesLine` (set via `SetSalesLine` before the sales line itself is inserted, during BOM explosion in Codeunit 8069 `"Sales Subscription Line Mgmt."`). However, local procedure `GetDate(): Date` performed a hard `SalesLine.Get(Rec."Document Type", Rec."Document No.", Rec."Document Line No.")` instead of reusing the existing `GetSalesLine(var SalesLine2: Record "Sales Line")` caching helper (which falls back to `Get()` only when the cached line doesn't match).

`GetDate()` is invoked from the `Unit Cost (LCY)` field's `OnValidate` trigger, which fires while the sales line still only exists in memory (not yet inserted). For FCY customers this validation path is triggered, so the hard `Get()` fails. Non-FCY customers don't hit this code path, which is why the defect only reproduces with a foreign currency.

### Expected behavior

BOM explosion on a sales line for a foreign-currency customer should succeed and create the corresponding `"Sales Subscription Line"` records without errors, the same as it does for LCY customers.

### Steps to reproduce

1. Create a customer with a foreign currency (`Currency Code <> ''`).
2. Create an item with `Replenishment System = Assembly (BOM)`, where the BOM component has a Subscription Package/Service Commitment assigned.
3. Create a sales order line for that customer/item.
4. Run `Codeunit::"Sales-Explode BOM"` on the sales line (for example, via the `Explode BOM` action, or programmatically).
5. Observe the error: `Sales Line does not exist`.

### Additional context

Root cause: Table 8068 `"Sales Subscription Line"` → local procedure `GetDate(): Date` called `SalesLine.Get(Rec."Document Type", Rec."Document No.", Rec."Document Line No.")` directly instead of `GetSalesLine(SalesLine)`.

Fix: replace the direct `Get()` call in `GetDate()` with `GetSalesLine(SalesLine)`, which reuses the in-memory cached sales line set via `SetSalesLine` during BOM explosion when it matches, and only falls back to `Get()` otherwise. This is consistent with every other call site in this table (`CalculateUnitCost`, `CalculateServiceAmount`, `CalculateCalculationBaseAmount`, `CreateVATAmountLineForSalesSubscriptionLine`).

pri-kise:
## What & why
Bugfix for a Sales Document when you explode the BOM of an Item when a BOM Component has subscription package lines.

## Linked work

Fixes #10441

## How I validated this

I fixed this test-driven.

## Risk & compatibility
Bugfix with test.

### DIFF
--- a/src/Apps/W1/Subscription Billing/App/Sales Service Commitments/Tables/SalesSubscriptionLine.Table.al
+++ b/src/Apps/W1/Subscription Billing/App/Sales Service Commitments/Tables/SalesSubscriptionLine.Table.al
@@ -818,7 +818,7 @@ table 8068 "Sales Subscription Line"
 
     local procedure GetDate(): Date
     begin
-        SalesLine.Get(Rec."Document Type", Rec."Document No.", Rec."Document Line No.");
+        GetSalesLine(SalesLine);
         exit(SalesLine.GetDate());
     end;
 
--- a/src/Apps/W1/Subscription Billing/Test/Service Commitments/SalesServiceCommitmentTest.Codeunit.al
+++ b/src/Apps/W1/Subscription Billing/Test/Service Commitments/SalesServiceCommitmentTest.Codeunit.al
@@ -1846,6 +1846,33 @@ codeunit 139915 "Sales Service Commitment Test"
     end;
 
     [Test]
+    [HandlerFunctions('StrMenuHandler')]
+    procedure TestTransferSalesServiceCommitmentsOnExplodeBOMInForeignCurrency()
+    var
+        Item2: Record Item;
+    begin
+        // [SCENARIO] Exploding a BOM on a sales line for a customer invoicing in a foreign currency must not
+        // fail with "'Sales Line' does not exist", because "Sales Subscription Line".GetDate() must reuse the
+        // cached sales line instead of a hard Get() on a sales line that has not been inserted yet.
+        Initialize();
+        LibraryAssembly.CreateItem(Item2, Item."Costing Method"::Standard, Item."Replenishment System"::Assembly, '', '');
+        CreateComponentItemWithSalesServiceCommitments(Item2."No.");
+
+        ContractTestLibrary.CreateCustomer(Customer, LibraryERM.CreateCurrencyWithRandomExchRates());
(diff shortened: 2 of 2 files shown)


alexei-dobriansky:

# **Recommendation: Accept**

<details>

This PR fixes BOM explosion for a foreign-currency sales customer when a BOM component creates a Sales Subscription Line. GetDate() now uses the same cached Sales Line helper as the other calculations instead of doing a hard Get().

The change matches the root cause from the linked issue. The FCY path in Unit Cost (LCY) still uses the same exchange-rate formula and unit-amount rounding precision; it only gets the date from the cached Sales Line when the line is not safely readable through Get(). The fallback to Get() remains for normal persisted-line cases, and the new test covers the reported BOM explosion path.
</details>

<details>

None.
</details>

<details>

**Risk:** This is a financial/currency path because Unit Cost (LCY) can recalculate Unit Cost for FCY sales documents. The regression surface is narrow because the conversion formula, currency factor, and rounding precision are unchanged. BaseApp's Sales-Explode BOM publisher was checked; the subscriber event still exists and this PR does not change that event surface.

**Necessity:** The linked bug is clear and important: FCY customers can hit a runtime error when exploding a BOM with subscription package lines. The fix is small and uses the table's existing helper pattern, so the scope is right.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=10443 round=1 by=alexei-dobriansky at=2026-08-24T01:15:53.2048369Z lastSha=02b8a7b1a4fb4b5a6769b7c225225ac58badd319 reviewKey=ca8fc7b8bf7c621d121645099f5578cbbf3a3edba1a9abc43139903fb31efbe7 suggestions=

PR #10451, opened 2026-08-20
[Bug 647233] Fix User Confirmed copied into Canceled when posting expense report lines

v-rohangarg20:
## Problem
Expense Report Line field 92 is User Confirmed (Boolean), while Posted Expense Report Line field 92 is Canceled (Boolean).

Posting uses PostedExpenseReportLine.TransferFields(ExpenseReportLine). TransferFields maps by field number and type, so a confirmed source line silently sets the posted line's Canceled field to true.

## Fix
- Obsoleted field 92 (Canceled) on Posted Expense Report Line with ObsoleteState = Pending, wrapped in #if not CLEAN29
- Added field 94 (Is Canceled) as the non-colliding replacement
- Updated Cancel Posted Expense Report codeunit to use Is Canceled
- Added upgrade logic in UpgradeExpenseAgentSetup to migrate existing data (old field 92 to new field 999)
- Added regression test verifying posting with User Confirmed = true does not set Is Canceled on the posted line

## Testing
- Regression test: PostingDoesNotCopyUserConfirmedIntoCanceled
- Existing test updated: CancelMarksPostedExpenseReportAsCanceled now uses Is Canceled

[AB#647233](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/647233)


BardurKnudsen (dismissed):

alexei-dobriansky:

<details>

This PR moves the posted expense report line cancel flag away from field 92, where it collided with Expense Report Line.User Confirmed during TransferFields. The internal cancel flow now reads and writes Is Canceled, and the new posting test covers the main regression. The fix addresses the root cause for new postings because field 94 has no matching source field, and the upgrade code protects already-canceled posted reports by copying the old value only when the header was canceled.
</details>

<details>

**S1** - *Make the obsolete caption distinct*
Both the obsolete Canceled field and the new Is Canceled field have the caption Canceled. Give the obsolete field a clear caption such as Canceled (Obsolete), so field pickers and metadata-driven tools do not show two identical columns.

**S2** - *Cover the upgrade migration path*
Add an upgrade test for an existing canceled posted report line. The new posting test covers future records, but it does not prove that old Canceled = true data is copied into Is Canceled during upgrade.
</details>

<details>

**Risk:** The changed code is in Expense Agent posting, cancellation, and upgrade data migration. The main regression risk is existing posted expense report data during upgrade and tools or extensions that can still see the obsolete table field. No BaseApp event publisher or Sales/Purchase/Service analog is involved in this PR.

**Necessity:** The change is needed because TransferFields can copy User Confirmed into an unrelated posted-line cancel flag. The scope is mostly right: it adds a replacement field, changes the cancellation code to use it, migrates old canceled data, and adds a focused regression test.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=10451 round=1 by=alexei-dobriansky at=2026-08-24T07:13:23Z lastSha=9fd3670d0255b763f9c5d0d8667751fbf589e575 reviewKey=3e132ef008cbfc1fa75b2817c0358354a84a97d26a768b4dd59efc004213222d suggestions=S1@dd9278ee,S2@93183771

src/Apps/W1/ExpenseAgent/app/src/ExpenseReport/Tables/PostedExpenseReportLine.Table.al
```diff
@@ -428,6 +428,16 @@ table 6916 "Posted Expense Report Line"
         {
             Caption = 'Canceled';
             Editable = false;
+#if not CLEAN29
+            ObsoleteState = Pending;
+            ObsoleteReason = 'Replaced by field "Is Canceled" to resolve TransferFields collision with Expense Report Line field 92 (User Confirmed).';
+            ObsoleteTag = '29.0';
+#endif
+        }
+        field(94; "Is Canceled"; Boolean)
```
encimita: Minor, nit:

This field is relevant only for the posted version. Should we move it to a higher field number, like 999 or something? We do a similar thing for field 1000 which is the job ledger entry.

I'm just thinking:
- The unposted version has already most of the field IDs before 90 taken
- AI adds fields in order, so fills up all ids in the unposted version without noticing the posted version/transferfields
- In a few months, we find ourselves with field 94 in the un-posted version, which will transfer into the posted version, and we will have a similar bug (which is not covered by your test)

It can still of course happen with field 999 or similar, but it's less likely IMHO.

encimita (dismissed):

src/Apps/W1/ExpenseAgent/app/src/ExpenseReport/Tables/PostedExpenseReportLine.Table.al
v-rohangarg20: Resolved. Thanks!

encimita (approved):

djukicmilica (approved):

PR #10405, opened 2026-08-19
[Extensibility requests] New template, workflows and README update

AleksandricMarko:
## What & why

New template for extensibility requests
extensibility requests included in readme file  

Adds two BCApps dispatcher workflows:
- Runs extensibility request triage four times daily, processing eligible issues independently in parallel.
- Records /not-accurate feedback through the private BCAppsTriage telemetry workflow.

## Linked work

Fixes [AB#646910](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/646910)

PR #10485, opened 2026-08-21
Add extensibility request implementation workflow

AleksandricMarko:
## Summary

- add the GitHub Actions workflow for implementing labeled extensibility requests
- add the `bc-extrequest-implement` skill with its extension and code-surgery guidelines
- add PowerShell orchestration and Kusto telemetry scripts

## Repository configuration

The workflow expects these Actions variables:

- `EXT_REQ_AZURE_CLIENT_ID`
- `EXT_REQ_AZURE_TENANT_ID`
- `EXT_REQ_AZURE_SUBSCRIPTION_ID`
- `EXT_REQ_KUSTO_CLUSTER_URI`
- `EXT_REQ_KUSTO_DATABASE`
- `EXT_REQ_KUSTO_IMPLEMENT_TABLE`

Fixes [AB#646910](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/646910)


.github/scripts/bc-extrequest-implement/invoke-implementation.ps1
```diff
@@ -0,0 +1,118 @@
...
+    [long] $IssueNumber,
+
+    [Parameter(Mandatory = $true)]
+    [string] $IssueUrl,
+
+    [Parameter(Mandatory = $true)]
+    [string] $Repository,
+
+    [string] $Model = 'gpt-5.6-sol',
+    [bool] $TelemetryEnabled = $false
```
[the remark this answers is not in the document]
AleksandricMarko: Fixed in d70ee63092. `TelemetryEnabled` is now used directly when constructing the telemetry context, and the nested telemetry function reads the context value.

.github/scripts/bc-extrequest-implement/invoke-implementation.ps1
```diff
@@ -0,0 +1,118 @@
+[CmdletBinding()]
+param(
+    [Parameter(Mandatory = $true)]
+    [long] $IssueNumber,
+
+    [Parameter(Mandatory = $true)]
+    [string] $IssueUrl,
```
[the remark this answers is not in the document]
AleksandricMarko: Fixed in d70ee63092. `IssueUrl` is now used directly when constructing the telemetry context, which is then consumed by `Send-ImplementationTelemetry`.

.github/workflows/bc-extrequest-implement.yml
```diff
@@ -0,0 +1,179 @@
...
+          "enabled=$($enabled.ToString().ToLowerInvariant())" | Out-File -FilePath $env:GITHUB_OUTPUT -Append
+          if (-not $enabled) {
+            Write-Warning 'Telemetry configuration is incomplete. The implementation will continue without telemetry.'
+          }
+
+      - name: Azure login for telemetry
+        id: azure-login
+        if: steps.telemetry-config.outputs.enabled == 'true'
+        continue-on-error: true
+        uses: azure/login@v3
```
aholstrup1: We should be using pinned actions everywhere (i.e. pinned to a specific commit sha)

.github/workflows/bc-extrequest-implement.yml
```diff
@@ -0,0 +1,179 @@
...
+          node-version: 24
+
+      - name: Install Copilot CLI
+        run: npm install --global @github/copilot
+
+      - name: Implement extensibility request
+        env:
+          COPILOT_GITHUB_TOKEN: ${{ github.token }}
+          GH_TOKEN: ${{ secrets.EXT_REQUEST_PR_PAT }}
+          GITHUB_TOKEN: ${{ secrets.EXT_REQUEST_PR_PAT }}
```
aholstrup1: Why is there both a GH_Token and a GITHUB_Token? GH_Token should be enough I guess

.github/workflows/bc-extrequest-implement.yml
```diff
@@ -0,0 +1,179 @@
...
+        with:
+          node-version: 24
+
+      - name: Install Copilot CLI
+        run: npm install --global @github/copilot
+
+      - name: Implement extensibility request
+        env:
+          COPILOT_GITHUB_TOKEN: ${{ github.token }}
+          GH_TOKEN: ${{ secrets.EXT_REQUEST_PR_PAT }}
```
aholstrup1: We shouldn't be using PATs. We need to use a GitHub app

.github/scripts/bc-extrequest-implement/invoke-implementation.ps1
```diff
@@ -0,0 +1,122 @@
...
+
+    [string] $Model = 'gpt-5.6-sol',
+    [bool] $TelemetryEnabled = $false
+)
+
+$ErrorActionPreference = 'Stop'
+Set-StrictMode -Version 3.0
+
+$runId = "$env:GITHUB_RUN_ID-$env:GITHUB_RUN_ATTEMPT"
+$branch = "bc-extrequest-implement/ext_issue-$IssueNumber"
```
aholstrup1, line 20: I'm still a little bit concerned about this tbh. One PR per event request might end up stealing a lot of resources away from other jobs in the repo which means slower builds for everyone 🤷‍♂️

.github/workflows/bc-extrequest-implement.yml
AleksandricMarko: Fixed in f3af3f0192. Every action in this workflow is now pinned to a full commit SHA, with the corresponding version retained as an inline comment.

.github/workflows/bc-extrequest-implement.yml
AleksandricMarko: Fixed in f3af3f0192. The duplicate `GITHUB_TOKEN` variable was removed. `GH_TOKEN` now carries the GitHub App token for GitHub CLI operations, while `COPILOT_GITHUB_TOKEN` remains separate for Copilot CLI access.

.github/workflows/bc-extrequest-implement.yml
AleksandricMarko: Fixed in f3af3f0192. The PAT validation and usage were removed. The workflow now mints a scoped GitHub App token with the same `actions/create-github-app-token` approach used in #10405, and uses that token for checkout, pushes, and PR operations.

.github/scripts/bc-extrequest-implement/invoke-implementation.ps1
AleksandricMarko: Addressed in f3af3f0192 by reducing the maximum implementation runtime from 120 to 60 minutes. We intentionally kept the label-triggered runs immediate rather than moving them to a scheduled serial queue, accepting that separate eligible requests can still run concurrently.

aholstrup1: Yeeeah no.. That's not what I meant. I'm not concerned about the time it takes to run the triage. I'm concerned with the amount of compute we'll need to process each event request individually. 

Could we batch event requests together? Could we have some MAX amount of open event-request PR limit?

AleksandricMarko: Sorry, this was AI respond. 
I would preferer to start with single issue for week or two and i will run agent after 4 PM
It will be easier for me to track execution and accuracy (also I am not expecting that many issues in first few days)
After that we can go with batches

aholstrup1 (approved):

JesperSchulz (approved):

PR #10493, opened 2026-08-21
[Master]-The Available to Take logic is incorrect or inaccurate in the Pick Worksheet as it includes Ship and Receive bin inventory

neeleshsinghal:


alexei-dobriansky:

<details>

This PR fixes the "Available Qty. to Take" calculation on the Pick Worksheet so it no longer counts stock sitting in the location's configured receipt and shipment bins (AB#647203). In `Warehouse Availability Mgt.CalcQtyAvailToTakeOnWhseWorksheetLine`, it skips the receipt bin (when "Require Put-away" and a receipt bin is set) and the shipment bin (when "Require Pick" and a shipment bin is set) while summing bin content. It also stops subtracting picked-not-shipped quantity when the shipment bin is excluded (to avoid removing the same units twice), and caps the result with `Minimum(AvailQtyBase, CalcInvtAvailQty(...))`. Six tests are added.

The change addresses the root cause: stock in the receipt bin is not yet put away, and stock in the shipment bin has effectively been picked, so neither should be reported as available to pick. The `<> ''` guards mean directed put-away and pick locations (which early-exit anyway) and locations without configured special bins are unaffected. The tests cover the receipt-bin, shipment-bin, not-configured, double-removal (result is 0 not negative), and partial put-away cases, which matches the risky paths.
</details>

<details>

**S1** - *Minimum cap applies to all bin-mandatory basic locations*
The new `Minimum(AvailQtyBase, CalcInvtAvailQty(...))` cap runs for every bin-mandatory (non-directed) location, not only ones with a receipt or shipment bin. Please confirm it never lowers "Available to Take" in existing setups where bin content can legitimately differ from `CalcInvtAvailQty` (for example due to how reservations are counted), so this does not become an over-eager reduction.

**S2** - *First new test has no scenario comment*
`AvailableToPickExcludesConfiguredReceiptBinInBasicWarehouse` has no `[FEATURE]`/`[SCENARIO]` comment block, while the other five new tests do. Add one so its intent and the bug id are documented like the others.
</details>

<details>

**Risk:** Medium. Availability drives how much the pick worksheet offers to pick, so an incorrect value causes over- or under-picking, which is a data-integrity concern. The exclusion is well guarded and directed warehouses early-exit, so the blast radius is basic (bin-mandatory) warehouses that configure receipt/shipment bins. The added `Minimum` cap and the "skip picked-not-shipped when shipment bin excluded" branch are the widest behavioral changes; the tests guard the double-subtraction case, but the cap affects a broader set of locations than the reported bug (see S1).

**Necessity:** Justified. Counting receipt-bin and shipment-bin stock as available to pick lets the worksheet suggest picking units that are not actually pickable, which is a real workflow defect. The fix is targeted and stays within bug-fix scope, and it ships with focused regression tests.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=10493 round=1 by=alexei-dobriansky at=2026-08-23T00:10:47Z lastSha=ee329b1b895af320c647062a1a73f7a5d64fb129 reviewKey=3f25aa6d0b407bb8a0b6c991bfa6ee9ad0bb499c009d90528ef33df676c59836 suggestions=S1@6d41e34e,S2@66e562ca

PR #10499, opened 2026-08-21
[Bug Fix] #210: Post item charge to Value Entry when receipt and charge invoice are posted separately

ventselartur:
## Bug Reference
Work item microsoft/BCAppsBugFix#210 — Fixes microsoft/BCAppsBugFix#210

## Summary
When a purchase order has an item line and an item-charge line assigned to that item line, and the **receipt** and the **item-charge invoice** are posted in **separate** steps (receipt first, then the charge invoice with the item line's `Qty. to Invoice = 0`), the item charge cost was never transferred to a Value Entry. This left `Qty. Assigned` at `0` while `Quantity Invoiced` was `1`, so the purchase order could never be deleted ("Qty. Assigned must be equal to '1' in Purchase Line..."). The fix makes the standalone item-charge invoice post the charge to a Value Entry against the already-posted receipt's item ledger entry.

## Root Cause
In `Codeunit 90 "Purch.-Post"` an item charge assigned to an item line of the same order was only transferred to a Value Entry as a side-effect of posting that item line's item journal line during an Invoice run — `PostItemJnlLine` calls `PostItemJnlLineItemCharges` (→ `PostItemChargePerOrder`) only when `(PurchLine.Type = Item) and PurchHeader.Invoice`. When the receipt is posted first (Invoice = false) and the charge is invoiced separately with the assigned item line's `Qty. to Invoice = 0`, the item line's `PostItemJnlLine` never runs, so no Value Entry is created. Posting the charge line runs `PostItemChargeLine`, but for an `Applies-to Doc. Type::Order`/`Invoice` assignment it only called `CheckItemCharge` (validation, no posting). The charge line therefore became `Quantity Invoiced = 1` while `Qty. Assigned` stayed `0`, and `PurchaseLine.DeleteChargeChargeAssgnt` (`TestField("Qty. Assigned","Quantity Invoiced")`) blocked deletion. Posting the receipt and charge invoice together worked because the item line was received in the same run, so the charge Value Entry was created.

## Changes Made
- `src/Layers/W1/BaseApp/Purchases/Posting/PurchPost.Codeunit.al`: Extended the `Order`/`Invoice`/`Return Order`/`Credit Memo` case in `PostItemChargeLine` to also call a new local procedure `PostItemChargeFromReceivedOrderLine`. When the assigned item line was already received (`"Qty. Rcd. Not Invoiced (Base)" <> 0`), it locates the corresponding `Purch. Rcpt. Line` and posts the charge via the existing receipt distribution helpers (`PostDistributeItemCharge` / `PostItemCharge`) — the same path as `PostItemChargePerRcpt` — creating the Value Entry and correctly updating `Qty. Assigned`. Guarded with `if TempItemChargeAssgntPurch.Mark() then exit(false)` so a charge already handled by the combined receipt+invoice flow is never double-posted. Minimal and additive; the combined-posting path, return/credit-memo semantics, and existing `CheckItemCharge` validations are unchanged.
- Regression test added to `src/Layers/W1/Tests/ERM/O365PurchItemChargeTests.Codeunit.al` (codeunit 135300).

## Implementation Process

- Fix iterations: 1 of 5
- Compilation: ✅ All projects compile successfully
- Publish: ✅ Modified app published successfully
- Validation: ✅ Automated tests passing

#### Pre-Fix Test Results (Baseline)
Tests run before implementing the fix (expected to fail):

```
❌ ItemChargeValueEntryCreatedWhenChargeInvoicedSeparately: FAILED
   Error: Assert.TableIsNotEmpty failed. Table <Value Entry> with filter <Item No.: GL00000001, Item Charge No.: GU00000002, Entry Type: Direct Cost> must contain records.
```

#### Post-Fix Test Results (Final)
Tests run after implementing the fix (expected to pass):

```
✅ OnRun: PASSED
✅ SetupItemChargeTest: PASSED
✅ CreateCorrectiveCreditMemoWithItemChargeTest: PASSED
✅ CreateCorrectiveCreditMemoWithItemChargesTest: PASSED
✅ CreateCorrectiveCreditMemoFromLargeInvoiceWithItemChargeTest: PASSED
✅ CreateCorrectiveCreditMemoWithItemChargeAndCurrencyTest: PASSED
✅ CreateCorrectiveCreditMemoWithoutItemChargeTest: PASSED
✅ CancelInvoiceWithChargeItemAssignedToMultipleShipmentLines: PASSED
✅ TestRemovePostedReceiptWithChargeItemAssigned: PASSED
✅ VariantCodeIsMandatoryDoesntDisruptItemChargePosting: PASSED
✅ VerifyQtyOnItemChargeAssignmentLinesForPurchInvoiceLinesFromPurchRcptsFromPurchOrder: PASSED
✅ ItemChargeValueEntryCreatedWhenChargeInvoicedSeparately: PASSED (new test for bug scenario)
```

**Total Tests Run:** 12
**All Tests Passing:** ✅ Yes

#### Iteration Summary

- **Iteration 1**: ✅ Added `PostItemChargeFromReceivedOrderLine` and extended the `PostItemChargeLine` Order/Invoice case; all tests passing.

## Validation Coverage

- [x] Existing tests pass
- [x] New tests added for bug scenario
- [x] Regression test for work item microsoft/BCAppsBugFix#210
- [x] Automated tests: All passing
- [x] Build and publish validation completed

## Miapp Propagation

- **Layers propagated to**: APAC, BE, CA, CH, ES, FI, GB, IT, NA, RU, US
- **Files changed by propagation**: 11
- **VerifyMiappSync**: ✅ No files still need to be integrated

## Review Notes
Focus on `PostItemChargeFromReceivedOrderLine` and the receipt-line resolution: the charge is posted against the posted `Purch. Rcpt. Line` for the assigned order line, reusing the existing distribution helpers. The double-post guard relies on the `TempItemChargeAssgntPurch.Mark()` flag set by the combined receipt+invoice flow.

🤖 Generated by the bc-fix-bug skill

---
Promoted from microsoft/BCAppsBugFix#266: https://github.com/microsoft/BCAppsBugFix/pull/266

### DIFF
--- a/src/Layers/APAC/BaseApp/Purchases/Posting/PurchPost.Codeunit.al
+++ b/src/Layers/APAC/BaseApp/Purchases/Posting/PurchPost.Codeunit.al
@@ -1475,7 +1475,11 @@ codeunit 90 "Purch.-Post"
                   TempItemChargeAssgntPurch."Applies-to Doc. Type"::Invoice,
                   TempItemChargeAssgntPurch."Applies-to Doc. Type"::"Return Order",
                   TempItemChargeAssgntPurch."Applies-to Doc. Type"::"Credit Memo":
-                        CheckItemCharge(TempItemChargeAssgntPurch);
+                        begin
+                            CheckItemCharge(TempItemChargeAssgntPurch);
+                            if PostItemChargeFromReceivedOrderLine(PurchHeader, PurchaseLineBackup) then
+                                TempItemChargeAssgntPurch.Mark(true);
+                        end;
                 end;
 
                 OnPostItemChargeLineOnAfterPostItemCharge(TempItemChargeAssgntPurch, PurchHeader, PurchaseLineBackup, PurchLine);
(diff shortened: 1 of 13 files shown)


alexei-dobriansky:

<details>

This change fixes a case where an item charge assigned to an item line of a purchase order is never posted to a Value Entry when the receipt and the item-charge invoice are posted in separate steps. It adds a new local procedure `PostItemChargeFromReceivedOrderLine` in `Purch.-Post` (codeunit 90). For an `Order`/`Invoice` charge assignment whose target line is already received, it finds the posted `Purch. Rcpt. Line` and posts the charge through the existing distribution helpers (`PostDistributeItemCharge` / `PostItemCharge`), the same helpers used by `PostItemChargePerRcpt`.

The direction is right and the structure is good: it reuses the receipt distribution path, restricts to `Order`/`Invoice` types (so return/credit-memo sign handling is untouched), and guards against double-posting with `TempItemChargeAssgntPurch.Mark()`. The concern is in how the receipt line is resolved. `PostItemChargePerRcpt` handles one specific receipt line per assignment entry, but the new procedure serves an order-line assignment and picks a single receipt line with `PurchRcptLine.FindFirst()`. When the target order line was received in more than one partial receipt, this does not distribute the charge across all receipts (see S1).
</details>

<details>

**S1** - *Charge posted against only the first receipt line*
`PostItemChargeFromReceivedOrderLine` resolves the receipt with `SetRange("Order No.")` + `SetRange("Order Line No.")` then `FindFirst()`, and posts the full `Qty. to Assign` and `Amount to Assign` against that one line's `Quantity (Base)` and item ledger entry. If the order line was received in two or more partial receipts, only the first receipt is charged and the rest get no charge, so the item cost is spread wrong. Loop over all matching `Purch. Rcpt. Line` records and post the charge per receipt line, like the per-receipt path does.

**S2** - *Test only checks the value entry exists*
The new test posts a single receipt of quantity 1 and then asserts only `Assert.RecordIsNotEmpty(ValueEntry)`. Please assert the posted charge amount and that `Qty. Assigned` now equals `Quantity Invoiced` on the charge line (the original blocking symptom), and add a case with two partial receipts on the same order line to cover the distribution in S1.
</details>

<details>

**Risk:** This is a posting/data-integrity path (item charge to Value Entry, inventory valuation), so an incorrect distribution posts a wrong cost. The single-receipt scenario in the test works, but the multi-receipt case in S1 can misallocate the charge amount and quantity. The change is propagated identically to all ten BaseApp layers (APAC, BE, CH, ES, FI, GB, IT, NA, RU, US, W1 - verified byte-identical), so any correctness gap ships to every layer at once. No public signature or event change; the new `OnPostItemChargeLineOnAfterPostItemCharge` context is unchanged.

**Necessity:** The scenario is valid and important: without the fix the charge line reaches `Quantity Invoiced = 1` while `Qty. Assigned` stays `0`, which blocks deleting the purchase order (`TestField("Qty. Assigned","Quantity Invoiced")`). The fix is worth making; it just needs to cover the partial-receipt case before merge.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=10499 round=1 by=alexei-dobriansky at=2026-08-21T12:06:36Z lastSha=203e4eae7945e47f6ac24e67184dd2c35c85c693 reviewKey=885c9924e84e3bd4f8c4aea25272761903c21c525322a20d61b80cc960ea77a4 suggestions=S1@8682eba8,S2@97cbfae1

alexei-dobriansky (changes requested):
Needs improvement

alexei-dobriansky:
Won't fix.

PR #10510, opened 2026-08-21
Slice 611704: With Autoincrement Tax Transaction Value ID field changed to BigInteger to prevent IDENTITY exhaustion

v-ajaztabrez:
[AB#611704](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/611704)
[Slice 611704](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/611704): [Repair Item][IN] Change "Tax Transaction Value".ID to be BigInteger

**Issue:**
The primary key field ID of table Tax Transaction Value (20261) was Integer with AutoIncrement. On a production environment (IN localization, ~525 GB database) the SQL IDENTITY sequence exceeded the integer limit (2,147,483,647), causing the error: "Arithmetic overflow error converting IDENTITY to data type int." The actual row count (~285M) was well below the limit, but Posting Preview and failed/rolled-back transactions kept consuming AutoIncrement IDs without committing rows, exhausting the sequence far beyond the actual data.

**Cause:**
SQL Server IDENTITY (AutoIncrement) is non-transactional — every insert attempt permanently increments the counter, even when the transaction is rolled back (Posting Preview, validation errors). Heavy use of Posting Preview consumed the integer range over time. The Integer type has a maximum of only 2,147,483,647.

**Solution:**
Changed field ID from Integer to BigInteger, extending the range to 9.2 × 10¹⁸. Introduced the platform NumberSequence API for ID generation while retaining AutoIncrement on the field. Centralized ID generation in the existing helper Transaction Value Helper (20236) as GetNextTransactionValueID() — no new object introduced. The sequence is lazily seeded from the current MAX(ID) on first use (collision-free even on an upgraded database), with a TryFunction + ClearLastError guard for concurrent creation. The table OnInsert trigger assigns the ID from the sequence (skipping temporary records and explicitly-set IDs). Live-table insert call sites (TaxRateComputation, TaxDocumentGLPosting) use Insert(true) so the trigger fires. The temp-table NextID variable in TaxPostingBufferMgmt was changed to BigInteger to match the field type. Temporary table inserts manage their own IDs via simple counters and are unaffected by this change.

### DIFF
--- a/src/Apps/W1/INTaxEngine/app/TaxEngine-PostingHandler/src/PostingManagement/codeunit/TaxDocumentGLPosting.Codeunit.al
+++ b/src/Apps/W1/INTaxEngine/app/TaxEngine-PostingHandler/src/PostingManagement/codeunit/TaxDocumentGLPosting.Codeunit.al
@@ -70,8 +70,7 @@ codeunit 20341 "Tax Document GL Posting"
                 ToTaxTransactionValue.Init();
                 ToTaxTransactionValue := FromTaxTransactionValue;
                 ToTaxTransactionValue."Tax Record ID" := ToRecID;
-                ToTaxTransactionValue.ID := 0;
-                ToTaxTransactionValue.Insert();
+                ToTaxTransactionValue.Insert(true);
             until FromTaxTransactionValue.Next() = 0;
     end;
 
--- a/src/Apps/W1/INTaxEngine/app/TaxEngine-PostingHandler/src/PostingManagement/codeunit/TaxPostingBufferMgmt.Codeunit.al
+++ b/src/Apps/W1/INTaxEngine/app/TaxEngine-PostingHandler/src/PostingManagement/codeunit/TaxPostingBufferMgmt.Codeunit.al
@@ -317,7 +317,7 @@ codeunit 20343 "Tax Posting Buffer Mgmt."
         InvoiceQty: Decimal)
     var
         TaxTransactionValue: Record "Tax Transaction Value";
-        NextID: Integer;
+        NextID: BigInteger;
     begin
         TempTransactionValue.Reset();
         NextID := TempTransactionValue.Count();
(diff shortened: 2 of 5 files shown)


alexei-dobriansky:

<details>

This PR changes `Tax Transaction Value`.`ID` from Integer to BigInteger and adds NumberSequence-based ID assignment when live records are inserted. The direction matches the reported exhaustion problem, and the changed live insert call sites now run table triggers so the ID generation path can execute. The main issue is that the field still has `AutoIncrement = true` while the trigger assigns `ID` manually, so the new sequence path is not a safe replacement for the old SQL identity behavior.
</details>

<details>

**Fit:** Partial

The change targets the right table and widens the key range. It does not yet make primary key ownership unambiguous, so inserts may still depend on the identity field behavior or fail when the trigger writes the ID itself.
</details>

<details>

**S1** (🔴 High): *Do not assign an AutoIncrement field*
The new `OnInsert` trigger sets `ID`, but the field still has `AutoIncrement = true`. This can keep the old SQL identity behavior in control, or write a manual value to an identity field, so the new NumberSequence path is not safe. Remove `AutoIncrement` when the sequence owns the value, and only assign a new ID when `ID = 0`.

**S2** (🟠 Moderate): *Cover the new ID allocation path*
This change moves primary key assignment for a tax posting table to `OnInsert`, but no test covers that path. Add a test that inserts live `Tax Transaction Value` records through the changed callers and verifies unique BigInteger IDs, including when existing rows seed the sequence.
</details>

<details>

**Risk:** This is a tax posting data-integrity path. A wrong ID allocation can stop posting, create duplicate keys, or leave transaction value rows detached from the posted document. The `Sequence No. Mgt.` BigInteger additions follow the existing Integer pattern, but the table field still needs one clear owner for ID assignment.

**Necessity:** The change is necessary because the existing Integer key can be exhausted by rollback-heavy posting and preview flows. The scope is right, but the primary key ownership must be fixed before merge.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10510 round=1 by=alexei-dobriansky at=2026-08-27T19:35:43Z lastSha=93ece28efdb9569ef68345439971de6c6a5bc810 reviewKey=47a8460e655ccb0c89c5f8085125d810d67f033755aae579b799ce418c92ac94 suggestions=S1@2976f554,S2@e4027404

PR #10521, opened 2026-08-21
[Master]-Inconsistent Validation of Non-Deductible VAT % for Purchase Lines Sharing the Same VAT Identifier

neeleshsinghal:


alexei-dobriansky:

<details>

This PR fixes the Non-Deductible VAT purchase-line case where one line can be set to 0 while another line with the same VAT Identifier keeps a non-zero Non-Deductible VAT %. The existing validation already rejects the reverse order, and the new tests cover both entry orders. The rule itself is correct, but the new table-level call is placed outside the existing handled validation path.
</details>

<details>

**S1** - *Keep the check inside the existing validation*
The new table-level call runs after `OnBeforeCheckNonDeductibleVATPctIsAllowed`, so subscribers that set `IsHandled` can still get this new error when the value is 0. Move the zero-value branch into `NonDedVATImpl.CheckNonDeductibleVATPctIsAllowed` after the existing `IsHandled` check, or add an equivalent handled event before this new error. Apply the same shape to the layer copies.
</details>

<details>

**Risk:** This is purchase-line VAT validation, so a wrong change can block purchase documents and API writes for a financial field. The same check was added consistently to W1 and the regional BaseApp Purchase Line copies, and no new BaseApp publisher dependency was added; I verified the existing handled event in `NonDeductibleVAT.Codeunit.al`. CI still had some unit-test jobs running and two integration jobs marked failed without logs available from the in-progress run, so those results should be checked before merge.

**Necessity:** The scenario is valid: without the change, a document can mix 0 and non-zero Non-Deductible VAT % values under the same VAT Identifier depending on line entry order. The scope is targeted and the tests cover both order directions, but the fix should preserve the existing handled validation contract.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=10521 round=1 by=alexei-dobriansky at=2026-08-21T18:11:27Z lastSha=5794e0cf14cf2215060b59024d444a53150b1f16 reviewKey=da273117122186a6f1c90c3e24db6b21fa5624b9f3c8ffe342b8e5fdf2463867 suggestions=S1@44c8aee0

PredragMaricic (changes requested):
Requesting changes for three high-confidence issues:

**S1 - Complete the setup-level fix.** ADO bug #647365 explicitly requires preventing creation of the conflicting VAT Posting Setup configuration. This PR changes only Purchase Line validation. `NonDedVATImpl.CheckVATPostingSetupChangeIsAllowed` still exits for a `Do Not Allow` setup, so the reported non-deductible/normal setup pair with the same VAT Identifier can still be created. Enforce the invariant when the setup is created, as required by the work item.

**S2 - Preserve the handled validation contract.** The new zero-percent check runs directly from table 39 after `NonDeductibleVAT.CheckNonDeductibleVATPctIsAllowed`. A subscriber that handles `OnBeforeCheckNonDeductibleVATPctIsAllowed(..., IsHandled)` can suppress the existing check but still receives this new error. Move the zero-value branch into `NonDedVATImpl.CheckNonDeductibleVATPctIsAllowed` after the same `IsHandled` gate rather than duplicating the query, label, and error across 14 table copies.

**S3 - Resolve the demonstrated test regression.** Twenty-one `LegacyTestsBucket1` jobs fail. The concrete failure is `CombinedVATAmountLineForTwoPurchLineFirstNonDedVATSecondNormalVAT`, which intentionally creates non-deductible and normal VAT lines sharing one VAT Identifier; the new validation now aborts that scenario. Since this work item intentionally changes the invariant, revise the existing W1/AU/IT/RU scenarios to assert the new behavior and add setup-level regression coverage. The PR status check is currently failed for this reason.

This is high-risk financial validation across all BaseApp layers. The two added tests cover both line-entry orders, but the implementation does not complete the linked acceptance scope, bypasses the established extensibility event, and breaks the existing suite.

PR #10524, opened 2026-08-21
Add a purchase order matching API

mynjj:
Adding public codeunits for creating "Matched Order Line" records. The first consumer of this API will be the E-Documents app to allow for creating 1-M matches and that can be received on invoice from the Payables Agent flows.

All the changes are additive and no existing BaseApp flow with the existing logic has been disturbed. We have yet to decide whether to update our own existing `MatchedOrderLineMgmt` with it. In essence, this change is only a facade for inserting "Matched Order Line" records.

The one novel bit, is an internal usage of this new API in the codeunit `PO Matching`, where we add a surface to suggest receipts in a match: given the order and the proposed match, if there are receipts that could cover the remaining amount to invoice, they are suggested in the match. This new procedure is not uptaken internally in the BaseApplication, but it will be used in the E-Document app (see second PR on the stack) as a convenience when receiving invoices from that flow that didn't have receipts specified for matching.

The new public API encapsulates a new match under the `PO Matching Group`, that holds a buffer of the matches to be inserted. At every point in time that a new match is added to the buffer, we run validations to ensure that it makes sense quantity-wise, and the other validations we have (like no prepayment, or partial receipts for item-tracked), additionally we discover the already existing matches that have been persisted in the database and load the whole match as well so that the validations are applied considering those as well. The edges to be loaded are discovered in a BFS fashion, and appended running the validations.

A iterator is implemented for `PO Matching Group` to allow consuming the matches, additionally, another iterator to encapsulate the possible receipts to suggest (`Invoiceable Receipts`) was also added. See how they are used to suggest the receipts on the match.

Fixes [AB#625392](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/625392)


alexei-dobriansky:

<details>

This PR adds a public BaseApp API for building and saving purchase order matching groups. It creates invoice-to-order, order-to-receipt, and invoice-to-receipt edges, validates quantity caps, loads connected persisted matches, and adds tests for the main happy paths and several error paths.

The feature is coherent, and the tests cover many important flows. However, the public API trusts that the passed SystemId values belong to the expected purchase document types and unit of measure shape. The posting code later treats these rows as invoice/order/receipt matches, so these validations need to be strict before rows are persisted.
</details>

<details>

**S1** - *Validate the purchase document types*
Add checks that Match."Document Line SystemId" points to a purchase invoice line and that Match."Matched Order Line SystemId" points to a purchase order line. Today any purchase line with matching item, vendor, and currency can be saved as a match, which can leave persisted rows that the posting code later reads with the wrong document role.

**S2** - *Reject mismatched units of measure*
Also compare the invoice and order line unit of measure before accepting an invoice-order edge. The current check only compares type and item number, so an invoice in one UOM can be matched to an order or receipt in another UOM and still pass the non-base quantity checks used during posting.
</details>

<details>

**Risk:** This is new public BaseApp surface in POMatchingGroup.Codeunit.al and it writes Matched Order Line records that are later used by purchase posting. The regression surface is narrow until consumers call the API, but a bad persisted match can affect invoice-to-order and order-to-receipt allocation quantities.

**Necessity:** The API is useful for E-Documents and similar consumers because it hides the matching persistence details. The scope is right for a feature slice, but the API boundary must reject wrong document roles and UOM mismatches before it becomes a public integration point.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=10524 round=1 by=alexei-dobriansky at=2026-08-21T18:10:38Z lastSha=7d58e292bf50ef4481d9c46ebb3f146153f793ff reviewKey=eb3802795d92a26d9d79f10b35aa83e030fcea931112e4fcbea4f3aef20431c8 suggestions=S1@419220ac,S2@b98fb3b7

alexei-dobriansky:

# **Recommendation: Accept**

<details>

This PR adds a public BaseApp API for building and saving purchase order matching groups (invoice-to-order, order-to-receipt, and invoice-to-receipt edges), with quantity-cap validation and tests.

The commits since round 1 harden the API boundary. Both invoice-order and order-receipt paths now run through one shared `ValidateInvoiceOrderCompatibility` procedure, so the document roles and unit of measure are checked before any row is persisted. The `Document Line SystemId` used on the order-receipt path is always resolved (passed in, or inferred via `InferInvoiceForOrder`, which errors when it cannot find exactly one invoice), so requiring a valid invoice line there does not break the order-receipt flow. The changes are correct and well targeted.
</details>

<details>

| ID  | Title                              | Status    | Author response |
|-----|------------------------------------|-----------|-----------------|
| S1  | Validate the purchase document types | Addressed | `ValidateInvoiceOrderCompatibility` now checks `InvoiceLine."Document Type"::Invoice` and `OrderLine."Document Type"::Order`; tests assert "must be an invoice line" / "must be an order line". |
| S2  | Reject mismatched units of measure | Addressed | Same procedure now compares `"Unit of Measure Code"` between invoice and order; two tests cover the UOM mismatch on the invoice-order and order-receipt paths. |
</details>

<details>

*None - the changes only addressed the prior suggestions (shared validation extracted, telemetry event id filled in, unused `PO Matching Posting Action` table removed, variable ordering fixed).*
</details>

<details>

**Risk:** This is new public BaseApp surface in `POMatchingGroup.Codeunit.al` that persists `Matched Order Line` rows later consumed by purchase posting. The regression surface stays narrow (no consumers yet), and the round-2 changes only tighten validation before rows are written, so they reduce risk rather than add it. The new invoice-line lookup on the order-receipt path is safe because `Document Line SystemId` is always resolved before validation.

**Necessity:** Unchanged from round 1. The API is a useful boundary for E-Documents and similar consumers, and the added document-role and UOM checks are exactly what a public integration point needs before it can be trusted by the posting code.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=10524 round=2 by=alexei-dobriansky at=2026-08-23T12:09:34Z lastSha=31940f62dbcd3c2e9c5ef211bc578535bc7de235 reviewKey=6643cf52c0dde8075ccfd97d68d30b0a93efa6e6e63eee3631ff201d6d28f996 suggestions=S1@419220ac:addressed,S2@b98fb3b7:addressed parentRound=1

alexei-dobriansky (dismissed):

alexei-dobriansky:

# **Recommendation: Accept**

<details>

This PR adds a public API for building purchase order matching groups and adds receipt suggestions for invoice/order matches. The commit since round 2 simplifies the edge factories by removing caller-supplied base quantities, so PO Matching Group derives Qty. to Invoice (Base) from the order line before validation and persistence.

The round-3 changes keep the earlier document type and unit of measure checks in place. Deriving the base quantity inside the group is safer than trusting external callers, and the updated tests still cover base-quantity derivation, quantity caps, receipt splits, and error paths.
</details>

<details>

| S1 | Validate the purchase document types | Addressed | Still addressed. The shared compatibility check still requires invoice lines and order lines to have the right document types. |
| S2 | Reject mismatched units of measure | Addressed | Still addressed. The same shared check still requires invoice and order lines to use the same unit of measure before base quantity is derived. |
</details>

<details>

*None - the new commit only simplifies the edge factory signatures, derives base quantity inside PO Matching Group, makes the receipt iterator internal, and updates the tests to use the new API shape.*
</details>

<details>

**Risk:** The main risk is the new public PO Matching API because it persists Matched Order Line records that purchase posting can later use. The latest commit reduces the API surface by removing a base-quantity parameter, and this is safe before the new API has merged. CI has passed many build checks; some unit-test buckets were still running when reviewed.

**Necessity:** The change is needed for the Payables Agent PO matching slice. The latest simplification is also justified because callers should not have to provide a base quantity that the group can derive and validate from the order line.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=10524 round=3 by=alexei-dobriansky at=2026-08-24T12:24:40Z lastSha=64e07b4e0a8a43bffbce40440b70cf8f2d6865fd reviewKey=81fe3384c0e6fad1750c4b44816bfef74ca72ebf89088d89d41ccc700a77266b suggestions=S1@419220ac:addressed,S2@b98fb3b7:addressed parentRound=2

alexei-dobriansky (dismissed):

PredragMaricic (changes requested):
Requesting changes for two data-integrity issues in the new matching API:

1. `ValidateOrderReceipt` accepts negative allocation quantities because every cap check is upper-bound-only. A negative receipt edge is persisted and reduces the aggregate pinned/consumed quantities, allowing a later positive edge to exceed the invoice-order budget or receipt availability while the signed total still matches the invoice quantity at posting. Reject wrong-sign/non-positive allocations and cover this case with regression tests.

2. `SaveMatchingGroups` revalidates an in-memory component loaded earlier by `AddMatch`, then persists without locking and reloading the shared component. Concurrent sessions matching different invoices to the same order or receipt can both validate against the same remaining quantity and insert distinct rows, over-allocating it. Lock the shared table, reload/merge the latest persisted component under that lock, then revalidate and persist atomically.

The functional test coverage is otherwise strong, and the reported checks are green, but these races/validation gaps can produce invalid matching data.

src/Layers/W1/BaseApp/Inventory/Tracking/InvoiceableReceipts.Codeunit.al
```diff
@@ -0,0 +1,57 @@
...
+    Access = Internal;
+
+    // A cursor over the posted receipt lines of a single order line that still have quantity received not
+    // invoiced. It nets a receipt's remaining quantity against a "PO Matching Group" so consumers see the
+    // capacity still available for covering, accounting for edges the group already holds.
+    var
+        TempPurchRcptLine: Record "Purch. Rcpt. Line" temporary;
+
+    /// <summary>Loads the invoiceable receipts of the order line and positions at the first, returning false if none.</summary>
+    internal procedure Load(OrderLine: Record "Purchase Line"): Boolean
```
PeterConijn, line 22: The scope for the codeunit is _internal_, so having separate access modifiers on internal procedures is redundant.

src/Layers/W1/BaseApp/Inventory/Tracking/InvoiceableReceipts.Codeunit.al
```diff
@@ -0,0 +1,57 @@
...
+    var
+        PurchRcptLine: Record "Purch. Rcpt. Line";
+    begin
+        TempPurchRcptLine.Reset();
+        TempPurchRcptLine.DeleteAll();
+
+        PurchRcptLine.SetRange("Order No.", OrderLine."Document No.");
```
PeterConijn, line 30: Key2 might be better for this combination of filters.

src/Layers/W1/BaseApp/Inventory/Tracking/InvoiceableReceipts.Codeunit.al
```diff
@@ -0,0 +1,57 @@
...
+{
+    Access = Internal;
+
+    // A cursor over the posted receipt lines of a single order line that still have quantity received not
+    // invoiced. It nets a receipt's remaining quantity against a "PO Matching Group" so consumers see the
+    // capacity still available for covering, accounting for edges the group already holds.
+    var
+        TempPurchRcptLine: Record "Purch. Rcpt. Line" temporary;
+
+    /// <summary>Loads the invoiceable receipts of the order line and positions at the first, returning false if none.</summary>
```
PeterConijn, line 20: XmlDocumentation in this codeunit is missing elements for parameters and return values.

src/Layers/W1/BaseApp/Inventory/Tracking/POMatching.Codeunit.al
```diff
@@ -0,0 +1,130 @@
...
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See License.txt in the project root for license information.
+// ------------------------------------------------------------------------------------------------
+namespace Microsoft.Inventory.Tracking;
+
+using Microsoft.Inventory.Ledger;
+using Microsoft.Purchases.Document;
+using Microsoft.Purchases.History;
+
+codeunit 5831 "PO Matching"
```
PeterConijn, line 11: Does this not need to follow a façade pattern?

src/Layers/W1/BaseApp/Inventory/Tracking/POMatching.Codeunit.al
```diff
@@ -0,0 +1,130 @@
...
+    end;
+    #endregion
+
+    #region Covering receipts
+    /// <summary>
+    /// Enriches the group by covering each invoice-order allocation with the order line's posted receipts that still have quantity received not invoiced, by adding or growing receipt edges.
+    /// </summary>
+    procedure SuggestCoveringReceipts(var POMatchingGroup: Codeunit "PO Matching Group")
+    var
+        TempIntendedEdge: Record "Matched Order Line" temporary;
```
PeterConijn: Not 100% sure about this one, but I do think it is customary to place _temporary_ variables under the normal ones.

src/Layers/W1/BaseApp/Inventory/Tracking/POMatching.Codeunit.al
```diff
@@ -0,0 +1,130 @@
...
+
+codeunit 5831 "PO Matching"
+{
+    Access = Public;
+
+    #region Edge factory
+    // Edges are unpersisted "Matched Order Line" records
+
+    /// <summary>Builds an invoice-to-order edge.</summary>
+    procedure InvoiceOrderEdge(InvoiceLineSystemId: Guid; OrderLineSystemId: Guid; QtyToAllocate: Decimal) Edge: Record "Matched Order Line"
```
PeterConijn, line 23: XmlDocumentation missing parameters and return values

PeterConijn (changes requested):

src/Layers/W1/BaseApp/Inventory/Tracking/InvoiceableReceipts.Codeunit.al
mynjj: It is redundant, but if there is a latter effort for making this codeunit public, I'd rather it would be then an explicit choice at that time to select which procedures should be public. 

If I already don't add the modifier, it may hide that decision later on.

src/Layers/W1/BaseApp/Inventory/Tracking/InvoiceableReceipts.Codeunit.al
mynjj: might end up being a table scan anyway because of the filter on `Qty Rcd. Not Invoiced` but I can be explicit anyway, I'll add it

src/Layers/W1/BaseApp/Inventory/Tracking/InvoiceableReceipts.Codeunit.al
mynjj: I often don't know what to write in those, usually the summary is good enough, but I'll add them

src/Layers/W1/BaseApp/Inventory/Tracking/POMatching.Codeunit.al
mynjj: I think it adds indirection that is not really needed, in general I'm not a fan of this pattern (but I know there are different opinions here), we could put it later on

src/Layers/W1/BaseApp/Inventory/Tracking/POMatching.Codeunit.al
mynjj: I hadn't heard of this one, we have an analyzer that validates sorting on some types (like records before codeunits, etc), but it doesn't have any consideration for `temporary` records.

I can do it regardless 😄 Do you think it helps readability? it's already prefixed with `Temp`.

mynjj:
> Requesting changes for two data-integrity issues in the new matching API:
> 
> 1. `ValidateOrderReceipt` accepts negative allocation quantities because every cap check is upper-bound-only. A negative receipt edge is persisted and reduces the aggregate pinned/consumed quantities, allowing a later positive edge to exceed the invoice-order budget or receipt availability while the signed total still matches the invoice quantity at posting. Reject wrong-sign/non-positive allocations and cover this case with regression tests.

Done, good catch, thanks. (We probably have already some gotchas with negative values in the existing PO Matching functionality that we could look at)

> 2. `SaveMatchingGroups` revalidates an in-memory component loaded earlier by `AddMatch`, then persists without locking and reloading the shared component. Concurrent sessions matching different invoices to the same order or receipt can both validate against the same remaining quantity and insert distinct rows, over-allocating it. Lock the shared table, reload/merge the latest persisted component under that lock, then revalidate and persist atomically.
> 

I prefer not to lock, the chance of having matches in this situation is very low, and regardless everything gets revalidated at posting, which is locked/"atomic-ish", so I won't add any further locks at creation (it's just asking for trouble IMO 😅), also with the reloading of persisted matches every time we add a match, the chance of this is even less.

@PredragMaricic

VolodySyn (approved):

PredragMaricic (approved):

alexei-dobriansky:
## Good Sense Reviewer - Round 4

<details>
<summary><strong>What changed since round 3</strong></summary>

The PO Matching files changed (the rest of the range is an unrelated merge from main). The relevant change hardens the new matching API against negative allocations and adds documentation:
- A new `CheckAllocationNotNegative(Match)` guard runs at the start of **both** `ValidateInvoiceOrder` and `ValidateOrderReceipt`, raising `NegativeAllocationErr` when `"Qty. to Invoice"` or `"Qty. to Invoice (Base)"` is negative.
- Two regression tests cover it: `AddInvoiceOrderMatchNegativeQuantityErrors` and `AddOrderReceiptMatchNegativeQuantityErrors`.
- XML doc comments were added to the edge factories and a receipt-key `SetCurrentKey` was added in `Invoiceable Receipts.Load`.
</details>

<details>

| ID  | Title | Status | Evidence |
|-----|-------|--------|----------|
| S1  | Validate the purchase document types | Addressed | Still enforced by `ValidateInvoiceOrderCompatibility`. |
| S2  | Reject mismatched units of measure | Addressed | Still enforced before base quantity is derived. |
</details>

<details>
<summary><strong>New observations (commits since round 3)</strong></summary>

**S3** (🟠 Moderate) - *Resolved:* negative/wrong-sign allocations are now rejected.
The earlier gap - upper-bound-only caps let a negative edge lower the aggregate consumed and free budget for a later positive edge - is closed. `CheckAllocationNotNegative` guards both validation paths and the two new tests assert both the invoice-order and order-receipt directions. This is the right fix and is properly covered. One small point: the guard rejects `< 0` but allows `0`; a zero-quantity edge is a harmless no-op for the budget, so this is fine, but reject `<= 0` if you want to forbid no-op edges outright.

**S4** (🟠 Moderate) - *Concurrent creation can transiently over-allocate; the no-lock decision is defensible but worth confirming.*
`SaveMatchingGroups` validates an in-memory component and persists without locking/reloading the shared component, so two sessions matching different invoices to the same order or receipt can each validate against the same remaining quantity and both insert rows. The stated rationale - the authoritative quantity consumption is revalidated at posting (which is locked), and matches are reloaded on each `AddMatch` - is reasonable, and the existing `Matched Order Line Mgmt.` posting path supports that this is the real gate. So this is not a blocker. The residual risk is a match that looks valid at creation but errors at posting under contention. Please confirm the posting-side revalidation rejects aggregate over-allocation under its lock (not only the per-line cap); if it does, the creation-time no-lock choice is sound.
</details>

<details>

**Risk:** Reduced since round 3. The negative-allocation hole was a genuine data-integrity gap and is now guarded and tested. The remaining concurrency consideration is a consciously accepted trade-off backed by posting-time revalidation, not an unguarded corruption path. The public API surface was already narrowed in round 3.

**Necessity:** The API is needed for the Payables Agent PO matching slice. The round-4 hardening is a clear improvement, and the concurrency decision is a legitimate engineering judgment given posting is the authoritative quantity gate.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10524 round=4 by=alexei-dobriansky at=2026-08-25T12:28:00Z lastSha=6c61059f2630da303a8769e1accdae07cbc73ce3 reviewKey=3f7c07f35f0cefd1044cd51fd16be3ef1534c98cd59219fc690a7110246ec900 suggestions=S1@419220ac:addressed,S2@b98fb3b7:addressed,S3@b3bccd84:addressed,S4@39fe17a4:new parentRound=3

alexei-dobriansky (approved):

alexei-dobriansky:
## Good Sense Reviewer - Round 5

<details>

This PR adds a public purchase order matching API: a matching-group facade and buffer with validation and persistence, an invoiceable-receipts iterator, and receipt-suggestion logic, with broad tests. Since the previous round the branch was force-updated, so this round reviews the full current change against the base. The validation flow is coherent for the feature: document-type, unit-of-measure, vendor/currency, receipt ownership, allocation caps, negative allocations, and item tracking are all guarded. No new High-impact defect was found in the current diff.
</details>

<details>

| S1 | Validate the purchase document types | Addressed | Document-type validation present |
| S2 | Reject mismatched units of measure | Addressed | Unit-of-measure validation present |
| S3 | Reject negative allocations | Addressed | `CheckAllocationNotNegative` guards both paths, with tests |
| S4 | Concurrent creation can transiently over-allocate | Not addressed | No locking/reload added at save time |
</details>

<details>
<summary><strong>New observations (changes since previous round)</strong></summary>

*None - the current changes did not introduce new issues.*
</details>

<details>

**Risk:** The main remaining risk is data integrity around persisted matched-order quantities and creation-time concurrency. `SaveMatchingGroups` validates an in-memory component and persists without locking or reloading the shared component, so two sessions matching different invoices to the same order or receipt can each validate against the same remaining quantity and both insert rows (S4). This is Moderate, not blocking: the authoritative quantity consumption is revalidated at posting under its own locked flow. It is worth confirming rather than leaving implicit.

**Necessity:** The API is a coherent additive feature increment and keeps existing BaseApp purchase flows undisturbed. The scope is broad but aligned with the feature goal and covered by tests for the key validation paths.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10524 round=5 by=alexei-dobriansky at=2026-08-27T06:17:24Z lastSha=999f19d925b7d705734a7dbe0faba5cb4793ae90 reviewKey=fffa477dae2f572a9b271ca9175c2feae37b9af7f79b08662bb0f8a9e5183f6b suggestions=S1@419220ac:addressed,S2@b98fb3b7:addressed,S3@b3bccd84:addressed,S4@39fe17a4:notaddressed parentRound=4

src/Layers/W1/BaseApp/Inventory/Tracking/POMatching.Codeunit.al
```diff
@@ -0,0 +1,148 @@
...
+    #endregion
+
+    #region Covering receipts
+    /// <summary>
+    /// Enriches the group by covering each invoice-order allocation with the order line's posted receipts that still have quantity received not invoiced, by adding or growing receipt edges.
+    /// </summary>
+    /// <param name="POMatchingGroup">The PO matches that we want to suggest receipts for. This parameter gets modified by this procedure by ading the suggested receipt matches.</param>
+    procedure SuggestCoveringReceipts(var POMatchingGroup: Codeunit "PO Matching Group")
+    var
+        OrderLine: Record "Purchase Line";
```
PeterConijn, line 76: I would call this variable "PurchaseLine" to keep it consistent with other variables and it makes it immediately clear what we are looking at.

src/Layers/W1/BaseApp/Inventory/Tracking/POMatching.Codeunit.al
```diff
@@ -0,0 +1,148 @@
...
+        Edge."Matched Rcpt./Shpt. Line SysId" := ReceiptLineSystemId;
+        Edge."Qty. to Invoice" := QtyToAllocate;
+    end;
+
+    /// <summary>Builds an invoice-to-receipt edge; the order line is derived when added.</summary>
+    /// <param name="InvoiceLineSystemId">SystemId of the Invoice Line.</param>
+    /// <param name="ReceiptLineSystemId">SystemId of the Receipt Line.</param>
+    /// <param name="QtyToAllocate">The quantity desired to allocate for the match.</param>
+    /// <returns>A match that can be used in AddMatch.</returns>
+    procedure InvoiceReceiptEdge(InvoiceLineSystemId: Guid; ReceiptLineSystemId: Guid; QtyToAllocate: Decimal) Edge: Record "Matched Order Line"
```
PeterConijn, line 61: Why is the return value called `Edge` and not `MatchedOrderLine`?

src/Layers/W1/BaseApp/Inventory/Tracking/POMatching.Codeunit.al
```diff
@@ -0,0 +1,148 @@
...
+        Edge."Qty. to Invoice" := QtyToAllocate;
+    end;
+
+    /// <summary>Builds an order-to-receipt edge with an explicit invoice, used to split a receipt across several invoices.</summary>
+    /// <param name="InvoiceLineSystemId">SystemId of the Invoice Line.</param>
+    /// <param name="OrderLineSystemId">SystemId of the Order Line.</param>
+    /// <param name="ReceiptLineSystemId">SystemId of the Receipt Line.</param>
+    /// <param name="QtyToAllocate">The quantity desired to allocate for the match.</param>
+    /// <returns>A match that can be used in AddMatch.</returns>
+    procedure InvoiceOrderReceiptEdge(InvoiceLineSystemId: Guid; OrderLineSystemId: Guid; ReceiptLineSystemId: Guid; QtyToAllocate: Decimal) Edge: Record "Matched Order Line"
```
PeterConijn, line 48: Why is the return value called Edge and not MatchedOrderLine?

src/Layers/W1/BaseApp/Inventory/Tracking/POMatching.Codeunit.al
```diff
@@ -0,0 +1,148 @@
...
+        Edge."Matched Order Line SystemId" := OrderLineSystemId;
+        Edge."Qty. to Invoice" := QtyToAllocate;
+    end;
+
+    /// <summary>Builds an order-to-receipt edge; the invoice is inferred when added if the order has a single invoice edge.</summary>
+    /// <param name="OrderLineSystemId">SystemId of the Order Line.</param>
+    /// <param name="ReceiptLineSystemId">SystemId of the Receipt Line.</param>
+    /// <param name="QtyToAllocate">The quantity desired to allocate for the match.</param>
+    /// <returns>A match that can be used in AddMatch.</returns>
+    procedure OrderReceiptEdge(OrderLineSystemId: Guid; ReceiptLineSystemId: Guid; QtyToAllocate: Decimal) Edge: Record "Matched Order Line"
```
PeterConijn, line 35: Why is the return value called Edge and not MatchedOrderLine?

src/Layers/W1/BaseApp/Inventory/Tracking/POMatching.Codeunit.al
```diff
@@ -0,0 +1,148 @@
...
+        if not POMatchingGroup.GetInvoiceOrderEdges() then
+            exit;
+
+        repeat
+            InvoiceLineSystemId := POMatchingGroup.GetInvoiceLine();
+            OrderLineSystemId := POMatchingGroup.GetOrderLine();
+            OrderLine.GetBySystemId(OrderLineSystemId);
+            Missing := POMatchingGroup.AllocatedInInvoiceOrderEdge() - POMatchingGroup.SumAllocatedInOrderReceiptEdges();
+
+            if (Missing > 0) and InvoiceableReceipts.Load(OrderLine) then begin
```
PeterConijn, line 92: Performance-wise, you might want to choose for a separated pattern, e.g.

```al
if Missing > 0 then
    if InvoiceableReceipts.Load(OrderLine) then begin
```

AL does not natively do lazy evaluation, so by using this pattern we can prevent unnecessary database calls if Misssing <= 0.

PeterConijn (changes requested):

AndersLarsenMicrosoft (approved):

alexei-dobriansky:
## Good Sense Reviewer - Round 6

<details>

This PR adds a public purchase order matching API: edge factories, a matching-group buffer with validation and persistence, an invoiceable-receipts iterator, and receipt-suggestion logic. The current net diff is a coherent feature slice and includes broad tests for caps, document roles, unit of measure, persisted-component reload, item tracking, and receipt suggestions.

Round detection found new branch commits after the previous review, but the API and test files are unchanged from the previous reviewed PR tip. I reviewed the full current net diff again because the previous head is no longer reachable. No new High-impact defect was found.
</details>

<details>

**Fit:** Strong

The requested behavior is a reusable way to create and validate purchase order matching rows for later consumers. The diff matches that goal without changing the existing BaseApp matching flow, and the API validates the main data-integrity rules before rows are saved.
</details>

<details>

| S1 | Validate the purchase document types | Addressed | No new reply; invoice lines and order lines are still checked for the right document types. |
| S2 | Reject mismatched units of measure | Addressed | No new reply; invoice and order lines must still use the same unit of measure. |
| S3 | Reject negative allocations | Addressed | No new reply; CheckAllocationNotNegative still guards both validation paths and tests cover both directions. |
| S4 | Concurrent creation can transiently over-allocate | Not addressed | No new reply; SaveMatchingGroups still validates the already loaded in-memory component and then persists without a save-time lock/reload. |
</details>

<details>
<summary><strong>Current suggestions</strong></summary>

**S4** (🟠 Moderate): *Confirm concurrent saves cannot over-allocate*
SaveMatchingGroups still revalidates only the already loaded in-memory component before persisting. If two sessions save different matches for the same order or receipt at the same time, both can pass creation-time checks; please confirm posting revalidates the aggregate matched quantities under lock, or reload and lock the component before saving if creation must be authoritative.
</details>

<details>

**Risk:** The main risk remains data integrity around Matched Order Line quantities because this is new public BaseApp surface used before purchase posting. The validation coverage is strong, and current CI reports all checks as passing, but creation-time concurrency can still leave a match that looked valid when saved and later fails when posted.

**Necessity:** The feature is needed as an additive integration boundary for purchase order matching. The scope is broad but aligned with the requested slice, and the remaining item is a non-blocking confirmation of the concurrency contract rather than a proven wrong posting result.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10524 round=6 by=alexei-dobriansky at=2026-08-27T18:16:49Z lastSha=1f12fa2cb8ae88ae9b028abd0959f1123b8d273f reviewKey=577a8e6b1f0ab36d322130243ea9a7de49b60f89e779c9192ec028a2d117792b suggestions=S1@419220ac:addressed,S2@b98fb3b7:addressed,S3@b3bccd84:addressed,S4@39fe17a4:notaddressed parentRound=5

src/Layers/W1/BaseApp/Inventory/Tracking/POMatching.Codeunit.al
mynjj: Purchase lines can be order or invoice lines, in this context it's more clear to know which kind of line is intended

src/Layers/W1/BaseApp/Inventory/Tracking/POMatching.Codeunit.al
mynjj: Taking this improvement in a separate workitem (648156)

PR #10530, opened 2026-08-22
[Main]- Mismatch between Purchase Invoice Total in Lines and Statistics after VAT adjustment

DeepsShukla:
[Bug 642172](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/642172): [master] [all-e]Mismatch between Purchase Invoice Total in Lines and Statistics after VAT adjustment.

Fixes [AB#642172](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/642172)

### DIFF
--- a/src/Layers/BE/BaseApp/Utilities/DocumentTotals.Codeunit.al
+++ b/src/Layers/BE/BaseApp/Utilities/DocumentTotals.Codeunit.al
@@ -7,6 +7,7 @@ namespace Microsoft.Utilities;
 using Microsoft.Finance.Currency;
 using Microsoft.Finance.GeneralLedger.Setup;
 using Microsoft.Finance.SalesTax;
+using Microsoft.Finance.VAT.Calculation;
 using Microsoft.Purchases.Document;
 using Microsoft.Purchases.History;
 using Microsoft.Purchases.Posting;
@@ -671,6 +672,8 @@ codeunit 57 "Document Totals"
         TotalPurchaseLine2: Record "Purchase Line";
         PurchaseLineWithReverseChargeVAT: Record "Purchase Line";
         VATAmountOfLinesWithRevChargeVAT: Decimal;
+        NonDeductibleVATAmount: Decimal;
+        GroupedVATAmount: Decimal;
         IsHandled: Boolean;
     begin
         IsHandled := false;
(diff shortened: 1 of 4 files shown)


alexei-dobriansky:

<details>

On the Purchase Invoice subform, the `Total VAT Amount` and `Total Amount Incl. VAT` did not update after a VAT adjustment on the Statistics page when the invoice had mixed VAT groups on the same G/L account with negative lines. The subform computed VAT as the simple sum of `Amount Including VAT - Amount`, which does not match the per-VAT-group rounding used by Statistics. This PR adds `GetGroupedVATAmount`, which recomputes the VAT via `CalcVATAmountLines` (the same per-group calculation Statistics uses) and, when it differs, overrides `VATAmount` and `Amount Including VAT` on the totals. The change is applied to W1, NA, and BE, and a W1 test is included.

The approach is sound: aligning the subform with the grouped VAT calculation is the correct way to reflect an allowed VAT difference. The guards are reasonable (`Allow VAT Difference` on, `Tax Liable` off to skip US sales tax, header `No.` present). The BE variant correctly also guards on no reverse-charge and no non-deductible VAT lines before overriding. The included test reproduces the exact scenario and asserts both totals, and its handler (`PurchaseInvoiceStatisticsUpdateVATAmountPageHandler`) and labels exist.
</details>

<details>

**S1** - *Check whether Sales needs the same fix*
Sales invoices use the parallel `CalculateSalesSubPageTotals` with the same simple `Amount Including VAT - Amount` VAT calculation. The same mismatch after a VAT-difference adjustment likely exists there too. Please check the Sales subform and, if it has the same defect, apply the analogous fix so Sales and Purchase stay consistent.

**S2** - *Redundant assignment to TotalPurchaseLine*
Inside the new block you set `TotalPurchaseLine."Amount Including VAT"`, but the procedure ends with `TotalPurchaseLine := TotalPurchaseLine2`, which overwrites it. Setting only `TotalPurchaseLine2."Amount Including VAT"` is enough; the extra line can be removed for clarity.
</details>

<details>

**Risk:** Medium. `Document Totals` (codeunit 57) runs on every purchase document subform refresh, so this is a hot path, but the recalculation is gated behind `Allow VAT Difference` and `Tax Liable = false`, which limits how often `CalcVATAmountLines` runs. The override only triggers when the grouped amount differs from the summed amount, so normal invoices are unaffected. The country variants (W1/NA/BE) keep their existing reverse-charge and non-deductible handling, and BE guards the override behind those being zero.

**Necessity:** Justified. A visible mismatch between the subform totals and Statistics after a legitimate VAT adjustment is a real, customer-facing correctness issue on a financial figure. The fix is targeted, guarded, and tested.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=10530 round=1 by=alexei-dobriansky at=2026-08-23T00:12:40Z lastSha=dfeb6337640266ac09835ac3e0da5118ec4945ff reviewKey=dc70afba7e41f5b4d2c5c637b3048885be2d8ee99a1abc27e00f4685b731429d suggestions=S1@8e65c88c,S2@41ddb0ca

PR #10540, opened 2026-08-23
[Extensibility Request] issue 29518: make document logo position extensible

AleksandricMarko:
## Summary
Extensions need to support additional company logo placements on sales documents, but the current setup field is a fixed option. This PR converts the field to an extensible, assignment-compatible enum while retaining the existing values and persisted ordinals.

Source issue repository: microsoft/AlAppExtensions; issue number: 29518

## Changes Made
- `Logo Position on Documents` - added an extensible enum with the existing logo positions and assignment compatibility.
- `Sales & Receivables Setup`.`Logo Position on Documents` - changed the field from Option to the new enum in W1 and all existing layer counterparts.

Fixes [AB#616632](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/616632)


AleksandricMarko:
@copilot it is necessary to check all places where field is used and to determine if it is needed to update usage
procedure  SetLogoPosition accepts option, it would be the best to create new one which will accept enum (and which will call existing one)

PR #10542, opened 2026-08-23
[Master] - Slice 638097: [Expense Agent] Enable simple interim approval scenario

v-rohangarg20:
## Summary
Enables a simple two-stage interim approval step for expense reports in agent mode.

## What changed
- Interim and final approver fields on the expense report header; the Final Approver No. is prepopulated from the expense user's designated approver.
- New **Interim Approved** status with two-stage routing: Pending Approval -> Interim Approved -> Approved (single-step Pending Approval -> Approved when no interim approver is set).
- "Assign Interim Approver" on the expense report card and via the API, gated to agent mode.
- Guards: the interim approver cannot be the submitter or the final approver, must be selected, is only assignable while Pending Approval, and cannot give the final approval.
- Activity logging for interim approver assignment and interim approval.
- Integration tests (codeunit 148346 "Expense Interim Approval Test").

## Work item
[AB#638097](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/638097)

### DIFF
--- a/src/Apps/W1/ExpenseAgent/app/src/APIs/ApproverViewAPI.Page.al
+++ b/src/Apps/W1/ExpenseAgent/app/src/APIs/ApproverViewAPI.Page.al
@@ -50,7 +50,7 @@ page 6969 "Approver View API"
                     EntityName = 'expenseReport';
                     EntitySetName = 'expenseReports';
                     SubPageLink = "Pending Approval By" = field("No."),
-                                    Status = const("Pending Approval");
+                                    Status = filter("Pending Approval" | "Interim Approved");
                 }
             }
         }
--- a/src/Apps/W1/ExpenseAgent/app/src/Approval/Codeunits/ExpenseReportApprovalMgmt.Codeunit.al
+++ b/src/Apps/W1/ExpenseAgent/app/src/Approval/Codeunits/ExpenseReportApprovalMgmt.Codeunit.al
@@ -27,6 +27,7 @@ codeunit 6901 "Expense Report Approval Mgmt"
         InterimApproverRequiredErr: Label 'Select an interim approver from the available approvers.';
         InterimApproverConflictErr: Label 'The %1 cannot be the same as the %2 %3.', Comment = '%1 = Interim Approver No. caption, %2 = conflicting field caption, %3 = conflicting field value';
         InterimApproverCannotFinalizeErr: Label '%1 %2 cannot give final approval. Final approval must be completed by a different approver.', Comment = '%1 = Interim Approver No. caption, %2 = Interim Approver No.';
+        ActorNotActiveApproverErr: Label 'This expense report is awaiting approval from %1. Only that approver can approve or reject it.', Comment = '%1 = Expense User No. of the approver the report is currently assigned to';
         InterimApproverAssignedCommentTxt: Label 'Interim approver set to %1 (%2).', Locked = true;
 
     procedure ProcessAction(var ExpenseReportHeader: Record "Expense Report Header"; ActionType: Enum "Expense Approval Action")
@@ -176,6 +177,8 @@ codeunit 6901 "Expense Report Approval Mgmt"
         ExpenseReportHeader.TestApprovalPending();
         CheckApproverPermissions(ExpenseReportHeader);
         ApproverExpenseUserNo := GetExpenseUserNo();
+        CheckActorIsNotInterimApprover(ExpenseReportHeader, ApproverExpenseUserNo);
+        CheckActorIsActiveApprover(ExpenseReportHeader, ApproverExpenseUserNo);
 
         SetApprovalStatusInExpenseReport(ExpenseReportHeader, ExpenseReportHeader.Status::Rejected, ApproverExpenseUserNo, CopyStr(UserId(), 1, 50));
         LogExpenseReportRejected(ExpenseReportHeader, ApproverExpenseUserNo, '');
@@ -190,6 +193,8 @@ codeunit 6901 "Expense Report Approval Mgmt"
 
         ExpenseUser.Get(ApproverExpenseUserNo);
         CheckApproverPermissions(ExpenseUser);
+        CheckActorIsNotInterimApprover(ExpenseReportHeader, ApproverExpenseUserNo);
+        CheckActorIsActiveApprover(ExpenseReportHeader, ApproverExpenseUserNo);
 
         UpdateApproverComment(ExpenseReportHeader, RejectReason);
         SetApprovalStatusInExpenseReport(ExpenseReportHeader, ExpenseReportHeader.Status::Rejected, ApproverExpenseUserNo, ExpenseUser."User Id For Approvals");
@@ -207,6 +212,7 @@ codeunit 6901 "Expense Report Approval Mgmt"
         CheckApproverPermissions(ExpenseReportHeader);
         ApproverExpenseUserNo := GetExpenseUserNo();
         CheckActorIsNotInterimApprover(ExpenseReportHeader, ApproverExpenseUserNo);
+        CheckActorIsActiveApprover(ExpenseReportHeader, ApproverExpenseUserNo);
 
         if ShouldRouteToFinalApprover(ExpenseReportHeader) then begin
             RouteToFinalApprover(ExpenseReportHeader, ApproverExpenseUserNo);
@@ -227,6 +233,7 @@ codeunit 6901 "Expense Report Approval Mgmt"
         ExpenseUser.Get(ApproverExpenseUserNo);
         CheckApproverPermissions(ExpenseUser);
         CheckActorIsNotInterimApprover(ExpenseReportHeader, ApproverExpenseUserNo);
+        CheckActorIsActiveApprover(ExpenseReportHeader, ApproverExpenseUserNo);
 
         if ShouldRouteToFinalApprover(ExpenseReportHeader) then begin
             RouteToFinalApprover(ExpenseReportHeader, ApproverExpenseUserNo);
@@ -351,6 +358,18 @@ codeunit 6901 "Expense Report Approval Mgmt"
 
         if ActingApproverExpenseUserNo = ExpenseReportHeader."Interim Approver No." then
             Error(InterimApproverCannotFinalizeErr, ExpenseReportHeader.FieldCaption("Interim Approver No."), ExpenseReportHeader."Interim Approver No.");
+    end;
+
+    local procedure CheckActorIsActiveApprover(ExpenseReportHeader: Record "Expense Report Header"; ActingApproverExpenseUserNo: Code[20])
+    begin
+        if ExpenseReportHeader."Interim Approver No." = '' then
+            exit;
+
+        if ActingApproverExpenseUserNo = '' then
+            exit;
+
+        if ActingApproverExpenseUserNo <> ExpenseReportHeader."Approver Expense User No." then
+            Error(ActorNotActiveApproverErr, ExpenseReportHeader."Approver Expense User No.");
     end;
 
     local procedure SetApprovalStatusInExpenseReport(var ExpenseReportHeader: Record "Expense Report Header"; ExpenseReportStatus: Enum "Expense Report Status"; ApproverExpenseUserNo: Code[20]; ApproverUserId: Code[50])
--- a/src/Apps/W1/ExpenseAgent/app/src/Approval/Codeunits/ExpenseReportApprovalMgmt.Codeunit.al
+++ b/src/Apps/W1/ExpenseAgent/app/src/Approval/Codeunits/ExpenseReportApprovalMgmt.Codeunit.al
@@ -22,6 +22,12 @@ codeunit 6901 "Expense Report Approval Mgmt"
         NotAuthorizedToRecallExpReportErr: Label 'Only the original submitter or a user with %1 can recall a submitted expense report.', Comment = '%1 = User Setup field caption';
         ApproverMustBeEnabledInExpenseUserErr: Label '%1 must be enabled to approve or reject expense reports in %2.', Comment = '%1 = Field Caption, %2 = Table Caption';
         UserIdForApprovalMustNotBeBlankInExpenseUserErr: Label '%1 must not be blank in %2.', Comment = '%1 = Field Caption, %2 = Table Caption';
+        InterimApproverAgentRequiredErr: Label 'An interim approver can only be assigned when the agent is enabled in %1.', Comment = '%1 = Expense Agent Setup table caption';
+        InterimApproverStatusErr: Label 'You can only assign an interim approver while the expense report is %1.', Comment = '%1 = Pending Approval status caption';
+        InterimApproverRequiredErr: Label 'Select an interim approver from the available approvers.';
+        InterimApproverConflictErr: Label 'The %1 cannot be the same as the %2 %3.', Comment = '%1 = Interim Approver No. caption, %2 = conflicting field caption, %3 = conflicting field value';
+        InterimApproverCannotFinalizeErr: Label '%1 %2 cannot give final approval. Final approval must be completed by a different approver.', Comment = '%1 = Interim Approver No. caption, %2 = Interim Approver No.';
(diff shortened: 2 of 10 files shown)


alexei-dobriansky:

<details>

This PR adds a two-step approval flow for expense reports in agent mode. It stores final and interim approvers, lets the submitter assign an interim approver, routes approval from Pending Approval to Interim Approved, and adds tests for the approval management flow.

The core approval code mostly follows the feature goal, and the tests cover the direct approve/reject paths. The main gap is in the approver-facing API view. It still finds reports from the submitter approval setup, not from the current active approver, so the new interim approver step is not routed correctly through that view.
</details>

<details>

**S1** - *Route approver view by active approver*  
The approver-view subpage now includes `Interim Approved`, but `Expense Reports API` still turns `Pending Approval By` into an `Expense User No.` filter from approval setup. Interim approvers are not in that setup, so they will not see reports waiting for their interim approval, while the final approver can still see them during the pending interim step. Filter by the active `Approver Expense User No.` or include interim approver assignments so the right approver gets the report.
</details>

<details>

**Risk:** The changed area controls who can see and act on an expense report approval. The new fields and API action are additive, and there is no BaseApp event dependency. The risk is that the API approver view can route the approval task to the wrong user, which breaks the two-stage approval flow.

**Necessity:** The slice is valid and useful because it adds an optional interim review before final approval. The scope is right for the Expense Agent app, but the approver view must match the new active-approver routing before merge.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=10542 round=1 by=alexei-dobriansky at=2026-08-23T19:10:25Z lastSha=0be547783a71de5656457a70bca4480228fa59ad reviewKey=90378536214beeb10e2392fc4cd6033cb844b20bce870f3f1b6cf549ac999645 suggestions=S1@16268985

src/Apps/W1/ExpenseAgent/app/src/Approval/Codeunits/ExpenseReportApprovalMgmt.Codeunit.al
```diff
@@ -22,6 +22,13 @@ codeunit 6901 "Expense Report Approval Mgmt"
         NotAuthorizedToRecallExpReportErr: Label 'Only the original submitter or a user with %1 can recall a submitted expense report.', Comment = '%1 = User Setup field caption';
         ApproverMustBeEnabledInExpenseUserErr: Label '%1 must be enabled to approve or reject expense reports in %2.', Comment = '%1 = Field Caption, %2 = Table Caption';
         UserIdForApprovalMustNotBeBlankInExpenseUserErr: Label '%1 must not be blank in %2.', Comment = '%1 = Field Caption, %2 = Table Caption';
+        InterimApproverAgentRequiredErr: Label 'An interim approver can only be assigned when the agent is enabled in %1.', Comment = '%1 = Expense Agent Setup table caption';
+        InterimApproverStatusErr: Label 'You can only assign an interim approver while the expense report is %1.', Comment = '%1 = Pending Approval status caption';
+        InterimApproverRequiredErr: Label 'Select an interim approver from the available approvers.';
+        InterimApproverConflictErr: Label 'The %1 cannot be the same as the %2 %3.', Comment = '%1 = Interim Approver No. caption, %2 = conflicting field caption, %3 = conflicting field value';
```
encimita: This will interpolate to 
```
"The Interim Approver No. cannot be the same as the Final Approver No. FOOBAR"
```
Should we instead do

```
InterimApproverConflictErr: Label 'The %1 cannot be the same as the %2. Value: %3.'
```

or similar?

src/Apps/W1/ExpenseAgent/app/src/Approval/Codeunits/ExpenseReportApprovalMgmt.Codeunit.al
encimita, line 91: Should we not update this as well?

src/Apps/W1/ExpenseAgent/app/src/APIs/ApproverViewAPI.Page.al
```diff
@@ -50,7 +50,7 @@ page 6969 "Approver View API"
                     EntityName = 'expenseReport';
                     EntitySetName = 'expenseReports';
                     SubPageLink = "Pending Approval By" = field("No."),
-                                    Status = const("Pending Approval");
+                                    Status = filter("Pending Approval" | "Interim Approved");
```
encimita, line 53: Widening the status filter here is right, but the same widening is missing on page 6913 `Accountant Expense Reports` (`src/Apps/W1/ExpenseAgent/app/src/Expense/Pages/AccountantExpenseReports.Page.al`), which this PR doesn't touch.

There, both `Approve` and `Reject` are gated on `Enabled = Rec.Status = Rec.Status::"Pending Approval"` (lines 131 and 148). Once a report moves to `Interim Approved`, the final approver can't act on it from the approver list page -- the surface they're most likely to use. They'd have to open the card instead.

`SetStatusStyle` (around line 255) also has no case for `Interim Approved`, so it falls through to `'Standard'` while `Pending Approval` renders as `'Attention'`. The report stops looking like it needs attention exactly when it still does.

Both should be updated the same way as this filter.

src/Apps/W1/ExpenseAgent/app/src/Approval/Codeunits/ExpenseReportApprovalMgmt.Codeunit.al
```diff
@@ -210,11 +232,146 @@ codeunit 6901 "Expense Report Approval Mgmt"
...
 
+    internal procedure AssignInterimApprover(var ExpenseReportHeader: Record "Expense Report Header"; NewApproverExpenseUserNo: Code[20]; ActorExpenseUserNo: Code[20])
+    var
+        InterimApprover: Record "Expense User";
+        ExpenseAgentSetup: Record "Expense Agent Setup";
+    begin
+        ExpenseAgentSetup.GetRecordOnce();
+        if not ExpenseAgentSetup."Enable Agent" then
+            Error(InterimApproverAgentRequiredErr, ExpenseAgentSetup.TableCaption());
+
```
encimita, line 258: This validates the *target* approver (`CheckApproverPermissions(InterimApprover)` further down) but never the *actor*. `ActorExpenseUserNo` arrives from the API caller, is trusted as-is, and is used only for activity logging. `Approve` and `Reject` both run `CheckApproverPermissions` against the acting user; this path doesn't.

I know the API is protected in the frontend, so this isn't exploitable as things stand. The concern is what happens when that assumption slips: a frontend bug passing the wrong submitter would silently reroute approval of someone else's report to an arbitrary user, and the activity log would record the wrong actor -- so the audit trail wouldn't help anyone spot it. That's the kind of thing that turns a frontend regression into an escalation.

A server-side check on the actor makes the backend independently correct instead of resting the whole defense on one layer.

src/Apps/W1/ExpenseAgent/app/src/Approval/Codeunits/ExpenseReportApprovalMgmt.Codeunit.al
```diff
@@ -210,11 +232,146 @@ codeunit 6901 "Expense Report Approval Mgmt"
...
+        InterimApprover: Record "Expense User";
+    begin
+        if ExpenseReportHeader."Interim Approver No." = '' then
+            exit;
+
+        if ExpenseReportHeader."Interim Approver No." = ExpenseReportHeader."Final Approver No." then
+            exit;
+
+        InterimApprover.Get(ExpenseReportHeader."Interim Approver No.");
+        ExpenseReportHeader."Approver Expense User No." := InterimApprover."No.";
```
encimita, line 330: This reroutes to the interim approver without rechecking that they still qualify. `AssignInterimApprover` runs `CheckApproverPermissions(InterimApprover)`, but that happens once, at assignment. Every reroute after that is blind. `RouteToFinalApprover` below has the same gap.

The `:=` assignment also means the field's `TableRelation = "Expense User"."No." where("Can Approve" = const(true))` never fires -- that only validates through `Validate()`.

A flow that wedges a report:

1. Alice submits, Bob is final approver, Carol is assigned as interim (passes the check -- Carol has `Can Approve`).
2. Carol rejects the report.
3. Carol moves teams; an admin clears `Can Approve` on her Expense User.
4. Alice fixes the report and resubmits. `Submit` calls this procedure, sees `Interim Approver No. = CAROL`, and routes to her anyway. Status is `Pending Approval`, active approver Carol.
5. Carol can't approve -- `CheckApproverPermissions` errors on `Can Approve`.
6. Bob can't approve either -- `CheckActorIsActiveApprover` errors, because the active approver is Carol.

The report is stuck until the submitter recalls it or someone with `Unlimited Expense Approval` intervenes. `ReopenApproved` reaches the same code path, so an approved report can be reopened into the same state.

Clearing `User Id For Approvals` is the quieter variant: `"Approver Expense User ID"` becomes blank, so `ApprovalActionsEnabled` on the card is false for everyone and the approval actions just silently don't render. No error, no report.

Running `CheckApproverPermissions` inside both route procedures should cover it -- either fall back to the final approver when the interim no longer qualifies, or fail with a message that says what's wrong.

alexei-dobriansky:

<details>

This slice adds a simple interim approval step to the Expense Agent. It introduces `Interim Approver No.` and `Final Approver No.` on the report header, a new `Interim Approved` status, and two-stage routing: Pending Approval to Interim Approved to Approved, with a single step when no interim approver is set. It adds an agent-gated "Assign Interim Approver" action on the card and API, guards (interim approver cannot be the submitter or final approver, is only assignable while Pending Approval, and cannot give final approval), activity logging, and integration tests. The commit since round 1 adds `CheckActorIsActiveApprover`, wires it plus `CheckActorIsNotInterimApprover` into the approve and reject paths, promotes the assign action, and adds three guard tests. These new guards correctly stop a non-active approver from acting. The remaining problem is discovery: the approver-facing API view still resolves the target approver from the submitter's approval setup, not from the report's active approver, so the two-stage flow is not routed correctly through the documented API.
</details>

<details>

| ID  | Title                              | Status        | Author response |
|-----|------------------------------------|---------------|-----------------|
| S1  | Route approver view by active approver | Not addressed | `ApproverViewAPI` still links `SubPageLink = "Pending Approval By" = field("No.")`, and `ExpenseReportsAPI` still turns that value into a submitter filter via approval setup. Interim approvers are not in that setup, so they still do not see reports waiting for their interim approval, and the final approver still sees them during the pending-interim step. The active `Approver Expense User No.` is still not filtered on or exposed. The new `CheckActorIsActiveApprover` blocks the wrong-approver action but does not fix the view routing. No author reply. |
</details>

<details>

**S2** - *No test exercises the API view routing*
The new tests call `AssignInterimApprover`, `PerformManualApproved`, and `PerformManualRejected` directly with an explicit user, so they bypass the approver API view. Nothing asserts that the interim approver actually sees the interim-pending report, or that the final approver does not, so the S1 routing defect is not caught by any test. Add a test that drives the report through the approver view for each stage.

**S3** - *Active-approver guard silently skips on blank actor*
`CheckActorIsActiveApprover` exits early when the acting approver code is blank, so if `GetExpenseUserNo()` ever returns blank the active-approver check is skipped. An earlier permission check likely blocks this today, so it is defensive only, but a short comment or a hard error on the blank case would make the intent clear.
</details>

<details>

**Risk:** The changed area controls who can see and act on an expense report approval. This round meaningfully lowers risk, because the new guards stop a non-active or already-acted approver from approving or rejecting, and those guards are tested. The residual risk is the unchanged approver-view routing: the API cannot route the interim step to the interim approver and shows the final approver reports they cannot yet act on, so the two-stage flow is not correctly discoverable through the API. The fields and action are additive with no BaseApp event dependency and no G/L posting path.

**Necessity:** The slice is valid and useful, adding an optional interim review before final approval, and it is scoped correctly to the Expense Agent app. The action-permission correctness is now in good shape. Before merge, the approver view must route by the active approver (or expose and filter the interim approver assignment) so the right approver is actually served the report, and a test should cover that routing.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=2 system=github pr=10542 round=2 by=alexei-dobriansky at=2026-08-24T13:27:19Z lastSha=14de858ac35714f6209aa5583e5f102c5d66a068 reviewKey=f763be6a783ff5de2007390fc5d9a63937634957818c2079c262db01424779b0 suggestions=S1@16268985:notaddressed,S2@9a078a25:new,S3@fe06fb34:new parentRound=1

src/Apps/W1/ExpenseAgent/app/src/APIs/ApproverViewAPI.Page.al
v-rohangarg20: Resolved. Thanks!

src/Apps/W1/ExpenseAgent/app/src/Approval/Codeunits/ExpenseReportApprovalMgmt.Codeunit.al
v-rohangarg20: Resolved. Thanks

src/Apps/W1/ExpenseAgent/app/src/Approval/Codeunits/ExpenseReportApprovalMgmt.Codeunit.al
v-rohangarg20: Done

src/Apps/W1/ExpenseAgent/app/src/Approval/Codeunits/ExpenseReportApprovalMgmt.Codeunit.al
v-rohangarg20: Resolved.

Rishukla90 (approved):

encimita (approved):

alexei-dobriansky:

<details>

This round keeps the simple interim approval flow and adds more guard work around resubmission, final-approver routing, and the approval actions. The new commits improve the direct approve/reject paths, but the approver-facing API still discovers reports from the submitter approval setup instead of the report's active approver, so the interim step is still not routed correctly through that view.
</details>

<details>

| S1 | Route approver view by active approver | Not addressed | No reply. `ApproverViewAPI` still links through `Pending Approval By`, and `ExpenseReportsAPI` still converts that value into a submitter filter from approval setup. Interim approvers still do not see reports waiting for their interim approval, while the final approver can still see the report during the interim step. |
| S2 | No test exercises the API view routing | Not addressed | No reply. The new tests still call the approval procedures directly and do not open or filter the approver API view for interim and final approvers. |
| S3 | Active-approver guard silently skips on blank actor | Not addressed | No reply. The helper still exits on a blank actor. The current approve/reject callers normally pass a nonblank value before this helper, so this remains a defensive concern rather than the main blocker. |
</details>

<details>

**S4** (🟠 Moderate): *Owner check can still be skipped*
The new owner check only runs when `ActorExpenseUserNo` is not blank. The page overload still calls interim assignment with a blank actor, and the API can also send a blank value. Require a real actor in the API path, or derive the actor on the server before assigning the interim approver.
</details>

<details>

**Risk:** The remaining risk is in approval routing and authorization for the new interim step. There is no BaseApp event dependency and no G/L posting path, but the wrong user can still be shown or allowed to drive the approval flow through the API surface.

**Necessity:** The feature is useful and the scope is right for an optional interim review before final approval. The routing view and actor validation still need to match that feature before the change is safe to merge.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10542 round=3 by=alexei-dobriansky at=2026-08-25T12:28:15Z lastSha=99fd82a068f47ae406c9834a8d8b1f3b92c75866 reviewKey=c51d3ad11919166300fe840c527af942dbd05e81759c593bd071694b4a3e469a suggestions=S1@16268985:notaddressed,S2@9a078a25:notaddressed,S3@fe06fb34:notaddressed,S4@98c9f862:new parentRound=2

djukicmilica (approved):

PR #10562, opened 2026-08-24
Bug 647211:  [Subcontracting] Install app by default in new online environments only

ChethanT:
## Linked work

Fixes [AB#647211](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/647211)

## How I validated this


alexei-dobriansky (approved):

build/projects.json
```diff
@@ -3708,7 +3708,6 @@
       "dvdFolder": "Applications\\Subcontracting\\Source",
       "supportedCountries": "All",
       "installOnEnvironmentUpdate": false,
-      "skipInstallOnPrem": true,
```
attilatoury, line 3711: The PR name says we want only for online environments, will this not also install the app in on-prem ones?

attilatoury (approved):

PredragMaricic (approved):

build/projects.json
ChethanT: installation on prem should be skipped ONLY for apps that ha sa dependency on SaaS environment.

aholstrup1 (approved):

alexei-dobriansky:

<details>

This change makes the Subcontracting app and its test app installable by default by removing `doNotInstall` from their build groups. It also keeps `installOnEnvironmentUpdate` set to `false`, which matches the requirement that existing online environments should not get the app just because they are updated.

The online update part is correct, and the Italian migration app remains opt-in. The problem is that the same diff also removes the on-premises install guards, so the change is broader than the reported online-only behavior.
</details>

<details>

**Fit:** Partial

The bug asks for automatic installation in new online environments while preserving the current behavior for existing environments and on-premises deployments. The diff handles the existing-online update path, but it also changes the on-premises groups and project metadata.
</details>

<details>

**S1** (🔴 High): *Keep on-premises installation opt-in*
Restore the on-premises guards while leaving the online group changes in place. `OnPrem`, `OnPremExtensions`, and the Subcontracting project metadata now opt into default installation, so on-premises packages can install the app by default even though the requested behavior is online-only.
</details>

<details>

**Risk:** The main risk is compatibility for on-premises deployments. In `build/groups.json`, a missing `doNotInstall` becomes `false`, and the install script skips only when that flag is true. Removing `skipInstallOnPrem` in `build/projects.json` also changes the on-premises metadata for the app. The JSON files parse successfully, and no BaseApp publisher check is needed because this is build metadata only.

**Necessity:** Enabling default installation for new online environments is required and the scope is small. The on-premises part is not required for this bug and should stay unchanged unless that behavior is approved separately.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10562 round=1 by=alexei-dobriansky at=2026-08-26T00:12:50Z lastSha=356cbe564b78115b5aac2db60635e31becc34d97 reviewKey=1a7d4e9fcb6dfcf63412445adb73f615b1c397a161568aae63c70acfeb1e3e80 suggestions=S1@e1ae4321

PR #10568, opened 2026-08-24
Send sample invoices from the demo guide directly

Explorer986:
## What & why

On the Payables Agent guide page, picking the email option only marked the sample invoices for sending. The real send happened later when setup was saved. So if you saved setup first and ran the guide afterwards, the invoices were never sent even though the guide reported success.

Now the guide sends the sample invoices right when you choose the email option, so it no longer depends on the order of steps. After a send the marks are cleared, so saving setup will not send them a second time, and the final screen reports success only when the email was actually sent.

Fixes [AB#647147](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/647147)

## How I validated this

- Saved Payables Agent Setup first, then ran the demo guide and chose "Send sample invoices to configured email": the invoices were sent and showed up as agent tasks. Previously nothing was sent in this setup order scenario.
- Ran the guide first, then saved setup, mail was sent once, not twice (flags cleared after the first send).
- Built the Payables Agent app with no new warnings.

## Risk & compatibility

Only the two PADemoGuide files are modified. SendDemoEmail returns a Boolean now rather than no return value, and since the codeunit is internal there is no public breaking change.

### DIFF
--- a/src/Apps/W1/PayablesAgent/app/Setup/PADemoGuide.Page.al
+++ b/src/Apps/W1/PayablesAgent/app/Setup/PADemoGuide.Page.al
@@ -6,7 +6,6 @@
 #pragma warning disable AS0007
 namespace Microsoft.Agent.PayablesAgent;
 
-using Microsoft.eServices.EDocument.Processing.Import.Purchase;
 using System.Environment;
 using System.Environment.Configuration;
 using System.Utilities;
@@ -187,6 +186,8 @@ page 3307 "PA Demo Guide"
         BackActionEnabled, FinishActionEnabled, NextActionEnabled, WelcomeStepVisible, ChooseDemoOption, ShowDemoFilesToDownloadVisible, TopBannerVisible : Boolean;
         DemoFilesToDownloadText, FinalStepHeadline, FinalStepText : Text;
         PayablesAgentTelemetryTok: Label 'Payables Agent', Locked = true;
+        DemoInvoicesSentTok: Label 'Sent demo sample invoices by email', Locked = true;
+        DemoInvoicesPreparedTok: Label 'Prepared demo sample invoices to send on agent activation', Locked = true;
         DemoFilesToDownloadLbl: Label 'Click here to select and download sample invoices (%1 file(s) available)', Comment = '%1 = number of files';
         FinalStepHeadlineEmailLbl: Label 'Sample invoices sent!';
         FinalStepTextEmailLbl: Label 'We have prepared the sample invoices and will send them when the agent is activated. If this is the first time you configure the agent, the agent is activated when you select "Update" on the configuration page. When the agent is activated you should see them appear as tasks on the agent''s avatar in the top right corner of the home screen';
@@ -227,11 +228,13 @@ page 3307 "PA Demo Guide"
 
     local procedure FinishAction();
     var
-        EDocSamplePurchInvFile: Record "E-Doc Sample Purch. Inv File";
         GuidedExperience: Codeunit "Guided Experience";
     begin
         if DemoOption = DemoOption::"Send Demo Email" then
-            EDocSamplePurchInvFile.ModifyAll("Send By Email", true);
+            if PADemoGuide.SendDemoInvoicesByEmail() then
+                Session.LogMessage('0000V8C', DemoInvoicesSentTok, Verbosity::Normal, DataClassification::SystemMetadata, TelemetryScope::All, 'Category', PayablesAgentTelemetryTok)
+            else
+                Session.LogMessage('0000V8D', DemoInvoicesPreparedTok, Verbosity::Normal, DataClassification::SystemMetadata, TelemetryScope::All, 'Category', PayablesAgentTelemetryTok);
         Session.LogMessage('0000PJV', 'Ran demo guide for payables agent', Verbosity::Normal, DataClassification::SystemMetadata, TelemetryScope::All, 'Category', PayablesAgentTelemetryTok);
         GuidedExperience.CompleteAssistedSetup(ObjectType::Page, Page::"PA Demo Guide");
         CurrPage.Close();
--- a/src/Apps/W1/PayablesAgent/app/Setup/PADemoGuide.Codeunit.al
+++ b/src/Apps/W1/PayablesAgent/app/Setup/PADemoGuide.Codeunit.al
@@ -55,20 +55,34 @@ codeunit 3309 "PA Demo Guide"
     /// <summary>
     /// Sends a demo email using the active PA setup configuration.
     /// </summary>
-    procedure SendDemoEmail()
+    /// <returns>True if the demo email was sent.</returns>
+    procedure SendDemoEmail(): Boolean
     var
         PayablesAgentSetup: Codeunit "Payables Agent Setup";
         PASetupConfiguration: Codeunit "PA Setup Configuration";
     begin
         PayablesAgentSetup.LoadSetupConfiguration(PASetupConfiguration);
-        SendDemoEmail(PASetupConfiguration);
+        exit(SendDemoEmail(PASetupConfiguration));
+    end;
+
+    /// <summary>
+    /// Sends the sample invoices by email if the agent is active, otherwise flags them to be sent at activation.
+    /// </summary>
+    /// <returns>True if the sample invoices were sent.</returns>
+    procedure SendDemoInvoicesByEmail(): Boolean
+    var
+        EDocSamplePurchInvFile: Record "E-Doc Sample Purch. Inv File";
+    begin
+        EDocSamplePurchInvFile.ModifyAll("Send By Email", true);
+        exit(SendDemoEmail());
     end;
 
     /// <summary>
     /// Sends a demo email using the provided PA setup configuration.
     /// </summary>
     /// <param name="PASetupConfiguration">An email account to send a demo email</param>
-    procedure SendDemoEmail(PASetupConfiguration: Codeunit "PA Setup Configuration");
+    /// <returns>True if the demo email was sent.</returns>
+    procedure SendDemoEmail(PASetupConfiguration: Codeunit "PA Setup Configuration"): Boolean
     var
         EDocSamplePurchInvFile: Record "E-Doc Sample Purch. Inv File";
         TempEmailAccount: Record "Email Account";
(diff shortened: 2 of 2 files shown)


src/Apps/W1/PayablesAgent/app/Setup/PADemoGuide.Page.al
```diff
@@ -227,13 +228,16 @@ page 3307 "PA Demo Guide"
...
     begin
         if DemoOption = DemoOption::"Send Demo Email" then
-            EDocSamplePurchInvFile.ModifyAll("Send By Email", true);
+            if PADemoGuide.SendDemoInvoicesByEmail() then
+                Session.LogMessage('0000V8C', DemoInvoicesSentTok, Verbosity::Normal, DataClassification::SystemMetadata, TelemetryScope::All, 'Category', PayablesAgentTelemetryTok)
+            else
+                Session.LogMessage('0000V8D', DemoInvoicesPreparedTok, Verbosity::Normal, DataClassification::SystemMetadata, TelemetryScope::All, 'Category', PayablesAgentTelemetryTok);
         Session.LogMessage('0000PJV', 'Ran demo guide for payables agent', Verbosity::Normal, DataClassification::SystemMetadata, TelemetryScope::All, 'Category', PayablesAgentTelemetryTok);
         GuidedExperience.CompleteAssistedSetup(ObjectType::Page, Page::"PA Demo Guide");
+        Commit(); // Persist pending writes and close the transaction so CurrPage.Close's client callback is allowed.
```
ventselartur: What exactly is going to fail if you remove this commit? I think that you are trying to insert this commit to get over the case where the payables agent configuration is not over yet and user is still on it's way to update the values.

src/Apps/W1/PayablesAgent/app/Setup/PADemoGuide.Page.al
```diff
@@ -227,13 +228,16 @@ page 3307 "PA Demo Guide"
 
     local procedure FinishAction();
     var
-        EDocSamplePurchInvFile: Record "E-Doc Sample Purch. Inv File";
         GuidedExperience: Codeunit "Guided Experience";
     begin
         if DemoOption = DemoOption::"Send Demo Email" then
-            EDocSamplePurchInvFile.ModifyAll("Send By Email", true);
+            if PADemoGuide.SendDemoInvoicesByEmail() then
```
ventselartur, line 234: I see what you are doing with this change. My only concern is how this is going to work in the case where the payables agent is still not activated. I think you are trying to solve this by adding the commit call which is something we need to look more precisely to not break the existing experience.

ventselartur (changes requested):

src/Apps/W1/PayablesAgent/app/Setup/PADemoGuide.Page.al
Explorer986: Thanks Artur, When Agent not activated, case is handled the same way it does today, the SendDemoInvoicesByEmail() is just added on top of SendDemoEmail().

When the agent is still not activated, CanSendDemoEmail returns false because the state is disabled, so SendDemoInvoicesByEmail flags the sample files but skips the actual send and returns false. The files stay flagged and the existing setup-save path (ApplyPayablesAgentSetup then SendDemoEmail) sends them once the agent is activated, exactly like before.

src/Apps/W1/PayablesAgent/app/Setup/PADemoGuide.Page.al
Explorer986: I was trying to solve the side observation "SuspendCallStackExecution is not allowed in write transaction", I have removed the commit() and let's handle it in different bug.

alexei-dobriansky:

<details>

This changes PADemoGuide.Page.al to call SendDemoInvoicesByEmail() from FinishAction(), and changes PADemoGuide.Codeunit.al so the send helpers return whether an email was actually sent. For the setup-already-saved flow, the new call flags the sample invoices, sends them through the active setup, clears the flags after success, and leaves the setup-save caller idempotent because no flagged rows remain.

The main path matches the reported bug: the guide no longer depends on a later setup save to dispatch the email. The fallback path still means "prepared for later", so the remaining risk is the user message around that fallback and the lack of automated coverage for the two setup orders.
</details>

<details>

**Fit:** Strong

The reported setup-saved-before-guide order is addressed directly because the guide now sends the sample invoice email itself. The existing guide-before-activation order is also preserved by leaving the records marked for the setup-save path when the agent cannot send yet.
</details>

<details>

**S1** (🟠 Moderate): *Show the actual send result*
The new branch knows whether SendDemoInvoicesByEmail() sent the email or only prepared it, but the finish page text is still chosen earlier and always uses the sent headline for the email option. Make the visible text match the prepared case, or use neutral text until the send result is known. Otherwise the guide can still say that invoices were sent when they were only queued for activation.

**S2** (🟠 Moderate): *Cover both setup orders*
Add a regression test for the saved-setup-before-guide path and the guide-before-activation path. The test should also prove that a later setup save does not send the same sample invoices twice. This is not a blocker for this UI flow, but it protects the new Boolean contract and flag clearing.
</details>

<details>

**Risk:** The change is limited to the Payables Agent demo guide and its internal helper codeunit, but it adds an immediate external email send to a page action and changes how the sample-invoice flags are cleared. No BaseApp publisher or event-ordering dependency is involved, and the build checks reported success.

**Necessity:** The change is needed because users can naturally save setup before running the guide, and that order previously left invoices flagged without sending them. The scope is right for the bug because it keeps the setup-save path for activation-time sending while adding the missing direct send from the guide.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10568 round=1 by=alexei-dobriansky at=2026-08-25T12:24:55Z lastSha=0766fe7e5c76f5423a81feeeb4f74ff0a6bb429e reviewKey=c89c145f3e31434e73cfbcd070d0672cbee2569a15b92ec6361b85203463bd52 suggestions=S1@dc74cf27,S2@67b2ae42

ventselartur (dismissed):

src/Apps/W1/PayablesAgent/app/Setup/PADemoGuide.Page.al
```diff
@@ -227,11 +228,13 @@ page 3307 "PA Demo Guide"
...
```
mynjj, line 234: minor: the name and comment on this new procedure is a bit misleading, because you are sending demo invoices in case it's active, but if it's not you are enqueueing them, so that they are sent at agent activation. Perhaps include that in the summary of the procedure, or name the procedure with something that reflects that.

mynjj (dismissed):

src/Apps/W1/PayablesAgent/app/Setup/PADemoGuide.Page.al
Explorer986: updated summary of procedure.

mynjj (approved):

alexei-dobriansky:

<details>

This round only changes the XML summary for `SendDemoInvoicesByEmail()` so it now describes both outcomes: sending immediately when the agent is active, or preparing the sample invoices for activation later.

The functional code is unchanged from the previous round. The main bug fix still sends sample invoices directly from the guide when setup was already saved, and it still leaves the activation-time flow in place when the agent cannot send yet.
</details>

<details>

| S1 | Show the actual send result | Not addressed | No author response. The new commit only updates the procedure summary; the page still sets the final headline and text before `FinishAction()` knows whether the email was sent. |
| S2 | Cover both setup orders | Not addressed | No author response. No test files changed, so the two setup orders and duplicate-send guard still depend on manual validation. |
</details>

<details>

*None - the new commit only updates documentation text and does not introduce new functional risk.*
</details>

<details>

**Risk:** The round-2 change itself is documentation-only. The PR still touches the Payables Agent demo guide path that sends external email and clears the sample-invoice flags, so misleading finish text and missing regression coverage remain the only review risks.

**Necessity:** The change remains needed because the guide must work when setup is saved before the sample-invoice guide is run. The latest commit improves the helper description, but it does not change the remaining non-blocking suggestions.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10568 round=2 by=alexei-dobriansky at=2026-08-26T13:10:41Z lastSha=5198a36f37397ecc8e8e8d732ce4a3b2fe822d76 reviewKey=8c813efa264d95bdaaeb74f433d77047c66aae9c42cc4788056e95a1dc88f202 suggestions=S1@dc74cf27:notaddressed,S2@67b2ae42:notaddressed parentRound=1

ventselartur (approved):

alexei-dobriansky:

<details>

This round adds telemetry that separates the two outcomes of SendDemoInvoicesByEmail(): the guide now logs one event when the sample invoices were sent immediately, and another event when they were only prepared for sending on agent activation.

The main fix still sends the sample invoices from the guide when the agent setup is already active, and keeps the activation-time path for the guide-first order. The new telemetry improves diagnostics, but it does not change the visible final page message.
</details>

<details>

| S1 | Show the actual send result | Not addressed | The new commit distinguishes sent vs prepared in telemetry only. The final page still sets the same headline and text before FinishAction() knows whether the email was actually sent. |
| S2 | Cover both setup orders | Not addressed | No test files changed. The saved-setup-before-guide path, guide-before-activation path, and duplicate-send guard are still covered only by manual validation. |
</details>

<details>

*None - the new commit adds outcome-specific telemetry and does not introduce a new functional risk.*
</details>

<details>

**Risk:** The round-3 change is limited to telemetry messages in the Payables Agent demo guide. The remaining risk is still user-facing: the final page can tell the user the wrong outcome when the helper returns the prepared path or when it sends immediately.

**Necessity:** The PR remains needed because the demo guide must work when setup is saved before the guide runs. The latest commit helps telemetry analysis, but the user-facing result text and regression coverage still need follow-up.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10568 round=3 by=alexei-dobriansky at=2026-08-27T19:24:50Z lastSha=37e8d99380e84c267d6cdb7def2ce8c7dd70e2c6 reviewKey=662247c36f73bee3bc7989a6780db640df57dc611bf32087fc96b9278cf87778 suggestions=S1@dc74cf27:notaddressed,S2@67b2ae42:notaddressed parentRound=2

PR #10571, opened 2026-08-24
[Master]-User is unable to filter the Deferral Summary-GL Report Not Filtering Totals by Global Dimensions.

neeleshsinghal:
Fixes [AB#647585](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/647585)

### DIFF
--- a/src/Layers/ES/Tests/ERM/TestREDSetupGenJnl.Codeunit.al
+++ b/src/Layers/ES/Tests/ERM/TestREDSetupGenJnl.Codeunit.al
@@ -21,6 +21,7 @@ codeunit 134803 "Test RED Setup Gen. Jnl."
         LibrarySetupStorage: Codeunit "Library - Setup Storage";
         LibraryTimeSheet: Codeunit "Library - Time Sheet";
         LibraryJournals: Codeunit "Library - Journals";
+        LibraryDimension: Codeunit "Library - Dimension";
         DeferralUtilities: Codeunit "Deferral Utilities";
         CalcMethod: Enum "Deferral Calculation Method";
         DeferralDocType: Option Purchase,Sales,"G/L";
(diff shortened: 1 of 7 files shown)


v-ankitgoyal (approved):

AndersLarsenMicrosoft (approved):

alexei-dobriansky:

<details>

This PR adds Global Dimension 1 and Global Dimension 2 filters to the Deferral Summary - G/L report and skips posted G/L deferral headers whose source G/L entry does not match those filters. The detail filtering is aimed at the reported case, but the skip happens before the report resets per-account total state, so a filtered-out account can still print stale totals from the previous matching account.
</details>

<details>

**Fit:** Partial

The bug asks for the report output and totals to respect global dimension filters. The detail rows are filtered, but the totals dataitem can still use old sum variables when all headers for the current account are skipped by the new filter.
</details>

<details>

**S1** (🔴 High): *Reset totals before dimension skip*
The new CurrReport.Skip() runs before SumAmtRecognized, SumRemainingAmtDeferred, and SumTotalAmtDeferred are reset for a new G/L account. If an account has only filtered-out headers after a matching account, the totals row can reuse the previous account's totals. Reset the per-account sums before this skip, or skip the totals row when no matching header was included.

**S2** (🟠 Moderate): *Assert filtered totals in the test*
The new tests only check which document numbers are present. Add an assertion for SumTotalAmtDeferred or the other total columns so the test fails if the report hides a row but keeps a wrong total.
</details>

<details>

**Risk:** This report is used for G/L deferral analysis, so wrong totals can make the report look filtered while still showing an incorrect amount. The change is narrow, but it changes report aggregation state and therefore affects financial reporting output.

**Necessity:** The change is needed because users must be able to analyze deferrals by global dimensions. The scope is right, but the aggregation path needs the same filtering guarantee as the detail rows.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10571 round=1 by=alexei-dobriansky at=2026-08-25T12:25:43.1323464Z lastSha=8e8a2d8a91dda093cce698beef4475101d843ea7 reviewKey=fff878d4f4e1249d9e159ce37ce4dd53ecc296659baa9e5cb0db531090d592e1 suggestions=S1@7f622c64,S2@605a2bcc

PR #10575, opened 2026-08-24
Calculate Inventory: scope Default Dimension reads by source table

### LINKED ISSUE #9100: [Bug]: Calculate Inventory applies default dimensions from an unrelated Location whose Code equals the Item No.
### Describe the issue

Report 790 "Calculate Inventory", procedure `CreateDimFromDefault()`, filters Default Dimension with two independent OR-filters:

```al
DefaultDimension.SetFilter("No.", '%1|%2', TempQuantityOnHandBuffer."Item No.", TempQuantityOnHandBuffer."Location Code");
DefaultDimension.SetFilter("Table ID", '%1|%2', DATABASE::Item, DATABASE::Location);
```

Because `"No."` and `"Table ID"` are filtered independently, the combinations are never paired per table. If an Item's `No.` happens to equal the `Code` of some unrelated Location (for example both are `10`), that unrelated Location's default dimensions pass both filters and leak onto the counting journal line, even when the inventory is calculated for a completely different location.

Verified still present in current main: `src/Layers/W1/BaseApp/Inventory/Counting/Journal/CalculateInventory.Report.al`, `CreateDimFromDefault()`.

The fix is to replace the cross-product filter with two scoped reads: one for `("Table ID" = Database::Item, "No." = Item No.)` and one for `("Table ID" = Database::Location, "No." = Location Code)`.

### Expected behavior

Only the default dimensions of the counted Item and of the Location actually being counted are applied to the physical inventory journal line. A Location that merely shares its Code with the Item's No. must not contribute dimensions.

### Steps to reproduce

1. Create Item with No. = 10.
2. Create Location with Code = 10 and add default dimensions to it.
3. Post some inventory for Item 10 on a different location (e.g. BLUE).
4. Run Calculate Inventory (report 790) for location BLUE.
5. The journal line receives the default dimensions of Location 10, which has nothing to do with the count.

### Additional context

This is a copy of microsoft/BusinessCentralApps#1542 (reported by Karel Koncel, labeled Approved/SCM in the contribution pilot). It was being fixed in microsoft/BusinessCentralApps#1908, which was reviewed (a regression test was requested and added) but was closed unmerged when the repository was retired. Re-filing here per the retirement notice to re-submit the fix against BCApps.

Franco111000:
## What & why

Report 790 "Calculate Inventory", procedure `CreateDimFromDefault()`, reads Default Dimensions with two independent OR-filters: one on `"No."` (Item No. or Location Code) and one on `"Table ID"` (Item or Location). Because the two filters are applied independently, the pairs are never scoped per table and the read expands into a cross-product. When an Item's `No.` happens to equal an unrelated Location's `Code`, that Location's default dimensions pass both filters and land on the physical inventory journal line, even though the count is being made for a completely different location.

This change replaces the cross-product read with two scoped passes, one per intended `(Table ID, No.)` pair: `(Item, Item No.)` and `(Location, Location Code)`. Each pass feeds the same `InsertDim` call, so the `TempDimBufIn` buffer, the `DimBufMgt.InsertDimensions()` call and the trailing buffer cleanup are untouched. When Item No. and Location Code do not collide, the set of dimensions applied is exactly the same as before.

The RU layer carries a full copy of the report with the same two filter lines, and receives the same change, as requested in the approval comment on the issue.

This is a re-submission of microsoft/BusinessCentralApps#1908 (pilot repository, retired). The fix and the regression test were reviewed there; the test was added at the reviewer's request. The pilot PR was closed unmerged by the repository retirement.

## Linked work

Fixes #9100

Original pilot review: microsoft/BusinessCentralApps#1908 (fixing microsoft/BusinessCentralApps#1542).

## How I validated this

- Verified against current main that the defect is still present: the two independent OR-filters are in `CreateDimFromDefault()` in both the W1 and the RU copy of the report, byte-identical apart from unrelated whitespace on a line this change does not touch.
- Verified that `CalculateInventory.Report.al` exists only in the W1 and RU layers, so no further layer propagation is needed, and that the two fixed `CreateDimFromDefault()` bodies are identical between the two layers. Codeunit 137275 "SCM Inventory Journals" exists only in W1, so the regression test has no layer copy to mirror.
- Swept the rest of the repository for the same shape, two independent OR-filters on `"No."` and `"Table ID"` against Default Dimension. Report 790 is the only occurrence. Other Default Dimension reads pin `"Table ID"` to a single value, which is safe, and the `'%1|%2'` form with a blank second value is the deliberate specific-or-blank idiom rather than a cross-product.
- New regression test `CalcInventoryDoesNotLeakDefaultDimWhenItemNoEqualsAnotherLocationCode` in codeunit 137275 "SCM Inventory Journals": the Item and an unrelated Location are both renamed to the same generated 10 character code, that Location is given a Default Dimension, stock for the Item is posted on a different target Location that carries its own Default Dimension, and Calculate Inventory is run. The test asserts both directions: the target Location's Default Dimension is present on the generated line's dimension set, and the unrelated Location's Default Dimension is not.
- The shared code comes from `LibraryUtility.GenerateRandomCode(Location.FieldNo(Code), DATABASE::Location)`, so it is unique and fits `Location.Code` by construction rather than by relying on the length the Item No. series happens to produce. This follows the rename pattern already used in `ERMDimensionLocations.Codeunit.al`.
- The two Locations carry Default Dimensions on different Dimension Codes on purpose. With the same Dimension Code, `if TempDimBufIn.Insert() then` would resolve the duplicate by insertion order and the test could pass on the unfixed code depending on how the values sort.
- The test reuses the helpers already in the codeunit (`CreateItem`, `UpdateItemInventory`, `CreateAndPostPhysInventoryJournal`, `FindItemJournalLine`); no new test infrastructure is added.
- I did not build the Base Application locally. The change is two scoped reads per report copy plus one test, and relies on CI for the compile and the test run.

## Risk & compatibility

- Low and localized: the change lives entirely inside the local `CreateDimFromDefault()` of report 790. No public surface, no signature change, no event change.
- Behavior is identical when an Item No. and a Location Code do not collide; only the colliding case changes, and it changes to the intended result.
- The `InsertDim` / `TempDimBufIn` flow, the `DimBufMgt.InsertDimensions()` call and the trailing `TempDimBufIn` cleanup are unchanged. The `TempDimBufIn.SetFilter("Table ID", '%1|%2', ...)` a few lines below the fix looks like the same shape but is not: it clears the temporary Dimension Buffer for both source tables at once, which is what that cleanup is meant to do, so it is deliberately left alone.
- Ordering is unaffected. "Dimension Buffer" is keyed on `("Table ID", "Entry No.", "Dimension Code")`, so Item and Location rows never collide in the buffer and it is iterated in key order rather than insertion order. The two passes therefore hand `DimBufMgt.InsertDimensions()` the same rows in the same order as the single pass did, and a Dimension Code held by both the Item and the Location resolves to the same winner as before.
- The two passes are kept inline, as reviewed in the pilot PR, where extracting a shared helper was raised as an optional, non-blocking suggestion.
- No schema, permission, or upgrade impact.

### DIFF
--- a/src/Layers/RU/BaseApp/Inventory/Counting/Journal/CalculateInventory.Report.al
+++ b/src/Layers/RU/BaseApp/Inventory/Counting/Journal/CalculateInventory.Report.al
@@ -962,29 +962,27 @@ report 790 "Calculate Inventory"
     end;
 
     local procedure CreateDimFromDefault() DimEntryNo: Integer
-    var
-        DefaultDimension: Record "Default Dimension";
     begin
-        DefaultDimension.SetRange("Table ID", DATABASE::Item);
-        DefaultDimension.SetRange("No.", TempQuantityOnHandBuffer."Item No.");
-        DefaultDimension.SetFilter("Dimension Value Code", '<>%1', '');
-        if DefaultDimension.FindSet() then
-            repeat
-                InsertDim(DefaultDimension."Table ID", 0, DefaultDimension."Dimension Code", DefaultDimension."Dimension Value Code");
-            until DefaultDimension.Next() = 0;
-
-        DefaultDimension.SetRange("Table ID", DATABASE::Location);
-        DefaultDimension.SetRange("No.", TempQuantityOnHandBuffer."Location Code");
-        DefaultDimension.SetFilter("Dimension Value Code", '<>%1', '');
-        if DefaultDimension.FindSet() then
-            repeat
-                InsertDim(DefaultDimension."Table ID", 0, DefaultDimension."Dimension Code", DefaultDimension."Dimension Value Code");
-            until DefaultDimension.Next() = 0;
+        InsertDefaultDimensions(DATABASE::Item, TempQuantityOnHandBuffer."Item No.");
+        InsertDefaultDimensions(DATABASE::Location, TempQuantityOnHandBuffer."Location Code");
 
         DimEntryNo := DimBufMgt.InsertDimensions(TempDimBufIn);
 
         TempDimBufIn.SetFilter("Table ID",'%1|%2',DATABASE::Item, Database::Location);
         TempDimBufIn.DeleteAll();
+    end;
+
+    local procedure InsertDefaultDimensions(TableID: Integer; No: Code[20])
+    var
+        DefaultDimension: Record "Default Dimension";
+    begin
+        DefaultDimension.SetRange("Table ID", TableID);
+        DefaultDimension.SetRange("No.", No);
+        DefaultDimension.SetFilter("Dimension Value Code", '<>%1', '');
+        if DefaultDimension.FindSet() then
+            repeat
+                InsertDim(DefaultDimension."Table ID", 0, DefaultDimension."Dimension Code", DefaultDimension."Dimension Value Code");
+            until DefaultDimension.Next() = 0;
     end;
 
     local procedure InsertDim(TableID: Integer; EntryNo: Integer; DimCode: Code[20]; DimValueCode: Code[20])
--- a/src/Layers/RU/BaseApp/Inventory/Counting/Journal/CalculateInventory.Report.al
+++ b/src/Layers/RU/BaseApp/Inventory/Counting/Journal/CalculateInventory.Report.al
@@ -965,8 +965,16 @@ report 790 "Calculate Inventory"
     var
         DefaultDimension: Record "Default Dimension";
     begin
-        DefaultDimension.SetFilter("No.", '%1|%2', TempQuantityOnHandBuffer."Item No.", TempQuantityOnHandBuffer."Location Code");
-        DefaultDimension.SetFilter("Table ID", '%1|%2', DATABASE::Item, DATABASE::Location);
+        DefaultDimension.SetRange("Table ID", DATABASE::Item);
+        DefaultDimension.SetRange("No.", TempQuantityOnHandBuffer."Item No.");
+        DefaultDimension.SetFilter("Dimension Value Code", '<>%1', '');
+        if DefaultDimension.FindSet() then
+            repeat
+                InsertDim(DefaultDimension."Table ID", 0, DefaultDimension."Dimension Code", DefaultDimension."Dimension Value Code");
+            until DefaultDimension.Next() = 0;
+
+        DefaultDimension.SetRange("Table ID", DATABASE::Location);
+        DefaultDimension.SetRange("No.", TempQuantityOnHandBuffer."Location Code");
         DefaultDimension.SetFilter("Dimension Value Code", '<>%1', '');
         if DefaultDimension.FindSet() then
             repeat
--- a/src/Layers/W1/BaseApp/Inventory/Counting/Journal/CalculateInventory.Report.al
+++ b/src/Layers/W1/BaseApp/Inventory/Counting/Journal/CalculateInventory.Report.al
@@ -959,8 +959,16 @@ report 790 "Calculate Inventory"
     var
         DefaultDimension: Record "Default Dimension";
     begin
-        DefaultDimension.SetFilter("No.", '%1|%2', TempQuantityOnHandBuffer."Item No.", TempQuantityOnHandBuffer."Location Code");
-        DefaultDimension.SetFilter("Table ID", '%1|%2', DATABASE::Item, DATABASE::Location);
+        DefaultDimension.SetRange("Table ID", DATABASE::Item);
+        DefaultDimension.SetRange("No.", TempQuantityOnHandBuffer."Item No.");
+        DefaultDimension.SetFilter("Dimension Value Code", '<>%1', '');
+        if DefaultDimension.FindSet() then
+            repeat
+                InsertDim(DefaultDimension."Table ID", 0, DefaultDimension."Dimension Code", DefaultDimension."Dimension Value Code");
+            until DefaultDimension.Next() = 0;
+
+        DefaultDimension.SetRange("Table ID", DATABASE::Location);
+        DefaultDimension.SetRange("No.", TempQuantityOnHandBuffer."Location Code");
         DefaultDimension.SetFilter("Dimension Value Code", '<>%1', '');
         if DefaultDimension.FindSet() then
             repeat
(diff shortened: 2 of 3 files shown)


alexei-dobriansky:

<details>

Report 790 "Calculate Inventory", procedure `CreateDimFromDefault()`, read Default Dimensions with two independent OR-filters: one on `"No."` (Item No. or Location Code) and one on `"Table ID"` (Item or Location). Applied together, these match a cross-product, so a Location whose `Code` equals the counted Item's `No.` had its Default Dimensions applied to the physical inventory journal line even though a different Location was counted.

The change replaces the single cross-product read with two scoped passes, one per intended pair: `(Item, Item No.)` and `(Location, Location Code)`. Both passes feed the same `InsertDim` / `TempDimBufIn` flow, and the same change is mirrored in the RU copy of the report.
</details>

<details>

**Fit:** Strong

The two scoped `SetRange("Table ID", ...)` + `SetRange("No.", ...)` passes read exactly the intended `(table, key)` pairs, so an unrelated Location that only collides on `No.` can no longer match. When Item No. and Location Code do not collide, the applied dimension set is unchanged. The regression test reproduces the collision (Item and an unrelated Location renamed to the same code) and asserts both directions - target Location dimension present, colliding Location dimension absent.
</details>

<details>

**S1** (🟡 Minor): *The two passes are near-identical; a small local helper would remove the duplication.*
The Item pass and the Location pass differ only in the `"Table ID"` and `"No."` values; the filter on `"Dimension Value Code"` and the `FindSet`/`repeat`/`InsertDim` body are the same. A small local procedure taking `(TableId, No.)` would remove the copy and keep the two report copies (W1 and RU) easier to keep in sync. This is optional and non-blocking - the pilot review already noted the same point as non-blocking.
</details>

<details>

**Risk:** Low and localized. The change lives entirely inside the local `CreateDimFromDefault()` of report 790 - no public surface, no signature change, no event change, no schema or upgrade impact. The non-colliding case is behavior-preserving, and the "Dimension Buffer" key (`"Table ID"`, `"Entry No."`, `"Dimension Code"`) keeps Item and Location rows from colliding in the buffer, so ordering into `DimBufMgt.InsertDimensions()` is unchanged. The trailing `TempDimBufIn.SetFilter("Table ID", '%1|%2', ...)` cleanup is correctly left alone - it clears both source tables at once, which is its intent.

**Necessity:** Justified. Leaking an unrelated Location's default dimensions onto a physical inventory line is a data-correctness defect (wrong dimensions posted), and the RU layer carries the same defect, so both copies are fixed. The regression test is appropriate for this dimension/data-integrity area.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10575 round=1 by=alexei-dobriansky at=2026-08-25T11:40:00Z lastSha=e230aca0e1663693e77a2516c696878bd558b9c0 reviewKey=722f83f38581873bc3a8e58c2dc580e980b6c2203811a53b4fd8275e9aaec282 suggestions=S1@492e9cf9

Franco111000:
Thanks for the review @alexei-dobriansky.

On S1, that is a fair point. I kept the two passes inline because it is the shape that was reviewed on the pilot pull request and the one named in the approval on #9100, so I wanted the diff to stay as close to the reviewed version as possible. If you would rather have the shared loop pulled into a small local procedure taking `(Table ID, No.)`, say the word and I will push it, it is behaviour neutral either way.

One note on the checks: the only red one is `For Microsoft: Validate link to internal work items`, which fails because issue #9100 has no `AB#` internal work item. As you pointed out on #9180, that one needs a maintainer, there is nothing I can set from the fork side.

alexei-dobriansky:
> Thanks for the review alexei-dobriansky.
> 
> On S1, that is a fair point. I kept the two passes inline because it is the shape that was reviewed on the pilot pull request and the one named in the approval on #9100, so I wanted the diff to stay as close to the reviewed version as possible. If you would rather have the shared loop pulled into a small local procedure taking `(Table ID, No.)`, say the word and I will push it, it is behaviour neutral either way.
> 
> One note on the checks: the only red one is `For Microsoft: Validate link to internal work items`, which fails because issue #9100 has no `AB#` internal work item. As you pointed out on #9180, that one needs a maintainer, there is nothing I can set from the fork side.

No problem, we'll handle AB# ourselves.

alexei-dobriansky (dismissed):
Please finish with S1.

Franco111000:
S1 done in c695ac1b7e. The shared filter and `InsertDim` loop now live in `InsertDefaultDimensions(TableID, No.)`, called once for the Item and once for the Location, so the two passes are gone. Same change in the RU copy, and the helper is identical in both. Behaviour is unchanged and the regression test is untouched.

Thanks for handling the `AB#` side.

alexei-dobriansky:

# **Recommendation: Accept**

<details>

The new commit extracts the repeated Default Dimension read into InsertDefaultDimensions(TableID, No.), called once for the Item pair and once for the Location pair. This keeps the fix scoped to the intended (Table ID, No.) combinations, so an unrelated Location with a code equal to the Item No. can no longer leak dimensions into the physical inventory line.

The change is behavior-neutral compared with the previous round. It keeps the same Dimension Value Code filter, the same InsertDim flow, the same cleanup after DimBufMgt.InsertDimensions(), and the same mirrored change in the RU report copy. The regression test still covers the collision case.
</details>

<details>

| S1 | The two passes are near-identical; a small local helper would remove the duplication. | Addressed | The shared filter and InsertDim loop now live in InsertDefaultDimensions(TableID, No.) in both report copies. |
</details>

<details>

*None - the new commit only addresses the previous suggestion.*
</details>

<details>

**Risk:** Low. The new helper is local to report 790 and only moves the same Default Dimension query and insert loop behind a small procedure. There is no public surface, event, schema, or upgrade impact.

**Necessity:** Still justified. The change prevents wrong default dimensions from being copied to physical inventory journal lines when Item and Location identifiers overlap, and the code now avoids the duplicated loop that made the two report copies harder to keep aligned.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10575 round=2 by=alexei-dobriansky at=2026-08-25T14:20:29.8876313Z lastSha=c695ac1b7e7d4904364c05d435e70bb86d092e32 reviewKey=7992f82a4e3670b59a593a653635849f21b111ca39be29f185c35d72255fd251 suggestions=S1@492e9cf9:addressed parentRound=1

alexei-dobriansky (approved):

PredragMaricic (approved):
The scoped Default Dimension reads fix the cross-product defect while preserving the existing dimension-buffer flow. The regression test covers both the intended counted-location dimension and exclusion of the unrelated same-code location dimension. No blocking findings.

PR #10600, opened 2026-08-25
[Master] - Slice 641282: [Sustainability][Value Chain] FA Reclassification Journal: add CO2e reclassification % field to split emissions like Acq. Cost %

v-rohangarg20:
## Description
Adds a new ` Reclassify CO2e % ` field to the FA Reclassification Journal so CO2e emissions can be reclassified independently of ` Reclassify Acq. Cost % `.

- ` SustFAReclassJournalLine.TableExt.al `: new ` Reclassify CO2e % ` field with validation requiring ` Sust. Account No. `/` New Sust. Account No. ` to be set.
- ` SustFAReclassJournal.PageExt.al `: exposes the new field on the page.
- ` SustFAJournalSubscriber.Codeunit.al `: uses ` Reclassify CO2e % ` (instead of ` Reclassify Acq. Cost % `) to compute the CO2e amount to reclassify, and requires the field to have a value.
- Tests: ` TestFAReclassJournalLineWithDifferentCO2ePercent ` and ` TestFAReclassJournalLineWithZeroCO2ePercent `.

Fixes [AB#641282](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/641282)

### DIFF
--- a/src/Apps/W1/Sustainability/app/src/FixedAssets/SustFAJournalSubscriber.Codeunit.al
+++ b/src/Apps/W1/Sustainability/app/src/FixedAssets/SustFAJournalSubscriber.Codeunit.al
@@ -116,6 +116,7 @@ codeunit 6283 "Sust. FA Journal Subscriber"
     begin
         FAReclassJournalLine.TestField("Sust. Account No.");
         FAReclassJournalLine.TestField("New Sust. Account No.");
+        FAReclassJournalLine.TestField("Reclassify CO2e %");
         FAReclassJournalLine.CheckSustainabilityAccount(FAReclassJournalLine."Sust. Account No.");
         FAReclassJournalLine.CheckSustainabilityAccount(FAReclassJournalLine."New Sust. Account No.");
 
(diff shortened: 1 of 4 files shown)


alexei-dobriansky:

<details>

This adds a dedicated CO2e reclassification percentage to the FA Reclassification Journal and uses it when generated journal lines receive sustainability CO2e. The BaseApp transfer events fire before the generated FA and general journal lines are inserted, so the subscriber can still set the sustainability account and Total CO2e; the explicit-percent path is covered by tests, but the default-to-acquisition-cost-percent path is not implemented.
</details>

<details>

**Fit:** Partial

The requested behavior is to let emissions split independently while defaulting to the acquisition-cost split. The diff supports an explicitly entered CO2e percentage, but a normal line where the new percentage is left at its default now fails instead of using the acquisition-cost percentage.
</details>

<details>

**S1** (🔴 High): *Default CO2e percent before transfer*
The new field is required, but nothing copies the acquisition-cost percentage into it when the line is created or when an acquisition cost amount later calculates the percentage. A normal reclassification can therefore fail with "Reclassify CO2e % must have a value" instead of using the default split described by the feature. Initialize the field from the acquisition-cost percentage before transfer, while still allowing an explicit different percentage.
</details>

<details>

**Risk:** This is in the fixed-asset reclassification flow that creates FA or general journal lines and posts sustainability ledger CO2e. If the default percentage is not set, valid reclassifications with sustainability accounts can be blocked; if users work around it manually, the split depends on extra manual entry.

**Necessity:** A dedicated CO2e percentage is needed for 1-to-many fixed-asset reclassification so emissions can follow a different split than acquisition cost. The scope is targeted, but it must keep the old default behavior before adding the independent override.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10600 round=1 by=alexei-dobriansky at=2026-08-25T12:24:15Z lastSha=0d009f45cae6eacac984d888ce817cf35f2fc025 reviewKey=bd59c5dbcf72b9c61628e700d583dc1c67ec6c91c826298b98a0527fff6fbdf5 suggestions=S1@1c06753a

djukicmilica (approved):

AndersLarsenMicrosoft (approved):

PR #10607, opened 2026-08-25
[Master]--Inconsistent Validation of Non-Deductible VAT % for Purchase Lines Sharing the Same VAT Identifier

neeleshsinghal:


alexei-dobriansky:

<details>

This PR blocks purchase lines with the same VAT Identifier from mixing zero and non-zero `Non-Deductible VAT %` values. The added table check covers the order where the current line is set to 0%, while the existing codeunit check still covers the non-zero order. The rule matches the bug, and the tests cover both directions, but the new check is placed outside the existing handled validation path.
</details>

<details>

**Fit:** Partial

The change addresses the reported inconsistent validation and updates the affected Purchase Line copies. The implementation should still preserve the existing validation contract for extensions that handle this check.
</details>

<details>

**S1** (🔴 High): *Keep the check inside the handled validation*
The new table-level call runs after `OnBeforeCheckNonDeductibleVATPctIsAllowed`, so subscribers that set `IsHandled` can still get this new error for 0%. Move the zero-percent branch into `CheckNonDeductibleVATPctIsAllowed` after the existing `IsHandled` check, or add an equivalent handled event before this error. This preserves the existing validation contract while still closing the bug.
</details>

<details>

**Risk:** This is Purchase Line VAT validation copied across many layers. A wrong placement can block purchase documents, API writes, imports, and extensions that use the existing handled event to replace this validation.

**Necessity:** The bug is valid because a document can otherwise keep mixed Non-Deductible VAT percentages under the same VAT Identifier depending on line entry order. The scope is needed, and the tests cover both orders, but the fix should use the existing validation extension point.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10607 round=1 by=alexei-dobriansky at=2026-08-25T18:06:11Z lastSha=75b2b22def5b793dcd3f97886754979bf7eff7d1 reviewKey=12040ba9aa09502642daadd4e7357a18905fbe0a72bc3fb67f2d7154940c2a89 suggestions=S1@44c8aee0

PR #10665, opened 2026-08-26
[SOA] Skip content extraction for image attachments

qasimikram:
## Summary

The Sales Order Agent runs every incoming attachment through the content extraction model (`SOA Attachment MLLM`). The extraction model rejects image attachments, so image-based attachments fail extraction and the attachment accuracy tests that rely on them do not pass.

This change skips the extraction call for image attachments and falls back to the text content already stored with the attachment. Relevance is still evaluated on that content, so image attachments continue to go through the normal relevance gate.

## Changes

**`src/Setup/SOASetup.Codeunit.al`**

Adds `IsImageAttachmentContentType`. It recognizes the standard image MIME types and also the non-standard `application/<extension>` values, because the platform derives the content type from the file extension and maps image extensions onto `application/<extension>`.

**`src/Integration/SOAAttachmentMLLM.Codeunit.al`**

For an image attachment, the canonical content is set to the content already stored with the attachment and the guarded extraction call is skipped. Non-image attachments are unaffected and still go through the existing guarded extraction path, including the system-ID marker check.

**`.resources/Prompts/SalesOrderAgent-AgentInstructions.json`**

Three instruction lines:

- `Check Requested Item Exists` - search every item requested in the request, including all items in the message body and all items in every attachment.
- Responsibilities - a request to create a sales order also means a sales quote is wanted first.
- Responsibilities - an attached document is processed regardless of its own document type, title, or file name. Relevance is already decided before the document reaches the agent.

## Temporary by design

The image skip is explicitly marked temporary in the code comment. It should be removed once the extraction model accepts image attachments, at which point image attachments should go back through the same guarded extraction path as documents.

## Testing

Test-side changes live in the NAV repo (`App/Internal/Apps/SalesOrderAgent/test`) and are submitted separately against the same work item.

[AB#647838](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/647838)

### DIFF
--- a/src/Apps/W1/SalesOrderAgent/app/.resources/Prompts/SalesOrderAgent-AgentInstructions.json
+++ b/src/Apps/W1/SalesOrderAgent/app/.resources/Prompts/SalesOrderAgent-AgentInstructions.json
@@ -15,6 +15,8 @@
 				"Handles requests for inquiries about items and services.",
 				"Handles requests for modifying the item lines on a sales quote created as part of the current conversation.",
 				"Handles requests for converting a sales quote created as part of the current conversation into a sales order.",
+				"Handles requests for creating a sales order. A request to create a sales order also means that a sales quote is wanted first, so always create the sales quote first and then continue with the sales order instructions. Never reply that creating a sales order is not supported.",
+				"Handles requests that arrive as an attached document. Any document attached to the request has already been assessed as relevant before it reached you, so never judge, question, or re-evaluate whether a document is relevant or intended for you. Whenever an attached document contains content you can process, process it as a sales quote request regardless of the document's own type, title, or file name, for example a purchase order, purchase quote, request for quotation, service order, order confirmation, delivery note, or invoice. Never reply that the document type is not supported and never ask the customer to confirm that the document is a sales quote request.",
 				"Request to modify an existing sales order is not supported."
 			]
 		},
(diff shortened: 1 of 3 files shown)


AleksandricMarko (approved):

VolodySyn (approved):

alexei-dobriansky:

<details>

The change avoids the extraction model for image attachments and lets the existing attachment text continue through the relevance check. The main flow is correct for supported image attachments once they reach `SOA Attachment MLLM`: missing files still take the old extraction/error path, image content is not sent to the extraction model, and non-image attachments keep the provenance-checked canonical extraction. The prompt changes also align the agent with processing relevant attached documents and searching all requested items.
</details>

<details>

**Fit:** Partial

The fix addresses the failure at the extraction step for images that reach the attachment analysis code. One upstream supported-type check still appears narrower than the new image-type helper, so some non-standard image MIME values can be ignored before the new skip logic runs.
</details>

<details>

**S1** (🟠 Moderate): *Align supported image MIME checks*
`IsImageAttachmentContentType` accepts `application/png`, `application/jpg`, `application/jpeg`, and `application/gif`, but `SupportedAttachmentContentType` still only accepts the standard image MIME types used before. If an image enters through the email attachment path with one of the non-standard values, it can be marked unsupported before this new helper is called. Please reuse this helper there or keep the supported-type list in sync.
</details>

<details>

**Risk:** The regression surface is limited to Sales Order Agent attachment handling and the prompt behavior for attached documents. Non-image extraction still uses the existing guarded path, while image attachments now depend on the text already stored with the attachment and on the relevance check that follows.

**Necessity:** The change is needed because image attachments otherwise fail before the relevant content can be used. The scope is small and fits the problem, but the supported MIME-type gate should be kept in sync with the new image helper so the same image formats are handled consistently.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10665 round=1 by=alexei-dobriansky at=2026-08-26T18:11:47.1565458Z lastSha=8579e7bda40e7a34d6b3ebe282427f709c4d7f58 reviewKey=1f9a3e52c69a035df416744222cf133cc59d33435ff03eafe05cb3ea571608cc suggestions=S1@a4d786a5

PR #10670, opened 2026-08-26
Add Agent Task Log JSON export

nikolakukrika:
## Summary
- add a selected-row JSON export action to the Agent Task Log list
- expose the export through an internal stream-based API for test automation
- include detail-page calculated values and related task context, messages, sibling actions, page context, tools, memorized data, and page settings
- enforce agent-management access and redact serialized page snapshots without Troubleshoot All Agents
- share calculation helpers with the detail page and add focused JSON/context tests

## Validation
- reviewed the complete AL diff against the requested behavior
- checked the patch with `git diff --check`
- direct AL compiler parsing reported no diagnostics in the changed application files; a full BC 29 build requires the enlistment's matching local symbols

[AB#624671](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/624671)


src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogEntry.Codeunit.al
```diff
@@ -116,6 +117,387 @@ codeunit 4314 "Agent Task Log Entry"
         exit(TextEncoding::UTF8);
     end;
 
+    procedure ExportToJson(var SelectedAgentTaskLogEntry: Record "Agent Task Log Entry"; var ExportOutStream: OutStream)
```
nikolakukrika: Lets move all export to JSON code to the dedicated codeunit. No need to polute the AgentTaskLog entry codeunit.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogEntry.Codeunit.al
```diff
@@ -116,6 +117,387 @@ codeunit 4314 "Agent Task Log Entry"
...
+        Agent: Record Agent;
+        AgentTaskImpl: Codeunit "Agent Task Impl.";
+    begin
+        if AgentTaskImpl.TryGetAgentRecordFromTaskId(AgentTaskLogEntry."Task ID", Agent) then
+            exit(Agent."Display Name");
+
+        exit('');
+    end;
+
+    procedure TryGetSuccess(MemoryEntryDetailsTxt: Text; var Success: Boolean): Boolean
```
nikolakukrika: No need to Put try, GetSuccess is sufficient. The other methods do not follow the same naming.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogEntry.Codeunit.al
```diff
@@ -116,6 +117,387 @@ codeunit 4314 "Agent Task Log Entry"
...
+        PageStack: JsonArray;
+        ExportPageStack: JsonArray;
+        PageToken: JsonToken;
+        Index: Integer;
+    begin
+        PageStack := ContextRoot.GetArray(PagestackLbl, true);
+        for Index := PageStack.Count() - 1 downto 0 do begin
+            PageStack.Get(Index, PageToken);
+            if not PageToken.AsValue().IsNull() then
+                ExportPageStack.Add(PageToken.AsValue().AsText());
```
nikolakukrika: We should also add the order, so it is clear which page is on the bottom and which page is on top.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogEntry.Codeunit.al
```diff
@@ -116,6 +117,387 @@ codeunit 4314 "Agent Task Log Entry"
...
+        ContextJson.Add(AvailableToolsLbl, EmptyAvailableTools);
+        ContextJson.Add(MemorizedDataLbl, EmptyMemorizedData);
+        ContextJson.Add(TaskPageSettingsLbl, EmptyTaskPageSettings);
+
+        if (ContextTxt = '') or not ContextRoot.ReadFrom(ContextTxt) then
+            exit;
+
+        ContextJson.Replace(ValidLbl, true);
+        ContextJson.Replace(DecisionPointLbl, GetDecisionPoint(ContextRoot));
+        ContextJson.Replace(ExportPageStackLbl, BuildPageStackJson(ContextRoot));
```
nikolakukrika: Replace is very inefficient and error prone. Lets set the variables properly, check if it exists, if it does not do not add it to the context JSON to keep it smaller.

src/System Application/Test/Agent/src/Troubleshooting/AgentTaskLogPageTest.Codeunit.al
```diff
@@ -173,30 +177,100 @@ codeunit 133964 "Agent Task Log Page Test"
...
-        TempMemorizedDataRecords.SetJsonText('{"key":"key1","value":"value1"}');
-        TempMemorizedDataRecords.Id := 2;
-        TempMemorizedDataRecords.Insert();
-        TempMemorizedDataRecords.SetJsonText('{"key":"key2","value":"value2"}');
+        MemorizedDataObject.Add('key1', 'value1');
+        MemorizedDataObject.Add('key2', 'value2');
+        ContextRootObject.Add('memorizedData', MemorizedDataObject);
 
         ContextRootObject.Add('isDecisionPoint', true);
-        ContextRootObject.Add('success', true);
```
nikolakukrika, line 186: Why did you remove this line and removed the temp records from the test? Why it was needed?

src/System Application/App/Agent/Troubleshooting/AgentTaskLogEntryList.Page.al
```diff
@@ -201,6 +205,31 @@ page 4303 "Agent Task Log Entry List"
...
+                    AgentTaskLogEntry: Codeunit "Agent Task Log Entry";
+                    ExportInStream: InStream;
+                    ExportOutStream: OutStream;
+                    FileName: Text;
+                begin
+                    CurrPage.SetSelectionFilter(SelectedAgentTaskLogEntry);
+                    TempBlob.CreateOutStream(ExportOutStream, TextEncoding::UTF8);
+                    AgentTaskLogEntry.ExportToJson(SelectedAgentTaskLogEntry, ExportOutStream);
+                    TempBlob.CreateInStream(ExportInStream, TextEncoding::UTF8);
+                    FileName := StrSubstNo(ExportFileNameLbl, Format(CurrentDateTime(), 0, '<Year4><Month,2><Day,2>_<Hours24,2><Minutes,2><Seconds,2>'));
```
nikolakukrika: Can we use one of the built in formats instead?

src/System Application/App/Agent/Troubleshooting/AgentTaskLogEntry.Page.al
```diff
@@ -301,57 +300,19 @@ page 4312 "Agent Task Log Entry"
...
-            Rec.Type::"User Intervention Request":
-                IsAgentAction := true;
-            // Operations performed by a user or the agent.
-            Rec.Type::Stop:
-                IsAgentAction := User.Get(Rec."User Security ID")
-                    ? User."License Type" = User."License Type"::Agent
-                    : Default;
-            // By default, consider it was a user action.
-            else
-                IsAgentAction := Default;
```
nikolakukrika, line 353: Shouldn't we still set the IsAgentAction from this method?

src/System Application/App/Agent/Troubleshooting/AgentTaskLogEntry.Page.al
```diff
@@ -301,57 +300,19 @@ page 4312 "Agent Task Log Entry"
...
+        AgentTaskLogEntry: Codeunit "Agent Task Log Entry";
+        Success: Boolean;
     begin
-        if not RootObj.ReadFrom(MemoryEntryDetailsTxt) then
-            exit;
-
-        IsSuccess := RootObj.Contains(SuccessLbl)
-            ? Format(RootObj.GetBoolean(SuccessLbl, true))
+        IsSuccess := AgentTaskLogEntry.TryGetSuccess(MemoryEntryDetailsTxt, Success)
+            ? Format(Success)
```
nikolakukrika: This one  will not work, we need to split it in multiple statements with else. We should not call Format(Success)

src/System Application/App/Agent/Troubleshooting/AgentTaskLogEntryList.Page.al
```diff
@@ -240,4 +269,7 @@ page 4303 "Agent Task Log Entry List"
         IsFeedbackActionEnabled: Boolean;
         DetailsTxt: Text;
         TypeStyle: Text;
+        ExportFileNameLbl: Label 'AgentTaskLog_%1.json', Comment = '%1 is a timestamp.', Locked = true;
```
nikolakukrika: Can we put the agent name too?

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogEntry.Codeunit.al
```diff
@@ -116,6 +117,387 @@ codeunit 4314 "Agent Task Log Entry"
...
+
+    procedure GetDecisionPoint(ContextRoot: JsonObject): Boolean
+    begin
+        if ContextRoot.Contains(IsDecisionPointLbl) then
+            exit(ContextRoot.GetBoolean(IsDecisionPointLbl, true));
+
+        exit(false);
+    end;
+
+    procedure ReadContextWithFallback(AgentTaskLogEntry: Record "Agent Task Log Entry"; AgentTaskMemoryEntry: Record "Agent Task Memory Entry"; var ContextSource: Text): Text
```
nikolakukrika: ReadContextSafe

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogEntry.Codeunit.al
```diff
@@ -116,6 +117,387 @@ codeunit 4314 "Agent Task Log Entry"
...
+
+        CommunicationCulture := TaskPageContext.GetObject(OutgoingCommunicationCultureLbl, true);
+        TaskPageSettings.Add(CommunicationLanguageLbl, CommunicationCulture.GetText(CommunicationLanguageSourceLbl, true));
+        TaskPageSettings.Add(CommunicationDateFormatLbl, CommunicationCulture.GetText(CommunicationDateFormatSourceLbl, true));
+        TaskPageSettings.Add(CommunicationTimeFormatLbl, CommunicationCulture.GetText(CommunicationTimeFormatSourceLbl, true));
+        TaskPageSettings.Add(CommunicationFormattedNumberExampleLbl, CommunicationCulture.GetText(CommunicationFormattedNumberExampleSourceLbl, true));
+        exit(TaskPageSettings);
+    end;
+
+    local procedure ReplaceContextToken(var ContextJson: JsonObject; ContextRoot: JsonObject; SourcePropertyName: Text; TargetPropertyName: Text; ExpectArray: Boolean)
```
nikolakukrika: Same comment as with other places, we should not use replace

src/System Application/Test/Agent/src/Troubleshooting/AgentTaskLogPageTest.Codeunit.al
nikolakukrika: The previous test populated the temporary result records directly and then asserted those same records, so it did not exercise ExtractMemorizedData. The updated setup puts memorizedData into the source context and calls the production extraction method. The success value is separate from context parsing—it comes from memory-entry details—so it is now tested directly through GetSuccess.

src/System Application/App/Agent/Troubleshooting/AgentTaskLogEntry.Page.al
```diff
@@ -458,8 +428,8 @@ page 4312 "Agent Task Log Entry"
         IsLogPageVisible: Boolean;
         GlobalCurrentID: Integer;
         SerializedPageLbl: Label 'serializedPage', Locked = true;
-        IsDecisionPointLbl: Label 'isDecisionPoint', Locked = true;
         TaskPageContextLbl: Label 'taskPageContext', Locked = true;
-        SuccessLbl: Label 'success', Locked = true;
+        YesLbl: Label 'Yes';
```
nikolakukrika: This looks wrong. Shouldn't we use true and false that locked or we use Format(success, 0, 9)

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
```diff
@@ -0,0 +1,398 @@
...
+        TaskContextLbl: Label 'taskContext', Locked = true;
+        SourceLbl: Label 'source', Locked = true;
+        SerializedPageRedactedLbl: Label 'serializedPageRedacted', Locked = true;
+        SerializedPageRedactionReasonLbl: Label 'serializedPageRedactionReason', Locked = true;
+        PageStackLbl: Label 'pageStack', Locked = true;
+        OrderLbl: Label 'order', Locked = true;
+        TaskPageSettingsLbl: Label 'taskPageSettings', Locked = true;
+        CurrencyCodeLbl: Label 'currencyCode', Locked = true;
+        CurrencySymbolLbl: Label 'currencySymbol', Locked = true;
+        OutgoingCommunicationCultureLbl: Label 'outgoingCommunicationCulture', Locked = true;
```
nikolakukrika, line 616: lets have a nesting here. The parent element should be communication, then we should have culture, language, DateFormat, TimeFormat

src/System Application/App/Agent/Troubleshooting/AgentTaskLogEntryList.Page.al
```diff
@@ -235,9 +264,39 @@ page 4303 "Agent Task Log Entry List"
...
         IsFeedbackActionEnabled: Boolean;
         DetailsTxt: Text;
         TypeStyle: Text;
+        ExportFileNameLbl: Label 'AgentTaskLog_%1_%2.json', Comment = '%1 is the agent name and %2 is the export date.', Locked = true;
+        ExportDialogTitleLbl: Label 'Export agent task log';
+        JsonFileFilterLbl: Label 'JSON files (*.json)|*.json', Locked = true;
+        UnknownAgentTok: Label 'UnknownAgent', Locked = true;
+        MultipleAgentsTok: Label 'MultipleAgents', Locked = true;
+        InvalidFileNameCharactersTok: Label '\/:*?"<>|', Locked = true;
+        FileNameCharacterReplacementsTok: Label '_________', Locked = true;
```
nikolakukrika: No need for this kind of complexity, simply remove all characters that are not letters or numbers.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
```diff
@@ -0,0 +1,398 @@
...
+        Index: Integer;
+        Order: Integer;
+    begin
+        PageStack := ContextRoot.GetArray(PageStackLbl, true);
+        for Index := PageStack.Count() - 1 downto 0 do begin
+            PageStack.Get(Index, PageToken);
+            if PageToken.AsValue().IsNull() then
+                continue;
+
+            Order += 1;
```
nikolakukrika: INCSTR

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
```diff
@@ -0,0 +1,398 @@
...
+    var
+        PageStack: JsonArray;
+        ExportPageStack: JsonArray;
+        PageStackEntry: JsonObject;
+        PageToken: JsonToken;
+        Index: Integer;
+        Order: Integer;
+    begin
+        PageStack := ContextRoot.GetArray(PageStackLbl, true);
+        for Index := PageStack.Count() - 1 downto 0 do begin
```
nikolakukrika: Why do we reverse the order?

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
```diff
@@ -0,0 +1,398 @@
...
+    end;
+
+    local procedure BuildMessageArray(AgentTaskLogEntry: Record "Agent Task Log Entry"; InputMessages: Boolean; EarlierMessages: Boolean): JsonArray
+    var
+        AgentTaskMessage: Record "Agent Task Message";
+        Messages: JsonArray;
+    begin
+        AgentTaskMessage.SetRange("Task ID", AgentTaskLogEntry."Task ID");
+        if EarlierMessages then
+            AgentTaskMessage.SetFilter("Memory Entry ID", '<%1', AgentTaskLogEntry."Memory Entry ID")
```
nikolakukrika: Dont we need to set the Task ID too?

nikolakukrika: Check all set ranges and set filters in the PR, if we are missing an additional filter or if we need an extra.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
```diff
@@ -0,0 +1,398 @@
...
+
+        Clear(Messages);
+        Messages := BuildMessageArray(AgentTaskLogEntry, false, true);
+        if Messages.Count() > 0 then
+            MessagesJson.Add(EarlierLbl, Messages);
+
+        exit(MessagesJson);
+    end;
+
+    local procedure BuildMessageArray(AgentTaskLogEntry: Record "Agent Task Log Entry"; InputMessages: Boolean; EarlierMessages: Boolean): JsonArray
```
nikolakukrika: Lets pass records with var, it is faster

src/System Application/App/Agent/Troubleshooting/AgentTaskLogEntry.Page.al
nikolakukrika: Updated to use Format(Success, 0, 9), which emits the invariant Boolean value.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
nikolakukrika: Updated the structure to taskPageSettings.communication.culture, with language, dateFormat, timeFormat, and formattedNumberExample inside culture.

src/System Application/App/Agent/Troubleshooting/AgentTaskLogEntryList.Page.al
nikolakukrika: Simplified the filename component sanitization to retain only ASCII letters and numbers, with UnknownAgent as the empty fallback.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
nikolakukrika: Updated order to text initialized as 1 and incremented with IncStr.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
nikolakukrika: Removed the reversal. Page stack entries now preserve their stored order and include an explicit order value.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
nikolakukrika: Task ID was already applied before the Memory Entry ID filter and remains in place.

nikolakukrika: Reviewed all ranges and filters. Messages are bounded by Task ID and Memory Entry ID; sibling actions by Task ID and Memory Entry ID while excluding the current entry; task export by Task ID. I removed the additional message Type filter so the export matches the detail-page parts.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
nikolakukrika: Updated record parameters in the export helpers to pass by var.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogEntry.Codeunit.al
```diff
@@ -111,11 +111,84 @@ codeunit 4314 "Agent Task Log Entry"
         ContentInStream.ReadText(ContextTxt);
     end;
 
+    procedure ReadContextSafe(AgentTaskLogEntry: Record "Agent Task Log Entry"; AgentTaskMemoryEntry: Record "Agent Task Memory Entry"; var ContextSource: Text): Text
+    var
+        ContextTxt: Text;
+    begin
+        ContextTxt := ReadContext(AgentTaskLogEntry);
```
nikolakukrika: This seems wrong. If we have agent task entry we will read it, but if we do not then we will read the agent task memory entry. Why do we have such logic? Isn't it better to have 2 methods one to read from the AgentTaskLogEntry and one to read from AgentTaskMemoryEntry.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogEntry.Codeunit.al
```diff
@@ -111,11 +111,84 @@ codeunit 4314 "Agent Task Log Entry"
         ContentInStream.ReadText(ContextTxt);
     end;
 
+    procedure ReadContextSafe(AgentTaskLogEntry: Record "Agent Task Log Entry"; AgentTaskMemoryEntry: Record "Agent Task Memory Entry"; var ContextSource: Text): Text
+    var
+        ContextTxt: Text;
+    begin
```
nikolakukrika: Shouldn't we have a check here if the agent task log entry is ID is zero to exit early?

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
```diff
@@ -0,0 +1,444 @@
...
+    end;
+
+    procedure ExportToJsonFile(var SelectedAgentTaskLogEntry: Record "Agent Task Log Entry"; AgentName: Text)
+    var
+        TempBlob: Codeunit "Temp Blob";
+        ExportInStream: InStream;
+        ExportOutStream: OutStream;
+        FileName: Text;
+    begin
+        TempBlob.CreateOutStream(ExportOutStream, TextEncoding::UTF8);
```
nikolakukrika: Lets not hardcode UTF8 everywhere, lets add a method called GetDefaultEncoding. Update the other references.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
```diff
@@ -0,0 +1,444 @@
...
+    begin
+        AgentTaskMemoryEntry.CalcFields(Details);
+        AgentTaskMemoryEntry.Details.CreateInStream(ContentInStream, AgentTaskLogEntry.GetDefaultEncoding());
+        ContentInStream.Read(MemoryDetailsTxt);
+    end;
+
+    procedure BuildContextJson(ContextTxt: Text; ContextSource: Text; IncludeSerializedPage: Boolean; var ContextJson: JsonObject)
+    var
+        ContextRoot: JsonObject;
+        JsonToken: JsonToken;
```
nikolakukrika: Lets not name JsonToken but lets give it a name. It should not overlap with the built in types. Please update all other references where it is named JsonToken: JsonToken.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
```diff
@@ -0,0 +1,444 @@
...
+        Character: Text;
+        Index: Integer;
+    begin
+        for Index := 1 to StrLen(Value) do begin
+            Character := CopyStr(Value, Index, 1);
+            if (Character in ['0' .. '9']) or
+               (Character in ['A' .. 'Z']) or
+               (Character in ['a' .. 'z'])
+            then
+                AlphaNumericValue += Character;
```
nikolakukrika: Better pattern is like this:
procedure RemoveNonAlphanumeric(InputText: Text): Text
var
    AllowedChars: Text;
begin
    // Define all allowed uppercase, lowercase, and numeric characters
    AllowedChars := 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
    
    // Nested DELCHR removes any character NOT present in AllowedChars
    exit(DelChr(InputText, '=', DelChr(InputText, '=', AllowedChars)));
end;

src/System Application/App/Agent/Troubleshooting/AgentTaskLogEntry.Page.al
```diff
@@ -301,57 +300,29 @@ page 4312 "Agent Task Log Entry"
...
-        if not AgentTaskImpl.TryGetAgentRecordFromTaskId(Rec."Task ID", Agent) then begin
-            AgentName := '';
-            exit;
-        end;
-
-        AgentName := Agent."Display Name";
+        AgentName := AgentTaskLogEntry.GetAgentName(Rec);
     end;
 
     local procedure SetIsAgentAction()
```
nikolakukrika, line 314: This local procedure makes no sense any more. It is one liner. We can replace it wherever it is called.

src/System Application/Test/Agent/src/Troubleshooting/AgentTaskLogPageTest.Codeunit.al
```diff
@@ -173,30 +177,101 @@ codeunit 133964 "Agent Task Log Page Test"
...
+        AgentTaskLogEntryRecord: Record "Agent Task Log Entry";
+        AgentTaskLogExport: Codeunit "Agent Task Log Export";
+        TempBlob: Codeunit "Temp Blob";
+        ExportJson: JsonObject;
+        ExportInStream: InStream;
+        ExportOutStream: OutStream;
+        ExportTxt: Text;
+    begin
+        // [GIVEN] A selection filter that cannot match a log entry
+        AgentTaskLogEntryRecord.SetRange(ID, -1);
```
nikolakukrika: We cannot write test like this. There will be no AgentTaskLogEntryRecord in the database. We need to either mock or remove the test.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogEntry.Codeunit.al
nikolakukrika: There are already two typed ReadContext overloads, one for Agent Task Log Entry and one for Agent Task Memory Entry. ReadContextSafe intentionally defines the precedence used by both the detail page and exporter: prefer the log-entry troubleshooting snapshot when present, then fall back to the memory-entry context because older or incomplete log entries may not contain their own snapshot. Splitting this decision into both callers would duplicate the same fallback logic. I agree with the adjacent comment that we should avoid reading the log-entry BLOB when its ID is zero.

src/System Application/Test/Agent/src/Troubleshooting/AgentTaskLogPageTest.Codeunit.al
nikolakukrika: This test intentionally exercises the zero-matching-record path. SetRange(ID, -1) guarantees that FindSet returns false; no Agent Task Log Entry is expected to exist or be read. The assertion verifies that the programmatic API still returns a valid versioned document with an empty entries array, so mocking or inserting a record would change the behavior under test.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogEntry.Codeunit.al
```diff
@@ -111,11 +111,84 @@ codeunit 4314 "Agent Task Log Entry"
         ContentInStream.ReadText(ContextTxt);
     end;
 
+    procedure ReadContextSafe(AgentTaskLogEntry: Record "Agent Task Log Entry"; AgentTaskMemoryEntry: Record "Agent Task Memory Entry"; var ContextSource: Text): Text
```
nikolakukrika: As we have discussed offline we need to 
Export them independently:

•  context : projected from the log entry’s  "Troubleshooting Info" .
•  memory.context : projected from the memory entry’s  Context .
•  memory.details : retain existing raw details.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogEntry.Codeunit.al
nikolakukrika: Updated after our discussion. ReadContextSafe was removed: the detail page now performs its display fallback explicitly, while the exporter calls the two typed ReadContext overloads independently.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogEntry.Codeunit.al
nikolakukrika: Added early exits before reading both record BLOBs: log entries require a nonzero ID, and memory entries require a nonzero Task ID.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
nikolakukrika: Added GetDefaultEncoding to the export codeunit and updated its stream creation and memory-details reading to use it.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
nikolakukrika: Renamed the variables to SerializedPageToken and ContextPropertyToken so they no longer overlap with the JsonToken type name.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
nikolakukrika: Replaced the character loop with the nested DelChr pattern and a locked allowed-characters token.

src/System Application/App/Agent/Troubleshooting/AgentTaskLogEntry.Page.al
nikolakukrika: Removed SetAgentName and assigned the value directly at its only call site.

src/System Application/Test/Agent/src/Troubleshooting/AgentTaskLogPageTest.Codeunit.al
nikolakukrika: Removed the empty-selection test as agreed.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogEntry.Codeunit.al
nikolakukrika: Implemented independent export: top-level context is projected from the log entry Troubleshooting Info, memory.context is projected from the memory entry Context, and memory.details remains unchanged. Snapshot redaction is applied independently through the shared projection method.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
```diff
@@ -363,22 +367,17 @@ codeunit 4313 "Agent Task Log Export"
...
-            if (Character in ['0' .. '9']) or
-               (Character in ['A' .. 'Z']) or
-               (Character in ['a' .. 'z'])
-            then
-                AlphaNumericValue += Character;
-        end;
+        AlphanumericValue := DelChr(Value, '=', DelChr(Value, '=', AllowedAlphanumericCharactersTok));
+
+        if AlphanumericValue = '' then
+            AlphanumericValue := UnknownAgentTok;
```
nikolakukrika: We should not set the unknownagent if we have removed the alphanumeric values and got the blank string. These two are not related. Please return the blank value if the string is blank.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
nikolakukrika: Shouldn't we also read the MemoryContext JSON or is this done at a different place?

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
nikolakukrika: Updated the sanitizer to return only the filtered value. It no longer substitutes UnknownAgent when the result is blank.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
```diff
@@ -370,9 +370,6 @@ codeunit 4313 "Agent Task Log Export"
     local procedure RemoveNonAlphanumericCharacters(Value: Text) AlphanumericValue: Text
     begin
         AlphanumericValue := DelChr(Value, '=', DelChr(Value, '=', AllowedAlphanumericCharactersTok));
```
nikolakukrika, line 571: Nit, we should exit the aplhanumericvalue so we are clear about the  intent.

src/System Application/Test/Agent/src/Troubleshooting/AgentTaskLogPageTest.Codeunit.al
```diff
@@ -173,30 +178,236 @@ codeunit 133964 "Agent Task Log Page Test"
...
+    [Test]
+    procedure TestBuildContextJson_IncludesCalculatedDetails()
+    var
+        AgentTaskLogExport: Codeunit "Agent Task Log Export";
+        ContextJson: JsonObject;
+        PageToken: JsonToken;
+        PageStackEntry: JsonObject;
+        ContextTxt: Text;
+    begin
+        // [GIVEN] Troubleshooting context containing all calculated detail-page sections
```
ventselartur: Please, give it a try when implementing a new test :)
https://microsoft.ghe.com/bic/BC-agentic-workspace/blob/master/bc-ai-artifacts/application/common/skills/bc-test-implementor/SKILL.md

src/System Application/Test/Agent/app.json
```diff
@@ -40,6 +40,12 @@
       "publisher": "Microsoft",
       "name": "Library Variable Storage",
       "version": "30.0.0.0"
+    },
+    {
+      "id": "40860557-a18d-42ad-aecb-22b7dd80dc80",
```
nikolakukrika: Permission mock is not needed, lets remove the permissions from testing of this functionality

alexei-dobriansky:

<details>

This adds JSON export actions for agent task logs. The export builds task context, log entries, memory entries, messages, related entries, and page context, and it redacts page snapshots unless the current user has the troubleshooting permission.

The main flow is targeted and uses the selected log-entry filter or current task ID before building JSON. The shared helpers keep the detail page values consistent for success, decision point, and agent-action values. I found one completeness issue in the message export.
</details>

<details>

**Fit:** Strong

The requested behavior is clear: allow troubleshooting log data to be downloaded as JSON, with sensitive page content protected. The diff adds the UI actions, the internal stream API, redaction handling, and focused tests for the JSON shape.
</details>

<details>

**S1** (🟠 Moderate): *Calculate message creator names before export*
`Agent Task Message."Created By Full Name"` is read in code, but the message query does not auto-calculate it. This full-name field follows the same pattern as the other exported full-name fields, which are explicitly auto-calculated. Add `SetAutoCalcFields("Created By Full Name")` before `FindSet()`, and cover message export in the tests.
</details>

<details>

**Risk:** The change is in System Application agent troubleshooting. It exports troubleshooting data into a downloadable file and protects page snapshots through the permission check. No BaseApp event publisher or posting flow is involved.

**Necessity:** The capability is useful for troubleshooting and test automation. The scope is limited to new export code and small page actions. Tests cover context projection, redaction, ordering, memory separation, details parsing, and language restore.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10670 round=1 by=alexei-dobriansky at=2026-09-07T22:17:26Z lastSha=7152545228caaa79be084fe22d4671bd35fe24d3 reviewKey=277e29c69c5f81961c11351845c80d9b516dd5527ddcd936b5fbc9bcaea5efc7 suggestions=S1@65717cd6

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
```diff
@@ -0,0 +1,636 @@
...
+        EntryJson.Add(DescriptionLbl, AgentTaskLogEntryRecord.Description);
+        EntryJson.Add(ReasonLbl, AgentTaskLogEntryRecord.Reason);
+        EntryJson.Add(DetailsLbl, AgentTaskImpl.GetDetailsForAgentTaskLogEntry(AgentTaskLogEntryRecord));
+        IsAgentAction := AgentTaskLogEntry.IsAgentAction(AgentTaskLogEntryRecord);
+        EntryJson.Add(AgentActionLbl, IsAgentAction);
+        if AgentTaskLogEntryRecord."Task ID" > 0 then begin
+            AgentName := GetAgentName(AgentTaskLogEntryRecord."Task ID");
+            if AgentName <> '' then
+                EntryJson.Add(AgentNameLbl, AgentName);
+            if AgentTaskMemoryEntry.Get(AgentTaskLogEntryRecord."Task ID", AgentTaskLogEntryRecord."Memory Entry ID") then
```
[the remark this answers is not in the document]
nikolakukrika: We will not have that much data, it is OK.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
```diff
@@ -0,0 +1,636 @@
...
+        if SelectedAgentTaskLogEntry.FindFirst() then
+            exit(SelectedAgentTaskLogEntry."Task ID");
+    end;
+
+    local procedure GetIncludeSerializedPage(): Boolean
+    var
+        FeatureAccessManagement: Codeunit "Feature Access Management";
+        AgentSystemPermissionsImpl: Codeunit "Agent System Permissions Impl.";
+    begin
+        FeatureAccessManagement.AgentManagementAllowed(true);
```
nikolakukrika: I believe we cannot use this check, this is limiting the access to the Sandbox only. We need the access in the PRODUCTION too. Please check which permission check is done on the page and implement the same.

src/System Application/Test/Agent/src/Troubleshooting/AgentTaskLogPageTest.Codeunit.al
```diff
@@ -173,30 +178,291 @@ codeunit 133964 "Agent Task Log Page Test"
...
+
+    [Test]
+    procedure TestBuildContextJson_RedactsSerializedPage()
+    var
+        AgentTaskLogExport: Codeunit "Agent Task Log Export";
+        ContextJson: JsonObject;
+    begin
+        // [GIVEN] Troubleshooting context containing a sensitive page snapshot
+        // [WHEN] The context is projected without the troubleshooting permission
+        AgentTaskLogExport.BuildContextJson('{"serializedPage":"{\"secret\":\"value\"}"}', false, ContextJson);
```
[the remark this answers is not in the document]
nikolakukrika: We do not want permissions testing

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
```diff
@@ -0,0 +1,636 @@
...
+        ExportRoot: JsonObject;
+        IncludeSerializedPage: Boolean;
+        CurrentGlobalLanguage: Integer;
+        ErrorText: Text;
+    begin
+        IncludeSerializedPage := GetIncludeSerializedPage();
+        CurrentGlobalLanguage := GlobalLanguage();
+        GlobalLanguage(1033); // ENU
+
+        if not TryBuildExportJson(SelectedAgentTaskLogEntry, SelectedAgentTaskMemoryEntry, AgentTaskID, IncludeMemoryEntries, IncludeSerializedPage, ExportRoot) then begin
```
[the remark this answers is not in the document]
nikolakukrika: The target audience is techincal, we do not want to hide the error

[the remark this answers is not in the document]
nikolakukrika: Small table, we will not get any benefits

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
```diff
@@ -0,0 +1,636 @@
...
+            EntryJson.Add(SuccessLbl, Success);
+
+        ContextTxt := AgentTaskLogEntry.ReadContext(AgentTaskLogEntryRecord);
+        if (ContextTxt = '') and (AgentTaskMemoryEntry."Task ID" <> 0) then
+            ContextTxt := AgentTaskLogEntry.ReadContext(AgentTaskMemoryEntry);
+        AddTroubleshootingContext(EntryJson, ContextTxt, IncludeSerializedPage);
+        if IsAgentAction and not EntryJson.Contains(DecisionPointLbl) then
+            EntryJson.Add(DecisionPointLbl, false);
+
+        MessagesJson := BuildMessagesJson(AgentTaskLogEntryRecord);
```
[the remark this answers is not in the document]
nikolakukrika: Not changing this. The export is a user-initiated troubleshooting operation over small task logs, and both lookups are narrowly filtered by task and memory-entry IDs. Pregrouping would add state and complexity without a measured benefit; we can optimize later if telemetry shows this path is material.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
```diff
@@ -0,0 +1,636 @@
...
+    local procedure ReadMemoryEntryDetails(var AgentTaskMemoryEntry: Record "Agent Task Memory Entry") MemoryDetailsTxt: Text
+    var
+        ContentInStream: InStream;
+    begin
+        AgentTaskMemoryEntry.CalcFields(Details);
+        AgentTaskMemoryEntry.Details.CreateInStream(ContentInStream, GetDefaultEncoding());
+        ContentInStream.Read(MemoryDetailsTxt);
+    end;
+
+    procedure BuildContextJson(ContextTxt: Text; IncludeSerializedPage: Boolean; var ContextJson: JsonObject)
```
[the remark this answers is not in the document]
nikolakukrika: Addressed in 03ba74d393. The Boolean-controlled builder is now local, so callers can no longer request raw page content. The only exposed context projection helper always redacts serialized page data, while the production export computes permission internally.

src/System Application/App/Agent/Troubleshooting/AgentTaskLogEntryList.Page.al
```diff
@@ -201,6 +203,23 @@ page 4303 "Agent Task Log Entry List"
...
+                Image = ExportFile;
+                Scope = Repeater;
+
+                trigger OnAction()
+                var
+                    SelectedAgentTaskLogEntry: Record "Agent Task Log Entry";
+                    AgentTaskLogExport: Codeunit "Agent Task Log Export";
+                begin
+                    CurrPage.SetSelectionFilter(SelectedAgentTaskLogEntry);
+                    AgentTaskLogExport.ExportToJsonFile(SelectedAgentTaskLogEntry);
```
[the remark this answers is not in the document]
nikolakukrika: Not applying this suggestion. The action is explicitly “Export selected”; when no multi-selection exists, SetSelectionFilter intentionally scopes to the current row. Copying Rec when MarkedOnly is false would instead export every visible row, which would be surprising and could export more sensitive data than requested.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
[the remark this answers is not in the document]
nikolakukrika: Addressed in 03ba74d393. Export failures now use a stable classified label and include GetLastErrorText only as a substitution value. The earlier “Small table” reply was intended for the performance thread.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
```diff
@@ -0,0 +1,636 @@
...
+    end;
+
+    local procedure ExportToJson(var SelectedAgentTaskLogEntry: Record "Agent Task Log Entry"; var SelectedAgentTaskMemoryEntry: Record "Agent Task Memory Entry"; AgentTaskID: BigInteger; IncludeMemoryEntries: Boolean; var ExportOutStream: OutStream)
+    var
+        ExportRoot: JsonObject;
+        IncludeSerializedPage: Boolean;
+        CurrentGlobalLanguage: Integer;
+        ErrorText: Text;
+    begin
+        IncludeSerializedPage := GetIncludeSerializedPage();
```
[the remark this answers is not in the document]
nikolakukrika: The export was already authorized because every public export entry point reaches the local export pipeline, which called AgentManagementAllowed(true) before serialization. In 03ba74d393 this was split into an explicit CheckExportAccess procedure so the authorization is no longer hidden inside the page-content decision.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
nikolakukrika: Double-checked the platform implementation and existing Agent Task API. AgentManagementAllowed checks whether Agent management is available in the current environment; it is not a sandbox-only check. Agent Task.OpenAgentTaskLogEntries uses the same protection, so the export now keeps it explicitly in CheckExportAccess.

nikolakukrika:
Addressed the Good Sense Reviewer suggestion in 03ba74d393 by auto-calculating Agent Task Message "Created By Full Name" before serializing messages.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
```diff
@@ -544,12 +554,17 @@ codeunit 4313 "Agent Task Log Export"
             exit(SelectedAgentTaskLogEntry."Task ID");
     end;
 
-    local procedure GetIncludeSerializedPage(): Boolean
+    local procedure CheckExportAccess()
```
nikolakukrika: We do not need this one liner function

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
```diff
@@ -544,12 +554,17 @@ codeunit 4313 "Agent Task Log Export"
...
-        AgentSystemPermissionsImpl: Codeunit "Agent System Permissions Impl.";
     begin
         FeatureAccessManagement.AgentManagementAllowed(true);
+    end;
+
+    local procedure CanIncludeSerializedPage(): Boolean
+    var
+        AgentSystemPermissionsImpl: Codeunit "Agent System Permissions Impl.";
+    begin
         exit(AgentSystemPermissionsImpl.CurrentUserHasTroubleshootAllAgents());
```
nikolakukrika: No need to have this one liner function.

src/System Application/Test/Agent/src/Troubleshooting/AgentTaskLogPageTest.Codeunit.al
```diff
@@ -173,30 +178,289 @@ codeunit 133964 "Agent Task Log Page Test"
...
+        Assert.AreEqual('1', PageStackEntry.GetText('order'), 'The first page should have order 1.');
+        Assert.AreEqual('Customer List', PageStackEntry.GetText('pageCaption'), 'The page stack should preserve the stored order.');
+        Assert.AreEqual(1, ContextJson.GetArray('availableTools').Count(), 'The available tools should be exported.');
+        Assert.AreEqual('10000', ContextJson.GetObject('memorizedData').GetText('customerNo'), 'The memorized data should be exported.');
+        Assert.AreEqual('USD', ContextJson.GetObject('taskAndPageSettings').GetText('currencyCode'), 'The task and page settings should be exported.');
+        Assert.AreEqual('en-US', ContextJson.GetObject('taskAndPageSettings').GetObject('communication').GetObject('culture').GetText('language'), 'The communication settings should be nested.');
+    end;
+
+    [Test]
+    procedure TestBuildContextJson_RedactsSerializedPage()
```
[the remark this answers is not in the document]
nikolakukrika: Addressed in 6b3476e465. The test is now explicitly named and documented as testing the redacted projection helper; it no longer claims to verify the permission boundary.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
nikolakukrika: Addressed in 6b3476e465 by inlining AgentManagementAllowed(true) into the export pipeline.

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
nikolakukrika: Addressed in 6b3476e465 by inlining CurrentUserHasTroubleshootAllAgents() into the export pipeline.

src/System Application/App/Agent/Troubleshooting/AgentTaskLogEntryList.Page.al
```diff
@@ -201,6 +203,23 @@ page 4303 "Agent Task Log Entry List"
                     Page.Run(Page::"Agent Task Log Entry", Rec);
                 end;
             }
+            action(Export)
```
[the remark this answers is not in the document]
nikolakukrika: Rejecting this finding. `CurrPage.SetSelectionFilter` uses the current repeater record when there is no multi-row selection, so an invoked row action does not yield an empty record set merely because rows were not explicitly marked. The `TaskSelected` condition on the task list represents that page's task eligibility state and is not evidence that every repeater export action needs a separate selection Boolean. No production defect is demonstrated here.

src/System Application/Test/Agent/src/Troubleshooting/AgentTaskLogPageTest.Codeunit.al
```diff
@@ -5,12 +5,15 @@
...
 
+using System.Agents;
 using System.Agents.Troubleshooting;
 using System.TestLibraries.Utilities;
+using System.Utilities;
 
 codeunit 133964 "Agent Task Log Page Test"
 {
     Subtype = Test;
+    TestPermissions = NonRestrictive;
```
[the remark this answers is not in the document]
nikolakukrika: Rejecting this finding. Permission-boundary integration testing was deliberately removed from this test suite: the required permission is already declared by the test app, and adding Permissions Mock or `Library - Lower Permissions` would test the platform permission framework rather than this serializer. These tests intentionally cover the deterministic redacted projection; production authorization continues to use `AgentManagementAllowed(true)` and `CurrentUserHasTroubleshootAllAgents()`.

src/System Application/Test/Agent/src/Troubleshooting/AgentTaskLogPageTest.Codeunit.al
```diff
@@ -173,30 +178,289 @@ codeunit 133964 "Agent Task Log Page Test"
...
+        // [WHEN] The context is projected through the explicitly redacted helper
+        AgentTaskLogExport.BuildRedactedContextJson('{"serializedPage":"{\"secret\":\"value\"}"}', ContextJson);
+
+        // [THEN] The snapshot is not present and the redaction is explicit
+        Assert.IsFalse(ContextJson.Contains('pageContent'), 'The page content should not be included.');
+        Assert.IsTrue(ContextJson.GetBoolean('pageContentRedacted'), 'The page content should be marked as redacted.');
+        Assert.IsTrue(ContextJson.GetText('pageContentRedactionReason').Contains('Troubleshoot All Agents'), 'The redaction reason should identify the required permission.');
+    end;
+
+    [Test]
```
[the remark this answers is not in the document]
nikolakukrika: Rejecting this finding. `TryBuildExportJson` has no supported, deterministic invalid input that should fail during normal serialization, so forcing this branch would require an artificial production test hook or coupling the test to an incidental runtime failure. The language is explicitly restored on both success and failure paths. Absence of a synthetic negative test is not a product defect and does not justify adding test-only production behavior.

alexei-dobriansky:

# **Recommendation: Accept**

<details>

This adds JSON export actions for agent task logs and keeps page snapshot data protected unless the current user has the troubleshooting permission. The latest changes add the missing message full-name calculation, keep the access check in the export path, avoid live-table lookups while testing with temporary records, and keep the redacted context helper narrow.

The prior message-export gap is fixed at the right point, before the message records are read. The other changes are small and keep the exported JSON shape aligned with the detail page helpers. No BaseApp publisher, posting flow, or Sales/Purchase/Service analog is involved.
</details>

<details>

| S1 | Calculate message creator names before export | Addressed | The latest diff adds SetAutoCalcFields("Created By Full Name") before message records are exported. |
</details>

<details>

*None - changes only addressed the previous suggestion and tightened the export/test seams.*
</details>

<details>

**Risk:** The change is in System Application agent troubleshooting and exports diagnostic data to a file. The access check and page-content redaction remain in the export path, and the latest commits do not add a posting, ledger, or BaseApp event dependency.

**Necessity:** The capability is needed so agent task logs can be analyzed outside the UI. The scope remains focused on export actions, a stream-based internal API, shared calculation helpers, and tests for the exported JSON shape.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10670 round=2 by=alexei-dobriansky at=2026-09-09T22:33:36Z lastSha=6b3476e46555d90b7e406330624538652276ef94 reviewKey=4fd03814da4f78975c04313d269a8bda8323c591da03869a6315b0c9ce3fac52 suggestions=S1@65717cd6:addressed parentRound=1

src/System Application/App/Agent/Interaction/Internal/AgentTaskList.Page.al
```diff
@@ -165,6 +167,22 @@ page 4300 "Agent Task List"
                     AgentTaskImpl.ShowTaskLogEntries(Rec);
                 end;
             }
+            action(ExportTaskLogEntries)
+            {
+                ApplicationArea = All;
+                Caption = 'Export log entries';
+                ToolTip = 'Download all log entries and related troubleshooting memory for the selected task as a JSON file.';
```
kmhansen, line 174: Nit: "memory" is that accurate?

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
```diff
@@ -0,0 +1,640 @@
...
+
+        TaskPageSettings := BuildTaskPageSettingsJson(ContextRoot);
+        if TaskPageSettings.Keys().Count() > 0 then
+            ContextJson.Add(TaskAndPageSettingsLbl, TaskPageSettings);
+
+        RawSerializedPageJson := ContextRoot.GetText(SerializedPageLbl, true);
+        if RawSerializedPageJson = '' then
+            exit;
+
+        if not IncludeSerializedPage then begin
```
kmhansen, line 250: Looks like this is the only place that Boolean is checked. Resolve it here (based on permissions) rather than passing it around may be safer?

kmhansen (approved):

src/System Application/App/Agent/Interaction/Internal/AgentTaskList.Page.al
nikolakukrika: Yes

src/System Application/App/Agent/Troubleshooting/Internal/AgentTaskLogExport.Codeunit.al
nikolakukrika: We will have more configuration options in near future. We will give you the option of excluding the serialized page most likely as serialized page is taking a lot of space, it is quite heavy. Currently the code is safe and is correct.

thloke (approved):

PR #10740, opened 2026-08-27
[SOA] Accuracy test fixes: agent instructions

qasimikram:
## Summary

Fixes for Sales Order Agent accuracy test failures.

This PR is a rolling collection - more accuracy test fixes will be added to it as
further scenarios are investigated. Each change is listed below.

All changes are to the agent instructions only
(`SalesOrderAgent-AgentInstructions.json`). No test expectations were relaxed.

Fixes [AB#647980](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/647980)

### 1. SOA-IRRELEVANT_ATTACHMENTS - request assistance when nothing actionable was received

**Tests:** `IRRELEVANT_ATTACHMENTPDF01`, `IRRELEVANT_ATTACHMENTPNG01`.

**Symptom:** Both cases expect user intervention, but the agent replied to the
customer instead, stating that no attachment had been received and asking the
customer to resend it.

**Root cause:** The attachment analyzer correctly marks an attachment with no
sales-relevant content as `ignored: true` / `ignoredReason: "No relevant content"`.
The ignored attachment is then stripped from the conversation history handed to
the agent, which sees `"attachments":[]`. The agent concluded the attachment had
been lost in transit and asked for it again. That reply is factually wrong - the
attachment was received and deliberately set aside - and it leaks internal
filtering behavior to the customer.

The instruction covering this already existed under **Check Requested Item Exists**
("If there is no item to be searched ... then request for assistance"), but the
agent did not recognize that it applied, because it framed the situation as a
missing attachment rather than as having no item to search.

**Change:** Clarified that existing instruction rather than adding a new rule:

- Defines when there is no item to be searched - neither the message body nor any
  attachment identifies an item, product, or service - and states that assistance
  is requested instead of replying to the customer.
- States that the agent must not ask the customer to resend or re-attach a
  document, and must not claim an attachment was missing or unreadable, because
  it may have been received and set aside as not relevant.

Deliberately kept narrow: the rule stays scoped inside **Check Requested Item
Exists**, does not override other guidelines, and does not stop the task - if the
user chooses to continue, the agent proceeds as before. Existing behavior for
"items were requested but the search found nothing" and the specific-variant
exception is unchanged.

**Evidence:** agent run logs 147 (PDF) and 148 (PNG).

### 2. SOA-SHIPPING_ADDRESS - Custom Address was unreachable for unregistered addresses

**Tests:** `SHIPPING_ADDRESS_CUSTOM_01`, `SHIPPING_ADDRESS_CUSTOM_02`.

**Symptom:** Both expect `Custom Address` with the requested address entered
manually. Instead the agent selected an unrelated registered alternate address,
failing with `Actual: Ship-to Code has a value, Expected: Empty Ship-to Code`.

**Root cause:** The matching instruction required the agent to select a registered
address whenever it was even a weak match, and listed the country as a sufficient
partial match:

> "If the best option is only a partial or weak match of the requested address
> (e.g. only the street, city, state, postal code, **or country**), you MUST still
> select that one ... custom address should only be used as an exception, when it
> is obvious that none of the registered addresses can be used."

Any US address therefore matched any US alternate address, which made
`Custom Address` effectively unreachable. In task 99 the requested address was
*987 Skyline Boulevard, 94062 Woodside, California*, and the agent selected the
`WORK` address in *Stamford, CT* with the reasoning:

> "The requested address **is not registered**, so the closest available registered
> U.S. ship-to address should be selected **as required**."

It correctly identified that the address was not registered and selected one
anyway, because the instruction told it that it must. Task 100 is the same, with
*Boulder, CO 80302* resolving to the `HOME` address in *Jersey City, NJ*.

**Change:** Split the single ambiguous notion of "partial match" into two explicit
cases:

- **Case A - the requested address is less specific than a registered address**
  (only a city, region, site, or street, for example "our warehouse in Austin,
  TX"). Still a match; select the closest registered address. This preserves
  `ALTERNATIVE_SHIPPING_ADDRESS` and the `SHIPPING_ADDRESS_ALTERNATE_*` cases.
- **Case B - the customer provided a complete address** (a street together with a
  postal code or city) that conflicts with every registered address. Not a match;
  use `Custom Address`. A different street, city, or postal code is a conflict,
  and a shared country alone is explicitly never a match.

The `Custom Address` step is also reframed from a "last resort exception" to the
correct choice when a complete unregistered address was given, and now tells the
agent to select the "none of the other rows are applicable" row (or cancel the
list) first.

**Evidence:** agent run logs 99 and 100.

### 3. SOA-SHIPPING_ADDRESS - alternate address list was skipped when the default is Custom Address

**Test:** `SHIPPING_ADDRESS_ALTERNATE_03`.

**Symptom:** Expects the registered `HOME` alternate address to be selected, but
the agent produced an empty Ship-to Code, failing with
`Actual: Empty Ship-to Code, Expected: Alternate Shipping Address`.

**Root cause:** In this case the contact's default shipping address is already
`Custom Address`, so the quote opened with `Ship-to` set to `Custom Address` and
the `Shipping and Billing/*` fields directly editable. The agent typed the
requested address straight into those fields and never opened the Alternate
Shipping Addresses list, skipping the mandatory check:

```
set_field_value  Shipping and Billing/Address   -> "742 Ocean Avenue, Apt 15"
set_field_value  Shipping and Billing/City      -> "Jersey City"
set_field_value  Shipping and Billing/Post Code -> "07302"
```

The requested address matches the registered `HOME` alternate exactly, so opening
the list would have matched on the first row. This is also why
`SHIPPING_ADDRESS_ALTERNATE_01` passed - there the default is
`Default (Sell-to Address)`, the address fields are not editable, and the agent
is forced through the `Ship-to` option field.

**Change:** Extended the existing mandatory step so it explicitly applies when
`Ship-to` already shows `Custom Address` and the address fields are editable, and
forbids typing the requested address into the `Shipping and Billing` fields before
the list has been checked.

**Evidence:** agent run log 236.


AleksandricMarko (approved):

alexei-dobriansky:

# **Recommendation: Accept**

<details>

This changes the Sales Order Agent instructions in two areas: no actionable item content should request assistance without claiming a missing attachment, and ship-to address selection must always check registered alternates before custom entry while distinguishing partial place requests from complete conflicting addresses.

The changes sit in the prompt steps that drive item lookup and sales quote ship-to handling. The new ship-to wording preserves the registered-address path for vague requests and makes Custom Address reachable for complete unregistered addresses. The attachment wording redirects the no-item path to assistance and avoids an unsupported customer response.
</details>

<details>

**Fit:** Strong

The bug describes three prompt failures: irrelevant attachments led to a customer resend request, complete unregistered addresses mapped to unrelated registered addresses, and an existing Custom Address default let the agent skip alternate address search. The diff directly edits those instructions and keeps scope to the prompt resource, so it matches the reported failures without broad changes.
</details>

<details>

No suggestions.
</details>

<details>

**Risk:** The regression surface is limited to Sales Order Agent quote creation flows where there is no actionable item content or where requested ship-to addresses are matched against alternate addresses. There is no AL code, public API, event, posting, or BaseApp publisher change.

**Necessity:** The change is needed because the prior wording allowed wrong customer replies and incorrect ship-to selection. The scope is narrow: one prompt resource, and each added instruction targets one failed scenario.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10740 round=1 by=alexei-dobriansky at=2026-08-27T18:13:22.8063991Z lastSha=d358ecfbb02d139fe122c449a647693eb071871a reviewKey=a4cd9ff9ece3073b01aa5c67708012f6eebe26b3885c106d7627b307925d53ef suggestions=

PredragMaricic (approved):

PR #10744, opened 2026-08-27
[main] Bug 647991 Assembly-to-Order Item Incorrectly Blocks Shipment of Unrelated Sales Order Line

MSNehaNawal:
[Bug 647991](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/647991): [master][ALL-E] Assembly-to-Order Item Incorrectly Blocks Shipment of Unrelated Sales Order Line

Fixes [AB#647991](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/647991)

Issue
On a sales order that contains an Assemble-to-Order (ATO) line and a separate, unrelated line, posting the inventory pick for the unrelated line fails.

Root cause
Posting the unrelated line's inventory pick calls Whse.-Activity-Post.InitSourceDocument, which runs SalesLine.Validate("Qty. to Ship", 0) over every line on the order — including the ATO line that is not part of this pick.
Because an inventory-movement warehouse activity line exists for the ATO assembly consumption line (and the basic-warehouse location has Require Shipment = false), AssemblyLineVerifyChange unconditionally blocked any change to Quantity to Consume, throwing the error.

Solution
In Assembly Warehouse Mgt.AssemblyLineVerifyChange, enforce the guard's actual intent — protect the already-picked quantity , instead of blocking every change. The error is now raised only when the recalculated Quantity to Consume would drop below the quantity already picked:

### DIFF
--- a/src/Layers/NA/Tests/SCM-Warehouse/SCMWarehouseIII.Codeunit.al
+++ b/src/Layers/NA/Tests/SCM-Warehouse/SCMWarehouseIII.Codeunit.al
@@ -4700,18 +4700,6 @@ codeunit 137051 "SCM Warehouse - III"
         // [THEN] The quantity of the pick is 5
         WarehousePickLine.TestField(WarehousePickLine.Quantity, AssemblyLine.Quantity);
 
-        // [WHEN] Update Quantity to Consume on Assembly header
-        asserterror AssemblyHeader.Validate("Quantity to Assemble", 3);
-
-        // [THEN] Error: active warehouse line exists
-        Assert.ExpectedError('must not be changed');
-
-        // [WHEN] Update Quantity to Consume on Assembly Line
-        asserterror AssemblyLine.Validate("Quantity to Consume", 3);
-
-        // [THEN] Error: active warehouse line exists
-        Assert.ExpectedError('must not be changed');
-
         // [GIVEN] The order is posted without registering the created picks.
         LibraryAssembly.PostAssemblyHeader(AssemblyHeader, '');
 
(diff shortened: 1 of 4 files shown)


v-ankitgoyal (approved):

alexei-dobriansky:

<details>

The change narrows `AssemblyLineVerifyChange` so `Quantity to Consume` is blocked only when the new base quantity would be below the quantity already picked. This lets posting an inventory pick for a non-assembly sales line continue when an unrelated ATO line has only an unregistered inventory movement.

The main condition fits the reported failure: the failing posting path can touch the ATO assembly line while `Qty. Picked (Base)` is still zero, so the previous unconditional check was too broad. The existing validation still keeps the source identity, item, location, quantity, and quantity per component unchanged while warehouse lines exist, and the event after the guard still fires in the same place.
</details>

<details>

**Fit:** Strong

The reported scenario is clear and important: one sales line should not block shipment of another unrelated line on the same order. The diff targets the specific assembly warehouse validation that caused the false block, and the new test exercises that posting path.
</details>

<details>

**S1** (🟠 Moderate): *Cover the already-picked guard*
The new guard depends on `Qty. Picked (Base)`, but the added regression test only covers the zero-picked case. Please add a case where an assembly movement has been registered, then lowering `Quantity to Consume` below the picked quantity still gives the warehouse-line error.
</details>

<details>

**Risk:** This touches assembly warehouse validation that runs from assembly line changes and from inventory pick posting. The regression surface is mainly ATO component quantities with existing warehouse activity lines; the change is narrow and CI build checks passed, but the already-picked protection should be locked down with a direct test.

**Necessity:** The change is needed because, without it, users can be forced to complete or change the unrelated ATO movement before shipping an available non-assembly line. The scope is appropriate because it keeps the old protection for picked quantities instead of removing warehouse validation entirely.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10744 round=1 by=alexei-dobriansky at=2026-09-01T06:13:53.9574879Z lastSha=75c375fc67d1a5434d01d079fe4a5b86db939b9a reviewKey=9429418642040e416fca22c77f0d5cc14879d652696d4d65ab22fc3608170901 suggestions=S1@f222f6f2

alexei-dobriansky (changes requested):
Please consider adding the suggested test.

PR #10751, opened 2026-08-28
Backport #10568: [29.0] Send sample invoices from the demo guide directly

Explorer986:
## What & why

Backport of #10568 to 29.0.

On the Payables Agent guide page, the email option used to only mark the sample invoices, leaving the real send to happen when setup was saved. Saving setup before running the guide meant the invoices were never sent, yet the guide still reported success.

The guide now sends the invoices when you finish it, so the order of steps no longer matters. The marks are cleared after sending so the setup save path will not resend, and the finish step reports success only for the send path.

Fixes [AB#648144](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/648144)

## How I validated this

- change was built and validated on main
- No automated tests added, matching the original PR. This is a UI and orchestration change in the demo guide with no test hook for the mail send.

## Risk & compatibility

Backport only, scoped to the two PADemoGuide files. SendDemoEmail returns a Boolean now instead of nothing, but the codeunit is internal so there is no public break. No permission, upgrade, or telemetry impact.

### DIFF
--- a/src/Apps/W1/PayablesAgent/app/Setup/PADemoGuide.Codeunit.al
+++ b/src/Apps/W1/PayablesAgent/app/Setup/PADemoGuide.Codeunit.al
@@ -55,20 +55,34 @@ codeunit 3309 "PA Demo Guide"
     /// <summary>
     /// Sends a demo email using the active PA setup configuration.
     /// </summary>
-    procedure SendDemoEmail()
+    /// <returns>True if the demo email was sent.</returns>
+    procedure SendDemoEmail(): Boolean
     var
         PayablesAgentSetup: Codeunit "Payables Agent Setup";
         PASetupConfiguration: Codeunit "PA Setup Configuration";
     begin
         PayablesAgentSetup.LoadSetupConfiguration(PASetupConfiguration);
-        SendDemoEmail(PASetupConfiguration);
+        exit(SendDemoEmail(PASetupConfiguration));
+    end;
+
+    /// <summary>
+    /// Sends the sample invoices by email if the agent is active, otherwise flags them to be sent at activation.
+    /// </summary>
+    /// <returns>True if the sample invoices were sent.</returns>
+    procedure SendDemoInvoicesByEmail(): Boolean
+    var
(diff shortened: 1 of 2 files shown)


alexei-dobriansky:

<details>

This change makes the demo guide mark the sample invoice files and call the demo email send path when the email option is finished. For an already active setup, this addresses the old `IsEmpty()` path because the files are now flagged before `SendDemoEmail()` runs, and a successful send clears the flags so setup save will not send them again.
</details>

<details>

**Fit:** Partial

The reported setup-first scenario is fixed by sending directly from the guide. The remaining mismatch is that the guide can still show the user-facing `Sample invoices sent!` finish text before the send result is known, and also when the new false branch only prepared the files for later activation.
</details>

<details>

**S1** (🟠 Moderate): *Show prepared state when mail is not sent*
When `SendDemoInvoicesByEmail()` returns false, this code only writes telemetry. The guide can still show `Sample invoices sent!` even though the invoices were only prepared for activation. Set the finish headline or text from the same send result so the user sees what actually happened.
</details>

<details>

**Risk:** The code path is narrow and stays inside the Payables Agent demo guide and setup codeunit. It sends external email, so the main regression risk is user confusion or duplicate sample mail, but the successful path clears the flags and there is no public API or BaseApp event dependency.

**Necessity:** The change is needed because a natural setup order silently left sample invoices flagged and unsent. The scope is appropriate for a backport, with the only remaining issue being the user-facing message for the prepared-but-not-sent path.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10751 round=1 by=alexei-dobriansky at=2026-08-28T07:14:30Z lastSha=f519b1cf1bcc83d6f24e92c5fed42bac280474c5 reviewKey=f2d14efb6f453b8ae376ebee481ac218fe2f31cb23bce28dd37aa25467062fb0 suggestions=S1@076b8b27

AndersLarsenMicrosoft (approved):

ventselartur (approved):

PR #10752, opened 2026-08-28
Backport #10568: [29.x] Send sample invoices from the demo guide directly

Explorer986:
## What & why

Backport of #10568 to 29.x.

On the Payables Agent guide page, the email option used to only mark the sample invoices, leaving the real send to happen when setup was saved. Saving setup before running the guide meant the invoices were never sent, yet the guide still reported success.

The guide now sends the invoices when you finish it, so the order of steps no longer matters. The marks are cleared after sending so the setup save path will not resend, and the finish step reports success only for the send path.

Fixes [AB#648145](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/648145)

## How I validated this

- change was built and validated on main
- No automated tests added, matching the original PR. This is a UI and orchestration change in the demo guide with no test hook for the mail send.

## Risk & compatibility

Backport only, scoped to the two PADemoGuide files. SendDemoEmail returns a Boolean now instead of nothing, but the codeunit is internal so there is no public break. No permission, upgrade, or telemetry impact.

### DIFF
--- a/src/Apps/W1/PayablesAgent/app/Setup/PADemoGuide.Codeunit.al
+++ b/src/Apps/W1/PayablesAgent/app/Setup/PADemoGuide.Codeunit.al
@@ -55,20 +55,34 @@ codeunit 3309 "PA Demo Guide"
     /// <summary>
     /// Sends a demo email using the active PA setup configuration.
     /// </summary>
-    procedure SendDemoEmail()
+    /// <returns>True if the demo email was sent.</returns>
+    procedure SendDemoEmail(): Boolean
     var
         PayablesAgentSetup: Codeunit "Payables Agent Setup";
         PASetupConfiguration: Codeunit "PA Setup Configuration";
     begin
         PayablesAgentSetup.LoadSetupConfiguration(PASetupConfiguration);
-        SendDemoEmail(PASetupConfiguration);
+        exit(SendDemoEmail(PASetupConfiguration));
+    end;
+
+    /// <summary>
+    /// Sends the sample invoices by email if the agent is active, otherwise flags them to be sent at activation.
+    /// </summary>
+    /// <returns>True if the sample invoices were sent.</returns>
+    procedure SendDemoInvoicesByEmail(): Boolean
+    var
+        EDocSamplePurchInvFile: Record "E-Doc Sample Purch. Inv File";
+    begin
+        EDocSamplePurchInvFile.ModifyAll("Send By Email", true);
+        exit(SendDemoEmail());
     end;
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


alexei-dobriansky:

<details>

This changes the Payables Agent demo guide so the Finish action marks the sample invoice files and calls the send helper directly. For the setup-saved-before-guide flow, the guide now sends the sample invoice email itself, and a successful send clears the Send By Email marks so a later setup save does not send the same samples again.

The main path matches the reported bug because the guide no longer depends on a later setup save to dispatch the email. The fallback path is still a prepared-for-activation path, so the remaining concerns are making the visible result match that fallback and protecting both setup orders with regression coverage.
</details>

<details>

**Fit:** Strong

The reported setup-saved-before-guide order is addressed directly because the guide now sends the sample invoices when the email option is finished. The older guide-before-activation order is also preserved by leaving the invoices marked when the agent cannot send yet.
</details>

<details>

**S1** (🟠 Moderate): *Show the actual send result*
The Finish step already knows whether SendDemoInvoicesByEmail() sent the email or only prepared it, but the visible text is set earlier and still uses the sent headline for both outcomes. Set the headline or text after the send result, or use neutral text for the prepared case. Otherwise the guide can still say that invoices were sent when they were only queued for activation.

**S2** (🟠 Moderate): *Cover both setup orders*
Add a regression test for the saved-setup-before-guide path and the guide-before-activation path. It should also prove that clearing the Send By Email marks stops a later setup save from sending the same samples again. This is not a blocker for this UI flow, but it protects the new Boolean result and flag clearing.
</details>

<details>

**Risk:** The change is limited to the Payables Agent demo guide and its internal helper codeunit, but it adds an immediate external email send to a page action and changes when sample-invoice flags are set and cleared. There is no BaseApp publisher or event-ordering dependency, no public API break, and the diff has no whitespace errors.

**Necessity:** The change is needed because users can naturally save setup before running the guide, and that order previously left sample invoices marked without sending them. The scope is right for the bug because it keeps the setup-save path for activation-time sending while adding the missing direct send from the guide.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10752 round=1 by=alexei-dobriansky at=2026-08-28T07:16:13Z lastSha=e152fbd0e7bd550a75f30359c584523635947fb7 reviewKey=19371202bff1c37aa77311077d01ecce2994be389c1d687f4f29097a5fbd4c28 suggestions=S1@dc74cf27,S2@f23947e3

AndersLarsenMicrosoft (approved):

ventselartur (approved):

PR #10763, opened 2026-08-28
SOA uses Earliest Shipment Date on quote lines

VolodySyn:
Fixes [AB#647611](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/647611)

Enhance item availability handling to calculate and communicate the earliest shipment date when low inventory is detected. Update quote-line shipment dates accordingly, including when resuming after user assistance.

### DIFF
--- a/src/Apps/W1/SalesOrderAgent/app/.resources/Prompts/SalesOrderAgent-AgentInstructions.json
+++ b/src/Apps/W1/SalesOrderAgent/app/.resources/Prompts/SalesOrderAgent-AgentInstructions.json
@@ -228,7 +228,14 @@
 						},
 						{
 							"name": "item_availability",
-							"value": "If there is a notification on the page regarding low inventory, request assistance."
+							"value": "If there is a notification on the page regarding low inventory, request assistance.",
+							"steps_include_numbering": "true",
+							"steps": [
+								{
+									"name": "capable_to_promise",
+									"value": "Ignore this notification. Check whether an earliest shipment date was already provided for the item referenced by the low inventory notification before the sales quote was created (that is, it came from the item availability search results). If it was not, recalculate the earliest shipment date, and make sure you specify the new shipment date later in the email. After that, treat the item as available and set the 'Shipment Date' on that item's sales quote line to the earliest shipment date or the requested delivery date, whichever is later. If the requested delivery date is blank or in the past, use the earliest shipment date. If assistance was already requested for this notification and you are asked to continue the task, set the 'Shipment Date' as described here before proceeding."
+								}
+							]
 						},
 						"{% endif -%}",
 						{


alexei-dobriansky:

<details>

The change adds a capable-to-promise sub-step to the low-inventory notification handling at the quote-review stage. Before, when a low-inventory notification appeared on the created quote, the agent just requested assistance. Now the agent ignores the notification, ensures an earliest shipment date exists (reusing the one from the availability search or recalculating it), sets the line 'Shipment Date' to the earliest shipment date or the requested delivery date (whichever is later, using the earliest date when the requested date is blank or in the past), and also handles the resume-after-assistance case.

The change is prompt-only and additive. It reuses the same override pattern already used earlier in the file (a parent "request assistance" step with a capable-to-promise child that treats the item as available), and its date rule matches the search-phase rule, so it is internally consistent. The JSON is well-formed.
</details>

<details>

**Fit:** Strong

The reported problem is that the agent did not carry the earliest shipment date onto quote lines when a low-inventory notification showed up after the quote was created. The new step targets exactly that path and also covers the resume-after-assistance flow, so it addresses the whole described scenario.
</details>

<details>

**S1** (🟠 Moderate): *Recalculate step does not say how to recalculate*
The step tells the agent to "recalculate the earliest shipment date" but does not name the action or page to use, unlike the search-phase step which uses the "Item Availability" action. At the quote-review stage the agent may not know how to recalculate and could invent a date, which would put a wrong shipment date on the line and in the email. Point the agent to a concrete mechanism (for example, reopen the "Item Availability" action for that item) so the recalculated date is trustworthy.

**S2** (🟡 Minor): *Reply step does not mention the shipment date*
This step says to "specify the new shipment date later in the email", but the reply summary step lists the fields to include (item No., description, variant, quantity, unit of measure, price) and does not mention the shipment date. Consider adding the shipment date to that reply summary step so the recalculated date is reliably communicated to the customer.
</details>

<details>

**Risk:** Low and bounded. This is a prompt-only change in `SalesOrderAgent-AgentInstructions.json`; it touches no AL logic, posting, or ledger code. The behavior change is scoped to the low-inventory-notification branch at quote review and follows the established capable-to-promise override pattern, so regression surface is limited to that path. The main residual risk is the unclear recalculation mechanism (S1), which could lead to an incorrect shipment date being shown to the customer. Validation here is scenario/eval-based rather than unit-tested, which is appropriate for an agent instruction change.

**Necessity:** Justified. Without it, the agent stops and requests assistance on every low-inventory notification even when an earliest shipment date is available, so quotes do not get the correct shipment date and manual intervention is needed. The scope is right for a bug fix - a single, targeted instruction block.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10763 round=1 by=alexei-dobriansky at=2026-08-28T10:15:17Z lastSha=d42b3c94313e9cc92b633e6d5f3609ba168e56a1 reviewKey=dd090719d7b196bf037bf8ca60bfcc02d0b33bd7b36d0492fa8c0adf15918f6f suggestions=S1@c05353ca,S2@0bfc1b38

ChethanT (approved):

PredragMaricic (approved):

qasimikram (approved):

PredragMaricic (approved):

PR #10771, opened 2026-08-28
Fix early preview feedback OCV routing

JesperSchulz:
## What & why

Early preview feedback was sent to the parent Dynamics 365 Business Central area because the code used `EarlyAccessPreview`, while the registered One Customer Voice area identifier is `EarlyPreview`. This updates the shared locked token so both release and feature feedback are categorized under the dedicated Early Preview area.

## Linked work

Fixes [AB#643382](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/643382)

## How I validated this

- Confirmed both feedback actions use the corrected shared `EarlyPreview` routing token.
- Ran `git diff --check` successfully.
- No automated test was added because the change corrects a static external OCV area identifier. Local AL compilation was unavailable because the worktree does not contain the required `.alpackages` dependency cache.

## Risk & compatibility

Low risk. This changes only feedback categorization in One Customer Voice; there are no API, schema, permission, or upgrade impacts.


JesperSchulz:
This was presumably resolved through #9400.

PR #10774, opened 2026-08-28
fix rounding issue for SCA

AndersLarsenMicrosoft:
What & why
There is a rounding issue when Source Currency Amount calculation causes discrepancies with Foreign Currency G/L Entries whit Reversal Charge VAT.

See 
image

The fix is simply to with the amount in currency. i dd "TDT" by publishing and run the test that filed until i published the fix 

## Linked work

Fixes [AB#647818](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/647818)

## How I validated this

### DIFF
--- a/src/Layers/APAC/BaseApp/Finance/GeneralLedger/Posting/GenJnlPostLine.Codeunit.al
+++ b/src/Layers/APAC/BaseApp/Finance/GeneralLedger/Posting/GenJnlPostLine.Codeunit.al
@@ -1136,19 +1136,26 @@ codeunit 12 "Gen. Jnl.-Post Line"
 
     local procedure CreateReverseChargeVATGLEntries(GenJnlLine: Record "Gen. Journal Line"; VATPostingSetup: Record "VAT Posting Setup"; VATPostingParameters: Record "VAT Posting Parameters")
     var
+        FullVATAmountSrcCurr: Decimal;
         LastNextEntryNo: Integer;
     begin
         if VATPostingParameters."Unrealized VAT" or not (NonDeductibleVAT.IsNonDeductibleVATEnabled()) then begin
+            // Preserve the source-currency VAT amount from the document for system-created entries so the two
+            // offsetting reverse charge VAT entries balance to zero instead of being recalculated from rounded LCY.
+            if GenJnlLine."System-Created Entry" then
+                FullVATAmountSrcCurr := GenJnlLine."Source Curr. VAT Amount"
+            else
+                FullVATAmountSrcCurr := CalcAmountSrcCurr(GenJnlLine, VATPostingParameters."Full VAT Amount");
             OnInsertVATOnBeforeCreateGLEntryForReverseChargeVATToPurchAcc(
                 GenJnlLine, VATPostingSetup, VATPostingParameters."Unrealized VAT", VATPostingParameters."Full VAT Amount", VATPostingParameters."Full VAT Amount ACY", true);
             CreateGLEntry(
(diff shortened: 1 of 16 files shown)


alexei-dobriansky:

<details>

This changes reverse charge VAT G/L entry creation so system-created entries use the source-currency VAT amount from the document instead of recalculating it from rounded LCY. That is the right root cause for the reported foreign-currency rounding mismatch, and the added tests target a real purchase invoice scenario.

The implementation is not complete in all localized variants. BE and NO still recalculate the balancing reverse charge entry, and BE also uses the new source-currency variable before it is assigned in one branch.
</details>

<details>

**Fit:** Partial

The change matches the reported problem and fixes the intended path in several variants, but it does not preserve both offsetting reverse charge VAT entries everywhere it changed the posting code. The remaining country-specific gaps can still create wrong source-currency VAT amounts.
</details>

<details>

**S1** (🔴 High): *BE balancing entry still recalculates source currency*
In the BE branch, `FullVATAmountSrcCurr` is used for the purchase VAT account, but the reverse charge account entry just below still calls `CalcAmountSrcCurr(...)`. That keeps one side calculated from rounded LCY and can leave the source-currency VAT balance wrong. Use the preserved source-currency value for this entry too, including the existing non-deductible source-currency part.

**S2** (🔴 High): *BE unrealized branch uses zero source amount*
The BE unrealized VAT branch passes `-FullVATAmountSrcCurr`, but that variable is only assigned inside the earlier `if not NonDeductibleVAT.IsNonDeductibleVATEnabled()` block. When non-deductible VAT is enabled, the source amount for this reverse charge entry becomes 0. Set the value before both branches, or keep the old calculation in this branch.

**S3** (🔴 High): *NO balancing entry still recalculates source currency*
In the NO branch, the reverse charge account entry still uses `CalcAmountSrcCurr(GenJnlLine, GetReverseChargeVATAmount(...))` after the first entry was changed to use `FullVATAmountSrcCurr`. That recreates the same rounding error on the balancing side; the new test shows -144.86 instead of -144.85. Use the preserved source-currency value, adjusted for proportional deduction when that applies.

**S4** (🟠 Moderate): *New test opens unhandled confirmations*
The new test sets the posting date to `WorkDate() + 1` and has no handler for confirmation dialogs. In some country runs it stops before the VAT assertions because posting asks for confirmation. Use `WorkDate()` or add the existing confirm handler so the test proves the VAT fix.
</details>

<details>

**Risk:** High. This is posting and financial code that creates G/L entries and source-currency amounts. A wrong fix can leave source-currency VAT amounts out of balance, and the localized variants are not identical, so BE and NO need separate verification. No BaseApp event publisher dependency was involved.

**Necessity:** The change is needed. Without it, reverse charge VAT entries for foreign-currency purchase invoices can be posted with source-currency amounts recalculated from rounded LCY, causing small but real mismatches and non-zero source-currency balances. The scope is right if every changed localized variant preserves both offsetting entries and the new tests run cleanly.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10774 round=1 by=alexei-dobriansky at=2026-09-01T19:22:41Z lastSha=22e37463fb7cd019367b74856d316bc4508bc8dd reviewKey=3fd64c18598860ef9604cce59b104bf9d1f5dfb199228ccc8e093917fdf212be suggestions=S1@27b67e10,S2@d4c67ba5,S3@352c274f,S4@d15769a8

PR #10775, opened 2026-08-28
[Shopify] Optimize unmapped cue counts

onbuyuka:
## Summary

The Shopify Activities cues for unmapped customers, products, and companies currently count records through lookup FlowFields. Business Central cannot push those predicates into a SQL aggregate, so it materializes the source tables and evaluates the lookup row by row on every Role Center refresh.

## Change

- Count empty stored Customer/Item SystemId fields so SQL can execute a direct `COUNT(*)`.
- Add a key led by `Item SystemId` so the global unmapped-product count can seek without an unconstrained `Shop Code` prefix.

## Behavioral note

A record with a nonempty SystemId that points to a deleted Customer or Item is no longer included in the cue. Such dangling mappings are a separate data-integrity concern.

Fixes [AB#648153](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/648153)

### DIFF
--- a/src/Apps/W1/Shopify/App/src/Base/Tables/ShpfyCue.Table.al
+++ b/src/Apps/W1/Shopify/App/src/Base/Tables/ShpfyCue.Table.al
@@ -26,13 +26,13 @@ table 30100 "Shpfy Cue"
         }
         field(2; "Unmapped Customers"; Integer)
         {
-            CalcFormula = count("Shpfy Customer" where("Customer No." = const('')));
+            CalcFormula = count("Shpfy Customer" where("Customer SystemId" = const('00000000-0000-0000-0000-000000000000')));
             Caption = 'Unmapped Customers';
             FieldClass = FlowField;
         }
         field(3; "Unmapped Products"; Integer)
         {
-            CalcFormula = count("Shpfy Product" where("Item No." = const('')));
+            CalcFormula = count("Shpfy Product" where("Item SystemId" = const('00000000-0000-0000-0000-000000000000')));
             Caption = 'Unmapped Products';
             FieldClass = FlowField;
         }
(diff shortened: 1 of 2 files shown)


alexei-dobriansky:

<details>

This changes the Shopify activity cues for unmapped customers, products, and companies to count stored SystemId fields instead of lookup FlowFields. It also adds an Item SystemId key so the product cue can use an index when the count is not scoped by shop.

The change matches the performance issue: Customer and Company already have keys on Customer SystemId, and the new product key covers the remaining count. Counting the zero GUID avoids per-row lookup of Customer No. or Item No. The main behavior change is that records with a stale nonzero SystemId no longer contribute to these cues.
</details>

<details>

**Fit:** Strong

The change covers all three slow unmapped cues and uses stored fields that can be filtered directly. The scope is narrow and matches the Role Center refresh path; the remaining stale-mapping behavior is separate from the performance bottleneck but should stay explicit.
</details>

<details>

**S1** (🟠 Moderate): *Make stale mappings visible*
With this change, a record that points to a deleted Customer or Item no longer appears in the unmapped count, even though the lookup value is blank. Please either clean these links when the target record is deleted or surface them separately, so broken mappings do not become hidden.

**S2** (🟠 Moderate): *Add coverage for the cue counts*
No test covers these cue FlowFields. Please add a small Shopify test that creates mapped and unmapped customer, company, and product records and checks that the cue counts only the intended records.
</details>

<details>

**Risk:** The fix touches visible Role Center counts only. It does not change posting or synchronization writes, but it changes what the count means for stale SystemId values and may affect users who rely on the cue to find broken mappings.

**Necessity:** The change is justified because the old FlowField predicates can turn a simple count into row-by-row lookup work on each Role Center refresh. The scope is small, and the key added for products closes the only obvious indexing gap.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10775 round=1 by=alexei-dobriansky at=2026-08-28T22:12:13.123Z lastSha=9d6ea5bd1295502c310e6538a7f825de9d5c60a9 reviewKey=d43f121600292351fc8774148aed1a77a22da65fd031a5ba55357ae33d05ebad suggestions=S1@78135424,S2@b9419f4f

darjoo (approved):

gggdttt (approved):

PR #10777, opened 2026-08-28
[Backport 29.0] Harden Contact Sync delta URL and ownership checks (#9966)

Naveen26423:
## What & why

What it does:

Delta URL validation — Any URL stored as a Contact Sync delta URL is now validated against an approved https://graph.microsoft.com/v1.0/ prefix before being persisted. If an invalid URL is already stored, the sync falls back to a full sync instead of using it.

Record ownership enforcement — New EnforceRecordOwnership / EnforceRecordOwnershipOnModify procedures on Contact Sync User table prevent a user from inserting or modifying another user's sync record, and prevent changing the User ID of an existing record.

Event subscriber codeunit (ContactSyncUserSubscriber.Codeunit.al) — Wires the two checks above into OnBeforeInsert and OnBeforeModify table events, so the guards fire on every write path. Install/upgrade execution contexts are exempt.

Test coverage — Adds TestRenameUser tests across the W1, CZ, and DACH test layers covering the rename/ownership scenarios.

Why:

Fixes ADO work item [AB#645036](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20VS%20Code/e4c7e7b1d6/resources/app/out/vs/code/electron-browser/workbench/workbench.html) — closing a security gap where a malicious delta URL or cross-user record manipulation could be exploited in the Contact Sync flow.

## Linked work

Fixes # [AB#645036](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/645036)

## How I validated this

Ran Contact Sync in a fastprod environment with a valid delta URL (https://graph.microsoft.com/v1.0/...); sync completed normally with no errors.
Manually set an invalid/tampered delta URL directly in the database; on next sync the invalid URL was detected, a warning was logged, and the sync fell back to a full sync as expected.
Attempted to insert a Contact Sync User record for a different user; received the expected ownership error and the record was not saved.
Attempted to modify the User ID field on an existing Contact Sync User record; received the expected error blocking the change.
Verified install/upgrade execution context skips ownership checks (no regressions on company copy/upgrade scenarios).

## Risk & compatibility
None

### DIFF
--- a/src/Layers/CZ/Tests/User/TestUserTasks.Codeunit.al
+++ b/src/Layers/CZ/Tests/User/TestUserTasks.Codeunit.al
@@ -16,6 +16,7 @@ codeunit 134769 "Test User Tasks"
         UserTaskGroupMember: Record "User Task Group Member";
         Assert: Codeunit Assert;
         LibraryUtility: Codeunit "Library - Utility";
+        RecordsWithNewUsernameErr: Label 'Records with new username should exist in %1.', Comment = '%1 = table name';
 
     [Test]
     [Scope('OnPrem')]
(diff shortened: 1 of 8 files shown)


darjoo (approved):

VolodySyn (approved):

PredragMaricic (dismissed):
Request changes

S1 - HTTPS validation can be bypassed

`ContactSyncUser.Table.al` and `O365BidirectionalSync.Codeunit.al` rely on `Uri.ValidateIntegrationURL`, which compares only the host and does not enforce the scheme, port, or required path boundary. An `http://graph.microsoft.com/v1.0/...` URL can therefore pass validation and receive the bearer authorization header. Require the `https` scheme, exact approved host, an approved/default HTTPS port, and the required path boundary. Add negative tests for HTTP, custom ports, malformed URLs, and paths outside the accepted scope.

S2 - Rejected delta URLs expose tokens through telemetry

`ContactSyncUser.Table.al` passes the complete rejected delta URL into telemetry. Microsoft Graph delta URLs contain opaque query parameters such as `$deltatoken`, so this can disclose synchronization tokens to extension-publisher telemetry. Never log the complete URL. Record only non-sensitive validation metadata such as the rejection reason or sanitized host, removing the query, fragment, user-info, and token-bearing path values.

S3 - Delete remains an unguarded cross-user write path

`ContactSyncUserSubscriber.Codeunit.al` guards insert and modify but has no delete subscriber, while standard permission sets grant delete access to `Contact Sync User`. A permitted user can delete another user's synchronization rows, causing state loss or a full resynchronization. Enforce ownership in `OnBeforeDeleteEvent`, including bulk-delete behavior, while preserving narrowly defined trusted install or upgrade operations. Add tests for both `Delete` and `DeleteAll`.

S4 - Ownership enforcement breaks legitimate user rename propagation

`ContactSyncUser.Table.al` rejects every change to `User ID`, and the subscriber invokes that guard during modification. The standard user-rename path propagates changes through related fields, but the changed W1, CZ, and DACH tests explicitly exclude this table instead of verifying it. Provide a narrowly authorized user-rename path that updates ownership without enabling arbitrary reassignment, remove the exclusions, and add a regression test with an existing Contact Sync row.

S5 - The owner-change test expects the wrong error

`ContactSyncTest.Codeunit.al` changes `User ID` and expects `CannotModifyOtherUsersSyncErr`, but production checks owner changes first and raises `CannotChangeRecordOwnerErr`. The test will fail when executed and does not establish behavioral coverage. Assert the actual error, run the targeted tests, and add executable coverage for non-HTTPS URLs, invalid stored-link fallback, untrusted next and delta links, deletion, and the clear-sync action.

PR #10567, opened 2026-08-24
SOA : Create Task from TaskPane respects review configuration for known senders

VolodySyn:
Fixes [AB#640666](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/640666)

Tasks created via the TaskPane "Create Task" action (and the "Try it out" flow on SOA Setup) always came in as "Needs attention", ignoring the configured review setting for known/unknown senders. Only the email-originated path applied the setting.

Changes:
- SOACreateTaskImpl.CreateTask now sets RequiresReview from SOATaskMessage.MessageRequiresReview(...), mirroring the email path, so manually created tasks honor the "No review / First message / All messages" configuration.
- Refactored SOATaskMessage.MessageRequiresReview to take the sender address (Text) instead of an Email Inbox record so both the email and manual paths can share it; updated the two email-path callers. Removed the now-unused System.Email using.
- SOA Create Task page: typing a sender email now resolves and initializes the matching contact/customer (per the Sender option), consistent with picking one from the lookup — so sample-message personalization, location, and language are populated. Clears any prior selection first and skips lookup on empty input to avoid stale/blank matches.

PR #10793, opened 2026-08-30
[29.x] SOA : Create Task from TaskPane respects review configuration for known senders

VolodySyn:
Fixes [AB#648306](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/648306)

Backport of https://github.com/microsoft/BCApps/pull/10567

### DIFF
--- a/src/Apps/W1/SalesOrderAgent/app/src/Integration/SOACreateTask.Page.al
+++ b/src/Apps/W1/SalesOrderAgent/app/src/Integration/SOACreateTask.Page.al
@@ -54,10 +54,32 @@ page 4409 "SOA Create Task"
 
                             trigger OnValidate()
                             var
+                                Contact: Record Contact;
+                                Customer: Record Customer;
+                                SOAFiltersImpl: Codeunit "SOA Filters Impl.";
                                 MailManagement: Codeunit "Mail Management";
+                                ContactCount: Integer;
                             begin
-                                if SenderEmail <> '' then
-                                    MailManagement.CheckValidEmailAddresses(SenderEmail);
+                                SOACreateTaskImpl.ClearSelectedSender();
+                                if SenderEmail = '' then
+                                    exit;
+
+                                MailManagement.CheckValidEmailAddresses(SenderEmail);
+
+                                case Sender of
+                                    Sender::Contact:
+                                        if SOAFiltersImpl.FindContactByEmail(Contact, SenderEmail, ContactCount) then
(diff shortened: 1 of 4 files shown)


PredragMaricic (approved):

alexei-dobriansky:

<details>

This change makes tasks created from the Sales Order Agent TaskPane use the same sender-review configuration as messages that arrive from email. The manual create-task path now calls the shared review-decision helper with the sender address and first-message flag before it creates the task message.

The core fix is well aligned with the bug. The helper still makes the same known-vs-unknown sender decision for the email path, and the manual path now supplies the only value that helper needs: the sender address. The new page validation also tries to resolve a typed sender email to a contact or customer so sample-message defaults stay in sync.
</details>

<details>

**Fit:** Strong

The reported problem is that manually created tasks ignore the configured review behavior while email-created tasks honor it. The diff addresses that directly by sharing the review calculation between both paths without changing the email path's effective input.
</details>

<details>

**S1** (🟡 Minor): *Typed sender lookup can pick an arbitrary duplicate*
The typed sender-email path uses the first matching contact or customer when several records share the same email. This can choose the wrong defaults for the sample message, language, or location. Consider keeping the selection empty when there is more than one match, or opening the lookup so the user can choose the right record.

**S2** (🟡 Minor): *Add a regression test for the manual path*
The main fix is that a manually created task now honors the review configuration. A test that creates a manual task for a known sender with review disabled and checks the resulting message would prevent this path from drifting away from the email path again. This is not a blocker because the change is not in a posting, financial, or data-integrity path.
</details>

<details>

**Risk:** Low. The change is limited to Sales Order Agent task creation and sender resolution. It does not touch posting, ledger data, schema, or public event surfaces, and the email path keeps the same sender-address based decision.

**Necessity:** Justified. Manual task creation should respect the same review settings as email ingestion, especially when the setup allows trusted senders to bypass review. Sharing the helper keeps those paths aligned.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10793 round=1 by=alexei-dobriansky at=2026-08-30T18:17:16Z lastSha=ec2538f8bda730333b4ed1e8aff68fdd1d006101 reviewKey=0c4f52e208e23aba97f1e7d5d59f1741418d96f3c7469c9a5cb43e6478b08cfc suggestions=S1@64e43dcf,S2@e0fd4d18

ChethanT (approved):

PR #10794, opened 2026-08-30
[29.0] SOA : Create Task from TaskPane respects review configuration for known senders

VolodySyn:
Fixes [AB#648307](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/648307)

Backport of https://github.com/microsoft/BCApps/pull/10567

### DIFF
--- a/src/Apps/W1/SalesOrderAgent/app/src/Integration/SOACreateTask.Page.al
+++ b/src/Apps/W1/SalesOrderAgent/app/src/Integration/SOACreateTask.Page.al
@@ -54,10 +54,32 @@ page 4409 "SOA Create Task"
 
                             trigger OnValidate()
                             var
+                                Contact: Record Contact;
+                                Customer: Record Customer;
+                                SOAFiltersImpl: Codeunit "SOA Filters Impl.";
                                 MailManagement: Codeunit "Mail Management";
+                                ContactCount: Integer;
                             begin
-                                if SenderEmail <> '' then
-                                    MailManagement.CheckValidEmailAddresses(SenderEmail);
+                                SOACreateTaskImpl.ClearSelectedSender();
+                                if SenderEmail = '' then
+                                    exit;
+
+                                MailManagement.CheckValidEmailAddresses(SenderEmail);
+
+                                case Sender of
+                                    Sender::Contact:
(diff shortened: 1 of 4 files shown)


alexei-dobriansky:

<details>

Tasks created from the Sales Order Agent task pane now use the same review decision as messages that arrive from email. The manual path calls MessageRequiresReview with the sender address and first-message flag, while the email path still passes the same sender address through the refactored helper.

The typed sender email path now also resolves a matching contact or customer so the sample message can use the same sender details as the lookup path. The core fix is targeted and the shared helper keeps the two task-creation paths aligned.
</details>

<details>

**Fit:** Strong

The reported problem is that manually created tasks ignored the configured review setting for registered senders. The change applies that setting at the point where the manual task message is built, and it preserves the existing email-path behavior by passing the same sender address value the old code read from the email record.
</details>

<details>

**S1** (🟠 Moderate): *Add coverage for manual review settings*
The main behavior change is that CreateTask now sets RequiresReview from the setup. A test that creates a manual task for known and unknown sender settings would lock in this path and keep it from drifting away from the email path again.

**S2** (🟡 Minor): *Avoid silently choosing among duplicate senders*
The typed email path uses the first matching contact or customer when several records share the same email. That can fill the sample message with details from a different record than the user expected. Consider only auto-selecting when the match is unique, or leaving the user to pick from the lookup when there are multiple matches.
</details>

<details>

**Risk:** Low. The change is limited to Sales Order Agent task creation and sender resolution. It does not touch posting, ledger data, schema, or public event contracts; the main regression risk is task review state or sample sender details being wrong.

**Necessity:** Justified. A manually created task should respect the same review configuration as an email-created task, and sharing the same review-decision helper is the right scope for that fix.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10794 round=1 by=alexei-dobriansky at=2026-08-30T18:15:55Z lastSha=772f3a1b747dff299aeed283cca3dc49588de721 reviewKey=d6c89b283c458851b65628736d440758974ba6d2b1c3d50356663a63c13bc40e suggestions=S1@e0fd4d18,S2@64e43dcf

ChethanT (approved):

PredragMaricic (approved):

PR #10805, opened 2026-08-31
[29.0] Extend OnBeforeCheckBankAcc to skip specific bank reversal checks

Explorer986:
Backport of #10630.

## What & why

This change lets extensions independently skip the Open and Statement No. validations when reversing a bank account ledger entry. The remaining standard reversal checks continue to run, avoiding the need for extensions to handle the entire validation through the existing IsHandled parameter.

The new controls are added as trailing parameters to the existing OnBeforeCheckBankAcc event, preserving compatibility with existing subscribers.

Fixes [AB#648338](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/648338)

## How I validated this

- Added a test that closes a bank account ledger entry, skips only the Open check, and verifies that the entry is reversed.
- Added a test that assigns a statement number, skips only the Statement No. check, and verifies that the entry is reversed.
- Verified that the publisher and test subscriber have no AL editor diagnostics.
- The tests were not run locally because the enlistment build and test environment could not be initialized.

## Risk & compatibility

The event change is additive. The existing parameters and IntegrationEvent attributes remain unchanged, and the new Boolean parameters are appended to the event signature.

### DIFF
--- a/src/Layers/APAC/BaseApp/Finance/GeneralLedger/Reversal/ReversalEntry.Table.al
+++ b/src/Layers/APAC/BaseApp/Finance/GeneralLedger/Reversal/ReversalEntry.Table.al
@@ -728,9 +728,13 @@ table 179 "Reversal Entry"
         BankAccount: Record "Bank Account";
         CheckLedgerEntry: Record "Check Ledger Entry";
         IsHandled: Boolean;
+        SkipBankAccLedgEntryOpenCheck: Boolean;
+        SkipBankAccLedgEntryStatementNoCheck: Boolean;
     begin
         IsHandled := false;
-        OnBeforeCheckBankAcc(BankAccountLedgerEntry, IsHandled);
+        SkipBankAccLedgEntryOpenCheck := false;
+        SkipBankAccLedgEntryStatementNoCheck := false;
+        OnBeforeCheckBankAcc(BankAccountLedgerEntry, IsHandled, SkipBankAccLedgEntryOpenCheck, SkipBankAccLedgEntryStatementNoCheck);
         if IsHandled then
             exit;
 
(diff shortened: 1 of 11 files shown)


alexei-dobriansky:

# **Recommendation: Accept**

<details>

This change appends two Boolean `var` parameters to `OnBeforeCheckBankAcc` in the bank reversal table copies and uses them to skip only the Open or Statement No. validation when a subscriber opts in. The publisher is present in the changed BaseApp table copies and is raised before those validations; both flags default to false, `IsHandled` keeps its existing full-exit behavior, and the posting-date, blocked-account, already-reversed, and check-ledger-entry checks still run on the normal path.
</details>

<details>

**Fit:** Strong

The requested behavior is to avoid handling the whole validation procedure when only one of the two bank ledger checks needs to be bypassed. The diff implements that narrow contract in each changed table copy and adds tests for both positive and negative combinations in the affected reverse bank ledger test apps.
</details>

<details>

*None - I did not find any issues that need changes.*
</details>

<details>

**Risk:** This touches bank reversal validation in table 179 across W1 and country-layer copies, so a wrong flag could let entries reverse when they should not. The risk is contained because the new flags are opt-in, default to false on every call, preserve event compatibility by appending parameters, and the tests cover each skip flag without skipping the other check.

**Necessity:** The change is needed because the existing single `IsHandled` flag is too broad for extensions that need the standard validation with only one specific exception. The scope is right for this kind of change: it adds no new event and leaves the rest of the reversal logic unchanged.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10805 round=1 by=alexei-dobriansky at=2026-08-31T13:18:27.0812704Z lastSha=e67da41e882a0eb89f8bd0c8efb245d2aa641bd5 reviewKey=a8ced09f44d691c295c16be589f2f16565751ed47fd4d70adb22d3175619c6da suggestions=none

ventselartur (approved):

AndersLarsenMicrosoft (approved):

PR #10806, opened 2026-08-31
[29.x] Extend OnBeforeCheckBankAcc to skip specific bank reversal checks

Explorer986:
Backport of #10630.

## What & why

This change lets extensions independently skip the Open and Statement No. validations when reversing a bank account ledger entry. The remaining standard reversal checks continue to run, avoiding the need for extensions to handle the entire validation through the existing IsHandled parameter.

The new controls are added as trailing parameters to the existing OnBeforeCheckBankAcc event, preserving compatibility with existing subscribers.

Fixes [AB#648337](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/648337)

## How I validated this

- Added a test that closes a bank account ledger entry, skips only the Open check, and verifies that the entry is reversed.
- Added a test that assigns a statement number, skips only the Statement No. check, and verifies that the entry is reversed.
- Verified that the publisher and test subscriber have no AL editor diagnostics.
- The tests were not run locally because the enlistment build and test environment could not be initialized.

## Risk & compatibility

The event change is additive. The existing parameters and IntegrationEvent attributes remain unchanged, and the new Boolean parameters are appended to the event signature.

### DIFF
--- a/src/Layers/APAC/BaseApp/Finance/GeneralLedger/Reversal/ReversalEntry.Table.al
+++ b/src/Layers/APAC/BaseApp/Finance/GeneralLedger/Reversal/ReversalEntry.Table.al
@@ -728,9 +728,13 @@ table 179 "Reversal Entry"
         BankAccount: Record "Bank Account";
         CheckLedgerEntry: Record "Check Ledger Entry";
         IsHandled: Boolean;
+        SkipBankAccLedgEntryOpenCheck: Boolean;
+        SkipBankAccLedgEntryStatementNoCheck: Boolean;
     begin
         IsHandled := false;
-        OnBeforeCheckBankAcc(BankAccountLedgerEntry, IsHandled);
+        SkipBankAccLedgEntryOpenCheck := false;
+        SkipBankAccLedgEntryStatementNoCheck := false;
+        OnBeforeCheckBankAcc(BankAccountLedgerEntry, IsHandled, SkipBankAccLedgEntryOpenCheck, SkipBankAccLedgEntryStatementNoCheck);
         if IsHandled then
             exit;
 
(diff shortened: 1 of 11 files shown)


alexei-dobriansky:

# **Recommendation: Accept**

<details>

This change extends `OnBeforeCheckBankAcc` in each BaseApp layer copy with two trailing `var Boolean` flags. Both flags default to false before the event is raised, and they guard only the Open and Statement No. errors. `IsHandled` still skips the whole validation when requested, while the posting-date, blocked-account, already-reversed, and related check-ledger-entry validations still run.
</details>

<details>

**Fit:** Strong

The requested behavior is to let subscribers skip either of two specific bank reversal checks without taking over the full validation procedure. The implementation does that directly and keeps existing subscribers compatible because the new event parameters are appended and default to false.
</details>

<details>

*None.*
</details>

<details>

**Risk:** The changed code is in a financial reversal path, but the regression surface is narrow. Existing behavior is preserved when no subscriber sets the new flags, and the W1, ES, and SE tests cover both successful skips and the case where skipping one check must not skip the other. The remaining layer copies have the same production change.

**Necessity:** The change is useful because it lets extensions reuse standard bank reversal validation while bypassing only the check that does not apply to their controlled scenario. The scope is appropriate: it adds targeted opt-in controls to the existing event instead of requiring duplicated reversal logic or a full `IsHandled` bypass.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10806 round=1 by=alexei-dobriansky at=2026-08-31T13:20:23Z lastSha=4db8f5ecc5d0d0b1dbf7f5af375939585ba90cbe reviewKey=b8f369776f0a3f2a3782e23697e06f3c4599ad06b30d0621911f27a91e0670df suggestions=

ventselartur (approved):

AndersLarsenMicrosoft (approved):

PR #10807, opened 2026-08-31
[29.x][Expense Agent] Confirm before closing Expense User Card without Employee No. 

v-rohangarg20:
Fixes [AB#648361](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/648361)

Issue: On the Expense User card (page 6949) the mandatory Employee No. field could be left blank and the card closed anyway. The card closed silently and the expense user was saved without being linked to an employee — no warning and no confirmation — so an unlinked expense user could be created unintentionally.

Cause: Employee No. was only flagged as mandatory for display (ShowMandatory); the page had no OnQueryClosePage guard. Nothing intercepted the close action while the field was empty, so the record was committed with a blank Employee No.

Solution: Added an OnQueryClosePage trigger on page 6949 that uses Codeunit "Confirm Management". When Employee No. is blank the user is prompted "Employee No. is blank. The expense user will not be linked to an employee. Are you sure you want to exit?"; choosing No cancels the close and keeps the card open. Added [HandlerFunctions('ConfirmHandlerYes')] to the two tests that close the card with a blank Employee No. (EmployeeIsNotCreatedFromExpenseUserWhenNameIsEmpty, ...WhenEmailIsEmpty) so they still assert the mandatory-field errors.

### DIFF
--- a/src/Apps/W1/ExpenseAgent/app/src/MasterData/Pages/ExpenseUser.Page.al
+++ b/src/Apps/W1/ExpenseAgent/app/src/MasterData/Pages/ExpenseUser.Page.al
@@ -260,7 +260,7 @@ page 6949 "Expense User"
     var
         NoFieldVisible: Boolean;
         IsCreateEmployeeVisible: Boolean;
-        CloseWithoutEmployeeNoQst: Label '%1 is blank. Do you want to close the card without linking an %1?', Comment = '%1 = Employee No. field caption';
+        CloseWithoutEmployeeNoQst: Label '%1 is blank. The expense user will not be linked to an employee.\\Are you sure you want to exit?', Comment = '%1 = Employee No. field caption';
 
     local procedure SetCodeFieldVisible()
     var
--- a/src/Apps/W1/ExpenseAgent/app/src/MasterData/Pages/ExpenseUser.Page.al
+++ b/src/Apps/W1/ExpenseAgent/app/src/MasterData/Pages/ExpenseUser.Page.al
@@ -5,6 +5,7 @@
 namespace Microsoft.ExpenseAgent;
 
 using Microsoft.Projects.Resources.Resource;
+using System.Utilities;
 
 page 6949 "Expense User"
 {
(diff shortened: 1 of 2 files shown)


alexei-dobriansky:

<details>

This adds a close check on the Expense User card when Employee No. is blank. The user gets a warning and choosing No keeps the card open, while choosing Yes keeps the old close behavior.

The fix targets the right boundary: the field was only marked mandatory in the UI, so the close path still needed a guard. The condition is narrow, and the decline path returns false before the page closes; records with an Employee No. and users who confirm continue through the existing path.
</details>

<details>

**Fit:** Strong

The bug says an Expense User could be saved without an employee link and without warning. The new close confirmation directly covers that case and lets the user either accept the unlinked record or stay on the card.
</details>

<details>

**S1** (🟠 Moderate): *Cover declined confirmation path*
The changed tests answer Yes only, so they do not prove that choosing No keeps the Expense User card open. Add a small TestPage case with ConfirmHandlerNo around ExpenseUserPage.Close() to cover the new cancel path.
</details>

<details>

**Risk:** The regression surface is limited to closing page 6949 when Employee No. is blank. This is UI behavior only; it does not touch posting, ledger entries, amounts, public APIs, or event contracts.

**Necessity:** The change is needed because ShowMandatory only shows the field as required and does not stop an unlinked Expense User from being saved. The scope is small and matches the reported problem.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10807 round=1 by=alexei-dobriansky at=2026-09-01T07:16:14Z lastSha=59e3ef43ae662f36ad7d7e8a42770e2290a1a160 reviewKey=41b35c1d515623c5706d2ad507952b036651d31d5c964ae7318d85e945754d8f suggestions=S1@9bbfb151

t-prda (approved):

ViditGupta1277 (approved):

PR #10819, opened 2026-08-31
[Pmt. Recon. Journal] Re-fire OnAfterGetLedgEntryInfo on proposal fast path

dcenic:
Re-firing an event that stopped firing because of a bug fix.

## What & why

The [#10498](https://github.com/microsoft/BCApps/pull/10498) perf refactor replaced GetLedgEntryInfo() with LoadLedgEntryInfoAndRemainingAmount(), which stopped raising Applied Payment Entry.OnAfterGetLedgEntryInfo for Customer/Vendor/Employee/Bank Account during proposal creation. Re-fire the legacy event via a temp round-trip (no second ledger read) so existing subscribers keep updating proposal fields.

## Linked work

Fixes [AB#648382](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/648382)

## How I validated this

## Risk & compatibility

Low, it re-fires an event that used to fire before a refactoring that broke it.

### DIFF
--- a/src/Layers/W1/BaseApp/Bank/Reconciliation/AppliedPaymentEntry.Table.al
+++ b/src/Layers/W1/BaseApp/Bank/Reconciliation/AppliedPaymentEntry.Table.al
@@ -895,6 +895,11 @@ table 1294 "Applied Payment Entry"
         OnAfterGetLedgEntryInfo(Rec);
     end;
 
+    internal procedure RunOnAfterGetLedgEntryInfo()
+    begin
+        OnAfterGetLedgEntryInfo(Rec);
+    end;
+
     local procedure GetCustLedgEntryInfo()
     var
         CustLedgEntry: Record "Cust. Ledger Entry";
--- a/src/Layers/W1/BaseApp/Bank/Reconciliation/PaymentApplicationProposal.Table.al
+++ b/src/Layers/W1/BaseApp/Bank/Reconciliation/PaymentApplicationProposal.Table.al
@@ -494,6 +494,7 @@ table 1293 "Payment Application Proposal"
         LedgerRemainingAmount: Decimal;
         IsHandled: Boolean;
         RemainingAmountHandled: Boolean;
+        LedgEntryInfoLoaded: Boolean;
     begin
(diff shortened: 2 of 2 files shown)


ventselartur (approved):

Explorer986 (approved):

alexei-dobriansky:

<details>

This restores the legacy `OnAfterGetLedgEntryInfo` extension point when payment application proposals use the optimized ledger-entry loading path. The fast path already fills the same customer, vendor, employee, and bank account fields as the old temp `Applied Payment Entry` path, then round-trips those fields through a temp record to run the existing event and copy subscriber changes back. The placement is before the remaining amount update, which matches the old order and keeps the later remaining-amount override event available.
</details>

<details>

**Fit:** Strong

The reported regression is narrow: proposal creation stopped calling an existing event after the ledger lookup was optimized. The diff targets that exact path, keeps the optimized single ledger read, and leaves the fallback path that already calls the event unchanged.
</details>

<details>

**S1** (🟠 Moderate): *Add fast-path event regression coverage*
Please add a test that creates a payment application proposal from a customer, vendor, employee, or bank account candidate and subscribes to `OnAfterGetLedgEntryInfo`. The test should prove the subscriber is called on the fast path and that a changed field is copied back to the proposal.
</details>

<details>

**Risk:** The change affects payment application proposal creation for customer, vendor, employee, and bank account candidates. The event signature is unchanged, and the new internal helper only reuses the existing publisher on a temporary record, so compatibility risk is low; the main risk is a future regression because the event is now reached indirectly.

**Necessity:** The change is needed to restore an existing extension point without undoing the performance improvement from loading the ledger entry once. The scope is small and matches the compatibility problem.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10819 round=1 by=alexei-dobriansky at=2026-08-31T13:21:17Z lastSha=bd19b027395ddb4e9423a7440ac42473148346c8 reviewKey=5360b19e517bb78686490f3729fa8151c1699da26e9062c278850f70a789c8cc suggestions=S1@f7e3d3d6

PR #10826, opened 2026-08-31
[Master] - Bug 648138: [master][NAV] Track failed BCApps validation

v-rohangarg20:
## What & why

Demo data generation fails when posting expense reports. The `Expense GL Account Names` labels are not translated for localizations, so the expense G/L account lookup returns nothing and `Expense Report-Post` errors out. This was identified during uptake.

This change guards the posting step with a `SkipPostingExpenseReport()` function in both historical demo data codeunits that `CreateHistoricalData` runs:

- codeunit 8217 `Create Posted Expense Report`
- codeunit 8232 `Create Posted Expense Reports`

Expense reports are still created; only the posting step is skipped. The function returns `true` today and should be removed once the translations are generated.

## Linked work

[AB#648138](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/648138)

## How I validated this

- No tests added: this touches demo data generation only and temporarily disables a step that currently fails. No posted expense reports were produced in the broken state, so existing coverage is unaffected.

## Risk & compatibility

- Contoso demo data will no longer contain posted expense reports until the skip is removed. Any demo or manual flow that relied on posted expense report history from demo data will see an empty list.
- Temporary by design. Follow-up: remove `SkipPostingExpenseReport()` once the `Expense GL Account Names` translations are generated.

### DIFF
--- a/src/Apps/W1/ExpenseAgent/app/src/Setup/Codeunits/ExpenseRuleValidation.Codeunit.al
+++ b/src/Apps/W1/ExpenseAgent/app/src/Setup/Codeunits/ExpenseRuleValidation.Codeunit.al
@@ -360,10 +360,11 @@ codeunit 6902 "Expense Rule Validation"
     local procedure ExistDuplicateInPostedExpenseReportLine(ExpenseReportLine: Record "Expense Report Line"): Boolean
     var
         PostedExpenseReportLine: Record "Posted Expense Report Line";
+        NegativeExpenseDateFormula: DateFormula;
     begin
         PostedExpenseReportLine.SetRange("Expense Ext. Doc. No.", ExpenseReportLine."Expense Ext. Doc. No.");
-        if Format(ExpenseAgentSetup."Do Not Allow Exp. Older Than") <> '' then
-            PostedExpenseReportLine.SetRange("Expense Date", CalcDate(StrSubstNo('<-%1>', ExpenseAgentSetup."Do Not Allow Exp. Older Than"), Today()), Today())
+        if TryGetNegativeExpenseDateFormula(NegativeExpenseDateFormula) then
+            PostedExpenseReportLine.SetRange("Expense Date", CalcDate(NegativeExpenseDateFormula, Today()), Today())
         else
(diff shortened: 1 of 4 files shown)


t-prda (dismissed):

alexei-dobriansky:

<details>

The change keeps expense reports being created in historical Contoso demo data, but it skips the posting step in both historical expense report codeunits. That matches the reported validation failure path because posting is where the blank expense posting accounts are checked. The guard is placed directly around `PostExpenseReport()`, so it avoids the failing code without changing the expense creation flow or the account lookup logic.
</details>

<details>

**Fit:** Partial

The bug says translated demo-data generation fails during expense report posting because required expense G/L accounts are not found. This change unblocks that path by not posting the demo reports, but it does not repair the account setup and it disables posted expense report history until the temporary gate is removed.
</details>

<details>

**S1** (🟠 Moderate): *Constant skip keeps valid posted demo data disabled*
Both `SkipPostingExpenseReport()` functions always return `true`, so posting stays disabled even after the G/L account setup is valid. Consider checking the missing account setup instead, or use another temporary switch that can fail closed when the setup is fixed.
</details>

<details>

**Risk:** The change is narrow and only touches historical Expense Agent demo-data generation. The regression surface is demo scenarios that expect posted expense report history; those records will no longer be produced while the guard returns `true`. There is no public API or event signature change.

**Necessity:** The change is justified as a short-term unblock for validation, because the failing path prevents translated demo databases from being generated. The scope is small, but the temporary skip needs a clear removal path so it does not become the long-term behavior.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10826 round=1 by=alexei-dobriansky at=2026-08-31T13:20:03.817Z lastSha=ada2555b00497e5d9ca3fbc9e6c321864401d68f reviewKey=343d4c4ef1108588146b98b9a3268c51a2003301f2db22f3ecb572c0d6f3ba02 suggestions=S1@31259f94

ViditGupta1277 (dismissed):

t-prda (approved):

v-ankitgoyal (approved):

PR #10866, opened 2026-09-01
[29.x]No Employee posting group for AT when new employee is created

sanjmaurya:
[Bug 648335](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/648335): [all-e][29.x]No Employee posting group for AT when new employee is created

Fixes [AB#648335](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/648335)

**Issue:**
The ADMINISTRATION and IT employee templates, and the demo employees created from them, did not have an Employee Posting Group assigned.

**Cause:**
The AT localization subscriber explicitly cleared the "Employee Posting Group" field while inserting these employee templates.

**Solution:**
Updated the subscriber to assign the EMPLEXP employee posting group using CreateEmployeePostingGroup.EmployeeExpenses(). Added tests to verify the posting group on both affected templates and all generated demo employees

### DIFF
--- a/src/Apps/AT/ContosoCoffeeDemoDatasetAT/app/DemoData/HumanResources/2. Master Data/CreateEmployeeTemplateAT.Codeunit.al
+++ b/src/Apps/AT/ContosoCoffeeDemoDatasetAT/app/DemoData/HumanResources/2. Master Data/CreateEmployeeTemplateAT.Codeunit.al
@@ -18,11 +18,12 @@ codeunit 11155 "CreateEmployee Template AT"
     [EventSubscriber(ObjectType::Table, Database::"Employee Templ.", 'OnBeforeInsertEvent', '', false, false)]
     local procedure OnBeforeInsertEmployeeTemplate(var Rec: Record "Employee Templ.")
     var
+        CreateEmployeePostingGroup: Codeunit "Create Employee Posting Group";
         CreateEmployeeTemplate: Codeunit "Create Employee Template";
     begin
         case Rec.Code of
             CreateEmployeeTemplate.AdminCode(), CreateEmployeeTemplate.ITCode():
-                Rec.Validate("Employee Posting Group", '');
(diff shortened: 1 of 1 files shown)


alexei-dobriansky:

<details>

This changes the AT employee-template insert subscriber so the ADMINISTRATION and IT templates keep the `EMPLEXP` employee posting group instead of being set to blank. The change is in the right place: it runs before those template records are inserted, uses the same `EmployeeExpenses()` source as the setup data, and keeps the scope to the two affected template codes.
</details>

<details>

**Fit:** Strong

The bug says new AT employees created from the ADMINISTRATION or IT template do not get an employee posting group. The diff changes the exact localization override that cleared the posting group for those templates and replaces it with the expected group.
</details>

<details>

**S1** (🟠 Moderate): *Add regression coverage for AT templates*
Please add a test that generates the AT human resources demo data and checks the ADMINISTRATION and IT templates use `EMPLEXP`. That would catch this bug if the subscriber clears or changes the value again.
</details>

<details>

**Risk:** The change is limited to AT Contoso Coffee demo data and only affects the two employee templates while they are generated. It does not change public API or posting logic, but a wrong value here can still leave demo users with employees that have no posting group.

**Necessity:** The change is needed because the localization subscriber was overriding the base template value with blank. The manual workaround is to fill the posting group on the template or employee, but the normal demo-data workflow should create usable templates.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10866 round=1 by=alexei-dobriansky at=2026-09-01T06:14:12.180Z lastSha=15e9cf41e021ac94614b70801a8e241f09687a91 reviewKey=313a10867b50f25145b6232df39aaa59ea6c1e3bf0d7b603f92597c05475fcb5 suggestions=S1@d417db1e

v-ankitgoyal (approved):

JesperSchulz (approved):

PR #10884, opened 2026-09-01
[Bug Fix] Fix recurring journal background posting for IRS 1099 invoice lines (US)

Explorer986:
## What & why

Fixes [AB#626641](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/626641)

## How I validated this

## Risk & compatibility

### DIFF
--- a/src/Apps/US/IRSForms/app/src/Extensions/IRS1099BaseAppSubscribers.Codeunit.al
+++ b/src/Apps/US/IRSForms/app/src/Extensions/IRS1099BaseAppSubscribers.Codeunit.al
@@ -269,7 +269,7 @@ codeunit 10032 "IRS 1099 BaseApp Subscribers"
 
     local procedure SaveChangesInGenJnlLine(var GenJnlLine: Record "Gen. Journal Line")
     begin
-        if GenJnlLine."Line No." <> 0 then
+        if (GenJnlLine."Line No." <> 0) and (GenJnlLine."Job Queue Status" = GenJnlLine."Job Queue Status"::" ") then
             if GenJnlLine.Modify(true) then;
     end;
 
--- a/src/Apps/US/IRSForms/test/src/IRS1099DocumentTests.Codeunit.al
+++ b/src/Apps/US/IRSForms/test/src/IRS1099DocumentTests.Codeunit.al
@@ -1380,6 +1380,91 @@ codeunit 148010 "IRS 1099 Document Tests"
         Assert.AreEqual(FormBoxNo[2], PurchCrMemoHdr."IRS 1099 Form Box No.", 'IRS 1099 Form Box No. in Purch. Cr. Memo Hdr. should be updated');
     end;
 
+    [Test]
+    procedure InvoiceLineScheduledForPostingSyncsIRSDataWithoutModifyError()
+    var
+        GenJournalLine: Record "Gen. Journal Line";
+        VendNo: Code[20];
+        FormNo: Code[20];
+        FormBoxNo: Code[20];
+        PeriodDate: Date;
+    begin
+        // [FEATURE] [Recurring Journal] [Background Posting]
+        // [SCENARIO 626641] Validating an IRS-relevant field on a Vendor Invoice journal line whose Job Queue Status
+        //                   is "Scheduled for Posting" must not fail. Background posting re-validates the Posting Date
+        //                   of recurring lines while the status is set; the IRS 1099 sync subscriber must persist its
+        //                   derived fields without re-entering the line's OnModify posting guard.
+        Initialize();
+        // [GIVEN] An IRS Reporting Period and a vendor with a 1099 form box setup in that period
+        PeriodDate := WorkDate();
+        LibraryIRSReportingPeriod.CreateOneDayReportingPeriod(PeriodDate);
+        FormNo := LibraryIRS1099FormBox.CreateSingleFormInReportingPeriod(PeriodDate);
+        FormBoxNo := LibraryIRS1099FormBox.CreateSingleFormBoxInReportingPeriod(PeriodDate, FormNo);
+        VendNo := LibraryIRS1099FormBox.CreateVendorNoWithFormBox(PeriodDate, FormNo, FormBoxNo);
+        // [GIVEN] A Vendor Invoice recurring journal line with Posting Date outside the IRS period, so validation will sync IRS data
+        CreateRecurringGenJnlLineForVendor(GenJournalLine, VendNo, GenJournalLine."Document Type"::Invoice, CalcDate('<-1D>', PeriodDate));
+        // [GIVEN] The line has been scheduled for background posting (status written without triggers)
+        GenJournalLine."Job Queue Status" := GenJournalLine."Job Queue Status"::"Scheduled for Posting";
+        GenJournalLine.Modify(false);
+
+        // [WHEN] The posting engine re-validates the Posting Date to a date inside the IRS period
+        GenJournalLine.Validate("Posting Date", PeriodDate);
+
+        // [THEN] No "scheduled for posting" error is raised and the IRS data is synced and persisted
+        GenJournalLine.TestField("Posting Date", PeriodDate);
+        VerifyIRS1099DataInGenJnlLine(GenJournalLine, PeriodDate, FormNo, FormBoxNo);
+        GenJournalLine.Find();
+        VerifyIRS1099DataInGenJnlLine(GenJournalLine, PeriodDate, FormNo, FormBoxNo);
+    end;
+
+    [Test]
+    procedure BlankDocTypeLineScheduledForPostingIsUnaffected()
+    var
+        GenJournalLine: Record "Gen. Journal Line";
+        VendNo: Code[20];
+        FormNo: Code[20];
+        FormBoxNo: Code[20];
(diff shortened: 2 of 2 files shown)


alexei-dobriansky:

<details>

This PR changes the IRS 1099 journal-line subscriber to save derived IRS fields with `Modify(false)`, and adds tests for scheduled recurring journal lines. The reported bug is a US recurring journal background-posting failure for invoice lines when the IRS subscriber re-enters the Gen. Journal Line `OnModify` trigger.

The change removes the trigger call that caused the failure, so the tested validation path can continue. However, `SaveChangesInGenJnlLine` is used by the Posting Date, Account No., Bal. Account No., Document Type, and Amount validation subscribers, not only by the posting batch. Because `Modify(false)` writes the whole line, the fix can save a scheduled journal line before the normal scheduled-for-posting guard runs.
</details>

<details>

**Fit:** Partial

The bug is clear and the code targets the failing IRS 1099 sync path. The solution is too broad because it bypasses `OnModify` for every IRS journal-line sync save, including normal field validation on lines already scheduled for posting.
</details>

<details>

**S1** (🔴 High): *Keep scheduled lines protected*
Do not use `Modify(false)` from the generic Gen. Journal Line validation subscribers while the line is scheduled. A user can change an invoice journal line after it is sent to the job queue, and this call can save the changed line before the normal scheduled-for-posting guard runs. Move the IRS save to the recurring posting flow, or add a guard that only allows the background posting update and leaves user edits blocked.
</details>

<details>

**Risk:** This is a posting and financial path. The changed helper is reached from date, account, document type, and amount validation; with a scheduled line it can bypass `CheckJobQueueStatus`, approval checks, `Check Printed`, and related `OnModify` logic. There are no public API or schema changes.

**Necessity:** The change is needed because the current behavior blocks a valid background-posting scenario, and the workaround is to remove the job queue or turn off background posting. The scope should be narrower than changing every IRS journal-line sync save to skip table triggers.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10884 round=1 by=alexei-dobriansky at=2026-09-01T13:19:03Z lastSha=9db273c1a55aeca68e71cfb43fdb45525e872db0 reviewKey=85d16b63b29d5ba153d9430d6f01af3983adb095459bd679e7cf6db7406994dc suggestions=S1@85feee94

src/Apps/US/IRSForms/app/src/Extensions/IRS1099BaseAppSubscribers.Codeunit.al
```diff
@@ -54,8 +54,6 @@ codeunit 10032 "IRS 1099 BaseApp Subscribers"
     [EventSubscriber(ObjectType::Table, Database::"Vendor Ledger Entry", 'OnAfterCopyVendLedgerEntryFromGenJnlLine', '', false, false)]
     local procedure UpdateIRSDataOnAfterCopyVendLedgerEntryFromGenJnlLine(var VendorLedgerEntry: Record "Vendor Ledger Entry"; GenJournalLine: Record "Gen. Journal Line")
     begin
-        if GenJournalLine."IRS 1099 Form Box No." = '' then
```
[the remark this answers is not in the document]
Explorer986: unrelated to latest changes

AndersLarsenMicrosoft (dismissed):

alexei-dobriansky:

# **Recommendation: Accept**

<details>

The new commit removes the IRS 1099 subscriber self-save from the Gen. Journal Line validation path and deletes the helper that performed it. It also updates the recurring background-posting tests so the subscriber only changes the record buffer, while the posting flow owns the final save.

This addresses the earlier concern. The BaseApp recurring posting flow validates the next Posting Date while the line has a job queue status, then saves the updated recurring line itself with `Modify()` without re-entering `OnModify`. With the subscriber no longer calling `Modify(true)` or `Modify(false)`, scheduled user edits remain blocked, and background posting can persist the intended IRS fields through the owner flow.
</details>

<details>

| S1 | Keep scheduled lines protected | Addressed | Fixed in d53cdfc |
</details>

<details>

*None - the new commit removes the unsafe self-save and adds coverage for the scheduled-line guard.*
</details>

<details>

**Risk:** This remains a sensitive US 1099 recurring-journal posting path because an incorrect save can affect scheduled journal lines and posted vendor reporting data. The current shape is narrow: it removes the extra save from the subscriber and relies on the existing posting owner to persist the line after validation. No public API, schema, permission, or upgrade behavior changes are introduced.

**Necessity:** The change is needed because recurring journals with invoice lines should post through background posting without failing on the scheduled-for-posting guard. The scope is now limited to the IRS 1099 Gen. Journal Line sync path that caused the failure.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10884 round=2 by=alexei-dobriansky at=2026-09-02T13:21:08Z lastSha=d53cdfcdc6c061595e5a2ce08881f7596ae50428 reviewKey=966386b3f95d4618e8773bdcf78aaba13f3fa793ea46a59c26f5e40300a18f09 suggestions=S1@85feee94:addressed parentRound=1

AleksanderGladkov (dismissed):

src/Apps/US/IRSForms/app/src/Extensions/IRS1099BaseAppSubscribers.Codeunit.al
```diff
@@ -267,10 +265,4 @@ codeunit 10032 "IRS 1099 BaseApp Subscribers"
         exit(GenJnlLine."Document Type" in [GenJnlLine."Document Type"::Invoice, GenJnlLine."Document Type"::"Credit Memo"]);
     end;
 
-    local procedure SaveChangesInGenJnlLine(var GenJnlLine: Record "Gen. Journal Line")
-    begin
-        if GenJnlLine."Line No." <> 0 then
-            if GenJnlLine.Modify(true) then;
```
ventselartur, line 273: Ok, so i just debugged the issue and the right solution would be to verify the GenJnlLine."Job Queue Status" before we do the modify:

if GenJnlLine."Line No." <> 0 and GenJnlLine."Job Queue Status" = '' then
  if GenJnlLine.Modify(true) then;

The problem comes from GenJnlPostBatch codeunit that calls this:
UpdateAndDeleteLine
begin
...
                    GenJnlLine2.Validate(
                      "Posting Date", CalcDate(GenJnlLine2."Recurring Frequency", GenJnlLine2."Posting Date"));

...
**GenJnlLine2.Modify();**
...

The line marked in black would call the modify without the trigger and we are fine with that. Then it means that in the 1099 code we could verify for the job queue status and does nothing.

ventselartur (changes requested):

ventselartur (approved):

src/Apps/US/IRSForms/app/src/Extensions/IRS1099BaseAppSubscribers.Codeunit.al
Explorer986: Thanks, Updated

mynjj (approved):

alexei-dobriansky:

<details>

The latest change brings back the IRS 1099 journal-line self-save, but only when `Job Queue Status` is blank. Scheduled background-posting lines are still left for the posting flow to save, so the earlier scheduled-line bypass concern is addressed.

The remaining problem is the exact status check. The Gen. Journal Line table allows `Modify(true)` when `Job Queue Status` is blank or `Error`, but the new helper only allows blank. That can leave recalculated IRS fields unsaved on failed background-posting lines that are being corrected.
</details>

<details>

| S1 | Keep scheduled lines protected | Addressed | The helper no longer saves lines while they are scheduled for posting. |
</details>

<details>

**S2** (🔴 High): *Allow failed scheduled lines to save IRS data*
The new guard only allows the self-save when `Job Queue Status` is blank. The table `OnModify` guard also allows `Error`, so a failed background-posting line can be edited but this subscriber will not persist the recalculated IRS fields. Include `Error` in the allowed statuses, or use the same condition as the table guard.

**S3** (🟠 Moderate): *Re-read the created journal line by key*
`GenJournalLine.Find()` does not target the line created by the test. If another line sorts before it, the persistence check can read the wrong record. Use `Get("Journal Template Name", "Journal Batch Name", "Line No.")` before verifying persisted IRS fields.
</details>

<details>

**Risk:** This remains a sensitive US 1099 recurring-journal posting path. A stale reporting period, form, box, or reporting amount can affect vendor reporting data. The change is narrow and has no public API, schema, permission, or upgrade impact.

**Necessity:** The change is needed because recurring journals with invoice lines should post through background posting without failing on the scheduled-for-posting guard. The scope is right, but the status check should match the platform guard so failed background-posting lines do not lose IRS sync updates.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10884 round=3 by=alexei-dobriansky at=2026-09-07T13:57:51Z lastSha=a06540b4726f6006da56a79aec68e64344e71026 reviewKey=2355a9c4b0d12bb27d542c688aabe4f79e0bac516d356419aa641c0195f3820a suggestions=S1@85feee94:addressed,S2@8d26d8bc:new,S3@3dbcd66f:new parentRound=2

PR #10923, opened 2026-09-02
Bug 626576: Surface AppSource install errors instead of failing silently

martinsrui-msft:
InstallAppsourceExtensionWithRefreshSession calls InstallAppsourceExtension, which is a [TryFunction]. When the installation raised an error the TryFunction swallowed it and the caller never read GetLastErrorText(), so the install dialog just closed with nothing shown to the user.

Both overloads now capture GetLastErrorText() immediately after the TryFunction returns false and surface it via Message(), matching the existing pattern in InvokeExtensionInstallation. Also adds telemetry for the failure and corrects the misleading "successful installation returns false" comment.

## Linked work

Fixes #

## How I validated this

image

### DIFF
--- a/src/System Application/App/Extension Management/src/ExtensionMarketplace.Codeunit.al
+++ b/src/System Application/App/Extension Management/src/ExtensionMarketplace.Codeunit.al
@@ -48,6 +48,8 @@ codeunit 2501 "Extension Marketplace"
         TelemetryTok: Label 'ExtensionManagementTelemetryCategoryTok', Locked = true;
         OperationResult: Option UserNotAuthorized,DeploymentFailedDueToPackage,DeploymentFailed,Successful,UserCancel,UserTimeOut;
         AppDoesntNeedSetupMsg: Label 'Your app is installed and ready to use.';
+        InstallationFailedMsg: Label 'The app could not be installed.\%1', Comment = '%1=the error that was reported by the installation';
+        InstallationFailedTelemetryTxt: Label 'AppSource app installation failed. App ID: %1. Error: %2', Comment = '%1=the app id of the extension, %2=the error that was reported by the installation';
 
     local procedure GetValue(JObject: DotNet JObject; Property: Text; ThrowError: Boolean): Text
     begin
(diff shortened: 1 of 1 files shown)


src/System Application/App/Extension Management/src/ExtensionMarketplace.Codeunit.al
```diff
@@ -260,39 +262,69 @@ codeunit 2501 "Extension Marketplace"
...
-            end else begin
-                ExtensionPendingSetup.SetRange("User Id", UserSecurityId());
-                ExtensionPendingSetup.DeleteAll();
-            end;
+        // InstallAppsourceExtension is a TryFunction, so it returns false when the installation raised an error.
+        if InstallAppsourceExtension(MarketplaceApplicationID, TelemetryURL) then
+            exit;
+
+        // Capture the error before any other statement resets it.
+        InstallErrorText := GetLastErrorText();
```
thloke, line 274: Are we sure that `GetLastErrorText` will give us the error we want? It gives us AL errors (though we may surface some platform errors too). Was the test-only installation an AL error or something in platform?

src/System Application/App/Extension Management/src/ExtensionMarketplace.Codeunit.al
```diff
@@ -260,39 +262,69 @@ codeunit 2501 "Extension Marketplace"
         ExtensionInstallationImpl: Codeunit "Extension Installation Impl";
         MySessionSettings: SessionSettings;
         AppId: Guid;
+        InstallErrorText: Text;
     begin
         ExtensionInstallationImpl.CheckPermissions();
 
-        if not InstallAppsourceExtension(MarketplaceApplicationID, TelemetryURL) then begin // successful installation returns false
```
thloke, line 266: The previous comment says that successful installation returns false, but now we're saying the inverse? Which is true? It feels like some behaviour is getting switched up here.

thloke (commented):
Have you tested this E2E?

PR #10642, opened 2026-08-26
[29.x] - [Bug 647827] Slice 641282: [Sustainability][Value Chain] FA Reclassification Journal: add CO2e reclassification % field to split emissions like Acq. Cost %

v-rohangarg20:
## Description
Adds a new `Reclassify CO2e %` field to the FA Reclassification Journal so CO2e emissions can be reclassified independently of `Reclassify Acq. Cost %`.

- `SustFAReclassJournalLine.TableExt.al`: new `Reclassify CO2e %` field with validation requiring `Sust. Account No.`/`New Sust. Account No.` to be set.
- `SustFAReclassJournal.PageExt.al`: exposes the new field on the page.
- `SustFAJournalSubscriber.Codeunit.al`: uses `Reclassify CO2e %` instead of `Reclassify Acq. Cost %` to compute the CO2e amount to reclassify, and requires the field to have a value.
- Tests: `TestFAReclassJournalLineWithDifferentCO2ePercent` and `TestFAReclassJournalLineWithZeroCO2ePercent`.

Fixes [AB#647827](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/647827)

## Backport
Backport of #10600 to `releases/29.x`.

Cherry-picked with `git cherry-pick -x`:
- `1a2ebcc8eb2a22e6f88ae26a5ea455d1023f0feb`
- `0d009f45cae6eacac984d888ce817cf35f2fc025`

Both commits applied cleanly with no conflicts or hand edits.

### Note on source PR state
The source PR #10600 was still open at the time of backporting. If it changes before merge, this backport must be refreshed.

PR #10932, opened 2026-09-02
[29.0] - [Bug 648663] Slice 641282: [Sustainability][Value Chain] FA Reclassification Journal: add CO2e reclassification % field to split emissions like Acq. Cost %

v-rohangarg20:
## Description

Adds a new `Reclassify CO2e %` field to the FA Reclassification Journal so CO2e emissions can be reclassified independently of `Reclassify Acq. Cost %`.

- `SustFAReclassJournalLine.TableExt.al`: new `Reclassify CO2e %` field with validation requiring `Sust. Account No.`/`New Sust. Account No.` to be set.
- `SustFAReclassJournal.PageExt.al`: exposes the new field on the page.
- `SustFAJournalSubscriber.Codeunit.al`: uses `Reclassify CO2e %` instead of `Reclassify Acq. Cost %` to compute the CO2e amount to reclassify, and requires the field to have a value.
- Tests: `TestFAReclassJournalLineWithDifferentCO2ePercent` and `TestFAReclassJournalLineWithZeroCO2ePercent`.

Fixes [AB#648663](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/648663)

## Backport

Backport of #10642 to `releases/29.0`.

Cherry-picked with `git cherry-pick -x`:

- `59b66b8fb5e2fca998138ee5020b9249177b2747`
- `0dd46958a2bed895a950e3ca13c566b70591648c`

Both commits applied cleanly with no conflicts or hand edits.

### DIFF
--- a/src/Apps/W1/Sustainability/app/src/FixedAssets/SustFAJournalSubscriber.Codeunit.al
+++ b/src/Apps/W1/Sustainability/app/src/FixedAssets/SustFAJournalSubscriber.Codeunit.al
@@ -116,6 +116,7 @@ codeunit 6283 "Sust. FA Journal Subscriber"
     begin
         FAReclassJournalLine.TestField("Sust. Account No.");
         FAReclassJournalLine.TestField("New Sust. Account No.");
+        FAReclassJournalLine.TestField("Reclassify CO2e %");
         FAReclassJournalLine.CheckSustainabilityAccount(FAReclassJournalLine."Sust. Account No.");
         FAReclassJournalLine.CheckSustainabilityAccount(FAReclassJournalLine."New Sust. Account No.");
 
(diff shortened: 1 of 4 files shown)


ViditGupta1277 (approved):

alexei-dobriansky:

<details>

This PR adds a separate `Reclassify CO2e %` field to the FA Reclassification Journal and uses it when sustainability emissions are moved between fixed assets. The new calculation is placed on the existing FA reclassification transfer events, which are raised after the BaseApp transfer logic has prepared each generated journal line. The split calculation itself is correct when the new field is manually filled, and the new tests cover that path, but the current change does not provide the expected default from `Reclassify Acq. Cost %`.
</details>

<details>

**Fit:** Partial

The reported scenario needs CO2e to use its own percentage while still following the acquisition-cost split by default. The diff covers manually-entered CO2e percentages, but a normal reclassification line with only the existing acquisition-cost percentage still stops with a missing-field error.
</details>

<details>

**S1** (🔴 High): *Default CO2e percentage is missing*
`GetAcquisitionTotalCO2e` now requires `Reclassify CO2e %`, but no product code copies the existing `Reclassify Acq. Cost %` into the new field. Set the new percentage from the acquisition-cost percentage when the user has not entered a value, and update the zero-value test to match that behavior. Without this, existing reclassification flows with sustainability accounts can fail instead of keeping the old split.
</details>

<details>

**Risk:** This touches sustainability fixed-asset reclassification and posting of generated FA and general journal lines. The regression surface is CO2e ledger amounts and validation for existing journal lines, especially 1:M reclassification where cost and emissions can split differently. The BaseApp publisher locations were checked; the event timing is appropriate, but the missing default can block established flows before posting.

**Necessity:** A separate percentage is needed to model emissions independently from acquisition cost, and the change is small enough for a bug fix. The scope is right once default and compatibility behavior is added, because the calculation and page/table surface are otherwise targeted.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10932 round=1 by=alexei-dobriansky at=2026-09-02T13:17:00Z lastSha=874a32d739cbc8dd31c04e28bf48bf7f8be78ad7 reviewKey=b1a79817238d96e66d5028e2b134a9f06fa806f868850b55461a73e4a9872c92 suggestions=S1@1c06753a

djukicmilica (approved):

PR #10937, opened 2026-09-02
[29.0] Hide currency fields in General Ledger Setup

Alexander-Ya:
## What & why

Hides the **Show Currency** and **Currency Symbol Position** fields on the General Ledger Setup page. These fields should not be exposed in the 29.0 release.

## Linked work

Approved issue: [AB#648670](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/648670)

## How I validated this

- Compared the branch with `releases/29.0`; the pull request contains one commit modifying only `ShowCurrencyGenLedgSetup.PageExt.al`.
- No automated test was added because this change only controls page field visibility.
- Local build and Business Central UI validation have not been run yet.

## Risk & compatibility

Low risk. The change only hides two existing setup fields and does not alter stored data, permissions, upgrade behavior, or public APIs.

### DIFF
--- a/src/Layers/W1/BaseApp/Finance/Currency/ShowCurrencyGenLedgSetup.PageExt.al
+++ b/src/Layers/W1/BaseApp/Finance/Currency/ShowCurrencyGenLedgSetup.PageExt.al
@@ -17,7 +17,8 @@ pageextension 60 ShowCurrencyGenLedgSetup extends "General Ledger Setup"
             {
                 ApplicationArea = Basic, Suite;
                 Caption = 'Show Currency';
-                importance = Additional;
+                Importance = Additional;
+                Visible = false;
 
                 trigger OnValidate()
                 begin
@@ -28,7 +29,8 @@ pageextension 60 ShowCurrencyGenLedgSetup extends "General Ledger Setup"
             {
                 ApplicationArea = Basic, Suite;
                 Caption = 'Currency Symbol Position';
-                importance = Additional;
+                Importance = Additional;
+                Visible = false;
 
                 trigger OnValidate()
                 begin


alexei-dobriansky:

<details>

This change hides the **Show Currency** and **Currency Symbol Position** fields on the General Ledger Setup page by setting `Visible = false` on both page fields. The placement is correct for hiding the setup UI in fresh companies, and it does not change table schema, permissions, or public APIs. However, the formatting code still reads the stored `Show Currency` value, so this does not fix companies where the value was already changed before the fields were hidden.
</details>

<details>

**Fit:** Partial

The reported problem is that this currency display setting should not affect amount formatting in this release. The diff hides the setting, but it does not undo values that may already have been set while the fields were visible, so it only prevents new changes.
</details>

<details>

**S1** (🔴 High): *Existing setup values stay active*
Please also reset existing non-default `Show Currency` values, or keep a supported way to set them back to `Never`. Hiding the fields stops new changes, but companies that already changed the value will still get currency symbols in formatted amounts and will no longer have a visible setup field to fix it.
</details>

<details>

**Risk:** The visible-property change itself is narrow and only affects the General Ledger Setup UI. The remaining risk is global because `Amount Auto Format` still uses `General Ledger Setup`.`Show Currency` for formatted amounts, including foreign-currency values used by reports, integrations, and XML export code. There are no public API or schema changes.

**Necessity:** Hiding the fields is needed if this currency display setting must not be exposed in 29.0. The scope is too small unless existing non-default setup data is also handled, because those companies can keep the breaking behavior after the UI is removed.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10937 round=1 by=alexei-dobriansky at=2026-09-02T13:10:29Z lastSha=37a0f6fd13c86559687a4fc5ff6eb1475ce5d9a8 reviewKey=9f3e9290c2e40be8fae114bd0076357a4c63a5ef6cc7d98e80caa0d7adf96240 suggestions=S1@aa76112c

AleksanderGladkov (approved):

mynjj (approved):

PR #10956, opened 2026-09-02
[ExpenseAgent] Refactor agent setup for policies

encimita:
## What & why

Expense Agent admins should be able to disallow user-run policy check

## Linked work

Fixes [AB#648726](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/648726)

## How I validated this

- [ ] I read the full diff and it contains only changes I intended.
- [ ] I built the affected app(s) locally with no new analyzer warnings.
- [ ] I ran the change in Business Central and confirmed it behaves as expected.
- [ ] I added or updated tests for the new behavior, or explained below why none are needed.

**What I tested and the outcome** *(required — be specific: scenarios, commands, screenshots for UI changes)*

PR #10959, opened 2026-09-02
[29.x][ExpenseAgent] Refactor agent setup for policies

encimita:
## What & why

Backports #10956 to `releases/29.x`.

Expense Agent admins should be able to disallow user-run policy checks.

## Linked work

Fixes [AB#648730](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/648730)

## How I validated this

- [ ] I built the affected app locally with no new analyzer warnings.

The backport applies cleanly, changes only the four files from #10956, has no whitespace errors, and has the same stable patch ID as the source commit. A local AL build wasn't run because BCContainerHelper and Docker aren't installed on this machine.

## Risk & compatibility

This is a direct backport of #10956 with no release-specific changes.

### DIFF
--- a/src/Apps/W1/ExpenseAgent/app/src/APIs/ExpenseAgentSetupAPI.Page.al
+++ b/src/Apps/W1/ExpenseAgent/app/src/APIs/ExpenseAgentSetupAPI.Page.al
@@ -42,6 +42,10 @@ page 6942 "Expense Agent Setup API"
                 {
                     Caption = 'Use Rules';
                 }
+                field(submitterRunEvaluation; Rec."Submitter-run Evaluation")
+                {
+                    Caption = 'Allow Submitters to Evaluate Policies';
+                }
                 field(enableAgent; Rec."Enable Agent")
                 {
                     Caption = 'Enable Agent';
--- a/src/Apps/W1/ExpenseAgent/app/src/Setup/Pages/ExpenseAgentSetup.Page.al
+++ b/src/Apps/W1/ExpenseAgent/app/src/Setup/Pages/ExpenseAgentSetup.Page.al
@@ -205,6 +205,9 @@ page 6996 "Expense Agent Setup"
                         end;
                     end;
                 }
+                field("Submitter-run Evaluation"; Rec."Submitter-run Evaluation")
+                {
+                }
                 field("Do Not Allow Expenses Older Than"; Rec."Do Not Allow Exp. Older Than")
                 {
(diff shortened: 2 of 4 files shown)


alexei-dobriansky:

<details>

This change adds an admin setting for submitter-run policy evaluation, exposes it through the setup API and setup pages, and moves the policy setup controls into their own wizard section.

The structure is clear, but the changed backend state is not enough to preserve the intended behavior. The new flag is not checked by the policy evaluation endpoints, and its default value can change existing behavior as soon as a client starts honoring it.
</details>

<details>

**Fit:** Partial

The reported scenario is to let admins stop submitters from running policy checks. The diff adds a stored option and UI/API exposure, but it does not fully enforce the option and may disable existing submitter-run checks by default.
</details>

<details>

**S1** (🔴 High): *Enforce the submitter setting on policy APIs*
The new flag is only exposed through the setup API. The policy-read and policy-result APIs still allow evaluation when the flag is false for any caller that already has the Expense Agent permission. Add a server-side check before returning policies to evaluate and before accepting evaluation results.

**S2** (🔴 High): *Preserve existing submitter-run behavior on upgrade*
A new Boolean starts as false for every existing company. If existing tenants already use submitter-run evaluation, honoring this flag will silently turn it off after upgrade. Set a default or upgrade value that preserves the current behavior, then let admins turn it off.
</details>

<details>

**Risk:** The regression surface is the Expense Agent setup record, setup API, and the policy evaluation APIs used before submission. No BaseApp event publisher or event ordering dependency is introduced, and the app build checks reported success, but the current diff leaves the new control split between UI state and server behavior.

**Necessity:** The option is needed because policy evaluation consumes AI credits and admins need a reliable way to control submitter-run checks. The scope is right for Expense Agent setup, but it needs backend enforcement and upgrade-safe defaults before merge.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10959 round=1 by=alexei-dobriansky at=2026-09-02T19:15:24Z lastSha=216bcf16201153d196caf50137d4e3260c605267 reviewKey=7aae3e81715e475927e832b2c943789eada158b7ee296e7f9a5b54db79f42330 suggestions=S1@53b75730,S2@5722bba0

t-prda (approved):

MonicaPAhuja (approved):

PR #10963, opened 2026-09-02
[Payables Agent] Match TTA candidates to line description

dayland:
## Bug Reference

[ADO work item 648714](https://dev.azure.com/dynamicssmb2/Dynamics%20SMB/_workitems/edit/648714)

## Summary

Restricts Text-to-Account candidates to mappings whose Mapping Text occurs in the currently selected invoice-line description. This prevents the Payables Agent from assigning the first visible, unrelated mapping to multiple lines.

## Root Cause

The E-Document draft action filtered Text-to-Account Mapping only by blank/current Vendor No. The agent therefore saw unrelated records before searching and could reuse the first row across different invoice lines.

## Changes Made

- Added `E-Doc. TTA Candidate Filter` to select case-insensitive description matches within the global/current-vendor scope.
- Updated `OpenTextToAccountMappings` to expose only the selected line's candidates, or an empty page when none match.
- Strengthened the Payables Agent instructions to validate Mapping Text and never reuse candidates between lines.

## Implementation Process

- Fix iterations: 1 of 5
- Compilation: E-Document Core and Payables Agent compile successfully
- Tests: 5/5 passing

### Pre-Fix

- `UnrelatedMappingIsNotACandidate`: failed, returning 2 candidates instead of 1.
- `CandidatesAreIndependentPerLineDescription`: failed, returning both mappings for each line.
- Three vendor/global-scope tests passed.

### Post-Fix

- `UnrelatedMappingIsNotACandidate`: passed
- `MatchingCurrentVendorMappingIsCandidate`: passed
- `MatchingGlobalMappingIsCandidate`: passed
- `OtherVendorMappingIsNotACandidate`: passed
- `CandidatesAreIndependentPerLineDescription`: passed

The E-Document Core, Payables Agent, and Payables Agent Tests apps were rebuilt and published before the final test run.

## Review Notes

Review the case-insensitive substring criterion and the temporary candidate-to-Line-No. page filter.

### DIFF
--- a/src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/EDocPurchaseDraftSubform.Page.al
+++ b/src/Apps/W1/EDocument/App/src/Processing/Import/Purchase/EDocPurchaseDraftSubform.Page.al
@@ -345,7 +345,7 @@ page 6183 "E-Doc. Purchase Draft Subform"
                     {
                         ApplicationArea = All;
                         Caption = 'Text-to-Account Mappings';
-                        ToolTip = 'Opens the Text-to-Account Mapping filtered for the current vendor.';
+                        ToolTip = 'Opens Text-to-Account Mappings showing only candidates whose Mapping Text is contained in the selected line description.';
                         Image = MapAccounts;
                         Scope = Repeater;
                         Visible = AgentDrivenLineMatchingEnabled;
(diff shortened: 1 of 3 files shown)


dayland:
Abandoning this approach: it replaces agent-led matching behavior with AL-enforced candidate filtering, which is contrary to the intended design.

PR #10967, opened 2026-09-02
[Bug 646558] Stabilize remittance advice test initialization

djukicmilica:
## Why

The E-Document remittance advice tests can still fail across localization runs because shared scenario initialization may change VAT reporting date usage after it was disabled. A failed test can also roll back fixture records while the codeunit-level `IsInitialized` flag remains set, leaving later tests with inconsistent setup.

## Summary

- **Moved** VAT reporting date normalization after shared E-Document scenario initialization so localization setup cannot overwrite it.
- **Committed** initialized fixture records before setting `IsInitialized` to keep database state aligned with the codeunit flag after test rollback.

Fixes [AB#646558](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/646558)

### DIFF
--- a/src/Apps/W1/EDocument/Test/src/Processing/EDocRemitAdviceTest.Codeunit.al
+++ b/src/Apps/W1/EDocument/Test/src/Processing/EDocRemitAdviceTest.Codeunit.al
@@ -614,12 +614,6 @@ codeunit 139547 "E-Doc. Remit. Advice Test"
         if BindSubscription(LibraryJobQueue) then;
         LibraryJobQueue.SetDoNotHandleCodeunitJobQueueEnqueueEvent(true);
 
-        GeneralLedgerSetup.Get();
-        if GeneralLedgerSetup."VAT Reporting Date Usage" <> GeneralLedgerSetup."VAT Reporting Date Usage"::Disabled then begin
-            GeneralLedgerSetup."VAT Reporting Date Usage" := GeneralLedgerSetup."VAT Reporting Date Usage"::Disabled;
-            GeneralLedgerSetup.Modify(false);
-        end;
-
         EDocumentService.DeleteAll();
 
         LibraryEDoc.SetupStandardVAT();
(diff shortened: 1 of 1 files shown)


alexei-dobriansky:

# **Recommendation: Accept**

<details>

This PR stabilizes the remittance advice integration tests by moving `VAT Reporting Date Usage` normalization after the shared E-Document setup and by committing the initialized fixture data before the codeunit marks itself initialized. This matches the failure pattern: localization setup can change VAT reporting date behavior, and a later rollback can remove fixture data while the codeunit global flag still skips full setup. The new order disables VAT reporting date usage after the setup that may overwrite it, and the commit before `IsInitialized := true` keeps the database setup aligned with that flag for later tests.
</details>

<details>

**Fit:** Strong

The bug describes cross-localization remittance advice test failures caused by unstable shared initialization and rolled-back setup. The diff targets that initialization path directly and does not change production behavior.
</details>

<details>

No suggestions.
</details>

<details>

**Risk:** The change is limited to one test codeunit. The main regression surface is shared fixture data for the remittance advice tests, especially the payment journal batch, E-Document service setup, and general ledger setup. There is no product runtime, public API, or BaseApp publisher/event timing impact; the added `Commit()` is a test-setup transaction boundary, and this pattern is used elsewhere when setup must survive later test transaction boundaries.

**Necessity:** The change is needed because later tests can trust `IsInitialized` only if the fixture records and GL setup survive the same rollback conditions as the flag. Without this, localization runs can keep failing with missing setup or VAT period errors. The scope is narrow: it reorders existing setup and commits it before setting the flag.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10967 round=1 by=alexei-dobriansky at=2026-09-02T22:13:13Z lastSha=427c7ed9f6ff872fd0ddd5c632da28433b392597 reviewKey=e48fe3f8f27d5f57477ad2db32fc75d469412f68e45e3ee14985533e3c5ae476 suggestions=none

dcenic (approved):

gggdttt (approved):

PR #10971, opened 2026-09-02
[29.x] bug 648733 - Add integration event for updating original document VAT date in General Ledger Setup CZL

DavidHolusa:
## What & why

The UpdateOriginalDocumentVATDateCZL procedure in the "General Ledger Setup" table currently does not allow extension with custom logic. This procedure is called from handlers in Codeunit 11744 when source fields change (posting date, document date, VAT date) and sets the value of the "Original Doc. VAT Date CZL" field in Purchase Header based on the "Default Orig.Doc. VAT Date CZL" setting. The call chain handler → UpdateOriginalDocumentVATDateCZL → validation provides no extensibility point to modify the behavior.

The integration event for updating original document VAT date in General Ledger Setup CZL has been added

## Linked work

Fixes [AB#648733](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/648733)

### DIFF
--- a/src/Apps/CZ/CoreLocalizationPack/app/Src/TableExtensions/GeneralLedgerSetupCZL.TableExt.al
+++ b/src/Apps/CZ/CoreLocalizationPack/app/Src/TableExtensions/GeneralLedgerSetupCZL.TableExt.al
@@ -180,6 +180,7 @@ tableextension 11713 "General Ledger Setup CZL" extends "General Ledger Setup"
     begin
         if ("Def. Orig. Doc. VAT Date CZL" = DefaultOrigDocVATDate) then
             OriginalDocumentVATDate := NewDate;
+        OnAfterUpdateOriginalDocumentVATDateCZL(NewDate, DefaultOrigDocVATDate, OriginalDocumentVATDate);
     end;
 
     procedure GetOriginalDocumentVATDateCZL(PostingDate: Date; VATDate: Date; DocumentDate: Date): Date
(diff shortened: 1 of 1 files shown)


alexei-dobriansky:

<details>

This change adds an integration event in GeneralLedgerSetupCZL.TableExt.al so extensions can change the original document VAT date calculated by UpdateOriginalDocumentVATDateCZL. The publisher is raised after the default assignment and before each caller validates the field. That placement lets subscribers change the final date, but it does not let them skip the standard assignment or see the previous value, so it does not fully provide the requested override hook.
</details>

<details>

**Fit:** Partial

The bug says this procedure needs an extension point for custom logic when purchase date fields update the original document VAT date. The diff adds a publisher on the right procedure, but the new event is an after-event without a handled flag, so it only partly matches the requested override behavior.
</details>

<details>

**S1** (🔴 High): *Add the handled override hook*
The new event runs after OriginalDocumentVATDate may already be overwritten and it has no IsHandled parameter. Add a handled event before the standard assignment, and exit when it is handled, so subscribers can keep or replace the date without losing the old value.
</details>

<details>

**Risk:** The changed path feeds purchase header and purchase advance letter validations before they validate the original document VAT date. Standard behavior is mostly unchanged when there are no subscribers, but the new integration event becomes a public contract, so wrong timing would be hard to correct later.

**Necessity:** The change is justified because the calculation is centralized and called from several purchase date changes. The scope is small and correct for an extensibility request, but the event must expose the requested override contract before it is merged.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10971 round=1 by=alexei-dobriansky at=2026-09-03T01:14:18Z lastSha=86d7eb8563f8e526f4740bdb504d1599ce1a3d88 reviewKey=f83e274a9c2dae135b5aca15c44447a406d700a64572f2b0af3430be9cbdefbe suggestions=S1@f0468e31

v-janpopr (approved):

AndersLarsenMicrosoft (approved):

PR #10972, opened 2026-09-02
[29.0] bug 648733 - Add integration event for updating original document VAT date in General Ledger Setup CZL

DavidHolusa:
## What & why

The UpdateOriginalDocumentVATDateCZL procedure in the "General Ledger Setup" table currently does not allow extension with custom logic. This procedure is called from handlers in Codeunit 11744 when source fields change (posting date, document date, VAT date) and sets the value of the "Original Doc. VAT Date CZL" field in Purchase Header based on the "Default Orig.Doc. VAT Date CZL" setting. The call chain handler → UpdateOriginalDocumentVATDateCZL → validation provides no extensibility point to modify the behavior.

The integration event for updating original document VAT date in General Ledger Setup CZL has been added

## Linked work

Fixes [AB#648733](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/648733)

### DIFF
--- a/src/Apps/CZ/CoreLocalizationPack/app/Src/TableExtensions/GeneralLedgerSetupCZL.TableExt.al
+++ b/src/Apps/CZ/CoreLocalizationPack/app/Src/TableExtensions/GeneralLedgerSetupCZL.TableExt.al
@@ -180,6 +180,7 @@ tableextension 11713 "General Ledger Setup CZL" extends "General Ledger Setup"
     begin
         if ("Def. Orig. Doc. VAT Date CZL" = DefaultOrigDocVATDate) then
             OriginalDocumentVATDate := NewDate;
+        OnAfterUpdateOriginalDocumentVATDateCZL(NewDate, DefaultOrigDocVATDate, OriginalDocumentVATDate);
     end;
 
     procedure GetOriginalDocumentVATDateCZL(PostingDate: Date; VATDate: Date; DocumentDate: Date): Date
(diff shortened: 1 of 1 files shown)


alexei-dobriansky:

<details>

The change adds an integration event in GeneralLedgerSetupCZL.TableExt.al after the default logic that updates the original document VAT date. The event runs before the callers validate the field, so subscribers can still change the final date before validation.

The placement is not enough for a true override point. If the default setting matches, the old value is replaced before the event runs, so a subscriber cannot choose to keep the previous value.
</details>

<details>

**Fit:** Partial

The requested behavior is an override point for UpdateOriginalDocumentVATDateCZL. The diff adds an after-event that can adjust the resulting date, but it does not let subscribers skip the default assignment when that assignment already changed the value.
</details>

<details>

**S1** (🔴 High): *Add a handled override point before assignment*
This event is raised after the default assignment. If the default setting matches, the old OriginalDocumentVATDate value is already lost, so a subscriber cannot choose to leave it unchanged. Add a before event with a ar IsHandled check before the assignment, or move this event before the assignment and honor IsHandled.
</details>

<details>

**Risk:** The standard behavior is unchanged, but the new event becomes a public extension contract. If it ships in this shape, subscribers that need to suppress the default assignment still cannot implement the scenario, and changing the event shape later would be harder.

**Necessity:** A customization point here is needed because several purchase date flows call this shared update procedure before validation. The scope is small and appropriate, but the event contract needs to support the override use case before merge.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=10972 round=1 by=alexei-dobriansky at=2026-09-03T01:15:18.7784271Z lastSha=8f2b6fdac4355a6dfab1dd05ec4503b15ee266c2 reviewKey=4aecb2d1296f70a7516160d736c5d8eb646e815f054a65ded38993122429cd09 suggestions=S1@f0468e31

v-janpopr (approved):

AndersLarsenMicrosoft (approved):

PR #11026, opened 2026-09-03
Service Line CalcVATAmountLines: count VAT Difference once and clear filters before OnAfterCalcVATAmountLines

### LINKED ISSUE #9102: [Bug]: Service Line CalcVATAmountLines double-counts VAT Difference and leaks filters onto OnAfterCalcVATAmountLines
### Describe the issue

`CalcVATAmountLines` in table 5902 "Service Line" diverges from the Sales Line / Purchase Line implementation, causing two defects visible on Service Statistics when invoice rounding and VAT difference are both enabled:

1. VAT Difference is double-counted when accumulating the rounding residual:

```al
TotalVATAmount += ServiceLine."Amount Including VAT" - ServiceLine.Amount + ServiceLine."VAT Difference";
```

Sales Line accumulates `SalesLine."Amount Including VAT" - SalesLine.Amount` only. Since `Amount Including VAT - Amount` already contains the VAT Difference, the Service version adds it twice, and the excess is dumped onto the 0% invoice-rounding VAT line, so the VAT specification shows an incorrect VAT amount on the rounding line.

2. The filters set by `FindVATAmountLine` in the invoice-rounding block are never reset, so `OnAfterCalcVATAmountLines` fires with a filtered `VATAmountLine`, and the VAT Specification subform on page 6030 "Service Statistics" can show a single VAT line instead of all lines. Sales Line clears these filters before the event.

Verified still present in current main in BOTH `src/Layers/W1/BaseApp/Service/Document/ServiceLine.Table.al` and the NA layer's full table copy `src/Layers/NA/BaseApp/Service/Document/ServiceLine.Table.al` (the only country layer that overrides table 5902).

### Expected behavior

Service document VAT statistics match the Sales/Purchase behavior: no double-counted VAT Difference on the invoice-rounding VAT line, and all VAT specification lines visible after recalculation.

### Steps to reproduce

1. Enable invoice rounding (Sales & Receivables Setup) and Allow VAT Difference with a Max. VAT Difference Allowed.
2. Create a Service Invoice whose amounts require an invoice rounding line.
3. Open Service Statistics and manually change the VAT amount (creating a VAT Difference) on a VAT line.
4. The rounding VAT line receives the VAT Difference amount again (incorrect VAT amount), and the VAT specification can collapse to a single line.

### Additional context

This is a copy of microsoft/BusinessCentralApps#1562 and microsoft/BusinessCentralApps#1563 (both labeled Approved/Finance in the contribution pilot). Both were being fixed together in microsoft/BusinessCentralApps#1913, which received a code-owner approval, but the internal merge attempt reported test failures and the PR was closed unmerged when the repository was retired. The likely cause of those failures has been identified: the fix patched the W1 layer only, while the NA layer's full copy of table 5902 kept the defects, so localization runs of the new W1 tests failed. The re-submitted fix will cover both layers.

Franco111000:
## What & why

`CalcVATAmountLines` on table 5902 "Service Line" diverges from the Sales Line and Purchase Line implementation in two places, and both show up on Service Statistics as soon as invoice rounding and a manual VAT difference meet:

1. The rounding residual accumulated `"Amount Including VAT" - Amount + "VAT Difference"` per line. `"Amount Including VAT" - Amount` already contains the VAT difference, so it was counted twice and the excess was settled on the invoice rounding VAT line, which then showed a VAT amount equal to the VAT difference. Sales Line and Purchase Line accumulate `"Amount Including VAT" - Amount` only.
2. When the rounding block ran, the filters set by `FindVATAmountLine` stayed on the `var VATAmountLine` when `OnAfterCalcVATAmountLines` fired and control returned to the caller. Page 6030 "Service Statistics" then ran `TempVATAmountLine.ModifyAll(Modified, false)` on one row and handed a filtered set to the VAT Specification subform, so only one VAT line was visible.

The fix drops the extra term, which makes the accumulation identical to Sales Line and Purchase Line, and adds `VATAmountLine.Reset()` after the rounding block, before the event, so the record is unfiltered on every path. For the record, Sales Line is not fully clean here either: `GetVATAmountLineOfMaxAmt` resets its filters only on its fall-through branch and leaves them set on the early `exit`. That narrower gap is out of scope for this change. Both layers that carry the procedure are changed: `src/Layers/W1` and the full table copy in `src/Layers/NA`. The two edited regions are byte-identical between the layers before and after the change.

## Linked work

Fixes #9102

Re-submission of microsoft/BusinessCentralApps#1913 (pilot issues #1562 and #1563), which received a code-owner approval and was closed unmerged when that repository was retired, with a note about internal test failures. Two things differ from that version and are the likely cause of those failures:

- The NA layer copy of table 5902 was not patched there. It is patched here.
- The pilot test helper validated `"VAT Identifier"` on a new VAT Posting Setup directly, which fails in the IT localization, where that field has a table relation to the VAT Identifier table. The tests here go through the codeunit's existing `CreateVATPostingSetup` helper, which the IT copy already overrides for exactly that reason.

## How I validated this

Two tests were added to codeunit 134028 "ERM Change VAT On VAT Amt Line", and mirrored into its full-replacement copies in the APAC, ES and IT test layers (those copies differ from W1 only in their pre-existing localized lines, verified by diffing each copy against W1 before and after the change).

- `ServiceCalcVATAmountLinesCountsVATDifferenceOnce`: a Service Invoice with an item line (10% VAT) and an invoice rounding G/L line (0% VAT, own VAT Identifier). The VAT amount of the item line is raised by a VAT difference the way the VAT Specification subform and Service Statistics do it (`Validate("VAT Amount")`, `Modified := true`, `UpdateVATOnLines` for General and Invoicing), then `CalcVATAmountLines` runs again. It asserts that the total VAT on the VAT Amount Lines equals the VAT on the Service Lines, and that the rounding VAT line stays at 0. Against the unpatched table the residual after `UpdateLines` is exactly the VAT difference, the rounding block adds it to the rounding VAT line, and both assertions fail.
- `ServiceCalcVATAmountLinesReturnsAllLinesAfterRoundingResidual`: same document, no VAT difference. A rounding residual is injected through `OnCalcVATAmountLinesOnBeforeUpdateVATAmountLine` from a manually bound instance of the test codeunit, so the rounding block runs without depending on the rounding arithmetic of a particular localization. It asserts that both VAT Amount Lines are visible on the record handed back to the caller, and that the residual was settled on a VAT Amount Line. With the first fix alone the block still leaves the `FindVATAmountLine` filters in place and `Count()` returns 1, so this test isolates the `Reset()`.

Neither test has executed yet: `.alpackages` is empty in this repository, so there is no local build, and the pull request build runs only after a maintainer approves the workflow. The pass and fail behaviour above comes from tracing the code paths by hand against `upstream/main`, not from a test run.

## Risk & compatibility

- Behaviour change for subscribers of `OnCalcVATAmountLinesOnAfterCalcLineTotals` on Service Line: `TotalVATAmount` no longer carries the duplicated VAT difference. That value was wrong; it now matches what the Sales and Purchase publishers pass.
- Behaviour change for subscribers of `OnAfterCalcVATAmountLines` on Service Line: `VATAmountLine` now arrives unfiltered on every path. Nothing in this repository subscribes to that event.
- No signature changes, no data or upgrade impact, no permissions, no feature flags.
- Out of scope, kept as is: Service Line settles the rounding residual on the VAT line of the last document line, where Sales Line uses `GetVATAmountLineOfMaxAmt`.

### DIFF
--- a/src/Layers/APAC/Tests/VAT/ERMChangeVATOnVATAmtLine.Codeunit.al
+++ b/src/Layers/APAC/Tests/VAT/ERMChangeVATOnVATAmtLine.Codeunit.al
@@ -2,6 +2,7 @@ codeunit 134028 "ERM Change VAT On VAT Amt Line"
 {
     Subtype = Test;
     TestPermissions = Disabled;
+    EventSubscriberInstance = Manual;
 
     trigger OnRun()
     begin
@@ -22,6 +23,7 @@ codeunit 134028 "ERM Change VAT On VAT Amt Line"
         LibraryUtility: Codeunit "Library - Utility";
         LibraryInventory: Codeunit "Library - Inventory";
         IsInitialized: Boolean;
+        RoundingResidual: Decimal;
         ErrorWithCurrency: Label '%1 for %2 must not exceed %3 = %4.', Comment = '%1=Field Caption;%2=Field Value;%3=Field Caption;%4=Field Value;';
         ErrorWithoutCurrency: Label '%1 must not exceed %2 = %3.', Comment = '%1=Field Caption;%2=Field Caption;%3=Field Value;';
         DefaultVATAmountLineTxt: Label 'VAT Amount';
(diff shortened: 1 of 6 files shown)


alexei-dobriansky:

# **Recommendation: Accept**

<details>

This change aligns Service Line VAT amount-line calculation with the Sales and Purchase implementations. It removes the extra `VAT Difference` addition from the service rounding total and resets the temporary VAT amount-line record before the after-event sees it. The tests cover both the double-counting case and the leaked-filter case, and they are copied to the affected localized test layers.
</details>

<details>

**Fit:** Strong

The reported problem is that service VAT statistics can show the VAT difference twice and can return a filtered VAT specification after invoice rounding. The diff changes exactly those two service calculation paths and includes tests for both visible failures.
</details>

<details>

No suggestions.
</details>

<details>

**Risk:** The regression surface is Service document VAT statistics when invoice rounding and VAT differences are involved, plus subscribers that read the totals or the VAT amount-line record from the service calculation events. The production change is small and makes Service behavior match the existing Sales and Purchase pattern.

**Necessity:** The change is required to stop an incorrect VAT amount from being assigned to the rounding VAT line and to make all VAT specification lines visible to callers. The scope is limited to the affected W1 and NA Service Line copies, with localized tests mirroring the shared behavior.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11026 round=1 by=alexei-dobriansky at=2026-09-04T09:37:19.3896003Z lastSha=88aafbc6f53bb295783dc6be827989daa307f08c reviewKey=ff624148510217befc0c0a1eb61f471769ff093ede9b7708750168a597d895a6 suggestions=

alexei-dobriansky (approved):

Issue #9977, opened 2026-08-05
[Bug]: [Quality Management] Warehouse Receipt Trigger "When Warehouse Receipt is created" never fires for receipt lines from inbound transfers or sales returns


Franco111000:
### Describe the issue

Inspection Generation Rules offer "Warehouse Receipt Trigger" = "When Warehouse Receipt is created" (`Qlty. Whse. Receipt Trigger::OnWarehouseReceiptCreate`). Neither the field tooltip (`QltyInspectionGenRule.Table.al`, field 22) nor the enum caption limits it to a source document type, and the rule page lets you configure it for any applicable rule. In practice it only works for warehouse receipt lines created from purchase orders.

The only subscriber for this trigger is `HandleOnAfterCreateRcptLineFromPurchLine` in codeunit 20411 "Qlty. Receiving Integration" (`src/Apps/W1/Quality Management/app/src/Integration/Receiving/QltyReceivingIntegration.Codeunit.al:116`), which subscribes to codeunit "Purchases Warehouse Mgt.", event `OnAfterCreateRcptLineFromPurchLine`, and calls the local helper `AttemptCreateInspectionWithReceiptLine` (`:280`).

Warehouse receipts are also created from inbound transfer orders and from sales return orders, and the Base Application publishes structurally identical events at those creation points:

- codeunit 5993 "Transfer Warehouse Mgt.", `OnAfterCreateRcptLineFromTransLine(var WarehouseReceiptLine; WarehouseReceiptHeader; TransferLine)`, raised at the end of `TransLine2ReceiptLine` (`src/Layers/W1/BaseApp/Inventory/Transfer/TransferWarehouseMgt.Codeunit.al:238`)
- codeunit "Sales Warehouse Mgt.", `OnAfterCreateRcptLineFromSalesLine(var WarehouseReceiptLine; WarehouseReceiptHeader; SalesLine)` (`src/Layers/W1/BaseApp/Sales/Document/SalesWarehouseMgt.Codeunit.al:419`)

Quality Management references neither event anywhere in its source.

The omission looks unintentional rather than by design, because the surrounding code is already prepared for these sources:

- The helper `AttemptCreateInspectionWithReceiptLine` is source agnostic. Its integration event `OnBeforeAttemptCreateInspectionWithReceiptLine` documents the optional source line as "purchase line, sales line, transfer line" (`QltyReceivingIntegration.Codeunit.al:509`).
[...]

PR #11028, opened 2026-09-03
[Quality Management] Add public rule engine facade for automatic inspection creation

### LINKED ISSUE #9978: [Extensibility Request][Quality Management] No supported way to create inspections through the rule engine; all creation procedures are internal
### Describe the issue

There is no supported way for a partner extension to ask Quality Management to "create an inspection for this record through the rule engine" with automatic-creation semantics. Every procedure involved in the app's own automatic creation flow is `internal`, and the manifest grants `internalsVisibleTo` only to Microsoft's Quality Management Test Library (`src/Apps/W1/Quality Management/app/app.json:33`).

Compiling a minimal extension against the shipped 28.3 symbols that reproduces the app's own receiving flow yields nine AL0161 protection-level errors, one per call. On current main the procedures are:

- `"Qlty. Inspection - Create"`: `CreateInspectionWithMultiVariants` (`QltyInspectionCreate.Codeunit.al:154`), `GetCreatedInspection` (`:781`), `IsLastInspectionNewlyCreated` (`:807`). In fact every creation procedure in this codeunit is `internal`.
- `"Qlty. Warehouse Integration"`: `CollectSourceItemTracking` (`QltyWarehouseIntegration.Codeunit.al:131`)
- `"Qlty. Batch Notif. Helper"`: `BeginBatch` (`:15`), `EndBatch` (`:21`), `TrackCreatedInspection` (`:36`), `ConfigureForBatch` (`:51`)
- `"Qlty. Management Setup"`: `GetSetupRecord` (`QltyManagementSetup.Table.al:527`)

The two accessible entry points do not substitute for automatic generation, because both hard-code `IsManualCreation = true`:

- API page `QltyCreateInspectionAPI.Page.al:94` and `:131` call `CreateInspection(AnyInputRecord, true)`
- Report `QltyCreateInspection.Report.al:326` calls `CreateInspectionWithMultiVariantsAndTemplate(..., true, ...)`

Manual creation applies the manual activation filter, ignores the per-source trigger fields, can open the inspection card, and raises a hard error when no rule matches. None of that is acceptable inside an automatic flow such as an event subscriber on a posting or document-creation routine.

The published integration events do not help either: each one is raised only from inside a path the app has already entered (purchase receipt, warehouse receipt from purchase, transfer posting, sales return posting). They allow modifying or replacing behavior at the moments the product ships, but cannot add a trigger moment the product does not ship.

Concrete consequence: to create inspections at warehouse receipt creation for transfer-sourced receipt lines (#9977), an extension currently has to re-implement rule matching, source field mapping, duplicate detection, and template expansion outside the app. That re-implementation drifts from the product on every update. This request stays relevant even if #9977 is fixed, because any next missing trigger moment hits the same wall.

### Expected behavior

Either a public, supported entry point on `"Qlty. Inspection - Create"` (or a facade codeunit) that creates an inspection for a given record through the rule engine with non-manual semantics and returns the created inspection, or widened access (`public`) on the procedures listed above.

### Steps to reproduce

1. Create an AL extension depending on Quality Management.
2. In a codeunit, declare `QltyInspectionCreate: Codeunit "Qlty. Inspection - Create"` and call `CreateInspectionWithMultiVariants(...)`.
3. Compilation fails with AL0161: cannot access internal procedure. The same happens for the other procedures listed above.

### Additional context

The pattern of the ask matches how other first-party apps expose their engines for extension scenarios. A single public procedure with the semantics "find matching generation rules for this record and create or reuse an inspection, non-manual, no UI" would cover the need; the batch notification helper and tracking collection only need to be public if the entry point does not handle them internally. A sketch of what would be sufficient, on `"Qlty. Inspection - Create"` or a small public facade codeunit:

```al
/// <summary>
/// Creates or reuses a quality inspection for the given record by matching the active
/// inspection generation rules, using automatic-creation semantics: no UI, the manual
/// activation filter is not applied, and no error is raised when no rule matches.
/// </summary>
/// <param name="RecordVariant">The record to create the inspection for.</param>
/// <param name="OptionalRelatedRecordVariant">Optional related record (for example a source document line) used for rule matching and source field mapping.</param>
/// <param name="QltyInspectionHeader">The created or reused inspection.</param>
/// <param name="IsNewlyCreated">True when the inspection was newly created rather than reused.</param>
/// <returns>True when an inspection was created or reused.</returns>
procedure CreateInspectionFromRuleEngine(RecordVariant: Variant; OptionalRelatedRecordVariant: Variant; var QltyInspectionHeader: Record "Qlty. Inspection Header"; var IsNewlyCreated: Boolean): Boolean
```

Item tracking handling could either be internal to the entry point (collect from the source and create one inspection per tracking line, matching the app's own receiving flow) or covered by an overload accepting a `Tracking Specification` buffer. Alternatively, simply widening access on the nine procedures listed above unblocks the scenario with no new surface, at the cost of exposing more implementation detail.

Franco111000:
## What & why

Quality Management has no supported way for an extension to create inspections through the generation rule engine with automatic semantics: every creation procedure on codeunit 20404 "Qlty. Inspection - Create" is `internal`, and the two accessible entry points (API page `Qlty. Create Inspection API` and report `Qlty. Create Inspection`) hard-code `IsManualCreation = true`, which applies the manual-only rule filter, can open the inspection card and raises an error when no rule matches. An extension that needs a trigger moment the app does not ship (for example warehouse receipt creation for transfer-sourced lines, #9977) currently has to re-implement rule matching, source traversal, field mapping, duplicate detection and template expansion outside the app.

This PR adds the narrow public facade recommended in the issue discussion, on the same codeunit, and keeps every implementation helper internal:

- `CreateInspectionFromRuleEngine(PrimaryRecordVariant, OptionalRelated2Variant, OptionalRelated3Variant, OptionalRelated4Variant, var TempFiltersQltyInspectionGenRule, var QltyInspectionHeader, var IsNewlyCreated): Boolean`
  Creates or reuses one inspection through `CreateInspectionWithMultiVariants` with `IsManualCreation = false`, suppresses the inspection-created notification for the duration of the call (restoring the instance flag afterwards), and returns the resolved header plus whether it was newly inserted. No page is opened and no error is raised when no rule matches. Up to four records are accepted because the app's own automatic paths pass up to four (receiving: receipt line, source line, receipt header, tracking line; manufacturing: ledger entry, routing line, journal line, prod. order line).
- `CreateInspectionsFromRuleEngine(PrimaryRecordVariant, OptionalRelated2Variant, OptionalRelated3Variant, var TempTrackingSpecification, var TempFiltersQltyInspectionGenRule, var NewlyCreatedQltyInspectionIds, var AllResolvedQltyInspectionIds): Boolean`
  Tracked-source overload: one creation call per tracking line within the buffer's current filters, with the tracking line in the fourth slot, and a single untracked call when the buffer is empty. This is the shape `AttemptCreateInspectionWithReceiptLine` in "Qlty. Receiving Integration" uses. Because the caller supplies the tracking lines, it serves sources the app's own collector (`CollectSourceItemTracking`, limited to purchase, transfer and sales lines) cannot. Several tracking lines can resolve to the same inspection depending on setup, so the result is returned as deduplicated newly-created and all-resolved lists, the same idiom as `CreateMultipleInspectionsWithoutDisplaying`; the existing `TrackResolvedInspection` now shares the dedup helper.

Two points where this deliberately differs from the sketches in the issue thread, both explained in the doc comments:

1. **The generation rule filter record is part of the contract.** Every automatic path in the app filters rules on its own trigger field before calling the engine (`HasWarehouseReceiptCreateGenRule` and siblings), and the engine applies those filters to the rule search, including through traversal. Without the parameter, a caller adding a "warehouse receipt created" moment would also fire rules configured for "warehouse receipt posted", since the trigger enums default to Never and the search otherwise matches by sort order alone. Passing an unfiltered temporary record considers every automatic rule, which is what the sketches did, so nothing is lost.
2. **The result lists are newly-created plus all-resolved rather than created plus reused**, because with the latter an inspection created by the first tracking line and reused by the second would land in both lists. The codeunit already uses the newly-created plus all-resolved shape internally, and the parameter names follow its `...Ids` naming.

## Linked work

Fixes #9978

## How I validated this

- Ten tests added to codeunit 139959 "Qlty. Tests - Create Inspect.", covering the six regression scenarios listed in the issue discussion plus the tracked overload and the rule filter: a matching rule creates and returns a new inspection; configured reuse returns the existing inspection with `IsNewlyCreated = false`; no matching rule returns false without an error; a manual-only rule is not selected; a related record is promoted to primary and its tracking line is mapped (tracking specification first, purchase line second, lot asserted on the inspection); rule filters select the second rule in sort order while the unfiltered call selects the first; a two-serial tracking buffer produces two inspections returned as newly created; an empty buffer falls back to one untracked creation; a reused inspection appears in the all-resolved list but not in the newly-created list; and, the case the buffer overload exists for, a source the internal collector does not handle (an item journal line) gets its item and lot from a caller-built buffer.
- The tests call the two public procedures directly from the test app rather than through the test library, so they also prove the procedures are reachable from outside the `internalsVisibleTo` boundary.
- I could not build or run the tests locally (no container in this environment), so I rely on the PR build for compilation, analyzers and test execution. Each test's arrangement was traced by hand through the test library (`CreatePrioritizedRule` disables all other rules and sorts the new one first; `Qlty. Pur. Order Generator` creates one reservation entry per serial; `EnsureSetupExists` commits, so it runs before any item is created) and through the engine.
- Notification suppression is not covered by a test: `NotifyInspectionCreated` exits on `not GuiAllowed()`, and the existing automatic-creation tests carry no notification handler, so a headless run cannot observe it. The facade uses the same flag the batch helper's `ConfigureForBatch` sets for the app's own automatic paths.
- No permission-lowered test was added; the facade runs under the same codeunit `Permissions` and the same automatic semantics as the existing subscribers on this codeunit, which are not permission-tested either. Happy to add one if you want it.

## Risk & compatibility

- Purely additive: two new public procedures and one local helper on codeunit 20404; no existing signature, event, enum or table changes. No new object, no permission set change, no translation impact. The new signatures become part of the supported surface from the next release on.
- Permissions: the facade runs with automatic semantics, so it does not call `VerifyCanCreateManualInspection`, exactly like the app's own event subscribers on the same codeunit. The codeunit's `Permissions` property covers only the Quality Management tables, so reading the caller's source record still runs under the caller's own permissions. The engine opens a temporary `RecordRef` on the table of the record the caller passed in; since the caller must already hold that record (or a `RecordId` it can read), no access is gained through the public entry point.
- Extensibility seam: `OnBeforeCreateInspection` with `IsHandled` already fires inside the path, so no new event is added around the facade; a handled creation returns false.
- The facade sets and restores `PreventShowingGeneratedInspectionEvenIfConfigured` on its own instance only; callers that share a codeunit instance with other code see no state change after the call.
- `Message(MultiRecordInspectionSourceFieldErr)` inside the engine (raised when several records are supplied and none maps a source field) and the workflow events raised on inspection creation are unchanged and apply to the facade like to every automatic path; the doc comment therefore promises only what the facade controls: no inspection page, no inspection-created notification, no error on no-match.
- Follow-up: #9977 (receipt-creation trigger for transfer and sales return lines) is the first consumer of this shape inside the app and stays a separate change.

### DIFF
--- a/src/Apps/W1/Quality Management/app/src/Document/QltyInspectionCreate.Codeunit.al
+++ b/src/Apps/W1/Quality Management/app/src/Document/QltyInspectionCreate.Codeunit.al
@@ -246,6 +246,112 @@ codeunit 20404 "Qlty. Inspection - Create"
     end;
 
     /// <summary>
+    /// Creates or reuses an inspection for the supplied records through the generation rule engine, with the same
+    /// automatic semantics as the app's own triggered creation paths. Intended for extensions that need to create
+    /// inspections at a moment the app does not trigger itself, for example from an event subscriber on a document
+    /// or posting routine.
+    ///
+    /// Automatic semantics:
+    /// - Only generation rules with an activation trigger of "Manual or Automatic" or "Automatic only" are considered.
+    /// - The filters on TempFiltersQltyInspectionGenRule are applied to the rule search, in the same way the app's own
+    ///   triggers filter on their trigger field (for example "Warehouse Receipt Trigger"). Pass an unfiltered temporary
+    ///   record to consider every automatic rule for the source table.
+    /// - No inspection page is opened and no inspection-created notification is raised. The caller receives the
+    ///   resolved inspection and decides whether to surface it. Workflow and integration events raised by inspection
+    ///   creation fire as for any automatic creation.
+    /// - No error is raised when no generation rule matches. The procedure returns false when no rule matches, when
+    ///   Quality Management Setup is missing, or when a subscriber to OnBeforeCreateInspection handles the creation.
+    ///
+    /// Up to four records can be supplied. The engine promotes each supplied record to primary in turn until an
+    /// inspection is resolved, and applies the source configuration of every supplied record to the inspection, so the
+    /// inspection may be created from a record other than the first. Unused slots can be left as unassigned variants.
+    /// A temporary "Tracking Specification" record positioned on one tracking line can be passed in any slot to supply
+    /// lot, serial and package information.
+    ///
+    /// Whether an existing inspection is reused or a new one is created follows the "Inspection Creation Option" and
+    /// "Inspection Search Criteria" in Quality Management Setup.
+    /// </summary>
+    /// <param name="PrimaryRecordVariant">The record to create the inspection for (Record, RecordRef, or RecordId).</param>
+    /// <param name="OptionalRelated2Variant">An optional related record used for rule matching and source field mapping, or an unassigned variant.</param>
+    /// <param name="OptionalRelated3Variant">An optional related record used for rule matching and source field mapping, or an unassigned variant.</param>
+    /// <param name="OptionalRelated4Variant">An optional related record used for rule matching and source field mapping, or an unassigned variant.</param>
+    /// <param name="TempFiltersQltyInspectionGenRule">A temporary record whose filters restrict which generation rules are considered; leave unfiltered to consider every automatic rule.</param>
+    /// <param name="QltyInspectionHeader">The created or reused inspection with a record filter applied, or a cleared record when nothing was resolved.</param>
+    /// <param name="IsNewlyCreated">True when the inspection was inserted by this call; false when an existing inspection was reused or nothing was resolved.</param>
+    /// <returns>True if an inspection was created or reused; otherwise, false.</returns>
+    procedure CreateInspectionFromRuleEngine(PrimaryRecordVariant: Variant; OptionalRelated2Variant: Variant; OptionalRelated3Variant: Variant; OptionalRelated4Variant: Variant; var TempFiltersQltyInspectionGenRule: Record "Qlty. Inspection Gen. Rule" temporary; var QltyInspectionHeader: Record "Qlty. Inspection Header"; var IsNewlyCreated: Boolean): Boolean
+    var
+        PreviousPreventShowingState: Boolean;
+        HasInspection: Boolean;
+    begin
+        Clear(QltyInspectionHeader);
+        IsNewlyCreated := false;
+
+        // The flag is assigned directly rather than through SetPreventDisplayingInspectionEvenIfConfigured so the
+        // previous value can be restored, leaving the instance unchanged for the caller after this call.
+        PreviousPreventShowingState := PreventShowingGeneratedInspectionEvenIfConfigured;
+        PreventShowingGeneratedInspectionEvenIfConfigured := true;
+        HasInspection := CreateInspectionWithMultiVariants(PrimaryRecordVariant, OptionalRelated2Variant, OptionalRelated3Variant, OptionalRelated4Variant, false, TempFiltersQltyInspectionGenRule);
+        PreventShowingGeneratedInspectionEvenIfConfigured := PreviousPreventShowingState;
+
+        if not HasInspection then
+            exit(false);
+
+        if not GetCreatedInspection(QltyInspectionHeader) then begin
+            Clear(QltyInspectionHeader);
+            exit(false);
+        end;
+
+        IsNewlyCreated := IsLastInspectionNewlyCreated();
+        exit(true);
+    end;
+
+    /// <summary>
+    /// Creates or reuses inspections for the supplied records and each line of a temporary tracking specification
+    /// buffer, through the generation rule engine with the same automatic semantics as CreateInspectionFromRuleEngine.
+    /// One creation call is issued per tracking line, with the tracking line supplied as the fourth record, which is
+    /// the same shape the app's own receiving triggers use for tracked sources. The buffer is iterated
+    /// within its current filters, so the caller can restrict the lines to process, and is left positioned on the last
+    /// processed line. When the buffer holds no lines, a single creation call is issued without tracking information.
+    /// The caller supplies the tracking lines, so any source can be served, including sources for which the app has no
+    /// tracking collector of its own.
+    ///
+    /// Several tracking lines can resolve to the same inspection depending on the "Inspection Creation Option" and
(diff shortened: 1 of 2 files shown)


alexei-dobriansky:

# **Recommendation: Accept**

<details>

This change adds a narrow public facade for creating or reusing Quality Management inspections through the rule engine with automatic semantics. It keeps the implementation helpers internal, lets callers pass rule filters and related records, and adds a tracked-source overload that iterates caller-supplied tracking lines. The tests cover creation, reuse, no-match behavior, manual-rule exclusion, rule filtering, tracking, and duplicate result handling.
</details>

<details>

**Fit:** Strong

The requested behavior is a supported way for extensions to use the same automatic inspection creation engine without opening UI or applying manual-only rules. The diff exposes that narrow entry point and preserves the existing internal engine shape instead of making broad implementation details public.
</details>

<details>

No suggestions.
</details>

<details>

**Risk:** The main risk is compatibility of the new public procedures, because their signatures and documented semantics become supported surface. Existing behavior is additive only; the internal helpers remain internal, and the tests exercise the main automatic and tracking paths.

**Necessity:** The change is justified because callers otherwise have to copy rule matching, source mapping, duplicate detection, and tracking handling outside the app. The scope is targeted to a facade on the existing creation codeunit plus tests for the new contract.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11028 round=1 by=alexei-dobriansky at=2026-09-04T09:37:26.7939074Z lastSha=5da7d941ecff5f154f57f07cbfcbbdd878831342 reviewKey=783e7e438b609168fa32e44d117be70b10f13971579951445b9ec76347cd682d suggestions=

alexei-dobriansky (approved):

PR #11098, opened 2026-09-05
[29.0] Prevent Sales Order Agent from skipping emails moved into monitored folders

tomasevicst:
## What & why

Stop advancing Earliest Sync At after email polling. Moved emails retain their original received timestamp, so advancing this value could permanently exclude them from later retrieval.

The configured Start Date remains the retrieval lower bound, while Graph excludes emails already categorized as processed. Added regression tests for empty and under-full polling.

## Linked work

Fixes [AB#649176](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/649176)

## How I validated this

## Risk & compatibility

Low risk. No schema, API, setup, or connector changes. Existing Start Date behavior during setup, reactivation, and mailbox changes remains unchanged. Previously processed emails continue to be excluded server-side by their Outlook category.

### DIFF
--- a/src/Apps/W1/SalesOrderAgent/app/src/Integration/SOARetrieveEmails.Codeunit.al
+++ b/src/Apps/W1/SalesOrderAgent/app/src/Integration/SOARetrieveEmails.Codeunit.al
@@ -93,7 +93,6 @@ codeunit 4582 "SOA Retrieve Emails"
         end;
 
         AddEmailInboxToSOAEmails(SOASetup, EmailInbox);
-        UpdateSOAEarliestSyncAt(SOASetup, EmailInbox.Count());
         Commit();
 
         SOAEmail.SetRange(Processed, false);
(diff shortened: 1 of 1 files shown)


alexei-dobriansky:

<details>

The change stops email polling from overwriting `Earliest Sync At` after retrieval, so folder monitoring keeps the configured lower bound instead of moving it to the time of an empty or under-full poll. This addresses the reported race where a message is moved into the monitored folder after polling, but keeps its older `receivedDateTime` and is then filtered out forever. The remaining category-exclude filter still prevents already imported messages from being fetched again, and the approach matches the analogous folder-polling pattern that uses a stable lower bound for moved messages.
</details>

<details>

**Fit:** Strong

The bug is caused by using a moving `receivedDateTime` watermark for folder polling. The diff removes that runtime advancement from the retrieval flow, while setup-time lower-bound behavior remains unchanged.
</details>

<details>

**S1** (🟠 Moderate): *Add tests for moved-folder polling*
No test file changed with this production change. Please add a regression test for a message that enters the monitored folder after an empty or under-full poll, and assert it is discovered without creating a duplicate for already processed mail.
</details>

<details>

**Risk:** Folder polling can now query from the configured `Earliest Sync At` on each run, so older unprocessed messages in that window may be seen again until the processed category is applied. The request is still bounded by the existing maximum count and category filter. There are no schema, public API, event publisher, or BaseApp dependencies in this change.

**Necessity:** The deleted update is the path that made eligible moved messages permanently invisible after a successful empty poll. Keeping a stable lower bound is required for monitored folders because moving a message does not change its original received time.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11098 round=1 by=alexei-dobriansky at=2026-09-06T00:17:16Z lastSha=01e6f258980de97ee4a2147a5f361d1a3389431b reviewKey=2196f6dd3acfe6a41834991e1d91984e5c7b96aaca249d09ae0b11981c86d1ae suggestions=S1@1f0e1b8e

ChethanT (approved):

VolodySyn (approved):

PR #11121, opened 2026-09-07
Fix generation rule source table validation 

6CRIPT:
Fixes [AB#642736](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/642736)

`OnModify` relied on `xRec` to determine whether a generation rule previously had a source table. During programmatic modifications, `xRec` can already contain the changed value, allowing an existing source table to be cleared without raising an error.

The validation now reads the persisted generation rule to preserve both intended behaviors:

- Clearing the source table of a valid rule raises an error.
- Legacy rules already stored without a source table remain editable.

Temporary records retain the previous behavior.

### DIFF
--- a/src/Apps/W1/Quality Management/app/src/Configuration/GenerationRule/QltyInspectionGenRule.Table.al
+++ b/src/Apps/W1/Quality Management/app/src/Configuration/GenerationRule/QltyInspectionGenRule.Table.al
@@ -291,13 +291,16 @@ table 20404 "Qlty. Inspection Gen. Rule"
     var
         PersistedQltyInspectionGenRule: Record "Qlty. Inspection Gen. Rule";
     begin
-        if Rec.IsTemporary() then begin
-            if xRec."Source Table No." <> 0 then
-                CheckSourceTableNoIsSet();
-        end else
-            if PersistedQltyInspectionGenRule.Get(Rec."Entry No.") then
-                if PersistedQltyInspectionGenRule."Source Table No." <> 0 then
+        if Rec."Source Table No." = 0 then
+            if Rec.IsTemporary() then begin
+                if xRec."Source Table No." <> 0 then
                     CheckSourceTableNoIsSet();
+            end else begin
+                PersistedQltyInspectionGenRule.SetLoadFields("Source Table No.");
+                if PersistedQltyInspectionGenRule.Get(Rec."Entry No.") then
+                    if PersistedQltyInspectionGenRule."Source Table No." <> 0 then
+                        CheckSourceTableNoIsSet();
+            end;
         UpdateSortOrder();
         if (xRec."Source Table No." <> Rec."Source Table No.") or (Rec.Intent = Rec.Intent::Unknown) or not GuiAllowed() then
             SetIntentAndDefaultTriggerValuesFromSetup();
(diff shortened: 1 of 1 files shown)


src/Apps/W1/Quality Management/app/src/Configuration/GenerationRule/QltyInspectionGenRule.Table.al
```diff
@@ -288,9 +288,19 @@ table 20404 "Qlty. Inspection Gen. Rule"
...
     begin
-        if xRec."Source Table No." <> 0 then
-            CheckSourceTableNoIsSet();
+        if Rec."Source Table No." = 0 then
+            if Rec.IsTemporary() then begin
+                if xRec."Source Table No." <> 0 then
+                    CheckSourceTableNoIsSet();
+            end else begin
+                PersistedQltyInspectionGenRule.SetLoadFields("Source Table No.");
+                if PersistedQltyInspectionGenRule.Get(Rec."Entry No.") then
```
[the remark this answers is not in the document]
6CRIPT: If Get() returns false, the target record no longer exists, so the enclosing Modify cannot succeed either

alexei-dobriansky:

# **Recommendation: Accept**

<details>

The change updates `Qlty. Inspection Gen. Rule` so a saved generation rule cannot be modified to clear `Source Table No.` without raising the existing table-required error. For non-temporary records it checks the persisted row instead of trusting `xRec`, which addresses the reported programmatic modify path where `xRec` may already contain the changed value.

The guard is narrow: it only runs when the current value is blank. It still allows legacy rows that were already saved without a source table to be edited, keeps the old temporary-record behavior, and leaves the later sort-order and intent refresh logic unchanged.
</details>

<details>

**Fit:** Strong

The reported bug is a missing validation when an existing valid rule is saved after its table number is cleared. The diff checks the old persisted value at modify time, so it restores that validation while preserving the intended legacy-data exception.
</details>

<details>

None.
</details>

<details>

**Risk:** Low. The change is limited to one configuration table trigger and adds one partial-record lookup only when a non-temporary rule is being saved with a blank `Source Table No.`. It does not change public APIs, event publishers, posting, ledger, or amount logic.

**Necessity:** Required. Without this change, programmatic modifies can leave an existing generation rule without a source table, which breaks the mandatory-table rule and the existing regression test scenario.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11121 round=1 by=alexei-dobriansky at=2026-09-07T12:16:31Z lastSha=47c64dfb949d2257b5f2169919f825894000156f reviewKey=32bc48202bbe67c1e30cca6ecf385b38682963b70ce9f0d17550374757b3f199 suggestions=none

src/Apps/W1/Quality Management/app/src/Configuration/GenerationRule/QltyInspectionGenRule.Table.al
```diff
@@ -288,9 +288,19 @@ table 20404 "Qlty. Inspection Gen. Rule"
     end;
 
     trigger OnModify()
+    var
```
attilatoury, line 291: The bug is about a disabled test, the fix looks fine, but have you also re-enabled the test (delete the corresponding line in the json file mentioned in the bug) and ran a stability job (there's an enlistment command to run all the relevant tests a few times so your fix is confirmed)?

6CRIPT: I thought I was supposed to do it in a separated PR, I did it this way previously. Should I do both things here?

alexei-dobriansky (approved):

JakovljevicDusan (approved):

PR #11136, opened 2026-09-07
[29.0][Bug Fix] Fix recurring journal background posting for IRS 1099 invoice lines (US)

Explorer986:
## What & why

Fixes [AB#649300](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/649300)

## How I validated this

## Risk & compatibility

### DIFF
--- a/src/Apps/US/IRSForms/app/src/Extensions/IRS1099BaseAppSubscribers.Codeunit.al
+++ b/src/Apps/US/IRSForms/app/src/Extensions/IRS1099BaseAppSubscribers.Codeunit.al
@@ -269,7 +269,7 @@ codeunit 10032 "IRS 1099 BaseApp Subscribers"
 
     local procedure SaveChangesInGenJnlLine(var GenJnlLine: Record "Gen. Journal Line")
     begin
-        if GenJnlLine."Line No." <> 0 then
+        if (GenJnlLine."Line No." <> 0) and (GenJnlLine."Job Queue Status" = GenJnlLine."Job Queue Status"::" ") then
             if GenJnlLine.Modify(true) then;
     end;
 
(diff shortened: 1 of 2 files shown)


alexei-dobriansky:

# **Recommendation: Accept**

<details>

This PR changes the IRS 1099 Gen. Journal Line subscriber so it only performs its immediate Modify(true) when the journal line is not in a job queue state. When a recurring journal line is already scheduled or posting, the subscriber still derives the IRS fields on the record buffer, but the posting flow keeps ownership of the database save.

The change matches the failing path. The background posting code marks the line with a job queue status before the recurring posting update, and the recurring update later saves with Modify() without running the table trigger. Skipping the subscriber self-save for non-blank job queue status removes the re-entry into OnModify while keeping normal unscheduled journal-line validation behavior unchanged.
</details>

<details>

**Fit:** Strong

The bug is clear: recurring journal background posting in the United States fails when IRS 1099 sync runs on invoice or credit memo lines with a job queue status. The diff targets that path and adds coverage for the failing invoice line, the unaffected blank document type, and the scheduled-line guard.
</details>

<details>

*None.*
</details>

<details>

**Risk:** This touches a financial posting path because it runs while recurring general journal lines are being posted and can affect IRS 1099 fields copied to vendor ledger entries. The change is narrow: it only suppresses the extra save when "Job Queue Status" is not blank, and it does not change schema, permissions, public events, or the Gen. Journal Line guard itself.

**Necessity:** The change is required because the workaround is to turn off background posting or remove the job queue for affected recurring journals. The scope is appropriate because it changes the IRS 1099 self-save instead of relaxing the scheduled-for-posting protection.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11136 round=1 by=alexei-dobriansky at=2026-09-07T13:29:15Z lastSha=d69976e66b9735548cd77b0eed208e806b23c304 reviewKey=38e10a64f27da09315e2073033fd43d4a4d5a9844abba55648d596419497f56f suggestions=none

mynjj (approved):

ventselartur (approved):

PR #11138, opened 2026-09-07
[29.x][Bug Fix] Fix recurring journal background posting for IRS 1099 invoice lines (US)

Explorer986:
## What & why

Fixes [AB#649301](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/649301)

## How I validated this

## Risk & compatibility

### DIFF
--- a/src/Apps/US/IRSForms/app/src/Extensions/IRS1099BaseAppSubscribers.Codeunit.al
+++ b/src/Apps/US/IRSForms/app/src/Extensions/IRS1099BaseAppSubscribers.Codeunit.al
@@ -269,7 +269,7 @@ codeunit 10032 "IRS 1099 BaseApp Subscribers"
 
     local procedure SaveChangesInGenJnlLine(var GenJnlLine: Record "Gen. Journal Line")
     begin
-        if GenJnlLine."Line No." <> 0 then
+        if (GenJnlLine."Line No." <> 0) and (GenJnlLine."Job Queue Status" = GenJnlLine."Job Queue Status"::" ") then
             if GenJnlLine.Modify(true) then;
     end;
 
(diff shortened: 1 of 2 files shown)


alexei-dobriansky:

<details>

This PR adds a Job Queue Status guard around the IRS 1099 Gen. Journal Line self-save, and adds tests for scheduled recurring journal lines. The reported failure is a US recurring-journal background-posting path where validating Posting Date re-enters OnModify while the line is scheduled for posting.

The main direction is sound. The BaseApp recurring posting flow validates the next posting date and then saves the line itself without running OnModify, so skipping the subscriber's extra Modify(true) avoids the scheduled-for-posting guard and still lets the posting owner persist the IRS 1099 fields. The new tests cover the scheduled invoice path, the blank document type path, and the guard that prevents a scheduled line from being silently saved by the subscriber.
</details>

<details>

**Fit:** Partial

The bug is clear, and the code targets the re-entrant save that causes background posting to fail. The guard may be a little wider than needed because it also skips the helper for lines in Error status, even though those lines are allowed to be edited.
</details>

<details>

**S1** (🟠 Moderate): *Keep error lines self-saving*
The new condition skips SaveChangesInGenJnlLine when Job Queue Status is Error, but the table trigger allows edits in that status. Please allow both blank and Error, or explain why IRS fields should stop self-saving on errored job queue lines. This avoids stale IRS 1099 data when a failed journal line is corrected before it is posted again.
</details>

<details>

**Risk:** This touches a US IRS 1099 recurring-journal posting path, so stale or wrongly saved IRS fields can affect vendor reporting data. The code change is narrow and does not change schema, permissions, public APIs, or event signatures, but the helper is shared by several Gen. Journal Line validation subscribers.

**Necessity:** The change is needed because background posting should be able to advance recurring invoice lines without failing on the scheduled-for-posting guard. The scope is close to the problem, but it should preserve the existing editable Error status behavior if there is no reason to change it.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11138 round=1 by=alexei-dobriansky at=2026-09-07T13:39:33Z lastSha=0e10d5fbe2fa8f1b8a903bd6bacb4dc6463696ab reviewKey=4587d7ee8be4a104f55fabd0d11a729c63c587a97cf77fff9e1039c2f15cddc8 suggestions=S1@a807a8a1

mynjj (approved):

ventselartur (approved):

PR #11139, opened 2026-09-07
The Word add-in shouldn't show the columns that are obsoleted in the data picker

mbaranovych:
## What & why

The Word add-in shouldn't show the columns that are obsoleted in the data picker.

Currently, if you open BC's Word data pane, you are able to select columns from **all** the columns in the dataset. Feedback from engineers and partners is that obsolete columns should not be shown in the data picker.
The idea behind this PR is that we want to prevent end users to use obsolete columns for their layouts. And this shouldn't be a problem because if they are obsolete, there should be a non-obsoleted column that they can use.

New logic on `TempWordTemplateField` prevents obsolete or deleted fields from corrupting the selected-field count.

## Linked work

[AB#611990](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/611990)

## How I validated this

## Risk & compatibility

As mentioned, hiding obsolete fields should not cause problem because there should be alternative non-obsolete columns

### DIFF
--- a/src/System Application/App/Word Templates/src/WordTemplateFieldSelection.Codeunit.al
+++ b/src/System Application/App/Word Templates/src/WordTemplateFieldSelection.Codeunit.al
@@ -165,6 +165,7 @@ codeunit 9989 "Word Template Field Selection"
         ExcludedFields: Integer;
     begin
         Field.SetRange(TableNo, TableId);
+        Field.SetRange(ObsoleteState, Field.ObsoleteState::No);
         TableFields := Field.Count();
 
         WordTemplateCustomField.SetCurrentTableId(TableId);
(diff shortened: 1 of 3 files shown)


src/System Application/App/Word Templates/src/WordTemplateFieldSelection.Codeunit.al
```diff
@@ -165,6 +165,7 @@ codeunit 9989 "Word Template Field Selection"
         ExcludedFields: Integer;
     begin
         Field.SetRange(TableNo, TableId);
+        Field.SetRange(ObsoleteState, Field.ObsoleteState::No);
```
darjoo, line 168: Should this include pending state? Those are still valid fields until they are actually obsoletestate=yes.

darjoo (approved):

PR #11145, opened 2026-09-07
Disable VAT period checks in E-Document item charge tests

djukicmilica:
## Why

The E-Document item charge tests fail under the Czech localization because sales posting validates the current VAT reporting date against localization-specific VAT periods. These W1 tests exercise item charge mapping, not VAT-date behavior, and must not depend on localized VAT periods covering the moving work date.

Cross-app review confirmed that disabling VAT Reporting Date is the established normalization for localization-neutral posting tests: it is used throughout E-Document core, PEPPOL, and connector test suites. Tests that exercise VAT reporting itself keep the feature enabled and create valid periods in their localization-specific test libraries.

## Summary

- **Disabled** VAT Reporting Date in the shared item charge test initializer so sales posting does not invoke country-specific VAT period validation.
- **Documented** the test boundary and why creating Czech VAT periods is not appropriate for this W1 test suite.
- **Aligned** the setup with the existing cross-localization pattern used by E-Document and connector tests.

Fixes
[AB#649216](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/649216)

### DIFF
--- a/src/Apps/W1/EDocument/Test/src/Processing/EDocItemChargeTests.Codeunit.al
+++ b/src/Apps/W1/EDocument/Test/src/Processing/EDocItemChargeTests.Codeunit.al
@@ -959,6 +959,8 @@ codeunit 139786 "E-Doc. Item Charge Tests"
         if IsInitialized then
             exit;
 
+        // Item charge tests do not exercise VAT date behavior. Disable it to avoid localization-specific
+        // VAT period validation during posting, such as the CZ check for the current work date.
         GeneralLedgerSetup.Get();
         GeneralLedgerSetup."VAT Reporting Date Usage" := GeneralLedgerSetup."VAT Reporting Date Usage"::Disabled;
         GeneralLedgerSetup.Modify();
--- a/src/Apps/W1/EDocument/Test/src/Processing/EDocItemChargeTests.Codeunit.al
+++ b/src/Apps/W1/EDocument/Test/src/Processing/EDocItemChargeTests.Codeunit.al
@@ -5,6 +5,7 @@
 namespace Microsoft.eServices.EDocument.Test;
 
 using Microsoft.eServices.EDocument;
+using Microsoft.Finance.GeneralLedger.Setup;
 using Microsoft.Finance.VAT.Setup;
 using Microsoft.Foundation.Enums;
 using Microsoft.Inventory.Item;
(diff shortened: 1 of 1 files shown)


gggdttt (approved):

alexei-dobriansky:

# **Recommendation: Accept**

<details>

This change disables VAT Reporting Date in the shared E-Document item charge test initialization. The failing scenario is a localization-neutral item charge posting test that should not depend on country-specific VAT period data for the current work date.

The setup runs before any test posts sales documents, so it removes the VAT period check before the failure point. The item charge tests still create and compare VAT posting setup data, but they do not validate VAT reporting-date behavior, so disabling that check is scoped to the right test boundary. Similar E-Document test setup already uses the same normalization.
</details>

<details>

**Fit:** Strong

The reported problem is a test failure caused by VAT period validation during posting, while the test goal is item charge mapping. The diff disables that unrelated validation for the whole test codeunit before posting starts, so it addresses the failure without changing production logic.
</details>

<details>

*None.*
</details>

<details>

**Risk:** The change only touches a test codeunit. It changes shared test setup for all E-Document item charge tests, so the main regression risk is hiding a VAT reporting-date behavior issue inside this suite; these tests do not assert that behavior, and dedicated VAT/date tests should keep covering it. No public API, event, posting implementation, or BaseApp publisher was changed.

**Necessity:** The change is justified because the suite should be stable across localizations and should test item charge mapping, not the current date's VAT period setup. The scope is narrow: one import, one setup record update, and no production code change.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11145 round=1 by=alexei-dobriansky at=2026-09-07T22:05:34Z lastSha=869b2693bdb49498cbc4a283e3599a540b85d699 reviewKey=9c1b89c193f63578183b6c940f5887b527e83ac2f6d1d7186ee005ed1af4cd59 suggestions=none

DavidHolusa (approved):
Hi. The VAT period validation can be disabled only for CZ localization but for that is needed to add the standard events from Library - Test Initialize codeunit to Initialize function. We subscribe this events in our CZ test applications.

You can see the usage in codeunit 148104 "Test Initialize Handler CZL" where is the subscriber UpdateRecordsOnBeforeTestSuiteInitialize and there are the following lines

```
...
            137161, // SCM Warehouse Orders
            139989: // Sub. Subcontracting Test
                DisableVATDateUsage();
...
```
so if you want then you can disable the VAT date usage only for CZ this way.

PR #11161, opened 2026-09-07
[29.x][BE][Purchase Ledger Report] Wrong VAT Amounts on the report

dcenic:
## What & why

Fixing an issue with Belgium-specific Purchase Ledger report that was:

using stale VAT amount values for certain lines
having wrong alignment of two RDLC cells

## Linked work

Fixes [AB#649339](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/649339)

## How I validated this

I added or updated tests for the new behavior

### DIFF
--- a/src/Layers/BE/BaseApp/Local/Purchases/Reports/PurchaseLedger.Report.al
+++ b/src/Layers/BE/BaseApp/Local/Purchases/Reports/PurchaseLedger.Report.al
@@ -802,10 +802,9 @@ report 11301 "Purchase Ledger"
         GLEntryVATEntryLink.SetRange("G/L Entry No.", GLEntryNo);
         if GLEntryVATEntryLink.FindSet() then
             repeat
-                if VATEntry.Get(GLEntryVATEntryLink."VAT Entry No.") then begin
-                    VATDetailBaseAmount += VATEntry.Base;
-                    VATDetailVATAmount += VATEntry.Amount;
-                end;
+                VATEntry.Get(GLEntryVATEntryLink."VAT Entry No.");
+                VATDetailBaseAmount += VATEntry.Base;
+                VATDetailVATAmount += VATEntry.Amount;
             until GLEntryVATEntryLink.Next() = 0;
     end;
 
--- a/src/Layers/BE/BaseApp/Local/Purchases/Reports/PurchaseLedger.Report.al
+++ b/src/Layers/BE/BaseApp/Local/Purchases/Reports/PurchaseLedger.Report.al
@@ -70,11 +70,11 @@ report 11301 "Purchase Ledger"
                     {
                         AutoFormatType = 1;
                     }
-                    column(VATDetailBase; VATDetail.Base)
+                    column(VATDetailBase; VATDetailBaseAmount)
                     {
                         AutoFormatType = 1;
                     }
-                    column(VATDetailAmount; VATDetail.Amount)
+                    column(VATDetailAmount; VATDetailVATAmount)
                     {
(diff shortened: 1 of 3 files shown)


AndersLarsenMicrosoft (approved):

AndersLarsenMicrosoft (approved):

alexei-dobriansky:

<details>

This change makes the Belgian Purchase Ledger report print VAT detail amounts from the G/L Entry - VAT Entry Link table, and it clears the report variables for each line. The new lookup matches each G/L entry to its linked VAT entries instead of walking VAT entries by transaction order, so the multi-rate case is handled directly. The added regression test covers a 21% line and a 0% line in the same purchase invoice, and the RDLC changes only align numeric cells.
</details>

<details>

**Fit:** Strong

The reported problem is that the report can print VAT base and amount from the wrong line, and the fix changes the data source to the deterministic link between the G/L entry and VAT entry. The scope is focused on the affected report and includes a test for the mixed-rate case that caused the stale or shifted amounts.
</details>

<details>

**S1** (🟠 Moderate): *Check the parallel Sales Ledger logic*
The same transaction-order VAT walk still exists in SalesLedger.Report.al. Please either apply the link-table lookup there too, or explain why sales cannot hit the same multi-rate transaction case. Otherwise the parallel report can still print VAT amounts from the wrong line.
</details>

<details>

**Risk:** The regression surface is the Belgian Purchase Ledger report detail dataset and RDLC layout. The change does not alter posting data, public APIs, or upgrade logic, but an incorrect lookup would affect report output for purchase documents with VAT. The dataset column names stay the same, so layout and consumer compatibility risk is low.

**Necessity:** The change is needed because matching VAT entries by transaction order can show the wrong VAT base or amount on a detail line. The link-table lookup is the right scope for the affected purchase report, and the remaining concern is whether the same fix should also cover the parallel sales report.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11161 round=1 by=alexei-dobriansky at=2026-09-07T20:01:29Z lastSha=5c0bae9921de3e8144d44bafebf81e4eb65b2da5 reviewKey=b169f94b126d47255181db9a1aa6946fd7d83d70011cb04941810d7060f8667c suggestions=S1@e15acd68

src/Layers/BE/BaseApp/Local/Purchases/Reports/PurchaseLedger.Report.al
```diff
@@ -811,4 +794,18 @@ report 11301 "Purchase Ledger"
...
 
+    local procedure CalculateVATDetailAmounts(GLEntryNo: Integer)
+    var
+        GLEntryVATEntryLink: Record "G/L Entry - VAT Entry Link";
+        VATEntry: Record "VAT Entry";
+    begin
+        GLEntryVATEntryLink.SetRange("G/L Entry No.", GLEntryNo);
+        if GLEntryVATEntryLink.FindSet() then
+            repeat
+                VATEntry.Get(GLEntryVATEntryLink."VAT Entry No.");
```
ventselartur, line 805: VATEntry.SetLoadsFields(...)

ventselartur (approved):

src/Layers/BE/BaseApp/Local/Purchases/Reports/PurchaseLedger.Report.al
dcenic: Won't fix in this PR

PR #11187, opened 2026-09-08
[Main] - User encounters a "record is not open" error when attaching a document to a posted sales shipment in Dynamics 365 (Base Application v28)

DhavalMore88:
[Bug 646549](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/646549): [master] [All-E] [SME][SaaS] User encounters a "record is not open" error when attaching a document to a posted sales shipment in Dynamics 365 (Base Application v28).

[AB#646549](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/646549)

Issue: When a user uploads a file from the Documents FactBox ("Doc. Attachment List Factbox", Page 1178) on a Posted Sales Shipment, the upload fails with the runtime error "The record is not open." The same happens on Posted Return Receipt, while the identical action works correctly on Sales Order, Purchase Order and Posted Sales Invoice. The AL call stack points to Table 1173 "Document Attachment".InsertAttachment / SaveAttachmentFromStream.

Cause: The GetRefTable procedure in DocumentAttachmentMgmt.Codeunit.al (Codeunit 1173) has a case DocumentAttachment."Table ID" statement with no branch for Database::"Sales Shipment Header" (110) or Database::"Return Receipt Header" (6661), even though both tables are already registered in TableHasNumberFieldPrimayKey. Because no branch matches, RecRef.Open() is never called, GetRefTable returns false, and the FactBox's AttachmentsUpload action ignored that result and passed the unopened RecordRef on to DocumentAttachment.SaveAttachment(files, RecRef), where RecRef.Find() in InsertAttachment throws "The record is not open."

Solution: Added the missing Database::"Sales Shipment Header" and Database::"Return Receipt Header" cases to GetRefTable, each opening the RecordRef on its table and calling GetTable when the posted document is found — matching the existing pattern used for Sales Invoice Header and Sales Cr.Memo Header. Additionally hardened all three GetRefTable call sites in DocAttachmentListFactbox.Page.al (Upload files, Show details, Attach from email) to raise a clear error when the source record still cannot be resolved after the OnAfterGetRecRefFail event, instead of letting an unopened RecordRef reach the platform and surface the cryptic message for any other unmapped table.

### DIFF
--- a/src/Apps/W1/EDocument/App/src/Processing/EDocAttachmentProcessor.Codeunit.al
+++ b/src/Apps/W1/EDocument/App/src/Processing/EDocAttachmentProcessor.Codeunit.al
@@ -158,11 +158,8 @@ codeunit 6169 "E-Doc. Attachment Processor"
     begin
         case DocumentAttachment."Table ID" of
             Database::"E-Document":
-                begin
-                    RecRef.Open(Database::"E-Document");
-                    if EDocument.Get(DocumentAttachment."No.") then
-                        RecRef.GetTable(EDocument);
-                end;
+                if EDocument.Get(DocumentAttachment."No.") then
+                    RecRef.GetTable(EDocument);
         end;
     end;
 
--- a/src/Apps/W1/ExpenseAgent/app/src/Expense/Codeunits/ExpenseAttachmentMgt.Codeunit.al
+++ b/src/Apps/W1/ExpenseAgent/app/src/Expense/Codeunits/ExpenseAttachmentMgt.Codeunit.al
@@ -85,23 +85,14 @@ codeunit 6989 "Expense Attachment Mgt."
     begin
         case DocumentAttachment."Table ID" of
             Database::Expense:
-                begin
-                    RecRef.Open(Database::Expense);
-                    if Expense.Get(DocumentAttachment."No.") then
-                        RecRef.GetTable(Expense);
-                end;
+                if Expense.Get(DocumentAttachment."No.") then
+                    RecRef.GetTable(Expense);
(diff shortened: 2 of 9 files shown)


v-ankitgoyal (dismissed):

PredragMaricic (changes requested):
S1: Exercise the actual FactBox subscriber fallback

`EnsureShowDetailsUsesRecRefResolvedBySubscriber` does not test the behavior named by the scenario. After binding the `OnAfterGetRecRefFail` subscriber, it creates a `Customer` `RecordRef` itself and calls `DocumentAttachmentDetails.OpenForRecRef` directly. This bypasses the `Doc. Attachment List Factbox`, `ResolveSourceRecRef`, `GetRefTable`, and the `OnAfterGetRecRefFail` event, so the test still passes if the FactBox stops publishing the event, ignores the subscriber-provided `RecordRef`, or raises the new error incorrectly.

Please drive the actual `OpenInDetail` action on a `TestPage "Doc. Attachment List Factbox"` filtered to the unmapped attachment, then verify that `DocumentAttachmentDetailsMPH` sees the customer attachment resolved by the subscriber. Apply the equivalent correction to both the W1 and ES test copies.

The direct mappings for Posted Sales Shipment and Posted Return Receipt are sound, but this PR also centralizes three FactBox actions behind a strict resolver and changes missing-record behavior across several attachment subscribers. That broader behavior needs a regression test which executes the real fallback event path.

PR #11215, opened 2026-09-08
Fix 647572: add ContractTemplateConfirmHandler to FI Company Field Report Test

Explorer986:
## What & why
The two service contract tests in codeunit 148150 fail on and off with an unhandled confirm, "Do you want to create the contract using a contract template?". Table 5965 Service Contract Header raises that confirm on insert only when a Service Contract Template record exists. Both tests create a contract through the shared helper CreateServiceContract but the codeunit does not handle the confirm, so whenever a template is left in the shared test database the dialog has nowhere to go. That is why it fails intermittently. This is a test isolation gap, not a product bug.

##Fix

Clear the Service Contract Template table at the start of CreateServiceContract so the confirm never fires. No template is applied, the tests behave as they did when green, and the assertions are unchanged. Both tests use this one helper, so the single change covers both.

Fixes [AB#647572](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/647572)

## How I validated this

## Risk & compatibility

Test-only, FI Core test app. No product code, schema, permissions or upgrade impact as nothing ships to customers.

### DIFF
--- a/src/Apps/FI/FICore/test/src/FICompanyFieldReportTest.Codeunit.al
+++ b/src/Apps/FI/FICore/test/src/FICompanyFieldReportTest.Codeunit.al
@@ -342,10 +342,12 @@ codeunit 148150 "FI Company Field Report Test"
     var
         ServiceContractHeader: Record "Service Contract Header";
         ServiceContractLine: Record "Service Contract Line";
+        ServiceContractTemplate: Record "Service Contract Template";
         Customer: Record Customer;
         ServiceItem: Record "Service Item";
         DocumentNumber: Variant;
     begin
+        ServiceContractTemplate.DeleteAll();
         LibrarySales.CreateCustomer(Customer);
         LibraryService.CreateServiceContractHeader(ServiceContractHeader, ServiceContractType, Customer."No.");
         LibraryService.CreateServiceItem(ServiceItem, Customer."No.");


alexei-dobriansky:
﻿## Good Sense Reviewer - Round 1

# **Recommendation: Accept**

<details>

This change wires a confirm handler into the two FI service contract report tests. The handler declines service contract template creation when a template exists, so the tests keep the same contract data they already assert against.

The fix matches the root cause. The confirm is raised from Service Contract Header insert only when a service contract template record exists, and both changed tests create the contract before opening the report request page. Replying false matches the default response in the product prompt and the older analogous FI test pattern, so the report assertions still test the intended company fields.
</details>

<details>

**Fit:** Strong

The reported problem is intermittent unhandled UI in these two service contract report tests when shared test data contains a contract template. The diff fixes exactly those two test entry points and does not change the report logic or the expected dataset assertions.
</details>

<details>

*None.*
</details>

<details>

**Risk:** The change is test-only in the FI Core test app. It does not change production code, schema, permissions, posting, ledger data, or public integration surfaces. The handler is broad, but it is attached only to the two affected tests, and the surrounding flow expects this template prompt before the report request page is handled.

**Necessity:** The change is needed because these tests can fail before reaching the report assertions when the database already has a service contract template. The scope is right because it only adds the missing UI handler and keeps the existing test behavior.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11215 round=1 by=alexei-dobriansky at=2026-09-08T14:55:17Z lastSha=3781d2eb763055c88a4ed2fe574696eb2d508e11 reviewKey=05cca68373b5f18a76880472bfd0548965c191bd19d726b734928796bf13d34e suggestions=none

alexei-dobriansky:

<details>

This change tries to stop two FI service contract report tests from failing when service contract templates exist in the test company. The current code avoids the confirm by deleting all service contract templates before creating the contract.

That makes the immediate dialog disappear, but it does not exercise the same path that fails when a template is present. The older matching FI test handles the confirm and replies `false`, which keeps the template data in place and matches the default answer used by the platform code.
</details>

<details>

**Fit:** Partial

The reported failure is clear and the changed area is the right helper used by both failing tests. The implementation only avoids the condition by removing all templates, so it is less targeted than handling the confirm that appears in the reported scenario.
</details>

<details>

*None.*
</details>

<details>

**S1** (🟠 Moderate): *Avoid deleting all service contract templates*
`ServiceContractTemplate.DeleteAll()` removes every template in the test company, and the helper commits later in the same flow. Please use a confirm handler that replies `false` instead, so these tests do not change unrelated template data.
</details>

<details>

**Risk:** Product risk is low because the change is test-only. The test-suite risk is that a broad delete can remove setup data created by another test and can hide the actual template-present path instead of proving it is handled.

**Necessity:** A fix is needed because the report tests can hit an unhandled confirm when a service contract template exists. The scope is small, but the safer shape is to handle the confirm rather than delete shared template data.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11215 round=2 by=alexei-dobriansky at=2026-09-09T01:09:07Z lastSha=b9feafbeda63bcd99f823bc2d673538d25d34d9a reviewKey=dc2dc3e9f16009294a20eaedbb77cfc2558422595b24e386c0701ad049a33948 suggestions=S1@e91d3bf3:new parentRound=1

ventselartur:
Are any of these unstable tests have been disabled? Do you need to enable them back?

ventselartur (approved):

mynjj (approved):

PR #11233, opened 2026-09-08
[MAIN]-Not enough inventory available at vendor location for this order

v-maheshsharma:
**Issue:**
Subcontracting receive/consume was failing with “Not enough inventory available at vendor location for this order”

**Cause:**
Failing due to Multiple ILE for larger quantity Flushing Method Backward and has

**Solution:**
Updated GetApplicationLines method by removing SetRange on "Entry No." so all ILE can be used during receiving.

### DIFF
--- a/src/Apps/IN/INGST/app/GSTSubcontracting/src/Codeunit/SubcontractingPost.Codeunit.al
+++ b/src/Apps/IN/INGST/app/GSTSubcontracting/src/Codeunit/SubcontractingPost.Codeunit.al
@@ -1623,7 +1623,6 @@ codeunit 18466 "Subcontracting Post"
         end else begin
             ItemLedgerEntry.Reset();
             ItemLedgerEntry.SetCurrentKey("Entry Type", "Location Code", "External Document No.", "Item No.");
-            ItemLedgerEntry.SetRange("Entry No.", AppDelChallan."Applies-to Entry");
             ItemLedgerEntry.SetRange("Entry Type", ItemLedgerEntry."Entry Type"::Transfer);
             ItemLedgerEntry.SetRange("Location Code", SubOrderCompVend."Vendor Location");
             ItemLedgerEntry.SetRange("External Document No.", AppDelChallan."Applied Delivery Challan No.");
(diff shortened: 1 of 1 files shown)


alexei-dobriansky:

<details>

The change lets the posting code use more than the one stored item ledger entry when receive or consume needs split remaining quantities. That matches the reported failure, but the new filter no longer scopes entries to the selected delivery challan line or production component. The surrounding challan logic uses order, order line, component line, and subcontracting order filters, so the posting code should keep that scope while removing only the single-entry limit.
</details>

<details>

**Fit:** Partial

The change addresses the need to use multiple open transfer entries, but it broadens the record set beyond the selected line and component. It is close to the right approach, but the missing filters can make the fix use unrelated inventory.
</details>

<details>

**S1** (🔴 High): *Keep the line-specific ILE filters*
The old single-entry filter was too narrow, but the new filter is now too broad. It can take open transfer entries from another challan line or production component with the same challan number, item, variant, and vendor location. Keep all matching entries, but add the order, order line, component line, and subcontracting order filters used when the applied line is created.

**S2** (🟠 Moderate): *Add split-entry posting coverage*
No test change covers split item ledger entries in this subcontracting posting path. This path moves inventory, so the corrected filter should be covered with multiple open entries for the same challan and item. Please add a regression test that proves only entries for the selected challan line and component are used.
</details>

<details>

**Risk:** This runs in GST subcontracting posting and selects item ledger entries for consume, vendor return, company return, and receipt flows. If it picks a sibling line's entry, inventory can move against the wrong source while the intended line remains open. No public API or event contract change was found.

**Necessity:** The change is necessary because a single stored entry can be insufficient when the vendor location has several open transfer entries for the selected delivery challan. The scope is small, but it must preserve the line and order identity already used by the challan records.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11233 round=1 by=alexei-dobriansky at=2026-09-09T01:09:42Z lastSha=f1a941c81b51e61de23e32a8cacb87ba5669a943 reviewKey=80d03a5b6ef4728b2e9872161bdb3eb3a70ae043f2960916ae0a6d4ced6feb3f suggestions=S1@2f98338d,S2@191bdc43

PR #11251, opened 2026-09-09
[29.x] bug 649305 - Add Report Address Source functionality to Company Information

DavidHolusa:
## What & why

Adds a new **Report Address Source** option to the Company Information table and page for the CZ Core Localization Pack. This allows users to choose whether the address printed on CZ sales and purchase document reports is taken from Company Information or from the Responsibility Center.

Previously, when a Responsibility Center was set on a document, its address was always used in printed reports. However, CZ legislation requires the billing address to be stated, and customers use Responsibility Centers in various ways — making it not always appropriate to override the company address. This change introduces a configurable setting so each company can control the behavior.

**Changes:**
- New enum `Report Address Source CZL` with options: `Company Information`, `Responsibility Center`
- New field `Report Address Source CZL` on the Company Information table (79)
- New field on the Company Information page (1), Communication FastTab (`Importance = Additional`)
- Updated CZ report address formatting logic to respect the new setting

## Linked work

Fixes [AB#649305](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/649305)

## How I validated this

- Set Report Address Source to "Company Information" → printed sales invoice shows company address regardless of Responsibility Center on the document.
- Set Report Address Source to "Responsibility Center" → printed sales invoice shows Responsibility Center address when one is assigned to the document.
- Verified default value preserves existing behavior (Responsibility Center) for backward compatibility.

## Risk & compatibility

- New enum and field addition — no breaking changes to existing functionality.
- Default value should be set to `Responsibility Center` to maintain backward compatibility with existing behavior after the previous redesign.
- No upgrade code needed as the default enum value covers existing installations.

### DIFF
--- a/src/Apps/CZ/CoreLocalizationPack/app/Src/Codeunits/FormatAddressHandlerCZL.Codeunit.al
+++ b/src/Apps/CZ/CoreLocalizationPack/app/Src/Codeunits/FormatAddressHandlerCZL.Codeunit.al
@@ -1,11 +1,24 @@
 namespace Microsoft.Foundation.Address;
 
+using Microsoft.Foundation.Company;
+
 codeunit 31144 "Format Address Handler CZL"
 {
-    [EventSubscriber(ObjectType::Codeunit, Codeunit::"Format Address", 'OnAfterGeneratePostCodeCity', '', false, false)]
+    [EventSubscriber(ObjectType::Codeunit, Codeunit::"Format Address", OnAfterGeneratePostCodeCity, '', false, false)]
     local procedure ClearCountyOnAfterGeneratePostCodeCity(var Country: Record "Country/Region"; var CountyText: Text[50])
     begin
         ClearCounty(Country, CountyText);
+    end;
+
+    [EventSubscriber(ObjectType::Codeunit, Codeunit::"Format Address", OnBeforeGetCompanyAddr, '', false, false)]
+    local procedure HandleOnBeforeGetCompanyAddr(var CompanyInfo: Record "Company Information"; var CompanyAddr: array[8] of Text[100]; var IsHandled: Boolean)
+    var
+        FormatAddress: Codeunit "Format Address";
+    begin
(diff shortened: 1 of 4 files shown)


AndersLarsenMicrosoft (approved):

alexei-dobriansky:

<details>

This change adds a Company Information setting for CZ report addresses and uses a Format Address subscriber to choose Company Information when that setting is selected. The subscriber is placed before the existing Responsibility Center fallback, so the new option can work. The problem is the new enum makes Company Information the default, which changes existing behavior instead of preserving it.
</details>

<details>

**Fit:** Partial

The change adds the requested setting and connects it to the shared report address flow. It does not fully fit because the default value should keep using Responsibility Center until a company chooses Company Information.
</details>

<details>

**S1** (🔴 High): *Default report address source flips existing behavior*
Because this enum field has no InitValue, value 0 becomes the default for Company Information records. That makes Company Information the default source and bypasses Responsibility Center when a document has one. Make Responsibility Center the default, either by assigning it value 0 or by setting and migrating the field default safely.
</details>

<details>

**Risk:** This affects CZ sales, purchase, and service reports that call Format Address.GetCompanyAddr. The W1 publisher was verified: it fires before the Responsibility Center branch and exits when IsHandled is true, so the new subscriber can change the address used by all those reports. With the current enum ordering, existing companies can silently print Company Information instead of Responsibility Center.

**Necessity:** The setting is needed because some companies must print the company billing address even when a document has a Responsibility Center. The scope is suitable, but the default must preserve the existing Responsibility Center behavior.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11251 round=1 by=alexei-dobriansky at=2026-09-09T22:17:22.3165338Z lastSha=ed6053c3a8991177d2d25d6601e2b51f73c3dd95 reviewKey=da32b59e7fd9151e78c3a624db1585132dfd22b86284f04edb64e762297c58b9 suggestions=S1@e163a4f8

v-janpopr (approved):

PR #11258, opened 2026-09-09
[GB] IdealPostcodes: fix provider selection and address retrieval through the Postcode Service Manager framework

### LINKED ISSUE #11254: [GB] IdealPostcodes provider is unusable through the Postcode Service Manager framework
**Describe the issue**
Reported by a BC partner who tried to integrate with IdealPostcodes as a postcode provider through the Base Application's Postcode Service Manager framework. Two defects in `src/Apps/GB/IdealPostcodes` make the provider unusable on that path (the app's own page extensions are unaffected — they call the provider directly):

1. **The provider cannot be selected.** `IPCProvider.Codeunit.al` (`OnDiscoverPostcodeServices`) registers the display name `IdealPostcodes` in `Name` and the key `IDEAL_POSTCODE_POSTCODE_SERVICE` in `Value` — swapped relative to `PostcodeServiceManager.RegisterService` and to the other provider (GetAddress.io). Page 9143 "Postcode Configuration Page W1" saves the selected row's **Name** as the service key, so every subscriber guard (`if ServiceKey <> MyServiceKeyTok then exit;`) rejects the stored key, `PostcodeServiceManager.IsConfigured()` returns false, and page 9143's `OnOpenPage` resets the stored selection to `Disabled` on every open.

2. **The selected address is never returned.** `OnRetrieveAddress` re-fetches the picked entry via `GetAddressDetails` from `https://api.ideal-postcodes.co.uk/v1/<id>` — a route the Ideal Postcodes API does not have (404; we are the API vendor and can confirm it never existed) — and `ParseAddressDetail` expects fields (`address`, `address2`, `city`, …) the API never returns. Every framework lookup ends with "Failed to retrieve address details."

**Expected behavior**
Selecting IdealPostcodes on the Postcode Service configuration page persists, `IsConfigured()` reports true, and confirming an entry on the "Postcode Select Address" page writes that address back.

**Proposed fix (implemented, verified live on a BC 28.2 sandbox against the production Ideal Postcodes API)**
- Accept both the token and the (locked) display name as the service key.
- In `OnRetrieveAddress`, re-run the search for the entered postcode/city and return the entry matching the selected `Address ID` (falling back to `Display Text`) instead of calling the nonexistent endpoint; this also makes the "Remove Organisation Name" setting apply to the written-back address.
- Remove the dead `GetAddressDetails`/`ParseAddressDetail`.
- Regression test that stores the key exactly the way page 9143 does and asserts `IsConfigured()`.

A PR with this change is ready on branch `sergey-idealpostcodes:fix/gb-idealpostcodes-service-key-and-retrieve-address` and will be opened once this issue is approved.

sergey-idealpostcodes:
Fixes #11254

A BC partner integrating with IdealPostcodes as a postcode provider through the Postcode Service Manager framework reported the lookup never becoming available. Two defects:

1. `OnDiscoverPostcodeServices` registers the display name in `Name` and the key in `Value`, while page 9143 saves the selected row's `Name` as the service key — every subscriber guard rejects the stored key, `IsConfigured()` stays false, and page 9143 resets the selection to `Disabled` on each open.
2. `OnRetrieveAddress` re-fetches the picked address from `https://api.ideal-postcodes.co.uk/v1/<id>` — a route the API does not have (we are the API vendor) — with a parser expecting fields the API never returns, so every framework lookup fails.

Changes:
- `IPCProvider`: new `IsMyServiceKey()` accepts both the token and the display name (label now `Locked` — it acts as a key); `OnRetrieveAddress` re-runs the search and returns the entry matching the selected `Address ID` (fallback: `Display Text`), which also makes the *Remove Organisation Name* setting apply to the written-back address.
- `IPCManagement`: removed dead `GetAddressDetails`/`ParseAddressDetail`.
- Tests: `TestIsConfiguredWithKeySavedByConfigurationPage` stores the key exactly the way page 9143 does and asserts `IsConfigured()`.

Verified live on a BC 28.2 sandbox against the production Ideal Postcodes API (postcode SW1A 2AE), deployed as a per-tenant copy of the app.

### DIFF
--- a/src/Apps/GB/IdealPostcodes/app/src/IPCManagement.Codeunit.al
+++ b/src/Apps/GB/IdealPostcodes/app/src/IPCManagement.Codeunit.al
@@ -57,40 +57,6 @@ codeunit 9400 "IPC Management"
         exit(false);
     end;
 
-    [NonDebuggable]
-    procedure GetAddressDetails(AddressId: Text; var TempIPCAddressLookup: Record "IPC Address Lookup" temporary; var ReasonCode: Integer; var ReasonPhrase: Text)
-    var
-        Config: Record "IPC Config";
-        TypeHelper: Codeunit "Type Helper";
-        FeatureTelemetry: Codeunit "Feature Telemetry";
-        AuditLog: Codeunit "Audit Log";
-        HttpClient: HttpClient;
-        HttpResponse: HttpResponseMessage;
-        ResponseText: Text;
-        RequestUrl: Text;
-    begin
-        if not GetConfiguration(Config) then
-            Error(ConfigNotSetupErr);
-
-        if not Config.Enabled then
-            exit;
-
-        FeatureTelemetry.LogUptake('0000RFF', 'IdealPostcodes', Enum::"Feature Uptake Status"::Used);
-        RequestUrl := Config.APIEndpoint() + '/' + TypeHelper.UriEscapeDataString(AddressId);
-        HttpClient.DefaultRequestHeaders().Add('Authorization', SecretStrSubstNo('IDEALPOSTCODES api_key="%1"', Config.GetAPIPasswordAsSecret(Config."API Key")));
-
-        if HttpClient.Get(RequestUrl, HttpResponse) then begin
-            ReasonCode := HttpResponse.HttpStatusCode();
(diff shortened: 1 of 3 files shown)


sergey-idealpostcodes:
@microsoft-github-policy-service agree company="Ideal Postcodes"

PR #11261, opened 2026-09-09
[Main]- Production order can be deleted with registered unconsumed warehouse picks

DeepsShukla:
Workitem [Bug 647854](https://dev.azure.com/dynamicssmb2/Dynamics%20SMB/_workitems/edit/647854): [All-E] [Repair Item] Production order can be deleted with registered unconsumed warehouse picks

Fixes [AB#647854](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/647854)

### DIFF
--- a/src/Layers/IT/BaseApp/Manufacturing/Document/ProdOrderComponent.Table.al
+++ b/src/Layers/IT/BaseApp/Manufacturing/Document/ProdOrderComponent.Table.al
@@ -1155,7 +1155,8 @@ table 5407 "Prod. Order Component"
                 TestField("Qty. on Transfer Order (Base)", 0);
             end;
 #endif
-            ConfirmDeletion();
+            if not CalledFromHeader then
+                CheckPickedQtyBeforeDeletion();
         end;
 
         ProdOrderWarehouseMgt.ProdComponentDelete(Rec);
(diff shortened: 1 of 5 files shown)


alexei-dobriansky:

<details>

This change blocks released production order, production order line, and component deletion when a component has picked quantity that has not been consumed. The new check runs before header deletion cascades, and the direct line and component paths are covered too. Using base quantities is the right direction, and the new tests cover the blocked header, line, and component paths. However, the guard still uses cumulative picked quantity only, so it does not prove the picked stock is still outstanding after a return movement.
</details>

<details>

**Fit:** Partial

The change fixes the main reported failure by preventing deletion while registered picked quantity is still unconsumed. It does not fully match the requested behavior because a returned picked quantity can still leave the stored picked quantity higher than consumed quantity.
</details>

<details>

**S1** (🔴 High): *Account for returned picked quantity*
The check uses cumulative `Qty. Picked (Base)` as if every unconsumed pick is still in the operation area. A supported return movement can move stock out without reducing this field, so deletion can still be blocked after the quantity is returned. Base the guard on outstanding picked quantity, or update the value this check uses when the return is registered.
</details>

<details>

**Risk:** The changed area protects manufacturing production orders that use warehouse picks. If the guard is too weak, warehouse entries and bin content can be left without the production document needed for reconciliation. If it is too broad, users can be blocked from deleting an order after they have already moved the stock back through a supported return flow. The tests cover the new blocked deletion paths, but not the return/success path.

**Necessity:** The change is needed because the old confirmation allowed a released production order to be deleted after a registered pick but before consumption. The scope is focused on the right delete paths, but it needs to recognize when picked quantity is no longer outstanding.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11261 round=1 by=alexei-dobriansky at=2026-09-09T12:04:44Z lastSha=333c7894b10f213f2778ddf4519119ec00e5bbc3 reviewKey=e3047268b5b4aa92d91b706ad36997b9c27ca77a62173b3583bef9dc5da9212c suggestions=S1@d50b13ad

PR #11277, opened 2026-09-09
Bug 649001: [29x][NAV] Track Translation update (AUTODETECTED)

BardurKnudsen:
Remove validation that causes failures if GLSetup isn't inserted yet. (translation builds)

Fixes [AB#649001](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/649001)

### DIFF
--- a/src/Layers/APAC/DemoTool/CreateCurrency.Codeunit.al
+++ b/src/Layers/APAC/DemoTool/CreateCurrency.Codeunit.al
@@ -218,7 +218,6 @@ codeunit 101004 "Create Currency"
         Currency.Validate("EMU Currency", CurrencyData."EMU Currency");
         Currency.Validate("Amount Decimal Places", CurrencyData."Amount Decimal Places");
         Currency.Validate("Unit-Amount Decimal Places", CurrencyData."Unit-Amount Decimal Places");
-        Currency.Validate(Symbol, Currency.ResolveCurrencySymbol(Currency.Code));
         Currency.Insert(true);
     end;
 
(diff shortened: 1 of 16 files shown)


mirouhl (approved):

JesperSchulz (approved):

alexei-dobriansky:

<details>

This removes `Currency.Symbol` validation from legacy DemoTool currency creation in the changed country layers. It avoids the early `General Ledger Setup` lookup during translated demo package creation, so currency rows can be inserted before setup exists.

The failing path is resolved because `Currency.OnValidate(Symbol)` no longer calls `CheckLCYSymbol()`. The tradeoff is that these demo currencies no longer get their automatic symbols, even though symbol resolution itself does not need General Ledger Setup; only the LCY-symbol warning check needs that setup record.
</details>

<details>

**Fit:** Partial

The change addresses the reported build failure in the affected DemoTool path. It also removes useful currency-symbol initialization, so the fix is broader than the failing check that needed to be made safe.
</details>

<details>

**S1** (🟠 Moderate): *Keep currency symbols while avoiding setup failure*
The change stops the setup error, but it also stops demo currencies from getting their normal symbols. Consider making `CheckLCYSymbol()` return when General Ledger Setup is not present, then keep the `Symbol` validation in the demo currency setup.
</details>

<details>

**Risk:** The touched path creates demo currency setup data, not posting entries. The main regression risk is visible demo data: standard currency symbols are no longer stored for those currencies. There is no public API or event-subscriber contract change.

**Necessity:** A change is needed because the setup flow can fail before demo-data packages are produced. The scope is small, but a narrower fix in the LCY-symbol warning check would preserve the symbol data while still allowing early setup.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11277 round=1 by=alexei-dobriansky at=2026-09-09T19:05:26Z lastSha=89ef49caaaae79f44000b7cdd41617a656aa421a reviewKey=7876fa22eaa34b59b6358abf1dd9b6c922922062e712d3037d2759d5f3584442 suggestions=S1@7edf6001

PR #11303, opened 2026-09-10
[Master]-Purchase invoice lines of type G/L Account where enabling 'Prices Including VAT' and changing the VAT Prod. Posting Group causes recalculation of the Direct Unit Cost Incl. VAT and Line Amount - Related to Bug  632380

neeleshsinghal:


alexei-dobriansky:

<details>

The change stops purchase G/L Account lines from recalculating Direct Unit Cost when the VAT Product Posting Group changes with Prices Including VAT enabled. It also updates the W1 and ES regression tests to expect Direct Unit Cost and Line Amount to stay fixed.

This matches the invoice-amount preservation part of the reported behavior and aligns purchase with the sales and service line pattern, where VAT changes only rescale item and resource prices. However, the purchase validation now skips the only default path that recalculates Unit Cost after the VAT % changes, so the fix does not fully update the cost side of the line.
</details>

<details>

**Fit:** Partial

The reported scenario is clear: changing VAT on a purchase G/L Account line should not change the supplier invoice amount, but the VAT distribution and unit cost still need to match the new VAT setup. The diff preserves Direct Unit Cost and Line Amount, but it does not prove or implement the required Unit Cost recalculation.
</details>

<details>

**S1** (🔴 High): *Unit cost keeps the old VAT rate*
After this branch skips `Validate("Direct Unit Cost")` for G/L Account lines, `UpdateAmounts()` recalculates the amounts but does not call `UpdateUnitCost()`. Recalculate the unit cost after the VAT % changes, and assert `Unit Cost` or `Unit Cost (LCY)` in the regression test, so posting and related checks do not keep the old VAT rate.
</details>

<details>

**Risk:** The regression surface is purchase invoice G/L Account lines with Prices Including VAT across all modified application layers. The public event remains in place, but the default flow can leave cost fields calculated from the previous VAT %, which is a financial/data-integrity risk for posting-adjacent logic and checks that use Unit Cost.

**Necessity:** The change is needed because the supplier invoice amount should stay stable when the VAT setup changes for this kind of expense line. The scope is small and well targeted, but it must also update the unit cost side to fully solve the scenario.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11303 round=1 by=alexei-dobriansky at=2026-09-10T12:18:20Z lastSha=059a8e382c78ab58bcc0839305bf95817635957d reviewKey=7a06bc920e892ed41df97b60a30f24f2fb4b3502251770f7ceffeb21c933119b suggestions=S1@df061cdb

PR #11321, opened 2026-09-10
Update default Pricing implementation (v.30)

Alexander-Ya:
## Linked work

Fixes #

## How I validated this

### DIFF
--- a/src/Layers/W1/BaseApp/Pricing/Calculation/PriceCalculationMgt.Codeunit.al
+++ b/src/Layers/W1/BaseApp/Pricing/Calculation/PriceCalculationMgt.Codeunit.al
@@ -110,11 +110,7 @@ codeunit 7001 "Price Calculation Mgt."
     begin
         if not IsExtendedPriceCalculationEnabled() then begin
             PriceCalculationSetup.Method := PriceCalculationSetup.Method::"Lowest Price";
-#if not CLEAN27
             PriceCalculationSetup.Implementation := PriceCalculationSetup.Implementation::"Business Central (Version 15.0)";
-#else
-            PriceCalculationSetup.Implementation := PriceCalculationSetup.Implementation::"Business Central (Version 16.0)";
-#endif
             exit(true);
         end;
 


alexei-dobriansky:

<details>

This PR removes the `CLEAN27` conditional from `Price Calculation Mgt.` when extended price calculation is disabled. In a clean v30 build, `CLEAN27` is already defined, so the current code uses the Version 16.0 pricing implementation; after this change, the same fallback always uses Version 15.0.

That is a behavior change in the price calculation engine, not only a preprocessor cleanup. The changed code does not update related setup logic or tests to show that moving the fallback back to the legacy implementation is intended and safe.
</details>

<details>

**Fit:** Unclear

The requested behavior is not clear enough to show why the fallback should change from Version 16.0 to Version 15.0 for clean v30 builds. The code change may be trying to clean old preprocessor symbols, but it keeps the branch that was not used by the clean build.
</details>

<details>

**S1** (🔴 High): *Keep the clean pricing default*
The clean v30 build already defines `CLEAN27`, so this path used `Business Central (Version 16.0)` before the cleanup. The PR removes that branch and leaves `Version 15.0`, which can change prices and discounts when the fallback path is used. Keep the Version 16.0 assignment unless this rollback is intentional and covered.
</details>

<details>

**Risk:** This is a pricing calculation path used by many sales, purchase, service, job, item, and journal callers through `Price Calculation Mgt.GetHandler`. Picking the wrong implementation can change price or discount results and can also change behavior for existing integrations that rely on the current clean-build default.

**Necessity:** The need for a v30 cleanup is reasonable, but this specific fallback change is not justified by the current code. The safer scope is to preserve the clean-build behavior while removing the obsolete preprocessor wrapper.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11321 round=1 by=alexei-dobriansky at=2026-09-10T12:19:35Z lastSha=76c6c10cebd01ed2ec7cddf4066e62652807e9d2 reviewKey=6981437ec184c57f267dc990478372eb19f07d8c5c21040388dca01380894c7f suggestions=S1@a4b8dd44

PR #11331, opened 2026-09-10
[29.0] - Bug 649843: Composite layout stress-test fixes, part resolution source and status action captions

v-mortenr:
Fixes [AB#649843](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/649843)

### Problem

Three defects in the composite layout feature, found by stress testing, plus one gap the same testing exposed. All present in `releases/29.0`:

- A duplicate part name silently overwrote an existing layout, because `InsertNewLayout` ended with an unconditional `Get` + `Delete(true)` of a same-named layout. Parts created from Report Theme and Header/Footer never reached the name check in the New Layout dialog.
- A layout could be moved out of Approved while a report was using it as its chosen default, and a non-approved layout could be made the default.
- An empty upload produced a platform message about an empty `InStream` object instead of a usable error.
- Header/footer parts could not be created blank, so the only way to a new part was exporting an existing one and stripping it.
- `Show layout parts` passed a blank layout name to `GetResolvedPartDisplays`, so it reported the report-default rows as `This layout` and never consulted a layout-scoped assignment. The message could contradict the Report Layout FactBox for the same layout.
- The status actions read `Set Approved` / `Set Draft` / `Set Pending Approval` / `Set Retired`.

### Changes

- `InsertNewLayout` takes a `ReplaceExisting` intent: New rejects a taken name before the file dialog opens, Replace still overwrites (after a successful upload) and carries the layout's status forward.
- `SetLayoutStatus` refuses to take a chosen default out of Approved, and reports the metadata-default fallback case instead of blocking it; the default-selection path validates that a layout is approved, from both routes.
- Uploads are buffered through a `Temp Blob` so an empty file is rejected with its own error, on the generated path too.
- `NewWordLayout` injects the report's `WordXmlPart` so a blank header/footer part can be created, exported and returned with Replace.
- `Show layout parts` resolves the report's effective layout through `GetDefaultReportLayoutSelection` and passes that layout's `CompositeLayoutKey`.
- Status action captions become `Set as ...` on both pages that expose them.

### Backport

Backport of #11211 to `releases/29.0`.

Cherry-picked commit(s):

- 4979d3868b5a06d31d3345479b7a955fecbca571

Covers Bug 649379, Bug 648842 and Bug 647373 — main squashed all three into the one commit above.

The five Base Application files applied unchanged (verified line by line against the main commit). One conflict, in `src/Layers/W1/Tests/Report/CompositeLayoutTests.Codeunit.al`: `releases/29.0` does not have the shipped-parts seeding tests, because codeunit `"Composite Report Parts Mgt."` only exists in main, so the file here is ~557 lines shorter and the conflict was main-only context rather than a real overlap. The commit's 332 added test lines (3 labels, 8 tests, 1 notification handler, 1 cleanup call, 8 helpers) were applied to the `29.0` version of the file and verified byte-identical to the main commit's additions; none of them reference main-only objects. No hand edits beyond placement.

### Validation

- [x] Diff reviewed against the main commit: 520 insertions, 18 deletions, same as main.
- [x] No main-only dependencies in the backported test code.
- [x] Built locally against 29.0 symbols: Base Application and Tests-Report both compile with **0 errors**, and the warning set is identical to a pristine `releases/29.0` build (647 in Base Application, 13 in Tests-Report — no new, none removed).

Build detail: `alc` 18.0.41.7630 from the enlistment (`Run/ALDevEnv/bin`), platform symbols `System 29.0.54137.0`, System Application and Business Foundation 29.0 from the local package cache. Tests-Report was compiled against the Base Application produced from this branch, not the shipped one.

### Risk & compatibility

Same intended behavioural changes as #11211: a duplicate part name is an error rather than a silent overwrite; a chosen default layout cannot be taken out of Approved in one step (a report's built-in fallback layout still can, with a message); a non-approved layout cannot be made the default; batch status changes are validated before anything is written. No schema, permission or telemetry changes.

### DIFF
--- a/src/Layers/W1/BaseApp/Foundation/Reporting/NewReportThemeHeaderFooter.Page.al
+++ b/src/Layers/W1/BaseApp/Foundation/Reporting/NewReportThemeHeaderFooter.Page.al
@@ -35,6 +35,13 @@ page 9668 "New Report Theme Header/Footer"
                 Caption = 'Description';
                 ToolTip = 'Specifies a description for the new theme or header/footer.';
             }
+            field(CreateEmptyLayoutField; CreateEmptyLayout)
+            {
+                ApplicationArea = Basic, Suite;
+                Caption = 'Create a Blank Layout';
+                Visible = IsHeaderFooter and not EditMode;
+                ToolTip = 'Specifies whether to create a blank header/footer that you can export and edit, instead of uploading a file.';
+            }
         }
     }
 
@@ -50,6 +57,8 @@ page 9668 "New Report Theme Header/Footer"
             DialogCaption := NewThemeCaptionLbl
         else
             DialogCaption := NewHeaderFooterCaptionLbl;
+
+        IsHeaderFooter := NewSubtype = Enum::"Report Layout Subtype"::HeaderFooter;
     end;
 
     internal procedure GetPartName(): Text[250]
@@ -60,6 +69,11 @@ page 9668 "New Report Theme Header/Footer"
     internal procedure GetPartDescription(): Text[250]
     begin
         exit(PartDescription);
+    end;
+
+    internal procedure GetCreateEmptyLayout(): Boolean
+    begin
+        exit(CreateEmptyLayout);
     end;
 
     /// <summary>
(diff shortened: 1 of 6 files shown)


PredragMaricic (approved):

PredragMaricic (approved):
No blocking findings. I verified that the five production files match the merged upstream implementation, including the final cross-company default-layout protection added during upstream review. The backport keeps the release-specific difference confined to test placement and adds regression coverage for effective layout-part resolution and status changes across current-company defaults, other-company defaults, stale company rows, and company-scoped overrides. The checks are green and the reported 29.0 compilation completed without errors.

PredragMaricic (approved):
No blocking findings. I verified that the five production files match the merged upstream implementation, including the final cross-company default-layout protection added during upstream review. The backport keeps the release-specific difference confined to test placement and adds regression coverage for effective layout-part resolution and status changes across current-company defaults, other-company defaults, stale company rows, and company-scoped overrides. The checks are green and the reported 29.0 compilation completed without errors.

darjoo (approved):

PR #11366, opened 2026-09-11
Filter top customer and top vendor Excel reports on the posting group

Explorer986:
## What & why

The Top Customer and Top Vendor Excel reports returned nothing when you filtered them on a posting group. Both balance queries filtered Detailed Cust. Ledg. Entry."Posting Group" and Detailed Vendor Ledg. Entry."Posting Group", which came in with v20 and are only written at posting time. Nothing backfills them on upgrade and cloud migration does not bring them across, so on many tenants they are blank. The ranking queries next to them filter the ledger entry posting group, which is always populated, so the two passes could also disagree about which accounts belong to the group. Fix is to join the detailed entry to its parent ledger entry and read the posting group from there.

In the suggested approach collecting the matching customers first and setting them as a number filter. I did not go that way. An empty match gives an empty filter string and SetFilter then clears the filter instead of matching nothing.

Fixes [AB#649289](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/649289)

## How I validated this

## Risk & compatibility

No breaking changes. Filter element names are unchanged so Transfer Filters is unchnaged. The join is on a mandatory key field and is many to one, so no rows are dropped and the Sum is unaffected.


alexei-dobriansky:

<details>

This change makes the customer and vendor top-list balance queries filter posting groups through the parent ledger entries instead of the detailed ledger entries. The query filter element names stay the same, so the reports still transfer filters through the existing names, and the one-to-one entry-number join matches the sales and purchase ranking queries. The production change fits the reported migrated-data case, but one new vendor test uses signs that do not match the query output.
</details>

<details>

**Fit:** Strong

The problem is clear: posting group filters can miss valid customer and vendor balances when detailed ledger entries do not carry posting group values. The diff addresses that root cause by reading the posting group from the parent ledger entry for both balance paths while keeping the report filter contract unchanged.
</details>

<details>

**S1** (🔴 High): *Fix the vendor test signs*
The vendor balance and purchase queries both have `ReverseSign = true`, and the report copies those query values directly into `AmountLCY` and `Amount2LCY`. With the current test data, the report returns the opposite signs from these assertions. Please use vendor ledger source signs that match real data, or update the expected values so the tests match the query output.
</details>

<details>

**Risk:** The app change is narrow to the two top-list balance queries and preserves the public query filter element names. The main regression risk is incorrect financial report output or a failing test path around vendor signs, because both vendor queries reverse signs before the report buffer is filled.

**Necessity:** The change is needed because upgraded or migrated data can have blank posting group values on detailed ledger entries. Joining to the parent ledger entry avoids relying on a field that may not be populated and keeps the balance pass aligned with the sales and purchase pass.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11366 round=1 by=alexei-dobriansky at=2026-09-11T09:35:45Z lastSha=2adb93b77d36e175c925692eb0a3e24bc5f24a85 reviewKey=61837aa70a49296da11a91276ba5395317edae3edcd70b3e8341ac3c9c594a90 suggestions=S1@17c3f1b1

alexei-dobriansky (dismissed):
Please take a look if the suggestion needs addressing.

alexei-dobriansky:

<details>

This update keeps the customer and vendor top-list balance filters on the same public filter names, but resolves posting groups through the parent ledger entries instead of the detailed ledger entries. That matches the reported migrated-data case and aligns the balance pass with the sales and purchase ranking pass. The vendor sign issue from the previous round is fixed, but the new test helpers now reuse entry numbers after `Init()` clears the record.
</details>

<details>

| S1 | Fix the vendor test signs | Addressed | The vendor test data now uses negative vendor ledger source amounts, which matches the `ReverseSign = true` query output. |
</details>

<details>

**S2** (🔴 High): *Keep entry numbers before Init*
The new helpers call `FindLast()`, then `Init()`, and then add 1 to `"Entry No."`. `Init()` clears the record, so the second customer or vendor inserted in each test tries to use entry number 1 again. Please store the last entry number before `Init()`, as the existing helper does, and use that saved value for both ledger and detailed ledger entries.
</details>

<details>

**Risk:** The production query change is narrow and keeps the existing report filter names. The remaining risk is in the new tests: they can fail before they validate the report because repeated helper calls reuse the same ledger entry numbers.

**Necessity:** The change is needed because migrated data can have blank posting group values on detailed ledger entries. Joining to the parent ledger entry is the right scope for the reported filter problem.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11366 round=2 by=alexei-dobriansky at=2026-09-11T12:34:50Z lastSha=83f04d2c35503bce40184185604e2d1f5d7ee705 reviewKey=ef01157d32ad6e4020affe0eeda770e2e04bcd694e52d68af9c8119bbc72aec7 suggestions=S1@17c3f1b1:addressed,S2@bb0cdebc:new parentRound=1

mynjj (approved):

alexei-dobriansky:

# **Recommendation: Accept**

<details>

This update keeps the customer and vendor top-list balance filter names unchanged, but reads the posting group through the parent ledger entries instead of the detailed ledger entries. That matches the reported upgraded and migrated data case where the detailed-entry posting group can be blank, and it aligns the balance pass with the sales and purchase ranking pass. The added tests cover both customer and vendor filtering with blank detailed-entry posting groups, and the prior entry-number concern does not hold because `Init()` does not clear primary key fields.
</details>

<details>

**Fit:** Strong

The reported problem is concrete: posting group filters can skip valid top customer or top vendor balances when detailed ledger entries do not carry the posting group. The diff addresses that root cause directly for both balance queries while preserving the report filter contract.
</details>

<details>

| S1 | Fix the vendor test signs | Addressed | The vendor test data now uses negative source amounts, so the reversed query output matches the expected positive balance. |
| S2 | Keep entry numbers before Init | Disputed | Rechecking the AL semantics shows that `Init()` does not clear primary key fields, so the value found by `FindLast()` is still used when adding one to the entry number. |
</details>

<details>

*None - no new blocking or non-blocking findings survived consensus.*
</details>

<details>

**Risk:** The production change is narrow to two ExcelReports balance queries and keeps the public query filter element names unchanged. The join uses the existing detailed-entry to ledger-entry number link and only changes where the posting group filter is evaluated, so the main regression surface is top-list report selection, not posting or persisted data.

**Necessity:** The change is needed because upgraded or migrated data can have blank posting group values on detailed ledger entries. Reading the posting group from the parent ledger entry is the right scope and keeps the balance filter consistent with the related sales and purchase top-list queries.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11366 round=3 by=alexei-dobriansky at=2026-09-14T18:04:13Z lastSha=9e7f91a080e8511a7e914198c305ca09e3a391d7 reviewKey=4e26b2a4ace24344f4713f7087cfdb80a38eeb2d99417a41c452e7b7e5cbc1ae suggestions=S1@17c3f1b1:addressed,S2@bb0cdebc:disputed parentRound=2

dcenic (approved):

PR #11374, opened 2026-09-11
Re-enable AutoAcceptICGenJnlTransactionUT: add Report 511 handler

Explorer986:
## What & why

Fixes test failure regression from PR #9038. When Page 615 licensing tier changed to IsBasicOnlyEnabled, Report 511 now runs during IC Accept flow and commits mid-test, causing Gen. Journal Line assertion to fail.

Solution adds RequestPageHandler to test, ensuring Report 511's request page is properly handled. Test isolation restored without modifying IC flow logic.

Fixes [AB#642679](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/642679)

## How I validated this

## Risk & compatibility

None. Test-only change using existing handler pattern. No IC flow modifications, no public API changes, no data impact.

### DIFF
--- a/src/Layers/W1/Tests/ERM/ERMIntercompanyIII.Codeunit.al
+++ b/src/Layers/W1/Tests/ERM/ERMIntercompanyIII.Codeunit.al
@@ -2265,6 +2265,7 @@ codeunit 134154 "ERM Intercompany III"
     end;
 
     [Test]
+    [HandlerFunctions('RequestPageHandler')]
     procedure AutoAcceptICGenJnlTransactionUT()
     var
         ICSetup: Record "IC Setup";


alexei-dobriansky:

<details>

This PR tries to restore an intercompany test by adding a request page handler for Report 511. The changed test still calls the report with the request window set to `false`, so the request page handler is not expected to run. The fix also updates only the W1 test copy while the synced layer copies still contain the old test.
</details>

<details>

**Fit:** Partial

The reported problem is clear, and the change targets the affected test. The implementation does not fully fit the execution path because the added handler is not reached by the current report call, and the parallel layer copies are not updated.
</details>

<details>

**S1** (🔴 High): *Handler is not reached*
The test runs Report 511 with the request window set to `false`, so this request page handler is not expected to run. Since listed handlers must run at least once, this can still leave the test failing instead of re-enabling it.

**S2** (🟠 Moderate): *Layer copies are not updated*
Apply the same test change to the CH, CZ, and NL copies, or run the propagation step. The required sync check reports these files are missing from the change.
</details>

<details>

**Risk:** Production risk is low because the PR changes test code only. Merge risk is high for the test suite because the changed handler is not reached by the current Report.Run call and the layer copies are incomplete.

**Necessity:** The disabled test should be restored, so the change is needed. The scope is too small as submitted because it does not cover all synced test copies and does not prove the handler is called.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11374 round=1 by=alexei-dobriansky at=2026-09-11T12:34:56Z lastSha=69554a5f46f2e7a07d09955dcdd84bed8209896a reviewKey=1dcd096104778ed177e53b43ebf4b7d8c34dcf2ec0d130d88fbb2910e73d02d7 suggestions=S1@73b3d9e9,S2@45bee23c

PR #11376, opened 2026-09-11
Export structured AI eval run data

nikolakukrika:
## Summary
- export a complete AI Eval run package from the command-line runner, one raw JSON file per action invocation
- write `results.json` as the analysis entry point and one self-contained evaluation result per AI Eval log entry
- include input, output, status, metrics, errors, dataset identifiers, and references to optional Agent Task troubleshooting details
- reuse the permission-protected System Application Agent Task export for detailed task, log, page snapshot, context, and memory data
- add `Invoke-AITTests -ExportAITRunData` with an optional `-AITRunDataFolder`; the default is `AITRunData` next to the runner module
- preserve persisted dataset group codes while removing known dataset extensions only from generated folder names
- percent-escape unsafe path characters and reject paths outside the selected output root or current suite/version
- delete only an existing export of the same suite version before rewriting it; preserve all other versions
- reuse the active test runner client context and form for export actions
- stop the total export loop after 20 minutes and write diagnostics for invalid, blank, timed-out, or failed exports without failing the test run

## Export structure
```text
<suite>/
  version-<version>/
    results.json
    <dataset-group-without-extension>-<test-input-code>/
      evaluation-result-<AI-Eval-log-ID>.json
      agent-task-details/
        task-<Agent-Task-ID>.json
```

`results.json` is intended for the first analysis pass. Agents only need to open `agent-task-details` for failures, uncertainty, or deeper causal analysis.

## Validation
- PowerShell module syntax parsing passes
- mocked export verified hierarchy creation, same-version cleanup, preservation of prior versions, raw UTF-8 writes without BOM, sentinel completion, invalid/blank-response diagnostics, cross-run path rejection, and caught action failures
- BCQuality AL review completed; its no-data finding was addressed with an `IsEmpty` guard
- `git diff --check` passes
- AI Test Toolkit compilation reaches dependency resolution; the available package cache lacks the new `Agent Task Log Export` symbol and a compatible `Temp Blob` symbol, and reports no diagnostics in the new run-data serializer before those dependency errors

## Notes
- the `Eng` files are copies of the NAV ALTestRunner modules for review and will be moved to the NAV repository separately

[AB#624671](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/624671)

### DIFF
--- a/src/Tools/AI Test Toolkit/src/AITCommandLineCard.Page.al
+++ b/src/Tools/AI Test Toolkit/src/AITCommandLineCard.Page.al
@@ -43,7 +43,7 @@ page 149042 "AIT CommandLine Card"
                         if AITCode <> xRec."Test Suite Code" then
                             Clear(LineNoFilter);
 
-                        Clear(AgentTaskLogs);
+                        ResetAgentTaskLogExport();
                         UpdateAITestMethodLines();
                     end;
                 }
@@ -77,7 +77,7 @@ page 149042 "AIT CommandLine Card"
 
                     trigger OnValidate()
                     begin
-                        Clear(AgentTaskLogs);
+                        ResetAgentTaskLogExport();
                         UpdateAITestMethodLines();
                     end;
                 }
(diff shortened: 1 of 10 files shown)


src/Tools/AI Test Toolkit/src/AITCommandLineCard.Page.al
```diff
@@ -363,4 +392,26 @@ page 149042 "AIT CommandLine Card"
...
+        NoLogEntriesFoundErr: Label 'No eval log entries were found for eval suite %1.', Comment = '%1 = Eval Suite Code';
+    begin
+        VerifyTestSuiteCode();
+        Clear(AgentTaskLogs);
+        AITLogEntry.SetRange("Test Suite Code", AITCode);
+        AITLogEntry.SetCurrentKey(Version);
+        if not AITLogEntry.FindLast() then
+            Error(NoLogEntriesFoundErr, AITCode);
+
+        Version := AITLogEntry.Version;
```
nikolakukrika: Lets not use built in keywords. Version of what?

PR #11383, opened 2026-09-11
[SOA] Fix "The Order Promising Setup does not exist." crash when setting quote quantity

VolodySyn:
## What & why

The Sales Order Agent could fail with **"The Order Promising Setup does not exist."**
when it entered or updated a quantity on a sales quote line.

`SOA Item Search`'s `OnAfterCheckItemAvailable` subscriber calls
`SOAShipmentDateMgt.Run()` as a bare statement. When **Include capable to promise**
is enabled but the company has no `Order Promising Setup` record,
`Capable to Promise.CalcCapableToPromiseDate` raises that error, and because the
`Run()` return value isn't checked the error propagates to the user. Customers who
don't use order promising hit this, and it was confirmed recurring in telemetry.

This change:
1. **Fixes the crash** — the capable-to-promise calculation is skipped when
   `Order Promising Setup` doesn't exist. The agent still sends the normal
   "inventory is lower than the entered quantity" notification instead of erroring.
   This resolves the crash for agents that are already configured this way.
2. **Prevents the misconfiguration** — enabling the *Include capable to promise*
   toggle on the SOA Setup page now checks for the setup and offers to create it;
   the toggle reverts if the user declines or lacks permission.

Both parts are needed: the affected customers already have the toggle on without the
setup, so the setup-page check alone would not stop their crash — the runtime guard does.

**Changed objects**
- `src/ItemSearch/SOAItemSearch.Codeunit.al` (codeunit 4591) - guard CTP call on
  `OrderPromisingSetupExists()` + helper.
- `src/Setup/SOASetup.Page.al` (page 4400) — `Include capable to promise` `OnValidate`
  confirmation, `VerifyOrderPromisingSetupExists()` helper, prompt label, and
  `using Microsoft.Inventory.Availability`.

## Linked work
Fixes [AB#622246](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/622246)

## How I validated this

- No AL compiler diagnostics in either changed file.
- Reviewed existing CTP accuracy tests: they set `"Incl. Capable to Promise"` directly
  on the `SOA Setup` record (via `SOATestProcessLibrary`), bypassing the page trigger,
  and run with `Order Promising Setup` present, so neither change affects them.
- Still to do before merge: container repro — enable *Include capable to promise*, remove
  the `Order Promising Setup` record, have the agent create/update a quote line, and confirm
  the availability notification appears with no error; and verify the setup-page prompt
  creates the record / reverts on cancel.

## Risk & compatibility

- Behavior changes **only** when `Order Promising Setup` is missing: the agent now degrades
  gracefully (skips the earliest-ship-date line) instead of erroring. When the setup exists,
  behavior is unchanged.
- New confirmation dialog on the SOA Setup page when enabling the toggle without the setup.
- `Order Promising Setup` is read (`IsEmpty`) from the agent session (codeunit 4591) and the
  admin session (page 4400); both flows already access this table.
- No schema/table changes, no upgrade or data impact, no new permissions, no telemetry changes.

### DIFF
--- a/src/Apps/W1/SalesOrderAgent/app/src/ItemSearch/SOAItemSearch.Codeunit.al
+++ b/src/Apps/W1/SalesOrderAgent/app/src/ItemSearch/SOAItemSearch.Codeunit.al
@@ -161,7 +161,8 @@ codeunit 4591 "SOA Item Search"
 
         Msg := StrSubstNo(NotificationMsg, Item.Description);
 
-        if SOASetup."Incl. Capable to Promise" then begin
+        // Capable to Promise reads the Order Promising Setup and errors if it is missing, so skip it when the setup is not configured.
+        if SOASetup."Incl. Capable to Promise" and OrderPromisingSetupExists() then begin
             SOAShipmentDateMgt.SetParamenters(Item."No.", SalesLine."Variant Code", SalesLine."Location Code", SalesLine."Unit of Measure Code", SalesLine."Shipment Date", SalesLine.Quantity);
             SOAShipmentDateMgt.Run();
             if SOAShipmentDateMgt.GetEarliestShipmentDate() <= SalesLine."Shipment Date" then
@@ -174,6 +175,13 @@ codeunit 4591 "SOA Item Search"
         QuoteAvailabilityCheckNotification.Message(Msg);
         QuoteAvailabilityCheckNotification.Scope(NotificationScope::LocalScope);
         NotificationLifecycleMgt.SendNotificationWithAdditionalContext(QuoteAvailabilityCheckNotification, SalesLine.RecordId, GetQuoteItemAvailabilityNotificationId());
+    end;
+
+    local procedure OrderPromisingSetupExists(): Boolean
+    var
+        OrderPromisingSetup: Record "Order Promising Setup";
+    begin
+        exit(not OrderPromisingSetup.IsEmpty());
     end;
 
     local procedure GetQuoteItemAvailabilityNotificationId(): Guid
--- a/src/Apps/W1/SalesOrderAgent/app/src/Setup/SOASetup.Page.al
+++ b/src/Apps/W1/SalesOrderAgent/app/src/Setup/SOASetup.Page.al
@@ -8,6 +8,7 @@
 
 namespace Microsoft.Agent.SalesOrderAgent;
 
+using Microsoft.Inventory.Availability;
 using System.Agents;
 using System.AI;
 using System.Email;
@@ -258,6 +259,8 @@ page 4400 "SOA Setup"
                             Editable = OnlyAvailableItemsActive;
                             trigger OnValidate()
                             begin
+                                if Rec."Incl. Capable to Promise" then
+                                    VerifyOrderPromisingSetupExists();
                                 ConfigUpdated();
                             end;
                         }
@@ -688,6 +691,21 @@ page 4400 "SOA Setup"
             TempAgentSetupBuffer.State := TempAgentSetupBuffer.State::Enabled;
     end;
 
+    local procedure VerifyOrderPromisingSetupExists()
+    var
+        OrderPromisingSetup: Record "Order Promising Setup";
+    begin
+        if not OrderPromisingSetup.IsEmpty() then
+            exit;
+
+        // Capable to Promise requires the Order Promising Setup; offer to create it so the agent does not fail later.
+        if Confirm(OrderPromisingSetupMissingQst, false) then
+            Page.RunModal(Page::"Order Promising Setup");
+
+        if OrderPromisingSetup.IsEmpty() then
+            Rec."Incl. Capable to Promise" := false;
+    end;
+
     local procedure EnabledAgentFirstConfig(): Boolean
     begin
         exit((TempAgentSetupBuffer.State = TempAgentSetupBuffer.State::Disabled) and IsFirstConfig() and CheckIsValidConfig());
@@ -868,4 +886,5 @@ page 4400 "SOA Setup"
         InboxFolderIdTok: Label 'inbox', Locked = true;
         NoFolderSelectedInboxWarningQst: Label 'There is no mail folder selected, so the agent will process emails from the inbox (%1 emails since %2). Do you want to continue?', Comment = '%1=email count, %2=start date';
         AgentArchivedNotificationMsg: Label 'This agent is archived, so its settings are read-only. Its tasks and logs remain available for auditing.';
+        OrderPromisingSetupMissingQst: Label 'The Order Promising Setup does not exist and is required to include capable-to-promise items. Do you want to create it now?';
 }


alexei-dobriansky:

<details>

The change skips capable-to-promise calculation in quote line availability when the order promising setup record is missing, and it warns users when they enable the capable-to-promise option without that setup. The BaseApp check confirms the calculation reads `Order Promising Setup` and fails before calculating dates when the setup is missing.

The missing-row guard fixes one crash path, but it does not cover setup records that exist without the required fields. Opening the setup page creates an empty record, and the calculation later requires the order promising template and worksheet, so the same quote-line flow can still stop with an error.
</details>

<details>

**Fit:** Partial

The reported crash is valid and the change addresses the exact missing-record case. It does not fully protect the capable-to-promise path because an empty or incomplete setup record is treated as valid.
</details>

<details>

**S1** (🔴 High): *Incomplete setup still reaches CTP*
Check that `Order Promising Setup` is fully configured before calling `SOAShipmentDateMgt.Run()`, not only that a row exists. The setup page creates an empty setup row on open, and `Capable to Promise` later requires the template and worksheet fields, so this can still error when the option is enabled.

**S2** (🟠 Moderate): *Missing regression for setup readiness*
Add a regression test for enabling capable-to-promise when the setup is missing or incomplete. The test should prove quote line availability does not stop with an error and the option is not left enabled for an unusable setup.
</details>

<details>

**Risk:** The changed runtime path is limited to Sales Order Agent quote availability checks and the setup page. The compatibility risk is low because no public signatures or schema changed, but the current guard can still leave users with a quote-line error after the setup page has created an incomplete setup record.

**Necessity:** The change is needed because a configured agent can hit this error while updating quote quantities. The scope is right, but the readiness check needs to match the fields that the capable-to-promise calculation actually requires.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11383 round=1 by=alexei-dobriansky at=2026-09-11T18:11:03.7543286Z lastSha=8e033bb60dc53d18899793763581f93c203dec68 reviewKey=29ea786005156c1597849946f640c6851887a9b6ef9ade151dc7ab4d46b68d21 suggestions=S1@ccb8b993,S2@82685b30

src/Apps/W1/SalesOrderAgent/app/src/ItemSearch/SOAItemSearch.Codeunit.al
```diff
@@ -161,7 +161,8 @@ codeunit 4591 "SOA Item Search"
 
         Msg := StrSubstNo(NotificationMsg, Item.Description);
 
-        if SOASetup."Incl. Capable to Promise" then begin
+        // Capable to Promise reads the Order Promising Setup and errors if it is missing, so skip it when the setup is not configured.
+        if SOASetup."Incl. Capable to Promise" and OrderPromisingSetupExists() then begin
```
JakovljevicDusan: A bit of benefit would be to use short-circuit evaluation. Ignore if not relevant.

alexei-dobriansky (changes requested):
Please check if the suggestion is valid.

JakovljevicDusan (dismissed):

JakovljevicDusan (dismissed):

alexei-dobriansky:

<details>

The latest changes keep quote availability from running capable-to-promise unless the order promising setup record exists and has the required template and worksheet. The setup page also checks the same readiness before leaving the capable-to-promise option enabled.

This addresses the prior blocker. The capable-to-promise code tests those same two fields before it calculates dates, and this Sales Order Agent path passes its own document id, so the number series field is not needed for this path.
</details>

<details>

**Fit:** Strong

The reported crash is covered for both a missing setup row and an incomplete setup row. The runtime guard now skips the date calculation when the setup cannot support it, and the setup-page check helps stop the unusable option from being saved.
</details>

<details>

| S1 | Incomplete setup still reaches CTP | Addressed | The check now requires the setup record plus the required template and worksheet before capable-to-promise runs. |
| S2 | Missing regression for setup readiness | Not addressed | No regression test was added; existing tests still do not cover missing or incomplete setup through this flow. |
</details>

<details>

*None - the new commits address the prior correctness blocker. The remaining item is the non-blocking test gap listed above.*
</details>

<details>

**Risk:** The remaining regression risk is limited to Sales Order Agent quote availability and setup validation. There is no schema change or public signature change, and the runtime behavior changes only when capable-to-promise setup is missing or incomplete.

**Necessity:** The change is needed because a configured agent can hit this error while updating quote quantities. The scope is now appropriate because it protects the runtime path and the setup path without changing unrelated availability logic.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11383 round=2 by=alexei-dobriansky at=2026-09-14T12:15:12.4668475Z lastSha=b00f0f54a76d07e33345f56147d70d78df384b9c reviewKey=43496a6031287f692c02bce3016a622898062de1545ae7dc1c1829444be22861 suggestions=S1@ccb8b993:addressed,S2@82685b30:notaddressed parentRound=1

alexei-dobriansky:

<details>

The latest change only cleans up the setup-page confirmation text and keeps the same runtime behavior. The full change guards capable-to-promise calculation unless the required setup record, template, and worksheet are present, and it prompts when the option is enabled without a usable setup.

The runtime path now matches the called codeunit's setup requirements before it enters capable-to-promise. That covers the missing and incomplete setup cases, and the prompt path gives the user a chance to create the setup before saving the option.
</details>

<details>

**Fit:** Strong

The reported crash happens because quote availability can enter capable-to-promise when required setup is missing or incomplete. The code now blocks that path and also helps prevent saving a configuration that would fail later.
</details>

<details>

| S1 | Incomplete setup still reaches CTP | Addressed | The guard still requires the setup record plus the required template and worksheet before capable-to-promise runs. |
| S2 | Missing regression for setup readiness | Not addressed | No regression test was added; the latest commit only changes confirmation formatting. |
</details>

<details>

*None - the latest commit only adjusts the confirmation call and removes a leading space from the label.*
</details>

<details>

**Risk:** The latest text-formatting change has no runtime risk. The full change touches Sales Order Agent quote availability and setup validation, and it changes behavior only when capable-to-promise setup is missing or incomplete. There is no schema change and no public or event signature change.

**Necessity:** The change is needed because configured agents can otherwise error while updating quote quantities. The scope remains narrow because it protects the runtime path and the setup path without changing unrelated availability logic.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11383 round=3 by=alexei-dobriansky at=2026-09-14T18:16:10Z lastSha=48dceec49c71f8ba7200ca6cd5174b183abee98b reviewKey=a23b05c2ed7c84394fb71caf5a736ce9f9095ea1b48c0ec641023a0005065a39 suggestions=S1@ccb8b993:addressed,S2@82685b30:notaddressed parentRound=2

PredragMaricic (approved):
The runtime guard addresses the reported missing/incomplete Order Promising Setup failure, while the setup-page validation prevents new configurations from enabling capable-to-promise without its required template and worksheet. The fallback preserves the normal inventory notification, and the existing configured path remains unchanged.

JakovljevicDusan (approved):

alexei-dobriansky (approved):

PR #11400, opened 2026-09-12
[Shopify] Handle missing variant mapping in product action

onbuyuka:
## What & why

The **Show product in Shopify** action could remain enabled when an item retained a Shopify Product mapping but its linked Shopify Variant had been deleted. Invoking the action then surfaced a raw database filter error.

This change enables the action only when the item has a Shopify Variant linked through an enabled shop. The shared product navigation methods also show clear guidance if the variant mapping disappears between rendering the page and invoking the action. The same behavior is applied to the Item Card and Item List.

## Linked work

Fixes [AB#649996](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/649996)

## How I validated this

- Built the Shopify Connector app successfully through the AL MCP build workflow.
- Added regression coverage for the Item Card action, Item List action, and the clear error returned when product navigation finds no linked variant.
- Focused test execution was blocked locally because the AL MCP server restored the v30 Test Toolkit symbols into the application cache instead of the test project cache and returned a generic compilation failure without diagnostics.

## Risk & compatibility

Low. There are no schema or data changes. Items with valid Shopify Variant mappings retain the existing navigation behavior.

### DIFF
--- a/src/Apps/W1/Shopify/App/src/Products/Codeunits/ShpfySyncProducts.Codeunit.al
+++ b/src/Apps/W1/Shopify/App/src/Products/Codeunits/ShpfySyncProducts.Codeunit.al
@@ -47,6 +47,7 @@ codeunit 30185 "Shpfy Sync Products"
         DialogMsg: Label '#1#########################', Locked = true;
         SyncInformationProgressLbl: Label 'Synchronizing information with Shopify.', Comment = 'Shopify is a product name.';
         FinishSyncProgressLbl: Label 'Finishing up synchronization with Shopify.', Comment = 'Shopify is a product name.';
+        NoProductVariantErr: Label 'No Shopify product variant is linked to this item. Synchronize products with Shopify, refresh the page, and try again.';
 
     /// <summary> 
     /// Export Items To Shopify.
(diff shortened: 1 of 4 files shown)


alexei-dobriansky:

# **Recommendation: Accept**

<details>

This change treats a Shopify product as showable from item pages only when a linked Shopify Variant exists in an enabled shop. It also changes the shared product navigation methods to show a clear message if the variant mapping disappears before navigation.

The page guards now check the same variant mapping that the action later uses. If a stale click or direct caller reaches the shared methods with an empty filtered variant set, FindFirst and FindSet now raise the clear mapping message instead of the raw filter error. The added tests cover both page actions and the empty-variant overview path.
</details>

<details>

**Fit:** Strong

The bug is a stale product mapping with no usable variant. The diff targets that exact gap by disabling the page action for that state and by keeping a clear error as a fallback if the state changes after the page was rendered.
</details>

<details>

*None.*
</details>

<details>

**Risk:** Risk is low and localized to Shopify item pages and the shared product navigation methods. A bad guard could hide valid product navigation or block items mapped only by product, but the action now uses the same variant mapping that the navigation path needs. There are no schema changes and no public signature changes.

**Necessity:** The change is needed because a stale product mapping without a usable variant can leave the action enabled and then show a raw database filter error. The scope is narrow, and the added tests cover the Item Card action, the Item List action, and the clear error path when navigation starts from an empty variant set.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11400 round=1 by=alexei-dobriansky at=2026-09-12T22:12:41.6530170Z lastSha=92a8166ff8049aea408c1241679b43cf50e68901 reviewKey=cc51690f5e349b018db56c718c33054a01e174949a4905a63426b043731b5098 suggestions=

darjoo (approved):

JesperSchulz (approved):

PR #11440, opened 2026-09-14
Fix remaining Standard Sales Invoice visibility expression

PredragMaricic:
## Summary

- Harden the remaining `Table_ShipToAdress` visibility expression so it always returns a Boolean.
- Apply the correction consistently to the W1, ES, FR, NA, and NO Standard Sales Invoice RDLC layouts.
- Add regression coverage that posts an invoice for a customer without name or address information and renders the RDLC layout as PDF.

## Root cause

The earlier fix hardened one scoped visibility expression, but an adjacent tablix member retained the original untyped `First(...) = First(...)` comparison. XML dataset tests do not evaluate RDLC visibility expressions, so the regression test renders the report as PDF.

ADO bug: [AB#592271](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/592271)


alexei-dobriansky:

# **Recommendation: Accept**

<details>

This change makes the remaining Ship-to tablix visibility check in each Standard Sales Invoice RDLC layout return an explicit Boolean. The expression now uses the same scoped and string-normalized comparison as the adjacent row group, so empty customer fields no longer leave the renderer with an invalid Hidden value. The added PDF render tests exercise the path that evaluates RDLC visibility expressions.
</details>

<details>

**Fit:** Strong

The reported failure is about the Standard Sales Invoice PDF render path, and the diff changes the exact Hidden expression that could still return an invalid type. The matching tests render the report as PDF with shipment information on all layers that carry the report test, so the fix is well aligned.
</details>

<details>

*None.*
</details>

<details>

**Risk:** The regression surface is the Ship-to section visibility in the Standard Sales Invoice report. The changed BaseApp layouts are the complete set of Standard Sales Invoice RDLC layouts in this repository, the RDLC files still parse as XML, and there is no posting, data, or public API change.

**Necessity:** The change is necessary because XML dataset checks do not evaluate RDLC Hidden expressions. A PDF-render test is an appropriate guard for this renderer-only failure, and the scope is limited to the affected report layouts.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11440 round=1 by=alexei-dobriansky at=2026-09-14T18:29:01.0027644Z lastSha=2131f15e9121e136e787175a124ec5f18e1ef118 reviewKey=b5de8f2e9a5f8a22d6462e2034517029599deb5f78f5eaee10fd74323524ed09 suggestions=none

alexei-dobriansky (approved):

PR #11484, opened 2026-09-15
[E-Documents Core] - Supported type direction configuration

GediminasGaubys:
## Implementation

Adds per-service direction control to the existing E-Document supported source
document type configuration. Administrators can allow each document type for
`Incoming`, `Outgoing`, or `Both`, preventing services configured for invoices
from implicitly exchanging other document types such as Remittance Advice.

Key additions:

- Adds a closed `E-Doc. Supp. Type Direction` enum and a `Direction` field to
  `E-Doc. Service Supported Type` (table 6122), exposed on the existing
  supported-types page from the E-Document Service card.
- Enforces the same configuration in both directions:
  - outbound document creation, recreate, and batch export require `Outgoing`
    or `Both`;
  - V1 and V2 inbound processing require `Incoming` or `Both` and log an
    E-Document error when the type is not permitted;
  - a missing supported-type row denies both inbound and outbound processing.
- Preserves document-type fallback relationships, such as Purchase Order and
  Purchase Invoice, while ensuring an explicit row for the requested type takes
  precedence over its fallback row. The upgrade migration (below) excludes the
  seven fallback-secondary types from its insert loop for this reason: leaving
  them unconfigured, rather than forcing an `Incoming`-only row, is what lets
  them keep resolving support through their sibling type after upgrade.
- Extends format defaults with direction:
  - Data Exchange seeds Sales Invoice, Sales Credit Memo, Service Invoice, and
    Service Credit Memo as `Outgoing`;
  - PEPPOL BIS 3.0 seeds the same four types plus Remittance Advice as
    `Outgoing`.
- Initializes defaults using the format selected on a new E-Document Service
  card, including before **Configure supported documents** is opened. Custom
  formats do not inherit Data Exchange defaults, and initialization runs only
  once during the new-service page session.
- Adds a per-company upgrade that converts historical implicit permissions into
  explicit direction rows:
  - existing supported-type rows become `Both`;
  - previously missing document types become `Incoming` for existing services,
    except the seven fallback-secondary types (Sales Order, Sales Return Order,
    Service Order, Finance Charge Memo, Reminder, Purchase Order, Purchase
    Return Order), which stay unconfigured to preserve their fallback behavior;
  - `None` is excluded.
- Registers the upgrade tag on first installation and guards registration and
  migration against reruns, so later administrator changes are not overwritten.
- Optimizes supported-type evaluation by using one lookup for non-fallback
  document types and caching the service record in the V2 rejection path.
- Seeds the built-in services that are created without going through the
  normal service-card flow — the PDF reader (`MSEOCADI`), the ForNAV bootstrap
  service, the Payables Agent service (`AGENT`), and the Contoso demo service
  (`E-DOC DEMO DATA`) — with the minimal `Incoming` supported-type rows they
  need, since the new default-deny model otherwise rejects every document they
  try to process from the moment they're created.
- Guards recurrent batch export against a fully-rejected document type: the
  working record is now `Reset()` (not just `Init()`) between document types
  so a type with zero exportable documents doesn't leave a stale empty-marks
  filter that silently skips later, actually-supported types.
- Guards the threshold batch-send path against an empty export selection,
  returning early instead of running an unconditional `FindSet()`/`repeat`
  against no current record.

## Test coverage

Adds `E-Doc. Serv. Supp. Type Test` coverage for:

- New-service initialization: defaults seeded before the card closes, and only
  once, so an intentionally emptied configuration stays empty.
- Outbound and inbound allow/deny behavior for all direction values and for a
  missing supported-type row.
- Purchase Order/Purchase Invoice fallback resolution, including the case
  where an explicit row on one side must not override the other's direction.
- V1 receive-flow and V2 `PrepareDraft` rejection with persisted error details.
- Recreate and batch-export enforcement after direction changes, and mixed
  batches where only the unsupported document is blocked.
- Data Exchange and PEPPOL default seeding, including Remittance Advice, and
  custom-format creation that must not seed Data Exchange defaults.
- Upgrade conversion of legacy rows to `Both`, exclusion of `None`, and
  migration idempotence.

Fixes
[AB#650228](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/650228)


alexei-dobriansky:

<details>

This change adds direction-specific supported document type setup for E-Document services and applies it to outbound export, recreate, batch export, inbound V1/V2 import, default seeding, and upgrade migration. The overall direction model is needed and most direct import/export paths are covered by tests, but the current implementation has two blocking issues: recurrent batch send can reuse state from a previous document type, and the new enum object ID collides with an enum already present on the target branch.
</details>

<details>

**Fit:** Partial

The reported bug is valid: services need per-document direction controls so one service does not implicitly send or receive unrelated document types. The solution fits that goal in the main direct paths, but it is not safe to merge until the recurrent batch state bug and object ID conflict are fixed.
</details>

<details>

**S1** (🔴 High): *Stale batch log entries cross types*
Clear `EntryNumbers` inside the per-document-type loop, together with `EDocumentListExportedFilter`. If one type is exported and a later type has no allowed documents, the old log entry numbers are still used and the later pass can update or send the wrong batch.

**S2** (🔴 High): *Enum object ID conflicts on merge*
Choose a free object ID for this enum. The target branch already has enum 6115, so this app will not compile when the branch is merged.
</details>

<details>

**Risk:** The change affects E-Document export, import, batch sending, service setup, seeded connector services, and upgrade behavior. The duplicate object ID blocks compilation after merge, and the recurrent batch issue can attach wrong data storage or send the wrong record set after a supported type is followed by a denied type.

**Necessity:** The change is necessary because the current supported-type setup is too broad for services that should allow only incoming or outgoing document types. The scope is justified for the bug, but the two blocking fixes are required before merge.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11484 round=1 by=alexei-dobriansky at=2026-09-15T22:15:28Z lastSha=6d7340bee873777cb42383f94aa082b0258edece reviewKey=7ffdf8fcac98a12da3d6d6f49f6f5db66ddd20a11e1a486463faa0ea2ad5fcca suggestions=S1@40c2a7ae,S2@fc8e7a20

djukicmilica (approved):

PR #11495, opened 2026-09-15
[29.x] Filter top customer and top vendor Excel reports on the posting group

Explorer986:
## What & why

The Top Customer and Top Vendor Excel reports returned nothing when you filtered them on a posting group. Both balance queries filtered Detailed Cust. Ledg. Entry."Posting Group" and Detailed Vendor Ledg. Entry."Posting Group", which came in with v20 and are only written at posting time. Nothing backfills them on upgrade and cloud migration does not bring them across, so on many tenants they are blank. The ranking queries next to them filter the ledger entry posting group, which is always populated, so the two passes could also disagree about which accounts belong to the group. Fix is to join the detailed entry to its parent ledger entry and read the posting group from there.

In the suggested approach collecting the matching customers first and setting them as a number filter. I did not go that way. An empty match gives an empty filter string and SetFilter then clears the filter instead of matching nothing.

Fixes [AB#650214](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/650214)

## How I validated this

## Risk & compatibility

No breaking changes. Filter element names are unchanged so Transfer Filters is unchnaged. The join is on a mandatory key field and is many to one, so no rows are dropped and the Sum is unaffected.


alexei-dobriansky:

# **Recommendation: Accept**

<details>

This changes the top customer and top vendor balance queries so the posting group filter is read from the parent ledger entry, not from the detailed entry. That matches the reported case where detailed entries can have a blank posting group.

The fix is targeted. The existing filter element names stay the same, the date, dimension, currency, and balance logic stay in place, and the new joins use the normal parent-entry relation for customer and vendor ledger entries. The added tests cover both customer and vendor filters with blank detailed-entry posting groups.
</details>

<details>

**Fit:** Strong

The change directly fixes the described mismatch for both affected reports. It also keeps the filter contract unchanged, so callers can keep using the same posting group filter names.
</details>

<details>

None.
</details>

<details>

**Risk:** The change is in read-only Excel report queries. The main regression surface is filtering or grouping in top customer and top vendor balance reports, but the join is one parent ledger entry per detailed ledger entry and no public filter names changed.

**Necessity:** The change is needed because some upgraded or migrated data can have blank posting groups on detailed ledger entries. Without this fix, valid customers or vendors can be hidden when the report is filtered by posting group.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11495 round=1 by=alexei-dobriansky at=2026-09-15T12:13:18.4917865Z lastSha=ac92f8c63136a64650379da21e283c7fb03a84e8 reviewKey=a94e9ab5aef2bfe3c7fe2e326e6b7357ecbde627fe185778236fad20d6e9f0b7 suggestions=none

mynjj (approved):

alexei-dobriansky (approved):

djukicmilica (approved):

PR #11500, opened 2026-09-15
[29.x] - Bug 650240: [Sustainability] Collect Amount from G/L Entries: track collected entries to prevent double counting

v-rohangarg20:
Fixes [AB#650240](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/650240)

Problem
In the Sustainability Journal, the Collect Amount from G/L Entries function sums matching G/L entries (by category account/dimension filters and a date range) and puts the total on the journal line. It stores only the aggregate — it keeps no record of which G/L entries were collected. Running collect again over an overlapping or adjacent period re-sums the same entries, causing double-counted emissions.

Goal
Give the collect function a memory of what it has already collected, so previously collected G/L entries are automatically excluded on later runs — without slowing down collection on large ledgers.

Proposed direction
Model it on the existing G/L ↔ Item Ledger relation: when BC posts inventory it writes a relation record linking each G/L entry to its item/value entry (G/L - Item Ledger Relation). Apply the same idea to sustainability:
A dedicated relation table links each collected G/L entry to the Sustainability entry created from it (plus the category it was collected under).
Collect shows/sums only entries not yet linked for that category, so users see only the new amount.
Links are written when the journal is posted, one per consumed G/L entry.
Reversing the sustainability entry removes its links, making those G/L entries collectable again (so first the Reverse functionality should be finished)
An indexed “collected” marker keeps the exclusion filter fast on large ledgers; the relation table stays the source of truth and audit trail.


alexei-dobriansky:

<details>

This change stores the G/L entries collected for a sustainability journal line, posts relation records from those collected entries to the created sustainability ledger entry, and removes posted relations when a sustainability ledger entry is reversed. The direction matches the reported double-counting problem, but the current implementation still has two data-integrity gaps: a cancel path can remove the pending entry list without replacing it, and posting does not enforce one collected relation per G/L entry and category.
</details>

<details>

**Fit:** Partial

The bug says repeated collection over overlapping periods must not count the same G/L entry again. The change adds the right kind of tracking, but it does not fully protect the stored links before posting or recheck uniqueness when posted relations are created.
</details>

<details>

**S1** (🔴 High): *Do not clear links before collection is confirmed*
Only clear the stored G/L links after the collect dialog returns OK or when the user really changes the custom amount. The collect action validates `Custom Amount` before the dialog opens, so canceling the dialog can delete the saved entry list while the journal amount remains. Posting that line then creates no posted relations, and the same G/L entries can be collected again.

**S2** (🔴 High): *Enforce one posted relation per category*
Add a posting-time check or unique constraint for `Account Category` and `G/L Entry No.` before inserting posted relations. The current primary key only prevents the same G/L entry from linking to the same sustainability ledger entry twice. Two stale or concurrent journal lines can still post the same G/L entry for the same category, which recreates the double-counting bug.
</details>

<details>

**Risk:** The regression surface is the sustainability journal collection, posting, and reversal flow. An incorrect relation set can persist wrong sustainability ledger data, and the new relation table is used as the source of truth for later exclusion.

**Necessity:** The change is needed because aggregate-only collection cannot tell which G/L entries were already consumed. The scope is appropriate for the bug, but the data-integrity protections need to be complete before merge.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11500 round=1 by=alexei-dobriansky at=2026-09-15T13:20:32Z lastSha=040f259a5b18afa53b592cbe908c871623d2d3f7 reviewKey=29ce66762446024595585764fedc974df647887b2536bb2ee77d1125562b0bcc suggestions=S1@c991af1b,S2@0670b479

djukicmilica (approved):

ViditGupta1277 (approved):

PR #11501, opened 2026-09-15
[29.0] - Bug 650241: [Sustainability] Collect Amount from G/L Entries: track collected entries to prevent double counting

v-rohangarg20:
Fixes [AB#650241](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/650241)

Problem
In the Sustainability Journal, the Collect Amount from G/L Entries function sums matching G/L entries (by category account/dimension filters and a date range) and puts the total on the journal line. It stores only the aggregate — it keeps no record of which G/L entries were collected. Running collect again over an overlapping or adjacent period re-sums the same entries, causing double-counted emissions.

Goal
Give the collect function a memory of what it has already collected, so previously collected G/L entries are automatically excluded on later runs — without slowing down collection on large ledgers.

Proposed direction
Model it on the existing G/L ↔ Item Ledger relation: when BC posts inventory it writes a relation record linking each G/L entry to its item/value entry (G/L - Item Ledger Relation). Apply the same idea to sustainability:
A dedicated relation table links each collected G/L entry to the Sustainability entry created from it (plus the category it was collected under).
Collect shows/sums only entries not yet linked for that category, so users see only the new amount.
Links are written when the journal is posted, one per consumed G/L entry.
Reversing the sustainability entry removes its links, making those G/L entries collectable again (so first the Reverse functionality should be finished)
An indexed “collected” marker keeps the exclusion filter fast on large ledgers; the relation table stays the source of truth and audit trail.


alexei-dobriansky:

<details>

This change adds staging and posted relation tables so Sustainability Journal collection can remember which G/L entries were used, exclude them from later collection, and remove the relation when the sustainability entry is reversed. The direction matches the reported double-counting problem, but the implementation still has several paths where links can be lost or reused, and the filter now changes existing event behavior before the final amount is calculated.
</details>

<details>

**Fit:** Partial

The change targets the right root cause: the collection process needs a durable link between collected G/L entries and sustainability entries. It does not yet cover all realistic journal lifecycles, so the same G/L entries can still be double counted or detached from the amount on the line.
</details>

<details>

**S1** (🔴 High): *Canceling collection deletes existing links*
Opening the collect action validates `Custom Amount` before the modal page is accepted. The new clear logic can delete the existing G/L-entry links even when the user cancels, while the old amount remains. Clear old links only after a new collection is accepted, or only when the user really types a manual amount.

**S2** (🔴 High): *Subscriber marks are cleared after event*
`OnAfterFilterGLEntry` subscribers can use marks to choose the exact G/L entries. The new `ClearMarks` call runs after that event and removes their selection, so the amount can be calculated from the wrong entries. Preserve existing marks or move the event to the final filtering point.

**S3** (🟠 Moderate): *Collect amount scans matching ledger entries*
The collect amount path now loops through every matching G/L Entry and marks rows before summing. Large account and date ranges can make the collect page slow. Keep the amount calculation set-based, or avoid scanning the full ledger range in AL.

**S4** (🔴 High): *Posting trusts stale staged G/L entries*
Posting uses the staged rows without checking whether the same G/L entry and category were already posted by another line. Two users, or two stale lines, can still create duplicate relations and double count emissions. Recheck under lock or enforce uniqueness per G/L entry and category before inserting the relation.

**S5** (🔴 High): *Recurring lines reuse collected entries*
A fixed recurring journal line remains after posting, and its staged G/L entry rows are not cleared. The next recurrence can post a new sustainability entry from the same collected G/L entries. Clear collection information after recurring posting or block fixed recurring lines from keeping collected rows.
</details>

<details>

**Risk:** This is a ledger/data-integrity path. If the collection memory is wrong, users can post incorrect sustainability amounts, extensions can change which entries are collected incorrectly, and large G/L ranges can become slower than before.

**Necessity:** The change is needed because repeated collection can overstate sustainability amounts. The scope is appropriate for the problem, but posting-time safeguards and safer lifecycle cleanup are required before this is safe to merge.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11501 round=1 by=alexei-dobriansky at=2026-09-15T13:20:31Z lastSha=2822f5eb9697079326a264ba4253e83f2649b331 reviewKey=87006c72ccf7b19989b2d9636bef1ef4e20c640dc33b3c3d12a13b928f89268e suggestions=S1@bccf80cd,S2@bea83dfb,S3@4e56a537,S4@bab058b0,S5@4ce7b524

djukicmilica (approved):

ViditGupta1277 (approved):

PR #11507, opened 2026-09-15
[Master] Fix Unix timestamp timezone conversion

haoranpb:
#### Summary

`TypeHelper.EvaluateUnixTimestamp` compensated for a locally created Unix epoch by adding the user's timezone offset. After the offset became daylight-saving-aware, this shifted the represented UTC instant by one hour for Central European users during summer time.

Delegate the conversion to the existing `System.DateTime` `"Unix Timestamp"` codeunit, which uses `DateTimeOffset.FromUnixTimeSeconds` and does not depend on the user's timezone.

Fix [AB#650271](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/650271)

#### Tests

- Verify timestamp `1789375237` evaluates to `2026-09-14T08:40:37Z` (`10:40:37` in Central European summer time).
- Verify evaluation returns the same DateTime with user time zones `UTC` and `Central Europe Standard Time`.


alexei-dobriansky:

# **Recommendation: Accept**

<details>

This PR changes `Type Helper`.EvaluateUnixTimestamp so it delegates Unix timestamp conversion to the existing `Unix Timestamp` codeunit instead of adding the current user's time zone offset. That matches Unix timestamp semantics because the input represents a UTC instant, not a local date and time. The change is narrow, the called code uses `DateTimeOffset.FromUnixTimeSeconds`, and the added tests cover both the concrete UTC value and independence from the user's configured time zone.
</details>

<details>

**Fit:** Strong

The reported bug is that evaluating a Unix timestamp can return a value shifted by the user's time zone. The diff removes that dependency from the only conversion point and keeps the public procedure shape unchanged, so callers now get a stable UTC-based result.
</details>

<details>

*None.*
</details>

<details>

**Risk:** The touched procedure is public, so callers that worked around the old shifted result may see different values. The regression surface is still limited to Unix timestamp conversion callers, and the new behavior matches the meaning of Unix timestamps. No event publisher or BaseApp event-order dependency is involved.

**Necessity:** The change is needed because Unix timestamps represent UTC instants. Keeping user time zone adjustment in this helper would keep producing wrong values in time zones with daylight-saving changes.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11507 round=1 by=alexei-dobriansky at=2026-09-15T22:11:11Z lastSha=6f80442739f1a711a3a87ca15fe772b0a4c1cfe1 reviewKey=235dc9f3889e72168bad9f0c9ce147d52909fea72f9b9704348c82b7a7c8eb5d suggestions=none

darjoo (approved):

Aleyenda (approved):

gggdttt (approved):

PR #11531, opened 2026-09-16
[Master] - Bug 648711: [Expense Agent] Changing an expense report line from Meals to Per-diem fails when adding Expense Location

v-rohangarg20:
Fixes [AB#648711](https://dynamicssmb2.visualstudio.com/Dynamics%20SMB/_workitems/edit/648711)

### Problem

Changing an existing expense report line from a Meals category to Per-diem succeeds initially, but entering the required Expense Location then fails with an out-of-date record error.

### Changes

- In `ExpenseReportLinePerDiem.Table.al`, removed the redundant `InvalidateParentPolicy()` call during Per Diem insert.
- In `ExpenseAutoPopulation.Codeunit.al`, incremented `"Policy Eval Version"` during per diem generation to keep policy version tracking in sync.


alexei-dobriansky:

<details>

The change tries to stop the stale page record error when a report line is changed to per diem and a location is entered. It does that by removing parent policy invalidation from every per-diem child insert and adding a version increment only in the auto-population loop. That avoids the child insert writing the parent row in the reported page path, but it moves a policy invariant from the child table to one caller and leaves other inserts uncovered.
</details>

<details>

**Fit:** Partial

The reported UI failure is valid and the changed path is related to it, but the fix is wider than the bug. Adding a per-diem child must still make the parent policy evaluation stale; this diff removes that guarantee for direct insert paths.
</details>

<details>

**S1** (🔴 High): *Keep per-diem insert invalidation complete*
Removing `InvalidateParentPolicy()` from the table trigger means other per-diem insert paths no longer mark the parent report line stale. `CopyPerDiemFromExpense()` and the existing child-record policy test insert per-diem rows without going through the auto-population loop, so a cleared policy evaluation can stay current after details are added. Keep a safe invalidation path for every per-diem insert, not only this caller.
</details>

<details>

**Risk:** The changed table is internal, but it owns expense report policy state. If a per-diem row is inserted by another codeunit or API path, `Policy Eval Version` can stay unchanged, and the line can continue to show a current or cleared evaluation for stale details.

**Necessity:** The stale page record bug is valid and should be fixed. The necessary scope is to avoid saving the parent through the child trigger while preserving policy invalidation for all child inserts, so the current approach needs adjustment.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11531 round=1 by=alexei-dobriansky at=2026-09-16T13:18:05Z lastSha=e2a088100803b22f2ca4f038c342bcf8d4604e8a reviewKey=90e2488cdb8089d3f3d83acc799b82755456a68c464f0e15b48b9e6981ba130b suggestions=S1@f90451f7

PR #11532, opened 2026-09-16
Fix Navigate override for Production Order documents by respecting IsHandled on OnAfterShowRecords of Document Entry

### LINKED ISSUE #11530: [Bug]: Mfg. Navigate Mgt. subscribes to OnBeforeShowRecords instead of OnAfterShowRecords, making the behavior impossible to override
### Describe the issue

Codeunit 99000994 `Mfg. Navigate Mgt.` subscribes to `Navigate.OnBeforeShowRecords` to show the Production Order List page:

```al
[EventSubscriber(ObjectType::Page, Page::Navigate, 'OnBeforeShowRecords', '', true, false)]
local procedure OnBeforeShowRecords(var TempDocumentEntry: Record "Document Entry"; DocNoFilter: Text; PostingDateFilter: Text; var IsHandled: Boolean; ContactNo: Code[250])
begin
    case TempDocumentEntry."Table ID" of
        Database::"Production Order":
            begin
                SetProdOrderFilters(DocNoFilter);
                Page.Run(0, ProductionOrderHeader);
            end;
    end;
end;
```

This subscriber does not follow the standard `IsHandled` guard pattern:
1. It never checks `IsHandled` before executing, so it runs even if another subscriber already handled the `Production Order` case.
2. It never sets `IsHandled := true` after running, so any logic that runs after this event (including the base `Navigate.OnBeforeShowRecords` publisher body) still executes for the "Production Order" table ID.

Because this logic sits in `OnBeforeShowRecords` (a "before" event) rather than `OnAfterShowRecords` (an "after" event) and doesn't set `IsHandled`, a partner/AppSource extension can no longer reliably override how Production Order document entries are opened from Navigate. Any subscriber added to `OnBeforeShowRecords` to customize this behavior would need to run *before* this one in an undefined subscriber order, and even then couldn't cleanly suppress this default handling since `IsHandled` is never respected here.

### Expected behavior

The logic should be moved to `OnAfterShowRecords`:

```al
    [EventSubscriber(ObjectType::Page, Page::Navigate, 'OnAfterShowRecords', '', true, false)]
    local procedure OnAfterShowRecords(var DocumentEntry: Record "Document Entry"; DocNoFilter: Text; PostingDateFilter: Text; ItemTrackingSearch: Boolean; ContactType: Enum "Navigate Contact Type"; ContactNo: Code[250]; ExtDocNo: Code[250])
    begin
        case DocumentEntry."Table ID" of
            Database::"Production Order":
                begin
                    SetProdOrderFilters(DocNoFilter);
                    Page.Run(0, ProductionOrderHeader);
                end;
        end;
    end;
```

### Steps to reproduce

1. Create an extension that subscribes to `Navigate.OnBeforeShowRecords`, checks for `TempDocumentEntry."Table ID" = Database::"Production Order"`, implements custom logic, and sets `IsHandled := true`.
2. Open the Navigate page for a document linked to a Production Order.
3. Observe that the base `Mfg. Navigate Mgt.` subscriber still runs (or races with the custom one), because it ignores `IsHandled` and is wired to the "before" event.

### Additional context

_No response_

IceOnly:
## What & why

`Mfg. Navigate Mgt.` handled the `Production Order` case in `Navigate.OnAfterShowRecords` unconditionally, without checking or setting `IsHandled`. This made it impossible for another extension to override how Production Order document entries are opened from the Navigate page, since this subscriber always ran regardless of whether another subscriber had already handled the case. This change adds the standard `IsHandled` guard so the behavior can be properly overridden by other subscribers.

## Linked work

Fixes #11530

## How I validated this

- Opened the Navigate page for a document linked to a Production Order and confirmed the Production Order List page still opens as before.
- Verified that a subscriber setting `IsHandled := true` before this one now prevents this default handling from running.
- No tests added: this is an event-subscriber wiring fix with no new business logic to unit test.

## Risk & compatibility

None — this only adds a guard clause around existing behavior; the default (unhandled) case behaves exactly as before.


alexei-dobriansky:

# **Recommendation: Accept**

<details>

The change moves the Production Order Navigate handler from the before event to the after event. This lets another handler set `IsHandled` and stop default handling, while the normal Production Order List still opens when nothing handles the case.

`Navigate.ShowRecords` exits before the after event when `IsHandled` is true. Because Production Order is not handled by the page's built-in `case`, the new after-event subscriber preserves the existing unhandled behavior and removes the override race in the before event.
</details>

<details>

**Fit:** Strong

The reported scenario is that custom handling for Production Order entries cannot reliably suppress the default opening behavior. The diff addresses that directly by moving the default handler behind the `IsHandled` exit instead of competing inside the before event.
</details>

<details>

None.
</details>

<details>

**Risk:** The change affects only the Production Order branch in `Mfg. Navigate Mgt.` when `Navigate.ShowRecords` opens records. Regression risk is low: unhandled cases still run `SetProdOrderFilters` and `Page.Run`, handled cases now exit earlier as intended, and no public event signature changes are introduced.

**Necessity:** The change is needed to make the standard `IsHandled` override pattern work for this Navigate scenario. The scope is narrow: one subscriber and one table ID, with no posting, financial, or data persistence path changed.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11532 round=1 by=alexei-dobriansky at=2026-09-16T22:09:13Z lastSha=1db09d4aa02c099a4b4914625fdf4f7c60ae73fe reviewKey=0fe1a7c9faf7dfbca6990603d6827986a57716045ca2e2df2f3397f913d4f0d0 suggestions=

Issue #11549, opened 2026-09-16
[CI] releases/28.x broken: System Application fails AL0133 (DotNet String[]& by-ref) after platform build 28.6.54594.0


mazhelez:
## Summary

CI/CD on `releases/28.x` has been failing since **2026-09-13**. The **System Application** build no longer compiles against the daily `bcinsider/Sandbox/28.6//latest` artifact, which fails the whole pipeline (all downstream jobs are skipped).

## Error

```
AL0133 Argument 4: cannot convert from 'DotNet "System.String[]"' to 'DotNet "System.String[]&"'
```

A platform .NET method changed its 4th parameter to **by-reference** (`System.String[]&`), so the System Application's existing DotNet interop declaration no longer matches.

## Root cause: platform regression (not a BCApps source change)

The branch `artifact` setting is unpinned (`bcinsider/Sandbox/28.6//latest`), so each scheduled run pulls the newest 28.6 insider build. `releases/28.x` HEAD has not moved since 2026-09-11 (`7fe5294e9b14698cb6c1802c441ff2074702e7bf`), yet the **same commit** flips from pass to fail purely based on the artifact:

| Date | Source commit | Artifact (application build) | System App build | AL0133 |
|------|---------------|------------------------------|------------------|--------|
| 2026-09-12 | `7fe5294` | `sandbox/28.6.54584.0` | ✅ success | no |
| 2026-09-13 | `7fe5294` | `sandbox/28.6.54594.0` | ❌ failure | **yes** |
| 2026-09-16 | `7fe5294` | `sandbox/28.6.54714.0` | ❌ failure | yes |

## Regression boundary

- **Last good platform build:** `28.6.54584.0`
- **First bad platform build:** `28.6.54594.0`

## Impact

- `releases/28.x` CI/CD red every scheduled run since 2026-09-13.
- Because System Application fails first, the earlier **Apps (W1) localization** build failures (2026-09-09 → 2026-09-12) are now masked and should be re-verified once the platform issue clears.

## Requested action

Route to the **BC platform team**: a by-ref signature change (`String[]` → `String[]&`) landed between platform builds `28.6.54584.0` and `28.6.54594.0` and breaks AL DotNet interop in the System Application. Either revert/adjust the signature, or advise the required AL interop change.

## Temporary mitigation

[...]

PR #11550, opened 2026-09-16
[28.x] Pin BCArtifact to 28.6.54584.0 to unblock CI

mazhelez:
## Problem

CI/CD on `releases/28.x` has been red since 2026-09-13. The **System Application** build fails to compile with:

```
AL0133 Argument 4: cannot convert from 'DotNet "System.String[]"' to 'DotNet "System.String[]&"'
```

This is **not** a BCApps source change. It is a BC platform regression in the daily insider artifact. The branch uses an unpinned artifact (`bcinsider/Sandbox/28.6//latest`), so every run pulls the newest 28.6 build — including the regressed one.

## Fix

Pin the artifact to the last-good build **28.6.54584.0**:

```diff
- "artifact": "bcinsider/Sandbox/28.6//latest",
+ "artifact": "bcinsider/Sandbox/28.6.54584.0//latest",
```

This is the same pinned-version convention already used on `main` (`bcinsider/Sandbox/<version>//latest`), so the format is precedented. No other settings are touched.

## Evidence

| Date | Artifact build | Result |
|------|----------------|--------|
| 9/12 | `28.6.54584.0` (last-good) | ✅ pass |
| 9/13 | `28.6.54594.0` (first-bad) | ❌ AL0133 fail |

## Note

This is a **temporary mitigation** to unblock CI. Revert this change back to `bcinsider/Sandbox/28.6//latest` once the platform regression is fixed upstream.

[AB#650351](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/650351)


mazhelez:
The **work-item validation** failure here is **not** about the linked issue. Issue #11549 *is* GitHub issue type `Task`, and the PR body references it correctly — so under the GitHub-Task exemption this PR should pass without an `AB#`.

Root cause: that exemption (#10743) was merged to `main` but was never ported to `releases/28.x`. This branch still runs the **old** `ValidateInternalWorkItemForPullRequest.ps1` that only accepts an ADO `AB#` reference and never inspects the linked issue's type, so it fails regardless of the Task link.

Fix: I opened #11553 to backport the exemption to `releases/28.x`. Once it merges, re-running validation here (or a synchronize/edit event) will pass via the Task link — no `AB#` needed. (`releases/27.x` / `26.x` / `25.x` are missing it too and can be backported the same way.)

aholstrup1 (approved):

SBalslev (approved):

PR #8318, opened 2026-05-26
SharePoint Graph Module - extend functionality with additional methods

Drakonian:
## What & why

The SharePoint Graph module currently supports creating files, folders, and list items ΓÇö but offers no way to modify them after creation. Renaming an uploaded file or updating metadata on a list item are among the most common SharePoint integration scenarios, and their absence forces consumers to drop down to raw HTTP calls.

This PR closes that gap by adding update operations for both drive items and list items, backed by the standard MS Graph PATCH endpoints. It also upgrades the test infrastructure to properly cover multi-step operations that were previously untestable.

## Linked work

Fixes #8317

## How I validated this

- [x] I read the full diff and it contains only changes I intended.
- [x] I built the affected app(s) locally with no new analyzer warnings.
- [x] I ran the change in Business Central and confirmed it behaves as expected.
- [x] I added or updated tests for the new behavior, or explained below why none are needed.
- [x] Compiled System Application, Test Library, and Test projects ΓÇö all three build clean, no new analyzer warnings.
- [x] Ran `UpdateDriveItem` and `RenameDriveItem` against a test SharePoint site ΓÇö file renamed successfully, response parsed into `SharePoint Graph Drive Item` record with updated name.
- [x] Ran `UpdateListItem` with a fields payload (`Title`, custom columns) ΓÇö PATCH succeeded, follow-up GET returned the full list item with updated values.
- [x] Ran `GetListItem` for a single item ΓÇö returned the expected record with all fields (ContentType, WebUrl, timestamps).
- [x] Verified `RenameDriveItemByPath` and `UpdateDriveItemByPath` ΓÇö path resolution + PATCH executed as two consecutive HTTP calls, both reflected in mock request log.
- [x] Confirmed `UpdateListItem_FollowUpGetFails` test ΓÇö when the GET after PATCH returns 404, the error is surfaced to the caller and request count is 2 (PATCH did fire).
- [x] Ran all existing SharePoint Graph tests ΓÇö no regressions. Sticky single-response mode works exactly as before.
- [x] 22 new tests pass: 10 for drive item updates, 8 for list item operations, 4 for multi-response handler infrastructure.

## Risk & compatibility

- All new methods are additive ΓÇö no existing signatures or behavior changed.
- The test handler refactor preserves the original single-response contract; existing tests work without modification.

Fixes [AB#640918](https://dynamicssmb2.visualstudio.com/1fcb79e7-ab07-432a-a3c6-6cf5a88ba4a5/_workitems/edit/640918)

PR #11555, opened 2026-09-17
[SharePoint Graph] Follow-up corrections for UpdateListItem and drive item telemetry

### LINKED ISSUE #10304: [Bug]: SharePoint Graph update methods (PR #8318) - follow-up corrections
Follow-up to #8318 (SharePoint Graph Module - extend functionality with additional methods), which was merged before these review points were addressed. @Drakonian, could you pick these up in a follow-up PR?

### Describe the issue

Code review of #8318 found one behavioural defect and several convention issues in the new `UpdateListItem` / `UpdateDriveItem` / `RenameDriveItem` code paths.

**1. Silent field loss in `UpdateListItem` (main issue)**

In `SharePoint Graph Client Impl.` (codeunit 9120), `UpdateListItem` does:

```al
GraphListItem.Init();
GraphListItem.Id := CopyStr(ItemId, 1, MaxStrLen(GraphListItem.Id));
GraphListItem.ListId := CopyStr(ListId, 1, MaxStrLen(GraphListItem.ListId));
SharePointGraphParser.ParseListItemFieldValueSet(JsonResponse, GraphListItem);
if not GraphListItem.Insert() then
    GraphListItem.Modify();
```

The PATCH to `.../items/{id}/fields` returns only a `fieldValueSet`, so it carries no `ContentType`, `WebUrl`, `CreatedDateTime` or `LastModifiedDateTime`. When a caller reuses a buffer that was previously populated by `GetListItem`, the `Init()` blanks those fields and the `Modify()` writes the blanks back over data that was already correct.

The XML doc says the record "is refreshed", which understates this - the record is effectively replaced, and four previously-valid fields are silently cleared.

**2. `ItemBufferCollisionTelemetryErr` is missing `Locked = true`**

```al
ItemBufferCollisionTelemetryErr: Label 'The record already contains item from a different list. Use a separate record variable per list.';
```

It is only ever passed to `Session.LogMessage`, so it should be locked. It is the only telemetry-only label in the file without it (compare `OperationSuccessTelemetryMsg` and `GraphSharePointCategoryLbl`). As-is it gets sent for translation and emits localized telemetry.

**3. Hand-rolled query string in `GetListItem`**

```al
Endpoint := SharePointGraphUriBuilder.GetListItemByIdEndpoint(ListId, ItemId);
if not GraphOptionalParameters.GetODataQueryParameters().ContainsKey(Format(Enum::"Graph OData Query Parameter"::expand)) then
    Endpoint += '?$expand=fields';
```

This is the only raw query-string concatenation in the module. It works, but the idiomatic form is `GraphOptionalParameters.SetODataQueryParameter(Enum::"Graph OData Query Parameter"::expand, 'fields')` - the parameter is passed by value, so mutating the local copy is safe and it keeps all URI assembly inside the builder.

**4. `ParseListItemFieldValueSet` has no return value**

Every other parser in `SharePoint Graph Parser` returns `Boolean` and callers check it. This one does not, so `UpdateListItem` cannot detect a malformed PATCH response and will report success on garbage input.

**5. Telemetry event IDs are shared across four public entry points**

`UpdateDriveItem`, `UpdateDriveItemByPath`, `RenameDriveItem` and `RenameDriveItemByPath` all funnel through `UpdateDriveItemAtEndpoint`, so all four log `0000UKR` / `0000UKS` / `0000UKT` with the operation name hard-coded as `'UpdateDriveItem'`. A `RenameDriveItemByPath` failure is indistinguishable from an `UpdateDriveItem` failure in telemetry.

Separately, it is still worth confirming that `0000UKB`-`0000UKV` are actually allocated to this area - this is the question @Groenbech96 raised on the PR in July that was never answered.

### Expected behavior

1. `UpdateListItem` should either preserve the fields the PATCH response does not return, or the XML doc on both the facade and the implementation should state plainly that the record is replaced and that `ContentType`, `WebUrl`, `CreatedDateTime` and `LastModifiedDateTime` are cleared. Preserving them is the better outcome.
2. `ItemBufferCollisionTelemetryErr` carries `Locked = true`.
3. `GetListItem` sets the default expand through `Graph Optional Parameters` rather than string concatenation.
4. `ParseListItemFieldValueSet` returns `Boolean` and `UpdateListItem` checks it, matching the surrounding parser contract.
5. Each public update/rename entry point is distinguishable in telemetry.

### Steps to reproduce

For the `UpdateListItem` field loss:

1. Call `GetListItem(ListId, ItemId, TempGraphListItem)` and observe `WebUrl`, `ContentType`, `CreatedDateTime` and `LastModifiedDateTime` are populated.
2. Call `UpdateListItem(ListId, ItemId, FieldsJson, TempGraphListItem)` with the **same** record variable.
3. Observe those four fields are now blank, while `Title` and the fields JSON reflect the update.

`TestUpdateListItem_BufferReusedFromGet` in codeunit 132984 covers this exact sequence but only asserts `Id`, `Title` and the request count, so the loss is not caught. Extending that test with assertions on `WebUrl` / `ContentType` reproduces it.

The remaining points are visible by inspection of `SharePointGraphClientImpl.Codeunit.al` and `SharePointGraphParser.Codeunit.al`.

### Additional context

Two further review points from #8318 that are no longer actionable in code, noted here only so the thread is complete:

- The PR description's validation checklist described a follow-up GET after the PATCH and a `UpdateListItem_FollowUpGetFails` test. Neither exists in the merged code - `UpdateListItem` issues a single PATCH. It also stated 22 new tests where there are 25. The description appears to have been written against an earlier iteration; worth keeping the checklist in sync with the final diff on future PRs, since reviewers lean on it.
- Roughly 837 of the 862 added README lines are a new architecture document (dependency graphs, object registry, test-coverage counts, pattern catalogue) unrelated to the linked issue. The object-registry and test-coverage sections in particular will drift out of date quickly - worth deciding whether they should be maintained or trimmed.

One consistency observation for whoever picks this up: the new methods use `if not Insert() then Modify()`, while the pre-existing `GetDriveItem` uses a bare `Insert()` and hard-errors on a reused buffer. `TestStickyModeStillWorks` works around this with a `TempDriveItem.DeleteAll()`. Not introduced by this PR, but the client now has two different buffer-reuse contracts, and aligning them would be worthwhile.

Original review context: #8318, fixes #8317.

Drakonian:
## What & why

Follow-up to #8318 addressing the review points collected in #10304.

- `UpdateListItem` no longer clears `ContentType`, `WebUrl`, `CreatedDateTime` and `LastModifiedDateTime` when the record already holds the item. The PATCH response is a fieldValueSet and does not carry those fields, so they are preserved and the XML docs say so.
- `ParseListItemFieldValueSet` now returns `Boolean`, matching the other parsers. It returns false only when the response has no column value at all, which cannot happen for a successful PATCH since the request always sends at least one field. `UpdateListItem` reports that as a parse error under new event ID `0000UKW`.
- `ItemBufferCollisionTelemetryErr` is `Locked`, since it is only sent to telemetry.
- The shared drive item update path receives the calling operation name and logs it as an `OperationName` dimension, so `UpdateDriveItem`, `UpdateDriveItemByPath`, `RenameDriveItem` and `RenameDriveItemByPath` are distinguishable without new event IDs.
- `GetListItem` keeps appending the default `$expand=fields` to the endpoint. Setting it through `Graph Optional Parameters` would mutate the caller's instance, because codeunits are reference types, and leak the expansion into later requests. A comment explains this and two tests protect it.

## Linked work

Fixes #10304

## How I validated this

- SharePoint, SharePoint Test Library and SharePoint Test compile locally with CodeCop, UICop, AppSourceCop and PTECop using the repo ruleset; no new diagnostics.
- `TestUpdateListItem_BufferReusedFromGet` now asserts the four preserved fields.
- New tests: `TestUpdateListItem_ResponseWithoutFieldValues`, `TestGetListItem_DefaultExpandNotLeakedToCallerParameters`, `TestGetListItem_CustomExpandOverridesDefault`.

## Risk & compatibility

- No public signature changes. The parser codeunit is internal.
- Telemetry: events `0000UKN` to `0000UKT` gain an `OperationName` custom dimension; `0000UKW` is new. Please confirm the ID is allocated to this area, as asked for `0000UKB` to `0000UKV` in #8318.
- Not changed: the pre-existing bare `Insert()` in `GetDriveItem` versus insert-or-modify elsewhere, mentioned as a side note in the issue.


alexei-dobriansky:

# **Recommendation: Accept**

<details>

This change tightens the SharePoint Graph update paths. It preserves list item details already present in the caller's temporary record when a PATCH response only returns a fieldValueSet, makes that parser report an empty fieldValueSet, keeps the default `$expand=fields` behavior without mutating caller parameters, and adds operation names to shared drive-item telemetry.

The code matches the reported behavior. It detects an existing list item before the PATCH, only initializes new records, parses field values before insert or modify, and the new tests cover the preserved fields, parse failure, and optional-parameter leak cases.
</details>

<details>

**Fit:** Strong

The bug describes field loss on a reused list-item buffer and indistinguishable update/rename telemetry. The diff directly addresses those paths, keeps the deliberate default expand behavior, and adds focused regression tests for the main record-preservation and optional-parameter behavior.
</details>

<details>

*None.*
</details>

<details>

**Risk:** The touched surface is the SharePoint Graph client in System Application. Regression risk is mainly consumers that reuse temporary SharePoint list item or drive item records, callers that reuse `Graph Optional Parameters`, and telemetry queries over events `0000UKN` through `0000UKT`. No public signatures or BaseApp event dependencies change.

**Necessity:** The change is needed because PATCH fieldValueSet responses do not contain `ContentType`, `WebUrl`, or timestamps, so the previous refresh behavior could silently clear already-correct values. The scope is appropriate for a follow-up bug fix: it changes only the affected parser/client paths, docs, and tests.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11555 round=1 by=alexei-dobriansky at=2026-09-21T22:20:14Z lastSha=1fb4f3d2563923c37f478cff1231ea7a1d6c0b88 reviewKey=c87e8269b887c9eaae62a785d18f9bc393ebe5e5c81e9bb72dbe6c376c9e70e4 suggestions=

PR #11612, opened 2026-09-18
Multiple runs of email retrieval on the same emails don't persist again

mynjj:

## Fix overview


ventselartur:
[Optional] The fix itself is great, but let's add a bit more telemetry to troubleshoot these kind of things easier in future.

ventselartur (approved):

AleksanderGladkov (approved):

alexei-dobriansky:

<details>

The change avoids storing the same retrieved email more than once by looking up an existing inbox entry by account and external message id. It also tries to load attachments later when the first stored copy did not include them.

The duplicate lookup is well placed in the Outlook retrieval flow, and the new key supports the lookup shape. The attachment path also reaches the existing message and inserts attachment rows, but it does so by adding a public caller-controlled bypass around the received-email modify guard. That makes the implementation wider than the bug requires.
</details>

<details>

**Fit:** Partial

The reported bug is clear, and the lookup prevents a second inbox row for the same account and provider id. The solution also changes the public attachment API in a way that allows more than the duplicate-retrieval scenario, so the fit is only partial until that bypass is narrowed or validated.
</details>

<details>

**S1** (🔴 High): *Do not expose the retrieval bypass*
The new overload lets any caller pass `true` and skip the `Modify()` path that enforces the received-email guard. Keep this bypass inside the retrieval flow, or validate that the message is already in the inbox and has no attachments before skipping `Modify()`.

**S2** (🟠 Moderate): *Add duplicate retrieval coverage*
The new tests call the lookup and attachment APIs directly, but they do not run the Outlook retrieval branch that finds an existing provider message. Add a regression test that retrieves the same provider message twice and verifies one inbox row, including the case where attachments are loaded only on the second retrieval.
</details>

<details>

**Risk:** The storage change affects email retrieval used by the Outlook-based connectors and downstream agent flows that process returned inbox entries. The table key is narrow, but the new public attachment overload expands the Email module API and can weaken the invariant that received messages are not edited after retrieval.

**Necessity:** The fix is needed because repeated retrieval of the same provider message can create duplicate persisted inbox entries and confuse consumers. The scope should stay focused on deduplication and controlled missing-attachment loading, not a general public bypass.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11612 round=1 by=alexei-dobriansky at=2026-09-18T22:33:13.1246529Z lastSha=78d0f8b1db0af62e8788b1e01930816bb7a01ef4 reviewKey=8f27cdd935b6106df9499605e50597dffb7721ac72864c5c7da9071b55d56fac suggestions=S1@33f9ec4c,S2@46ebc0d1

darjoo (approved):

PR #11616, opened 2026-09-18
[28.0] Multiple runs of email retrieval on the same emails don't persist again

mynjj:
Backport of #11612.

## Fix overview

> [!NOTE]


dcenic (approved):

AleksanderGladkov (approved):

alexei-dobriansky:

<details>

This adds a shared email-inbox lookup by account and external message id, plus an attachment path for messages that were already retrieved. The sequential duplicate case is covered by the new lookup and tests, but the changed public API surface leaves access, concurrency, and bypass gaps that can still produce unsafe behavior.
</details>

<details>

**Fit:** Partial

The change provides the reusable System Application part of the duplicate-retrieval fix, but this branch still needs the connector-side use to make the behavior complete. The current implementation also does not fully guard the duplicate lookup or the retrieved-message attachment path.
</details>

<details>

**S1** (🔴 High): *Keep retrieved-email lookup behind permissions*
Add the same permission check that `RetrieveEmails` uses before `FindRetrievedEmail` reads or copies `Email Inbox`. The new public method runs through an internal codeunit that has table permissions, so a caller can otherwise discover stored inbox rows without the normal email permission gate.

**S2** (🔴 High): *Make retrieved-email lookup concurrency safe*
The new key is only a lookup key and it does not include `Connector`, even though an account id is paired with a connector. Two retrieval sessions can still insert the same external message twice, and the lookup can pick a row for a different connector. Add a uniqueness or locking guard and include the connector in the lookup/key.

**S3** (🔴 High): *Restrict retrieved-message attachment bypass*
Do not expose `MessageAlreadyRetrieved` as a caller-controlled public bypass. Any caller can pass `true` and skip the modify check that protects retrieved messages, so keep the bypass internal to the retrieval path or verify the loaded message is the matching retrieved email.

**S4** (🟠 Moderate): *Avoid temporary inbox ID collisions*
In a temporary result set, `Id` values can already be used by new messages inserted earlier in the same retrieval. If a temp row has the same `Id` as the stored inbox row, `Get` succeeds, the stored row is not copied, and the wrong row is marked. Check that the temp row matches the stored row before marking it.
</details>

<details>

**Risk:** This touches public System Application email APIs and Email Inbox/Message storage. An incorrect fix can expose inbox metadata, mutate retrieved messages, or still create duplicates. The current validation run also fails to compile with AL0133, so the branch needs a clean build before merge.

**Necessity:** The change is needed because repeated retrieval should reuse the existing inbox entry and still allow missing attachments to be added. The scope is reasonable for the System Application part, but it must stay guarded and must be consumed by the connector-side change for the bug to be fixed end to end.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11616 round=1 by=alexei-dobriansky at=2026-09-18T22:38:08.1240065Z lastSha=929a477d236b773a9c46ad155c665059b025270c reviewKey=e8cc161efa301aaa3dd7e9310b21d42233afa325ad0d31426bc5c8939895e55e suggestions=S1@4d95880d,S2@86b79be2,S3@33f9ec4c,S4@d1cadb4c

ventselartur (dismissed):

darjoo (dismissed):

alexei-dobriansky:

<details>

This round only renames test variables. It does not change the email lookup, key, permission, or attachment-bypass logic from round 1, so the earlier correctness and safety concerns still apply.
</details>

<details>

| S1 | Keep retrieved-email lookup behind permissions | Not addressed | No reply. |
| S2 | Make retrieved-email lookup concurrency safe | Not addressed | No reply. |
| S3 | Restrict retrieved-message attachment bypass | Not addressed | No reply. |
| S4 | Avoid temporary inbox ID collisions | Not addressed | No reply. |
</details>

<details>

*None - the new commit only renames variables in the added tests.*
</details>

<details>

**Risk:** This still touches public System Application email APIs and persisted inbox/message data. If merged as-is, callers can bypass normal read and modify protections, duplicate retrievals can still happen under concurrency, and current validation is still red with AL0133.

**Necessity:** Reusing an already retrieved email is still needed, and adding missing attachments is a valid part of the fix. The scope remains reasonable, but the public API and storage changes need the same guards as the rest of the email retrieval path.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11616 round=2 by=alexei-dobriansky at=2026-09-21T22:08:16Z lastSha=638c35ed9f962f29a83291e88a98690755fd0ce7 reviewKey=170b89051ba57cbf5f517db0842cc1a0fad344f5b872f30dd5a03eaf85a39af4 suggestions=S1@4d95880d:notaddressed,S2@86b79be2:notaddressed,S3@33f9ec4c:notaddressed,S4@d1cadb4c:notaddressed parentRound=1

AleksanderGladkov (approved):

dcenic (approved):

PR #11619, opened 2026-09-18
[28.3] Multiple runs of email retrieval on the same emails don't persist again

mynjj:
Backport of #11612.

## Fix overview

> [!NOTE]


dcenic (approved):

AleksanderGladkov (approved):

alexei-dobriansky:

<details>

The change adds shared email-module support for detecting an already retrieved inbox entry by account and external message id. It also adds an attachment path that can complete a retrieved message that was stored without attachments.

The lookup shape matches the duplicate-retrieval scenario: it uses account plus external id, returns the existing inbox row, and marks it so callers can include it in the retrieve result. The attachment path is riskier because it exposes a public bypass around the retrieved-message modify guard, and the current branch cannot merge while the System Application app fails to generate.
</details>

<details>

**Fit:** Partial

The shared email-module change is a reasonable part of the duplicate-retrieval fix, and the tests cover the intended lookup and missing-attachment paths. The branch still needs a compiling build and a safer public API boundary before the solution can be considered complete.
</details>

<details>

**S1** (🔴 High): *Fix the System Application build failure*
The branch does not build in Default or Strict. Fix the AL0133 by-ref `String[]` mismatch and rerun the System Application build before merge.

**S2** (🟠 Moderate): *Keep the permission guard on lookup*
`FindRetrievedEmail` reads the inbox through `Email Impl` permissions without the usual guard. Add the same permission check used by `RetrieveEmails` before reading or marking the inbox entry.

**S3** (🟠 Moderate): *Do not trust the public bypass flag*
The new public overload lets any caller pass `true` and skip the `Modify()` path for retrieved messages. Keep the exception internal or verify that the message is the already-retrieved message being completed with missing attachments.
</details>

<details>

**Risk:** This is public System Application email surface, so a mistake affects all connectors and extensions that use retrieved messages. The main risks are a non-compiling app, permission bypass for inbox reads, and attachments being added to retrieved messages without the normal modify checks.

**Necessity:** The change is needed to avoid duplicate inbox rows when the same provider message is retrieved again. The scope is reasonable for the shared module, but it must be safe to call from public APIs and must build cleanly.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11619 round=1 by=alexei-dobriansky at=2026-09-18T22:31:17Z lastSha=0ee3aae1f26d01fcfc46b6f70fab046193ed4413 reviewKey=9a85286cd790f3a2dba0c1173dfe5150fcd0c5a518f681ef0158827857dbb644 suggestions=S1@a14cdaa9,S2@4d95880d,S3@33f9ec4c

ventselartur (dismissed):

darjoo (approved):

alexei-dobriansky:

<details>

The change adds shared email-module support for finding a previously retrieved inbox entry by account and external message id. It also adds a path for completing a retrieved message that was first stored without attachments.

The only author code change since round 1 renames the test account variable. That change targets the earlier compile issue, but the current Default and Strict System Application build checks still fail with AL0133, so the branch is still not mergeable. The permission and public-bypass concerns are also unchanged in the net diff.
</details>

<details>

| S1 | Fix the System Application build failure | Not addressed | The latest build still fails with AL0133. |
| S2 | Keep the permission guard on lookup | Not addressed | No change to the lookup permission guard. |
| S3 | Do not trust the public bypass flag | Not addressed | No change to the public bypass overload or validation. |
</details>

<details>

*None - the new commit only renames test variables and does not introduce a separate review finding.*
</details>

<details>

**Risk:** This remains public System Application email surface. A bad change can affect connectors and extensions that use retrieved messages, and the current head cannot be accepted while the app fails to generate.

**Necessity:** The duplicate-retrieval fix is still needed to avoid storing the same provider message more than once. The scope is reasonable for the shared module, but it must build and keep the public API boundary safe.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11619 round=2 by=alexei-dobriansky at=2026-09-21T22:20:12Z lastSha=128d66b5e286c8c8ba050ff13a4a463434e490e1 reviewKey=54c9b99bddde60f7bc74e2d0445cc1710c8b98b07ffbd077abf136bbd964bd88 suggestions=S1@a14cdaa9:notaddressed,S2@4d95880d:notaddressed,S3@33f9ec4c:notaddressed parentRound=1

AleksanderGladkov (approved):

dcenic (approved):

PR #11623, opened 2026-09-18
[29.0] Multiple runs of email retrieval on the same emails don't persist again

mynjj:
Backport of #11612.

## Fix overview


dcenic (approved):

AleksanderGladkov (approved):

alexei-dobriansky:

<details>

This change lets email retrieval detect an already stored provider message by account and external id, return the existing inbox entry, and only add missing attachments when the caller asks for them. The lookup is centralized in the email module, and the Outlook REST API helper exits before creating a second Email Message or Email Inbox row. The attachment path avoids modifying the received Email Message while still leaving queued and sent attachment guards in place.
</details>

<details>

**Fit:** Strong

The reported scenario is duplicate persistence when the same email is retrieved more than once. The diff checks for the existing inbox entry before creating records, marks and returns that entry, and handles missing attachments on a later retrieval, so it addresses the root cause instead of cleaning duplicates after insert.
</details>

<details>

**S1** (🟠 Moderate): *Cover the connector duplicate-retrieval path*
Please add a connector-level regression test in the Outlook REST API test app that calls `RetrieveEmails` twice with the same message. It should verify that the second run returns the existing inbox entry without inserting another one, and that a second run with `Load Attachments` fills attachments for a message first stored without them.
</details>

<details>

**Risk:** The regression surface is the shared email retrieval path used by apps that process Email Inbox rows, plus the new public Email Message overload used to add attachments to already retrieved messages. The queued and sent-message guards still run when inserting attachments, and there is no BaseApp event dependency.

**Necessity:** The change is necessary because repeated retrieval can otherwise persist duplicate inbox rows and trigger duplicate downstream processing. The scope is targeted to lookup-before-create and missing-attachment completion; the remaining gap is test coverage for the actual connector path.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11623 round=1 by=alexei-dobriansky at=2026-09-18T22:21:15Z lastSha=d860285dc2179a849872a2b0872239e147f71a16 reviewKey=075413743ed4bb9cbf3f9ca18e38a3415b0d5319046457108fe8682388deda16 suggestions=S1@a3204f1b

ventselartur (approved):

darjoo (approved):

PR #11624, opened 2026-09-18
[29.x] Multiple runs of email retrieval on the same emails don't persist again

mynjj:
Backport of #11612.

## Fix overview


dcenic (approved):

AleksanderGladkov (approved):

alexei-dobriansky:

<details>

The change prevents duplicate inbox entries when the same Outlook message is retrieved again. It adds a shared lookup for previously retrieved messages, reuses the existing inbox row, and fills attachments when the earlier retrieval did not load them.

The main idea addresses the reported bug, and the new tests cover the single-connector lookup and the retrieved-message attachment path. However, the lookup does not include the connector even though email accounts are keyed by account id plus connector, and the new attachment bypass is public instead of being limited to the retrieval repair flow. Those issues can make the fix return the wrong inbox entry and weaken the protection around external messages.
</details>

<details>

**Fit:** Partial

The fix targets the right symptom: repeated retrieval should reuse the existing inbox entry and only fill missing attachments. It does not fully match the email account model because the lookup ignores connector, and the attachment escape hatch is broader than the missing-attachment case needs.
</details>

<details>

**S1** (🔴 High): *Include connector in duplicate lookup*
Use both `Account Id` and `Connector` when finding an existing inbox entry. Email accounts are keyed by both fields, so the current lookup can return a message from another connector that happens to use the same account id and provider message id. That can skip a real new inbox row and mark the wrong message as retrieved.

**S2** (🔴 High): *Do not expose the retrieved-message bypass*
Do not put the `MessageAlreadyRetrieved` bypass on the public `Email Message.AddAttachment` API. Any extension can pass `true` and add attachments to a retrieved message while skipping the modify guard that normally protects external messages. Use a narrow helper for the retrieval flow that verifies the inbox row and only adds the missing provider attachments.
</details>

<details>

**Risk:** The changed code sits in System Email and the Outlook retrieval path, so a regression can affect any feature that retrieves mail and then processes inbox rows. The new public APIs also affect extensions; a wrong lookup can hide a real message, and an unconstrained bypass can change stored received mail.

**Necessity:** The change is needed because storing the same retrieved message more than once can make downstream processing handle the same email repeatedly. The intended scope is right, but it should be constrained to the exact account connector and the exact missing-attachment repair case.
</details>

---
> [AI-PR-REVIEW] version=1 promptVersion=4 system=github pr=11624 round=1 by=alexei-dobriansky at=2026-09-18T22:19:41Z lastSha=ef773cc273dd400401321b7547162ab0f8d9e0f7 reviewKey=ff06559411d2dacbe1edd85fc459dc4663655e0bedb2e0b3e0faa6cc24d30f09 suggestions=S1@6cc51827,S2@ae050aca

ventselartur (approved):

darjoo (approved):

